PR #965, opened 2014-11-13
HttpClientImpl.requestAbs() losing the query params

adrianluisgonzalez:
`HttpClientImpl.request()` was losing the query params if passing a string absoluteURI.  

From `URL getFile()` Java Docs:

_Gets the file name of this URL. The returned file portion will be the same as getPath(), plus the concatenation of the value of getQuery(), if any. If there is no query portion, this method and getPath() will return identical results._


adrianluisgonzalez:
I added query asserts to `HttpTest`.  This is for vert.x 3, not vert.x 2.1.x.

adrianluisgonzalez:
Any feedback?

adrianluisgonzalez:
Yes, I have signed the Eclipse CLA

PR #976, opened 2014-12-01
Fix sendHead() in http client request

untoldwind:
If the underlying connection is already establish (e.g. by a previous GET request with keep alive), the connected() callback is invoked in the very same thread as sendHead(). Since the writeHead flag is set after the connect the HTTP header is actually never written.
ATM the only workaround is to call sendHead() twice (which is very ugly).


purplefox:
Thanks, do you have a reproducer for this?

purplefox:
Ok, nvm. I'm going to fix this manually :)

PR #986, opened 2015-01-20
increase event bus bridge ping interval to 20s

davidphan:
Signed-off-by: David Phan <gphandavid@gmail.com>

We noticed unexpected connections closed using Safari and Firefox + OS X

In order to save battery life, these browsers increase minimal setInterval to 1000ms (just like Chrome).
But in some case, this interval in increase to 10s. For instance, if your browser is not in your current desktop.

You can find an example that demonstrate the issue here: https://github.com/davidphan/js-interval


purplefox:
Hi @davidphan thanks for the report. Vert.x 2.x is in bug fix only mode now as it has been superseded by 3.0, but thanks anyway.

PR #987, opened 2015-01-22
update to io.vertx~lang-scala_2.11~1.1.0-M1

akostadinov:
Updating `scala` requirement to `io.vertx~lang-scala_2.11~1.1.0-M1:org.vertx.scala.platform.impl.ScalaVerticleFactory`. With the old default it seems there is no `mod-mysql-postgresql` version module that would work.
Reference: https://github.com/vert-x/mod-mysql-postgresql/issues/51


akostadinov:
Is this failure caused by my change of scala version? The module worked for me so it shouldn't cause issues.

purplefox:
Hi @akostadinov is M1 a milestone version? We would normally only set the default lang mods to final version in this config file.

Going to close this for now.

PR #998, opened 2015-02-25
Bug fix : unset the TCCL on wrong thread when executing blocking handler

vietj:
@purplefox please review

### DIFF
--- a/src/main/java/io/vertx/core/impl/ContextImpl.java
+++ b/src/main/java/io/vertx/core/impl/ContextImpl.java
@@ -221,6 +221,10 @@ public abstract class ContextImpl implements Context {
           }
         } catch (Throwable e) {
           res.fail(e);
+        } finally {
+          if (blockingCodeHandler != null) {
+            unsetContext();
+          }
         }
         if (resultHandler != null) {
           runOnContext(v -> res.setHandler(resultHandler));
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/impl/ContextImpl.java
```diff
@@ -228,10 +232,6 @@ public EventLoop getEventLoop() {
       });
     } catch (RejectedExecutionException ignore) {
       // Pool is already shut down
-    } finally {
-      if (blockingCodeHandler != null) {
```
purplefox: agreed that this block should be removed

src/main/java/io/vertx/core/impl/ContextImpl.java
```diff
@@ -221,6 +221,10 @@ public EventLoop getEventLoop() {
           }
         } catch (Throwable e) {
           res.fail(e);
+        } finally {
+          if (blockingCodeHandler != null) {
```
purplefox: I'm thinking that this block is actually not necessary - we don't explicit unset context in other cases, only at close so I think that should be fine in this case too (?)

vietj: ok

src/main/java/io/vertx/core/impl/ContextImpl.java
vietj: that was leaving null thread contextclassloader of the junit thread and issues in subsequents jvm-npm.js deployment

PR #999, opened 2015-03-01
Fix unit tests so they pass on Windows (tested on 8.1).

rexorient:
Signed-off-by: Alain Penders <alain@rexorient.com>

### DIFF
--- a/src/test/java/io/vertx/test/core/DNSTest.java
+++ b/src/test/java/io/vertx/test/core/DNSTest.java
@@ -55,6 +55,8 @@ public class DNSTest extends VertxTestBase {
     assertNullPointerException(() -> dns.resolvePTR(null, ar -> {}));
     assertNullPointerException(() -> dns.resolveNS(null, ar -> {}));
     assertNullPointerException(() -> dns.resolveSRV(null, ar -> {}));
+
+    dnsServer.stop();
   }
 
   @Test
@@ -71,6 +73,7 @@ public class DNSTest extends VertxTestBase {
       testComplete();
     });
     await();
+    dnsServer.stop();
   }
 
   @Test
@@ -86,6 +89,7 @@ public class DNSTest extends VertxTestBase {
       testComplete();
     });
     await();
+    dnsServer.stop();
   }
 
   @Test
@@ -105,6 +109,7 @@ public class DNSTest extends VertxTestBase {
       testComplete();
     });
     await();
+    dnsServer.stop();
   }
 
   @Test
@@ -121,6 +126,7 @@ public class DNSTest extends VertxTestBase {
       testComplete();
     });
     await();
+    dnsServer.stop();
   }
 
   @Test
@@ -137,6 +143,7 @@ public class DNSTest extends VertxTestBase {
       testComplete();
     });
     await();
+    dnsServer.stop();
   }
 
   @Test
@@ -155,6 +162,7 @@ public class DNSTest extends VertxTestBase {
       testComplete();
     });
     await();
+    dnsServer.stop();
   }
 
   @Test
@@ -169,6 +177,7 @@ public class DNSTest extends VertxTestBase {
       testComplete();
     });
     await();
+    dnsServer.stop();
   }
 
   @Test
@@ -196,6 +205,7 @@ public class DNSTest extends VertxTestBase {
       testComplete();
     });
     await();
+    dnsServer.stop();
   }
 
   @Test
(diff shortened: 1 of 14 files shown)


purplefox:
Thanks!

src/test/java/io/vertx/test/core/DNSTest.java
```diff
@@ -55,6 +55,8 @@ public void testIllegalArguments() throws Exception {
     assertNullPointerException(() -> dns.resolvePTR(null, ar -> {}));
     assertNullPointerException(() -> dns.resolveNS(null, ar -> {}));
     assertNullPointerException(() -> dns.resolveSRV(null, ar -> {}));
+
+    dnsServer.stop();
```
purplefox: Would be better to do this in a tearDown.

src/test/java/io/vertx/test/core/FileSystemTest.java
```diff
@@ -538,6 +539,8 @@ public void testChownToOwnUser() throws Exception {
 
   @Test
   public void testChownToOwnGroup() throws Exception {
+    // Not supported in WindowsFileSystemProvider
+    Assume.assumeFalse(System.getProperty("os.name").startsWith( "Windows" ));
```
purplefox: Maybe create a method isWindows() containing:
System.getProperty("os.name").startsWith( "Windows" )
to avoid writing the same code many times

src/test/java/io/vertx/test/core/JsonObjectTest.java
```diff
@@ -1120,18 +1120,18 @@ public void testEncodePrettily() throws Exception {
...
-      "  \"mylong\" : 1234,\n" +
-      "  \"myfloat\" : 1.23,\n" +
-      "  \"mydouble\" : 2.34,\n" +
-      "  \"myboolean\" : true,\n" +
-      "  \"mybinary\" : \"" + strBytes + "\",\n" +
-      "  \"myobj\" : {\n" +
-      "    \"foo\" : \"bar\"\n" +
-      "  },\n" +
-      "  \"myarr\" : [ \"foo\", 123 ]\n" +
+    String expected = "{" + System.getProperty("line.separator") +
```
purplefox: Would be preferable to put in constant to avoid writing out same code multiple times.

src/test/java/io/vertx/test/core/RedeploymentTest.java
```diff
@@ -88,7 +88,13 @@ public void setUp() throws Exception {
 
   @Override
   public void tearDown() throws Exception {
-    vertx.fileSystem().deleteRecursiveBlocking(tempDir.getPath(), true);
+    /* One of the temp files is a jar which is added to the classpath to test verticle redeploy on jar
+     * changes.  This jar remains locked by the JVM until all references to this class loader have been
+     * removed and garbage collected.  The following clean up statement fails because either there is
```
purplefox: Maybe use Junit TemporaryFolder like we do in other tests (e.g. HttpTest)? Perhaps this works better on Windows?

src/test/java/io/vertx/test/core/DNSTest.java
rexorient: I started off doing that, but it seemed weird to clean up dnsServer inside a tearDown method when it is constructed inside the test.  Symmetry and all.  There is also a test which doesn't use it, so the tearDown would have to be intelligent.

src/test/java/io/vertx/test/core/FileSystemTest.java
rexorient: Turns out there already was one... Windows.isWindows().  Done.

src/test/java/io/vertx/test/core/JsonObjectTest.java
rexorient: Done.  I changed a few things.  The Windows class contained only one helper method (isWindows).  I renamed that class to Utils and then added LINE_SEPARATOR in there.  Found another place in the code that defined it's own instance, and pointed that at the Utils version too.

src/test/java/io/vertx/test/core/RedeploymentTest.java
rexorient: The problem with TemporaryFolder is that it fails silently.  I think we want to see failed tests if something changes that prevents the folder from being deleted (on Linux / in CI.)

purplefox:
Thanks

PR #1001, opened 2015-03-06
IsolatingClassLoader getResource() and getResources() should behave the same as loadClass()

adrianluisgonzalez:
loadClass() first tries to load from the current class loader URLs, resources should behave the same way to prevent resources being shared across isolation groups.


adrianluisgonzalez:
Yes I have signed the CLA.

purplefox:
Hi Adrian, I wonder if you could provide a bit more info on this, and a reproducer if at all possible :)

adrianluisgonzalez:
The use case I have run into is around slf4j and log4j.  I configure the vert.x logging to use slf4j and log4j and add the log4j2.xml file to the class path along with the jars.  This works well.  

When I run something as a service, it runs in an isolation group and the slf4j and log4j jars are added to the IsolatingClassLoader so I have new instances of the loggers however the log4j2.xml file is the same one used by vert.x.  This means my file logger tries to write to the same file and the vert.x logging and application logging end up overwriting each other.

My hope was to be able to specify one log4j2.xml file to be used by the vert.x instance and a separate file to be used in my service so they don't write to the same file and overwrite each other.

My current work around is to try to separate things by package name to write to different files, but this is not easy to do and quickly becomes messy.

Any other suggestions on how to work around this?  As long as there are two static instances of the log4j loggers between different class loaders I think there will be problems if they use the same log4j2 configuration.

I'll put together a running example.

adrianluisgonzalez:
Here is a project to reproduce:
https://github.com/adrianluisgonzalez/vertx-log4j-problem

You can see the problem in the log file: 
https://github.com/adrianluisgonzalez/vertx-log4j-problem/blob/master/app.log#L59

adrianluisgonzalez:
@purplefox, perhaps rather than trying to force resource files to work for my issue, a better question is whether there is a way to get at the same class instances of the slf4j and log4j loggers.  I would prefer to have one single log file anyway.

Would you consider allowing a way to add to the list of vert.x and system classes in [IsolatingClassLoader.isVertxOrSystemClass()](https://github.com/englishtown/vert.x/blob/feature/isolatingclassloader/src/main/java/io/vertx/core/impl/IsolatingClassLoader.java#L122) via a system property?

Then I can add `org.slf4j` and `org.apache.log4j` at application start and everything should have the same logger instance.

purplefox:
Hi Adrian-
I'm a little confused here. You state that "My hope was to be able to specify one log4j2.xml file to be used by the vert.x instance and a separate file to be used in my service so they don't write to the same file and overwrite each other.", but I can only see one log4j2.xml file in the repo.

adrianluisgonzalez:
Sorry Tim, I could have been a lot clearer.  Let me try again.

__Problem__
Two instances of the slf4j and log4j loggers exist and both use the same log4j2.xml configuration and therefore write to the same log file overwriting entries in the following scenario:
* Running as a service so the `IsolatingClassLoader` is used to deploy my verticle
* Vert.x is configured to use the `io.vertx.core.logging.impl.SLF4JLogDelegateFactory`
* My application has dependencies that use slf4j and access the slf4j LoggerFactory directly

__Cause__
The `SLF4JLogDelegateFactory` is used by vert.x and the slf4j and log4j jars are added to the main vert.x class loader.  When a Verticle is deployed with an `IsolatingClassLoader` the jars are added again to the new class loader.  The `IsolatingClassLoader` has different instances of the slf4j and log4j classes since they are not considered [vert.x or system classes](https://github.com/eclipse/vert.x/blob/3.0.0-milestone3/src/main/java/io/vertx/core/impl/IsolatingClassLoader.java#L42).  However, the log4j2.xml file is always the same one since `getResource()` and `getResources()` are __not__ overridden.  Even if a new log4j2.xml file is added as a resource in my application's jar, the `UrlClassLoader` getResourceAsStream() will try to retrieve the resource from the parent class loader first.

The sample project I posted was to show the problem when there are  two loggers overwriting each other.

__Solutions__
1. Allow adding additional classes/packages to the list of vert.x and system classes so a single instance of the loggers is used.  Something like this: https://github.com/englishtown/vert.x/blob/feature/isolatingclassloader2/src/main/java/io/vertx/core/impl/IsolatingClassLoader.java#L94
2. Override the getResource() and getResources() behavior.  This pull request is for this solution.
3. Use JUL logging as the vert.x logger and SLF4J as my application logger.  No change required.

I think #1 is the most flexible option (and I would prefer to have one instance of loggers), #3 will have problems as soon as any of my application's dependencies use JUL.  Or maybe you have a better approach you could recommend?


I hope that better explains my problem.

src/main/java/io/vertx/core/impl/IsolatingClassLoader.java
```diff
@@ -63,6 +66,59 @@
...
+    return url;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Enumeration<URL> getResources(String name) throws IOException {
+    @SuppressWarnings("unchecked")
+    Enumeration<URL>[] enums = (Enumeration<URL>[]) new Enumeration<?>[2];
```
purplefox: Wouldn't it be simpler to just put them in an ArrayList then return Collections.enumeration(list) ?

purplefox:
Thanks :) I understand now

adrianluisgonzalez:
What are your thoughts on how to resolve this?  Change IsolatingClassLoader to allow adding system classes? Or override getResource() and getResources()?  Or something else?

My preference is to allow adding system classes so I have only one instance of the logger class types.

Thanks for your input.

purplefox:
I prefer overridding getResources - as IsolatingClassloader is meant to isolate. Your PR looks ok, just that comment about enumeration...

adrianluisgonzalez:
Ok, I just followed the approach in ClassLoader.getResources(), but happy to simplify as you describe.

adrianluisgonzalez:
Updated the commit.  Is that what you are looking for?

purplefox:
Thanks

PR #1002, opened 2015-03-18
Bug fix : cannot use event bus from worker

vietj:
@purplefox please review


purplefox:
I fixed this in a somewhat different way. But thanks anyway :)

vietj:
I was really not sure of the fix, hence the review. the testcases were useful :-)

purplefox:
Yeah thanks, I copied the test cases :)

PR #1003, opened 2015-03-25
Eventbus metrics

vietj:
Changes to improve eventbus metrics reporting to provide additional and more accurate metrics.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -1076,7 +1076,14 @@ public class EventBusImpl implements EventBus, MetricsProvider {
         }
       } else {
         checkNextTick();
-        handler.handle(event);
+        metrics.beginHandleMessage(metric);
+        try {
+          handler.handle(event);
+          metrics.endHandleMessage(metric, null);
+        } catch (Exception e) {
+          metrics.endHandleMessage(metric, e);
+          throw e;
+        }
       }
     }
 
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -410,7 +410,7 @@ public class EventBusImpl implements EventBus, MetricsProvider {
                 socket.write(PONG);
               }
             } else {
-              receiveMessage(received, -1, null, null);
+              receiveMessage(received, -1, null, null, false);
             }
           }
         }
@@ -429,18 +429,27 @@ public class EventBusImpl implements EventBus, MetricsProvider {
       // Choose one
       ServerID sid = subs.choose();
       if (!sid.equals(serverID)) {  //We don't send to this node
+        metrics.messageSent(message.address(), false, false, true);
         sendRemote(sid, message);
       } else {
-        receiveMessage(message, timeoutID, asyncResultHandler, replyHandler);
+        metrics.messageSent(message.address(), false, true, false);
+        receiveMessage(message, timeoutID, asyncResultHandler, replyHandler, true);
       }
     } else {
       // Publish
+      boolean local = false;
+      boolean remote = false;
       for (ServerID sid : subs) {
         if (!sid.equals(serverID)) {  //We don't send to this node
+          remote = true;
           sendRemote(sid, message);
         } else {
-          receiveMessage(message, timeoutID, null, replyHandler);
+          local = true;
         }
+      }
+      metrics.messageSent(message.address(), true, local, remote);
+      if (local) {
+        receiveMessage(message, timeoutID, null, replyHandler, true);
       }
     }
   }
@@ -472,7 +481,6 @@ public class EventBusImpl implements EventBus, MetricsProvider {
   private <T> void sendOrPub(ServerID replyDest, MessageImpl message, DeliveryOptions options,
                              Handler<AsyncResult<Message<T>>> replyHandler) {
     checkStarted();
-    metrics.messageSent(message.address(), !message.send());
     Handler<Message<T>> simpleReplyHandler = null;
     long timeoutID = -1;
     if (replyHandler != null) {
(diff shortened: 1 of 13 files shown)


vietj:
@purplefox  can you review these changes please ?

purplefox:
Sure thing! :)

BTW... do you remember that user report of handler metric count not going down from some time back? Did you get a chance to look at that?

vietj:
btw there is a matching PR for metrics project here : https://github.com/vert-x3/vertx-metrics/tree/eventbus-metrics

vietj:
> do you remember that user report of handler metric count not going down from some time back? 

I'm not sure to understand can you clarify ?

vietj:
In the current implementation of metrics with Codahale, there is no blocking code, except perhaps JMX implementation but I don't think it's doing blocking stuff. The other reporting that uses IOs (ganglia, etc...) are asynchronous and scan periodically the Codahale metrics to send reports.

jtruelove:
Slightly on a tangent as far as codahale integration goes is there an
option to run it in a raw mode or use an implementation of an interface the
user provides? For example nothing gets aggregated it just gets collected
and batched forwarded out. One issue we noticed in some of dropwizard
projects is it tends to make assumptions about your data, ie it aggregates
everything up and you lose all granularity it's more about a server than a
fleet of servers. It would be nice to have a generic way to plug in your
own reporter. If I've got something wrong/misunderstood the implementation
please fill me in.

vietj:
sure thing @jtruelove actually this project is an implementation of Vert.x metrics SPI (https://github.com/eclipse/vert.x/tree/master/src/main/java/io/vertx/core/spi/metrics) and have the ability to support only reporting (like sending to collectd, etc...)

So one can come up with your own implementation of the Vert.x core metrics SPI easily.

Codahale is one of them, it does a bit more than reporting since you can obtain metrics snapshots in json format. We are going to rename this project "vertx-codahale-metrics" or something like that to remove the ambiguity.

jtruelove:
Yeah one of the most frustrating things about frameworks like dropwizard is
the metrics approach is hardcoded in and you can't do much other than just
try and turn it completely off if you want something else.

vietj:
@jtruelove do you have a particular metrics reporting system in mind ?

jtruelove:
I threw this together a bit back https://github.com/cyngn/mod-opentsdb, I
am generally interested in tracking service metrics at a granular level. If
the framework lets me easily access say HTTP timing data from start to
finish that's even better. I tend to find the more generic the system the
better at least for my purposes that way you can build any reporting from
the ground up outside the service itself, ie tracking 90th percentile of
request time on an API across all of my service instances while still being
able to dig into that same metric on an individual service instance if
needed.

vietj:
ok, if you can make a branch for vertx 3 and test the SPI it would be an awesome thing :-)

jtruelove:
Yeah I'll take a look, I haven't actually messed with vert.x 3 at all. Are
there updated docs yet?

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -736,17 +750,26 @@ private void checkStarted() {
...
+    final HandlerRegistration<T> handler;
     final boolean replyHandler;
     final boolean localOnly;
     final long timeoutID;
     boolean removed;
 
     // We use a synchronized block to protect removed as it can be unregistered from a different thread
-    synchronized void setRemoved() {
-      removed = true;
+    void setRemoved() {
```
purplefox: Was the synchronized deliberately removed?

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -1061,7 +1084,15 @@ public synchronized void handle(Message<T> event) {
         }
       } else {
         checkNextTick();
-        handler.handle(event);
+        MessageImpl abc = (MessageImpl) event;
+        metrics.beginHandleMessage(metric, abc.getSocket() == null);
+        try {
+          handler.handle(event);
+          metrics.endHandleMessage(metric, null);
```
purplefox: Why is it useful to measure begin and end of handling a message?

purplefox:
Other than the comments.. looks good to me :)

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: the synchronization is done in a synchronized block in the method to have the metrics.handlerUnregistered callback to be done out of this block.

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: it allows the metric implementation to measure the processing time of the callback.

purplefox: shouldn't the processing time of the callback be effectively zero (or very small), unless the user is blocking the event loop?

And, if they are blocking the event loop, the event loop blocking code will kick in anyway and warn them.

vietj: I see what you mean.

Your point is about detecting blocking code in the logs.

This provides the opportunity to the metrics implementation to report the precise behavior of the eventbus message delivery and log them in a time serie database to make queries.

purplefox: I still question its value. If the execution time is always approx. zero, what's the point?

I'd also be concerned about such fine grained metrics killing performance. Is there some way of enabling metrics but disabling the more fine grained stuff, or is metrics being enabled all or nothing?

vietj: Everything lies down in the implementation of the metrics SPI. The EventBusMetrics implementation with Codahale has been modified to care only about the handlers that match a certain address.

When an handler is registered, it is matched against a list of permitted address : https://github.com/vert-x3/vertx-metrics/blob/eventbus-metrics/src/main/java/io/vertx/ext/metrics/impl/EventBusMetricsImpl.java#L99

When nothing matches then it returns null.

Then when the handler metric is null then nothing is happens:

https://github.com/vert-x3/vertx-metrics/blob/eventbus-metrics/src/main/java/io/vertx/ext/metrics/impl/EventBusMetricsImpl.java#L123
https://github.com/vert-x3/vertx-metrics/blob/eventbus-metrics/src/main/java/io/vertx/ext/metrics/impl/EventBusMetricsImpl.java#L136

It is fine grained indeed and the metric SPI implementation should really care about it. A broken SPI implementation can certainly make vertx behave badly.

purplefox: OK, fair enough

PR #1004, opened 2015-03-26
Use correct thread pools

purplefox:
@vietj please review

Previously we weren't using the correct pools for different types of operations, this commit fixes it.


vietj:
looks good to me

PR #1005, opened 2015-03-26
Removed VertxSPI as is now redundant

purplefox:
@vietj please review


purplefox:
Pretty straightforward this one. Basically removed VertxSPI as Vertx.executeBlocking can now be used directly.

PR #1008, opened 2015-03-27
Check the message body for sessionID, not the message JSON itself

michalboska:
During message reply handling, vert.x Event bus bridge decides if the reply contains authentication data (sessionID) which it should cache and then keep using for that socket. The reply is stored as "sessionID" variable in the reply body JSON object.

The code however checks for presence of this property in the message itself (message contains properties such as "type":"send", "address":"xxxxx" etc, not the message body itself. Therefore, such session ID in a reply is never found.

This patch fixes the issue, so that vert.x searches for the session ID in the message body.

Signed-off-by: Michal Boska <michal@boska.me>

### DIFF
--- a/vertx-core/src/main/java/org/vertx/java/core/sockjs/EventBusBridge.java
+++ b/vertx-core/src/main/java/org/vertx/java/core/sockjs/EventBusBridge.java
@@ -374,7 +374,12 @@ public class EventBusBridge implements Handler<SockJSSocket> {
     }
     if (curMatch.doesMatch) {
       if (curMatch.requiresAuth) {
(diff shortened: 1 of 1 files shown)


michalboska:
Update: this applies for each message sent through the EventBus bridge to an endpoint that requires authentication (basically everything processed by org.vertx.java.core.sockjs.EventBusBridge#doSendOrPub )

michalboska:
I actually discovered that this behavior is not a bug and why it happens this way. Shortly - you need to set the sessionID directly on the client javascript eventbus object, e.g. eb.sessionID = "a123456.....";
Closing this pull request.

PR #1009, opened 2015-03-31
Websocket metrics

vietj:
@purplefox can you review this please ?

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/ClientConnection.java
@@ -80,8 +80,8 @@ class ClientConnection extends ConnectionBase {
   // Requests can be pipelined so we need a queue to keep track of requests
   private final Queue<HttpClientRequestImpl> requests = new ArrayDeque<>();
   private final Handler<Throwable> exceptionHandler;
-  private final HttpClientMetrics metrics;
   private final Object metric;
+  final HttpClientMetrics metrics;
 
   private WebSocketClientHandshaker handshaker;
   private HttpClientRequestImpl currentRequest;
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -637,6 +637,7 @@ public class HttpServerImpl implements HttpServer, Closeable, MetricsProvider {
         ServerWebSocketImpl ws = new ServerWebSocketImpl(vertx, theURI.toString(), theURI.getPath(),
           theURI.getQuery(), new HeadersAdaptor(request.headers()), wsConn, shake.version() != WebSocketVersion.V00,
           connectRunnable, options.getMaxWebsocketFrameSize());
+        ws.metric = metrics.connected(wsConn.metric(), ws);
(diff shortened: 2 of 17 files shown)


vietj:
can you review it @purplefox ?

src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -80,8 +80,8 @@
   // Requests can be pipelined so we need a queue to keep track of requests
   private final Queue<HttpClientRequestImpl> requests = new ArrayDeque<>();
   private final Handler<Throwable> exceptionHandler;
-  private final HttpClientMetrics metrics;
   private final Object metric;
+  final HttpClientMetrics metrics;
```
purplefox: I am not a huge fan of package protected members. Any chance of making private and providing getters?

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -637,6 +637,7 @@ private void handshake(FullHttpRequest request, Channel ch, ChannelHandlerContex
         ServerWebSocketImpl ws = new ServerWebSocketImpl(vertx, theURI.toString(), theURI.getPath(),
           theURI.getQuery(), new HeadersAdaptor(request.headers()), wsConn, shake.version() != WebSocketVersion.V00,
           connectRunnable, options.getMaxWebsocketFrameSize());
+        ws.metric = metrics.connected(wsConn.metric(), ws);
```
purplefox: Do we distinguish between client websockets connected and server websockets?

vietj: I'm not sure to understand, can you be more precise ?

src/main/java/io/vertx/core/http/impl/ClientConnection.java
vietj: no problems with that. (I don't mind package protected members when they are final)

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
purplefox: My question is: What does the websocket "connected" metric represent? Does it represent only server websockets that have been connected at the server side, or client websockets connected at the client side? Or both?

purplefox:
I assume the the documentation for the metrics needs to be updated too (list of metrics)?

https://github.com/vert-x3/vertx-dropwizard-metrics/blob/master/src/main/asciidoc/java/index.adoc

I'm thinking the above should really be in core, not in the dropwizard metrics proj?

vietj:
so the dropwizard impl has not yet been pushed. I'm waiting for validation in vertx core.

the doc for dropwizard is really specific to dropwizard

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
vietj: it represents a server web socket that connected.

there are two SPI, one for client (HttpClientMetrics.connected with WebSocket), one for server (HttpServerMetrics#connected with ServerWebSocket)

PR #1011, opened 2015-04-07
Resolved API/impl dependency for json module.

atsticks:
Signed-off-by: atsticks <atsticks@gmail.com>
There were circular dependencies between the json's api part and its impl. In most other modules these deps are resolved, so I resolved it here similarly by adding a JSonServices interface, registered using the ServiceLoader component. As an alternative one could refactor the JSON package completely to a util module (with the disadvantage of a hardcoded dep to the Jackson framework, of course).
The change should be backward compatible (I deprecated the former impl/Json class because now the JSonServices should be used instead of, where needed).
Nevertheless the API is relatively weak from an API design perspective. Maybe we should think on changing method signatures on JSonServices to be more explicit (Object can be anything). Feel free to comment, adapt etc.

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonServices.java
+++ b/src/main/java/io/vertx/core/json/JsonServices.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2011-2014 The original author or authors
+ * ------------------------------------------------------
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Apache License v2.0 which accompanies this distribution.
+ *
+ *     The Eclipse Public License is available at
+ *     http://www.eclipse.org/legal/epl-v10.html
+ *
+ *     The Apache License v2.0 is available at
+ *     http://www.opensource.org/licenses/apache2.0.php
+ *
+ * You may elect to redistribute this code under either of these licenses.
+ */
+package io.vertx.core.json;
+
+import io.vertx.core.ServiceHelper;
+
+import java.util.Objects;
+
+/**
+ * Abstraction for Json related functionality used by the API. Implementations must register using the
+ * {@link io.vertx.core.ServiceHelper} mechanism (by default based on {@link java.util.ServiceLoader}.
+ */
+public interface JsonServices {
+    /** The current instance loaded. */
+    static final JsonServices INSTANCE = ServiceHelper.loadFactory(JsonServices.class);
+
+    /**
+     * Accessor for getting the singleton instane.
+     * @return the services instance, not null.
+     */
+    public static JsonServices of(){
+        return Objects.requireNonNull(INSTANCE, "No JsonServices loaded.");
+    }
+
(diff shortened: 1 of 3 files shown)


atsticks:
Interesting, I signed the CLA just a few minutes ago. Seems that it requires some time to synchronize...

purplefox:
Could you provide a bit more information on the details of the circular dependency you found?

atsticks:
I dont remember if the deps really were circular (it is quite a long time ago, since I committed that pull request). What I remember is that I tried to adapt and exchange things within verty at that time, and I struggled over several refs where the implementation (AFAIK it was mainly the http server part) was casting the interfaces to the implementation types internally, rendering the usage of interfaces more or less useless, at least as far as I want to provide some alternate implementation of them. Looking at the API AFAIR all artifacts were separated, except this JSON part here, which directly depended on the Jackson library. 
The key point is that when I have an artifact that is part of a (hopefully decoupled) API the artifact should not depend on anything external to the API package if possible. A dependency to Jackson I would say, is quite a strong dep and IMO should to be part of a utitliy or implementation class (and conseuquently not be part of the API). That was the reason for the given separation.

I was looking briefly into the latest version of vertx also quite some time ago. I did not had time to have a very deep look, but I saw that at least some of the issues I identified were resolved ;) So I would see this ticket more some kind of feedback, than a one-to-one thing to be imported into the current source tree. 
If you need further info, I can go back the history of changes in the former Git tree and give you more details, just let me know.

purplefox:
Vert.x has strong dependencies to Jackson and Netty (actually very few 3rd party dependencies compared to most projects).

There's no intention to change Jackson at the moment, and if we did we would rewrite the impl. I'm not a huge fan of making every single thing super pluggable for the remote possibility that sometime in the future it might be changed (but probably never will).

So, I'm going to close this PR. But thanks anyway :)

atsticks:
You're welcome!
Anatole

2015-09-15 17:55 GMT+02:00 Tim Fox <notifications@github.com>:

> Vert.x has strong dependencies to Jackson and Netty (actually very few 3rd
> party dependencies compared to most projects).
>
> There's no intention to change Jackson at the moment, and if we did we
> would rewrite the impl. I'm not a huge fan of making every single thing
> super pluggable for the remote possibility that sometime in the future it
> might be changed (but probably never will).
>
> So, I'm going to close this PR. But thanks anyway :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/eclipse/vert.x/pull/1011#issuecomment-140440340>.
>



-- 
*Anatole Tresch*
Java Engineer & Architect, JSR Spec Lead
Glärnischweg 10
CH - 8620 Wetzikon

*Switzerland, Europe Zurich, GMT+1*
*Twitter:  @atsticks*
*Blogs: **http://javaremarkables.blogspot.ch/
<http://javaremarkables.blogspot.ch/>*

*Google: atsticksMobile  +41-76 344 62 79*

PR #1012, opened 2015-04-10
VertxImpl.getContext should not return a context create from another Ver...

vietj:
...tx instance


vietj:
@purplefox please review Tim

purplefox:
Looks good.

PR #1014, opened 2015-04-14
Allow Starter to run in its own classloader if -cp command line option i...

purplefox:
...s provided

@vietj please review


purplefox:
@vietj ping :)

PR #1019, opened 2015-04-18
exposed the number of instances of a verticle that were started

jtruelove:

### DIFF
--- a/src/main/java/io/vertx/core/Context.java
+++ b/src/main/java/io/vertx/core/Context.java
@@ -127,4 +127,9 @@ public interface Context {
    */
   boolean remove(String key);
 
+  /**
+   * Get the number of instances of the verticle that were started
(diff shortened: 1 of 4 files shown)


purplefox:
Jeremy,

DeploymentTest is probably a good place :)

Also, can you two spaces for tabs, not four?

jtruelove:
also in the case of a null deployment or no deployment options should the function return 1 or 0? Currently it's 0 but it should probably be implicitly 1 I'd think.

jtruelove:
updated with tests

jtruelove:
I can also re-branch and submit the pull so the checkin comments aren't janked, I tried remote amending a commit like a newb.

jtruelove:
new pull request here exposed the number of instances of a verticle that were started

PR #1020, opened 2015-04-22
Update Jackson to 2.5.2

marcust:
From what we've seen module jars can not override the core runtime jars. We are using vertx-mod-jersey on top of Vertx, which uses jackson for serialization and deserialization. Unfortunately the 2.2 Version has some issues when it comes to polymorphic serialization.
It would be really great if we could have a 2.1.6 Version of Vertx with the updated libraries.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,8 +26,8 @@ group=io.vertx
 
 # dependency versions
 hazelcastVersion=3.2.3
-jacksonCoreVersion=2.2.2
-jacksonDatabindVersion=2.2.2
+jacksonCoreVersion=2.5.2
+jacksonDatabindVersion=2.5.2
 nettyVersion=4.0.21.Final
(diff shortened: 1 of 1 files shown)


purplefox:
Thanks, I will deal with this manually as it needs a manual merge.

marcust:
I actually wanted to have that on the master branch and not in Vert.x latest. Any chance that you can backport that?

purplefox:
I think you mean you want this on the 2.x branch I guess? Yes I can do this too if it doesn't break anything :) (master is Vert.x 3)

marcust:
Yes, I mean the 2.x branch. 
Well, the tests still run, I used it. Don't know where that would actually break anything.

marcust:
@purplefox Any chance to get that into the next 2.x release?

PR #1021, opened 2015-04-23
exposed the number of instances of a verticle that were started

jtruelove:

### DIFF
--- a/src/main/java/io/vertx/core/impl/ContextImpl.java
+++ b/src/main/java/io/vertx/core/impl/ContextImpl.java
@@ -279,5 +279,11 @@ public abstract class ContextImpl implements Context {
     Thread.currentThread().setContextClassLoader(tccl);
   }
 
-
+  public int getInstanceCount(){
+    if(deployment == null || deployment.deploymentOptions() == null) {
+      // implicitly there is the current instance of the verticle running always
+      return 1;
+    }
+    return deployment.deploymentOptions().getInstances();
+  }
 }
--- a/src/main/java/io/vertx/core/impl/ContextImpl.java
+++ b/src/main/java/io/vertx/core/impl/ContextImpl.java
@@ -280,8 +280,11 @@ public abstract class ContextImpl implements Context {
   }
 
   public int getInstanceCount(){
-    if(deployment == null || deployment.deploymentOptions() == null) {
-      // implicitly there is the current instance of the verticle running always
+    // the no verticle case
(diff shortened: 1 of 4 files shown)


jtruelove:
anything else needed on this?

vietj:
Hi,

we will review and validate it for the next milestone4 hopefully.

src/main/java/io/vertx/core/impl/ContextImpl.java
```diff
@@ -279,5 +279,11 @@ private void setTCCL() {
     Thread.currentThread().setContextClassLoader(tccl);
   }
 
-
+  public int getInstanceCount(){
+    if(deployment == null || deployment.deploymentOptions() == null) {
+      // implicitly there is the current instance of the verticle running always
```
purplefox: If deployment == null then it should return zero as it's possible to have a context without a verticle (e.g. in the embedded case)

src/test/java/io/vertx/test/core/DeploymentTest.java
```diff
@@ -987,6 +987,46 @@ public void testUndeployAllNoDeployments() throws Exception {
...
+    vertx.eventBus().consumer("instanceCount", event -> {
+      messageCount.incrementAndGet();
+      totalReportedInstances.addAndGet((int)event.body());
+      if(messageCount.intValue() == 3) {
+        assertEquals(9, totalReportedInstances.get());
+      }
+    });
+
+    vertx.deployVerticle(TestVerticle3.class.getCanonicalName(), new DeploymentOptions().setInstances(3), ar -> {
+      assertTrue(ar.succeeded());
```
purplefox: Not sure this test is valid as the deployment handler could be called before the message consumer has got all the instanceCount messages.

Better to put the testComplete() in the message handler.

src/main/java/io/vertx/core/impl/ContextImpl.java
jtruelove: Gotcha, so should the logic then be

if(deployment == null) {
   return 0;
} else if ( deployment.deploymentOptions == null) {
  return 1;
}

src/test/java/io/vertx/test/core/DeploymentTest.java
jtruelove: kk, easy fix

jtruelove:
updated with feedback

PR #1024, opened 2015-04-27
Fixed typo which caused JDBC to be described as async

andyhedges:


purplefox:
Thanks Andy, I will fix this manually

PR #1025, opened 2015-04-27
SockJS Session must call super.close()

fleaflicker:
Must call super.close or writeHandler is never unregistered from EventBus. This means the Session object cannot be garbage collected.


fleaflicker:
Sorry for the confusion, I see where it is called.

PR #1028, opened 2015-05-03
bz466228:  BlockedThreadChecker doesn't have a config value for the time until a stack trace is written

alexlehm:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=466228

added config value warningExceptionTime to VertxOptions to set the threshold value when a stack trace
is included in the blocking warning. changed the log code to put the stack trace into the log statement
so that it is output by the logger and not to stdout
added a unit test for the BlockChecker class

### DIFF
--- a/src/main/java/io/vertx/core/VertxOptions.java
+++ b/src/main/java/io/vertx/core/VertxOptions.java
@@ -562,7 +562,7 @@ public class VertxOptions {
    * @return the warning exception time threshold
    */
   public long getWarningExceptionTime() {
-    return warningExceptionTime ;
+    return warningExceptionTime;
   }
 
   /**
--- a/src/main/java/io/vertx/core/VertxOptions.java
+++ b/src/main/java/io/vertx/core/VertxOptions.java
@@ -572,6 +572,9 @@ public class VertxOptions {
    * @return a reference to this, so the API can be used fluently
    */
   public VertxOptions setWarningExceptionTime(long warningExceptionTime) {
+    if (warningExceptionTime < 1) {
+      throw new IllegalArgumentException("warningExceptionTime must be > 0");
+    }
     this.warningExceptionTime = warningExceptionTime;
     return this;
   }
--- a/src/main/java/io/vertx/core/VertxOptions.java
+++ b/src/main/java/io/vertx/core/VertxOptions.java
@@ -77,7 +77,7 @@ public class VertxOptions {
   /**
    * The default value of max event loop execute time = 2000000000 ns (2 seconds)
    */
-  public static final long DEFAULT_MAX_EVENT_LOOP_EXECUTE_TIME = 2000l * 1000000;
+  public static final long DEFAULT_MAX_EVENT_LOOP_EXECUTE_TIME = 2l * 1000 * 1000000;
 
   /**
    * The default value of max worker execute time = 60000000000 ns (60 seconds)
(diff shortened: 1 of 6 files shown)


src/main/java/io/vertx/core/impl/BlockedThreadChecker.java
```diff
@@ -45,12 +46,14 @@ public void run() {
           long execStart = thread.startTime();
           long dur = now - execStart;
           if (execStart != 0 && dur > (thread.isWorker() ? maxWorkerExecTime : maxEventLoopExecTime)) {
```
purplefox: duplicate evaulation, better to do it once.

alexlehm: ok, changed it

src/main/java/io/vertx/core/VertxOptions.java
```diff
@@ -545,6 +556,26 @@ public VertxOptions setMetricsOptions(MetricsOptions metrics) {
...
+  }
+
+  /**
+   * Set the threshold value above this, the blocked warning contains a stack trace.
+   *
+   * @param warningExceptionTime
+   * @return a reference to this, so the API can be used fluently
+   */
+  public VertxOptions setWarningExceptionTime(long warningExceptionTime) {
+    this.warningExceptionTime = warningExceptionTime;
```
purplefox: Need to check argument is not < 1 as in other getters

src/test/java/io/vertx/test/core/BlockedThreadCheckerTest.java
```diff
@@ -0,0 +1,82 @@
...
+import io.vertx.core.AbstractVerticle;
+import io.vertx.core.DeploymentOptions;
+import io.vertx.core.Verticle;
+import io.vertx.core.Vertx;
+import io.vertx.core.VertxOptions;
+
+import org.junit.Test;
+
+/**
+ * @author <a href="http://tfox.org">Tim Fox</a>
```
purplefox: Did I write this test? ;)

src/test/java/io/vertx/test/core/BlockedThreadCheckerTest.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+public class BlockedThreadCheckerTest extends VertxTestBase {
```
purplefox: I don't really see the point of this test.

What we should do is have tests in VertxOptionsTest as we do with the other settings

src/test/java/io/vertx/test/core/BlockedThreadCheckerTest.java
alexlehm: sorry, I copied another test class and exchanged the test methods so the comments ended up there, should I change that?

src/test/java/io/vertx/test/core/BlockedThreadCheckerTest.java
alexlehm: main reason for the test was to get 100% coverage, actually asserting the logging calls is a bit difficult since it requires a log mock of some kind

alexlehm: i added tests to VertxOptionsTest

src/test/java/io/vertx/test/core/BlockedThreadCheckerTest.java
alexlehm: ok, i changed the author comment

src/main/java/io/vertx/core/VertxOptions.java
alexlehm: added

alexlehm:
ok, i think i implemented the necessary chanced, can you please do a new review

purplefox:
thanks alex

PR #1032, opened 2015-05-06
MessageConsumer#endHandler should be invoked with the registration context

vietj:
@purplefox please review this fix

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -1150,7 +1150,17 @@ public class EventBusImpl implements EventBus, MetricsProvider {
 
     @Override
     public synchronized MessageConsumer<T> endHandler(Handler<Void> endHandler) {
-      this.endHandler = endHandler;
+      if (endHandler != null) {
+        // We should use the HandlerHolder context to properly do this (needs small refactoring)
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -1150,7 +1150,17 @@ public synchronized boolean isRegistered() {
 
     @Override
     public synchronized MessageConsumer<T> endHandler(Handler<Void> endHandler) {
-      this.endHandler = endHandler;
+      if (endHandler != null) {
+        // We should use the HandlerHolder context to properly do this (needs small refactoring)
```
purplefox: I agree. So why not do that? ;)

vietj: I prefer to postpone until next week and open an issue to do be done carefully.

purplefox: ok fair enough. don't forget though ;)

vietj: that being said you can do it if you can

vietj: I won't forget

PR #1034, opened 2015-05-08
Added helper methods to determine whether running on worker or event …

purplefox:
…loop thread and now allow httpclient/server and netclient/server to run on workers

@vietj please review


vietj:
The contextData map is a plain HashMap and is accessed by different threads, it requires some kind of synchronization.

purplefox:
Julien, I don't think I changed anything in the context data in this PR... ?

vietj:
yes it was just a remark I found when reading the code in my PR in the IDE :-) . Besides that it's good PR and useful feature for developer.

PR #1035, opened 2015-05-11
Vertx end handler fixes

vietj:
@purplefox can you review this PR please ?

### DIFF
--- a/src/test/java/io/vertx/test/core/LocalEventBusTest.java
+++ b/src/test/java/io/vertx/test/core/LocalEventBusTest.java
@@ -1087,7 +1087,7 @@ public class LocalEventBusTest extends EventBusTestBase {
   }
 
   @Test
-  public void testUnregisterationOfRegisteredConsumerCallsEndHandlerWithMessaqgeStream() {
+  public void testUnregisterationOfRegisteredConsumerCallsEndHandlerWithMessageStream() {
     MessageConsumer<String> consumer = eb.consumer(ADDRESS1);
     testUnregisterationOfRegisteredConsumerCallsEndHandler(consumer, consumer);
   }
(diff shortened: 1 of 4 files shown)


src/test/java/io/vertx/test/core/TimerTest.java
```diff
@@ -163,26 +164,33 @@ public void testTimerStreamExceptionDuringHandle() throws Exception {
...
+    AtomicInteger count = new AtomicInteger();
     timer.handler(l -> {
-      fail();
-    });
-    timer.endHandler(v -> {
-      testComplete();
+      if (count.incrementAndGet() == 1) {
+        timer.handler(null);
+        vertx.setTimer(200, id -> {
+          assertEquals(1, count.get());
```
purplefox: If you want to verify a timer is only called once, then probably need to wait a bit for completing test

src/test/java/io/vertx/test/core/TimerTest.java
```diff
@@ -163,26 +164,33 @@ public void testTimerStreamExceptionDuringHandle() throws Exception {
...
+        fail();
+      }
     });
-    timer.handler(null);
     await();
   }
 
   @Test
   public void testTimerStreamCancellation() throws Exception {
     TimeoutStream timer = vertx.timerStream(10);
```
purplefox: I suspect this will cause non deterministic failures on CI as 10 ms is very short

src/test/java/io/vertx/test/core/LocalEventBusTest.java
```diff
@@ -1128,69 +1121,6 @@ public void testUnregisterThenUnsetEndHandler() {
     await();
   }
 
-  private void testUnregistrationOfUnregisteredConsumerCallsEndHandler(MessageConsumer<String> consumer, ReadStream<?> readStream) {
```
purplefox: Quite a few tests have been removed - are there corresponding tests for the new behaviour?

vietj: they were actually testing the previous behavior, one test was kept to check endHandler is not called when unregistering and only when eventbus is closing.

src/test/java/io/vertx/test/core/TimerTest.java
vietj: I think I can improve the test by instead wrapping it in a runOnContext, so we are sure that it will never be executed.

PR #1036, opened 2015-05-12
Code refactor 

gitplaneta:
I found that there are places in the code which do not use new features of java 7 and 8. I've committed  a handful of them. I've seen that static analysis found much more places that could be easily refactored. 

If you think it's worth to commit such improvements in the Vert.x project, I would be glad to spend some more time to help improve the code quality by refactoring.


purplefox:
Thank you

PR #1037, opened 2015-05-13
Metrics thread/context tests and fixes

vietj:
@purplefox please review this contrib

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -261,6 +261,9 @@ public class EventBusImpl implements EventBus, MetricsProvider {
   @Override
   public void close(Handler<AsyncResult<Void>> completionHandler) {
     unregisterAllHandlers();
+    if (metrics != null) {
+      metrics.close();
+    }
     if (server != null) {
       server.close(ar -> {
         if (ar.failed()) {
@@ -574,12 +577,20 @@ public class EventBusImpl implements EventBus, MetricsProvider {
       }
       if (subs != null && !replyHandler && !localOnly) {
         // Propagate the information
-        subs.add(address, serverID, registration::setResult);
+        subs.add(address, serverID, ar -> {
+          holder.context.runOnContext(v -> {
+            registration.setResult(ar);
+          });
+        });
       } else {
-        registration.setResult(Future.succeededFuture());
+        holder.context.runOnContext(v -> {
+          registration.setResult(Future.succeededFuture());
+        });
       }
     } else {
-      registration.setResult(Future.succeededFuture());
+      holder.context.runOnContext(v -> {
+        registration.setResult(Future.succeededFuture());
+      });
     }
 
     handlers.list.add(holder);
@@ -772,7 +783,9 @@ public class EventBusImpl implements EventBus, MetricsProvider {
         }
       }
       if (unregisterMetric) {
-        metrics.handlerUnregistered(handler.metric);
+        context.runOnContext(v -> {
+          metrics.handlerUnregistered(handler.metric);
+        });
       }
     }
 
@@ -991,6 +1004,7 @@ public class EventBusImpl implements EventBus, MetricsProvider {
     private final Queue<Message<T>> pending = new ArrayDeque<>(8);
     private boolean paused;
     private Object metric;
+    private Context context;
 
     public HandlerRegistration(String address, boolean replyHandler, boolean localOnly, long timeoutID) {
       this.address = address;
@@ -1065,8 +1079,12 @@ public class EventBusImpl implements EventBus, MetricsProvider {
       registered = false;
     }
 
+    /**
+     * This method must be invoked the handler registration context/thread.
+     */
     private synchronized void setResult(AsyncResult<Void> result) {
       this.result = result;
+      this.context = Vertx.currentContext();
       if (completionHandler != null) {
         if (result.succeeded()) {
           metric = metrics.handlerRegistered(address, replyHandler);
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -46,6 +46,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 
 /**
  * This class is thread-safe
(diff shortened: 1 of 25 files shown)


vietj:
is it ok for merge @purplefox ?

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -261,6 +261,9 @@ public EventBus unregisterDefaultCodec(Class clazz) {
   @Override
   public void close(Handler<AsyncResult<Void>> completionHandler) {
     unregisterAllHandlers();
+    if (metrics != null) {
+      metrics.close();
```
purplefox: this one not run on a context?

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -574,12 +577,20 @@ private String generateReplyAddress() {
...
         // Propagate the information
-        subs.add(address, serverID, registration::setResult);
+        subs.add(address, serverID, ar -> {
+          holder.context.runOnContext(v -> {
+            registration.setResult(ar);
+          });
+        });
       } else {
-        registration.setResult(Future.succeededFuture());
+        holder.context.runOnContext(v -> {
```
purplefox: Looks like its running on a context twice (registration.setrestult will also call context.runOnContext)

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -574,12 +577,20 @@ private String generateReplyAddress() {
...
+          registration.setResult(Future.succeededFuture());
+        });
       }
     } else {
-      registration.setResult(Future.succeededFuture());
+      holder.context.runOnContext(v -> {
+        registration.setResult(Future.succeededFuture());
```
purplefox: as above

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -772,7 +783,9 @@ void setRemoved() {
         }
       }
       if (unregisterMetric) {
-        metrics.handlerUnregistered(handler.metric);
+        context.runOnContext(v -> {
```
purplefox: as above

src/main/java/io/vertx/core/spi/metrics/EventBusMetrics.java
```diff
@@ -26,7 +26,10 @@
 public interface EventBusMetrics<H> extends Metrics {
 
   /**
-   * Called when a handler is registered on the event bus.
+   * Called when a handler is registered on the event bus.<p/>
+   *
+   * This method is invoked with {@link io.vertx.core.Context} and thread of the message handler and therefore
```
purplefox: ... and therefore MIGHT be different on....

src/main/java/io/vertx/core/spi/metrics/NetworkMetrics.java
```diff
@@ -20,7 +20,12 @@
 
 /**
  * An SPI used internally by Vert.x to gather metrics on a net socket which serves
- * as a base class for TCP or UDP.
+ * as a base class for TCP or UDP.<p/>
+ *
+ * All the methods on this object including the methods inherited from the super interfaces are invoked
+ * with {@link io.vertx.core.Context} and thread of the network server/client and therefore are the same than
```
purplefox: what if it's in a worker verticle?

vietj: I will add a test for this and correct the doc :-)

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: will check

src/main/java/io/vertx/core/spi/metrics/NetworkMetrics.java
vietj: I think actually I need to test worker verticle (context) but also an executeBlocking block used in an event loop verticle.

vietj: btw we are talking here of using an http/net client/server to be used with a worker that was added recently, right ?

vietj: so
- the SPI can determine at creation time if it's in a event loop verticle / worker verticle
- when it's a worker verticle, what is guaranteed is:
    - the life cycle of a request will always use the same thread : so any object created during the life cycle of the request will use the same thread
    - at most one thread will make a callback in the spi

therefore I think the super interface like TCPMetrics / NetworkMetrics should only specify to look at the doc of the _concrete_ interfaces and these should be updated with this info.

does it make sense ?

purplefox: If it's a worker verticle there is no guarantee that a context always uses the same thread. That guarantee you only get with standard verticles.

vietj: ok I see clearly now, so in case of a worker we can only guarantee that at most one thread makes the callback.

vietj: well actually I meant that for a given request/connection that thread will be the same as Netty assigns a Netty event loop (which can be Vert.x worker or not) to a channel handler.

vietj: actually forgot what I said previously, I was only seing the Netty side and missed the executeFromIO override in WorkerContext that redispatch on a worker thread

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: to which particular context should the EventBus run ? that's why I haven't done anything here.

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: confirmed, indeed.

vietj: actually it needs to be done when the context has just been created (embedded) example.

vietj:
@purplefox I updated the PR with testing + doc for worker + fixes for other remarks. Can you finish the review ?

vietj:
@purplefox can you continue the review please :-) ?

purplefox:
Yes sorry, I need to look at this in some detail, but it's not been too high up my priority list recently, hence the delay.

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
purplefox: I think we should just say in the documentation that metrics.close() can be called from any thread.

Then we don't need to run close on a context anywhere which will simplify things.

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -574,12 +584,14 @@ private String generateReplyAddress() {
       }
       if (subs != null && !replyHandler && !localOnly) {
         // Propagate the information
-        subs.add(address, serverID, registration::setResult);
+        subs.add(address, serverID, ar -> {
```
purplefox: Looks like you're effectively doing runOnContext twice here too

purplefox:
So.. this was merged? Weren't there outstanding issues from the comments to address first?

vietj:
yes it is merged, I removed the close/create constraints and the unnecessary runOnContext, I will revert the merge though.

PR #1041, opened 2015-05-19
Reworked RecordParser to be able "codegen"

LarsTimm:
Split the RecordParser into an interface annotated with VertxGen and an implementation class.

No logic changes and the existing unit-test still passes. I have tested with vertx-lang-scala that it can be translated.

Signed-off-by: LarsTimm <larsdtimm@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/parsetools/RecordParser.java
+++ b/src/main/java/io/vertx/core/parsetools/RecordParser.java
@@ -79,15 +79,14 @@ public interface RecordParser extends Handler<Buffer> {
 
   /**
    * Create a new {@code RecordParser} instance, initially in delimited mode, and where the delimiter can be represented
-   * by the {@code byte[]} delim.
+   * by the {@code Buffer} delim.
    * <p>
    * {@code output} Will receive whole records which have been parsed.
    *
-   * @param delim  the initial delimiter byte[]
+   * @param delim  the initial delimiter buffer
    * @param output  handler that will receive the output
    */
-  @GenIgnore
-   static RecordParser newDelimited(byte[] delim, Handler<Buffer> output) {
+   static RecordParser newDelimited(Buffer delim, Handler<Buffer> output) {
      return RecordParserImpl.newDelimited(delim, output); 
    }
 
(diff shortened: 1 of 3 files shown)


vietj:
what's your email Lars ?

vietj:
sorry, ignore that I found it in the signature :-)

src/main/java/io/vertx/core/parsetools/RecordParser.java
```diff
@@ -115,11 +86,10 @@ public static RecordParser newDelimited(String delim, Handler<Buffer> output) {
    * @param delim  the initial delimiter byte[]
    * @param output  handler that will receive the output
    */
-  public static RecordParser newDelimited(byte[] delim, Handler<Buffer> output) {
-    RecordParser ls = new RecordParser(output);
-    ls.delimitedMode(delim);
-    return ls;
-  }
+  @GenIgnore
+   static RecordParser newDelimited(byte[] delim, Handler<Buffer> output) {
```
vietj: I think we need to have a non @GenIgnore version with Buffer as delim.

src/main/java/io/vertx/core/parsetools/RecordParser.java
```diff
@@ -157,13 +122,8 @@ public void delimitedMode(String delim) {
...
    */
-  public void delimitedMode(byte[] delim) {
-    Objects.requireNonNull(delim, "delim");
-    delimited = true;
-    this.delim = delim;
-    delimPos = 0;
-    reset = true;
-  }
+  @GenIgnore
+  void delimitedMode(byte[] delim);
```
vietj: need polyglot version with Buffer

LarsTimm:
I will look into in asap. Do you think the buffer version should replace the byte[] or just an alternative?

vietj:
just keep the Buffer version that will be enough.

LarsTimm:
Something like this you had in mind Julien?

vietj:
yes, thanks

LarsTimm:
Just came to see. We don't have a way of initializing a Buffer with some form of byte list in the generated languages, only string.
You can do that in Java with byte[] but they are marked as GenIgnore also.

Do we need to do something about that?

vietj:
some languages don't support byte[] (like JS), however they can build a buffer with Buffer.appendByte method. I agree that the Groovy version could be "patched" to have direct byte[] factory method.

that's something to look after 3.0 maybe.

LarsTimm:
Ahh, okay. Scala could really benefit from having access to those methods also.

vietj:
yes indeed, thanks for the contribution btw :-)

PR #1042, opened 2015-05-20
Metrics thread contexts fixes

vietj:
same PR than before (so with removed create/close context/thread constraints)
+ race condition fixes for websocket client/server workers that were seen in metrics test


vietj:
@purplefox can you review please ?

vietj:
it is same review than before + bugfixes for websocket worker client/server

PR #1043, opened 2015-05-20
Client/server worker fixes

vietj:
@purplefox please review

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/ClientConnection.java
@@ -80,7 +80,6 @@ class ClientConnection extends ConnectionBase {
   // Requests can be pipelined so we need a queue to keep track of requests
   private final Queue<HttpClientRequestImpl> requests = new ArrayDeque<>();
   private final Handler<Throwable> exceptionHandler;
-  private final Object metric;
   private final HttpClientMetrics metrics;
 
   private WebSocketClientHandshaker handshaker;
@@ -88,6 +87,7 @@ class ClientConnection extends ConnectionBase {
   private HttpClientResponseImpl currentResponse;
   private HttpClientRequestImpl requestForResponse;
   private WebSocketImpl ws;
+  private Object metric;
 
   ClientConnection(VertxInternal vertx, HttpClientImpl client, Handler<Throwable> exceptionHandler, Channel channel, boolean ssl, String host,
                    int port, ContextImpl context, ConnectionLifeCycleListener listener, HttpClientMetrics metrics) {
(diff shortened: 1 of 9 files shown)


src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -146,6 +149,7 @@ synchronized void toWebSocket(String requestURI, MultiMap headers, WebsocketVers
       p.addBefore("handler", "handshakeCompleter", new HandshakeInboundHandler(wsConnect, version != WebSocketVersion.V00));
       handshaker.handshake(channel).addListener(future -> {
         if (!future.isSuccess() && exceptionHandler != null) {
+          // On good thread ?
```
purplefox: ?

src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -183,46 +187,48 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
 
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
-      context.executeFromIO(() -> {
-        synchronized (ClientConnection.this) {
-          if (handshaker != null && handshaking) {
-            if (msg instanceof HttpResponse) {
-              HttpResponse resp = (HttpResponse) msg;
-              if (resp.getStatus().code() != 101) {
+      synchronized (ClientConnection.this) {
```
purplefox: This doesn't look right. The synchronized block needs to be inside the context.executeIO - this is to ensure that there is a memory barrier over access to ClientConnection internal state.

src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -242,8 +248,12 @@ private void handshakeComplete(ChannelHandlerContext ctx, FullHttpResponse respo
       }
       ws = new WebSocketImpl(vertx, ClientConnection.this, supportsContinuation, client.getOptions().getMaxWebsocketFrameSize());
       handshaker.finishHandshake(channel, response);
+
       log.debug("WebSocket handshake complete");
-      wsConnect.handle(ws);
+      context.executeFromIO(() -> {
```
purplefox: handshake complete is executed from channelread which is already executed on the right thread so this appears to be doing it twice.

vietj: when I execute the test (WebsocketTest#testClientWorker), the channel thread is an eventloop thread, not sure to understand what you mean.

src/main/java/io/vertx/core/http/impl/ClientConnection.java
vietj: At this point the thread is an event loop thread. So it means we need it in the executeFromIO smaller blocks.

PR #1044, opened 2015-05-20
Metrics context/thread fixed and javadoc

vietj:
@purplefox here it is.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ServerConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/ServerConnection.java
@@ -109,7 +109,7 @@ class ServerConnection extends ConnectionBase {
     return metric;
   }
 
-  void setMetric(Object metric) {
+  synchronized void setMetric(Object metric) {
     this.metric = metric;
   }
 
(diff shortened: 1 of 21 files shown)


src/main/java/io/vertx/core/http/impl/ServerConnection.java
```diff
@@ -102,14 +102,17 @@
...
-    this.metric = metrics.connected(remoteAddress());
   }
 
   @Override
-  protected Object metric() {
+  protected synchronized Object metric() {
     return metric;
   }
 
+  void setMetric(Object metric) {
```
purplefox: not synchronized?

src/main/java/io/vertx/core/http/impl/ServerConnection.java
```diff
@@ -102,14 +102,17 @@
     this.server = server;
     this.handshaker = handshaker;
     this.metrics = metrics;
-    this.metric = metrics.connected(remoteAddress());
   }
 
   @Override
-  protected Object metric() {
+  protected synchronized Object metric() {
```
purplefox: shouldn't it be metrics(), not metric() ?

purplefox: Typed as Object. really?

purplefox:
Other than the comments, looks fine

src/main/java/io/vertx/core/http/impl/ServerConnection.java
vietj: good catch

src/main/java/io/vertx/core/http/impl/ServerConnection.java
vietj: yes, this is an handback object provided by the metrics SPI that contains data associated with the connection

PR #1045, opened 2015-05-20
Improve generic type usage

gitplaneta:
Previous version of Json class allows this instruction to fail during runtime instead of compilation time:

```
TypeA instance = Json.decodeValue(str, TypeB.class)
```

Signed-off-by: Radoslaw Busz <radoslaw.busz@gmail.com>


purplefox:
thanks

PR #1048, opened 2015-05-27
Network worker 3 reproducers + 2 fixes

vietj:


vietj:
extra reproducer with more low level server networking https://github.com/eclipse/vert.x/commits/websocket-client-bug-handsake

PR #1056, opened 2015-06-02
Enable pattern replacement (Slf4j / Logback like)

PatrickSauts:
The changes introduce what recent Loggers allow 

So know you can do

```
String myParam = "there"
LOG.debug("Hi {}.", myParam)
```
##### That will produce

```
Hi there.
``` 
##### Only if debug is enabled.

--------------------------

The main difference with 

```
LOG.debug("Hi "+ myParam + ".")
```

Is that in that case the string is always built at run time even if LOG.isDebugEnabled() == false

So 

```
LOG.debug("Hi {}.", myParam)
```

is equivalent to 

```
if(LOG.isDebugEnabled()){
  LOG.debug("Hi "+ myParam + ".")
}
```

You can write the Logs the way you're use to or use this neat way.

Example

```
LOG.debug("the message pattern {} {} {} {} {} {} ...", "param1", 2, 2.3, 200000000000000000l, 2.9999999999, new String[]{"param7", "param8", "param9", "param10"})
```

That will produce

```
the message pattern param1 2 2.3 200000000000000000 2.9999999999 [param7, param8, param9, param10] ...
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,7 @@
     <metrics.version>3.0.2</metrics.version>
     <log4j.version>1.2.17</log4j.version>
     <slf4j.version>1.7.7</slf4j.version>
+    <logback.version>1.1.3</logback.version>
     <junit.version>4.11</junit.version>
     <asciidoclet.version>1.5.1</asciidoclet.version>
     <apacheds-protocol-dns.version>1.5.7</apacheds-protocol-dns.version>
@@ -105,6 +106,12 @@
       <artifactId>slf4j-api</artifactId>
       <version>${slf4j.version}</version>
       <scope>provided</scope>
+    </dependency>
+    <dependency>
+   <groupId>ch.qos.logback</groupId>
+     <artifactId>logback-classic</artifactId>
+     <version>${logback.version}</version>
+     <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>io.vertx</groupId>
--- a/src/main/java/io/vertx/core/logging/helper/LogFormatter.java
+++ b/src/main/java/io/vertx/core/logging/helper/LogFormatter.java
@@ -1,0 +1,345 @@
+package io.vertx.core.logging.helper;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Inspired from LogBack and Slf4j<br>
+ * 
+ * @author Patrick Sauts
+ */
+final public class LogFormatter {
+  static final char PARAM_START = '{';
+  static final char PARAM_STOP = '}';
+  static final String PARAM_STR = "{}";
+  private static final char ESCAPE_CHAR = '\\';
+  private static final String SEPARATOR = ", ";
+
+  /**
+   * @param messagePattern
+   * @param arg
+   * @return
+   */
+  final public static LogTupple format(String messagePattern, Object arg) {
+    if(arg instanceof Object[]) return paramsFormat(messagePattern, (Object[]) arg);
+    return paramsFormat(messagePattern, new Object[] { arg });
+  }
+
+  /**
+   * @param messagePattern
+   * @param arg1
+   * @param arg2
+   * @return
+   */
+  final public static LogTupple format(final String messagePattern, Object arg1, Object arg2) {
+    return paramsFormat(messagePattern, new Object[] { arg1, arg2 });
+  }
+
+  /**
+   * @param messagePattern
+   * @param params
+   * @return
+   */
+  final public static LogTupple paramsFormat(final String messagePattern, final Object[] params) {
+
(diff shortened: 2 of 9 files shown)


src/main/java/io/vertx/core/logging/impl/Log4jLogDelegate.java
```diff
@@ -100,4 +86,146 @@ private void log(Level level, Object message) {
   private void log(Level level, Object message, Throwable t) {
     logger.log(FQCN, level, message, t);
   }
+
+  @Override
+  public void fatal(String format, Object arg) {
+    if (logger.isEnabledFor(Level.FATAL)) {
+      logger.fatal(LogFormatter.format(format, arg).getMessage());
+    }
+    if (logger.isEnabledFor(Level.FATAL)) {
```
purplefox: Is this deliberately in here twice?

pom.xml
```diff
@@ -107,6 +108,12 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+   <groupId>ch.qos.logback</groupId>
```
purplefox: This dependency doesn't actually seem to be used anywhere in the code

src/main/java/io/vertx/core/logging/impl/LogBackLogDelegateFactory.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * 
+ * @author Patrick Sauts
+ *
+ */
+public class LogBackLogDelegateFactory implements LogDelegateFactory
+{
+   public LogDelegate createDelegate(final String clazz)
+   {
+      return new SLF4JLogDelegate(clazz);
```
purplefox: This class returns a SLF4jLogDelegate so it's just the same as the SLF4JLogDelegateFactory, so what's the point?

src/main/java/io/vertx/core/logging/helper/LogFormatter.java
```diff
@@ -0,0 +1,361 @@
...
+    if(throwableCandidate != null) --length;
+    
+    if (messagePattern == null) {
+      return new LogTupple(join(params,length), throwableCandidate);
+    }
+
+
+    int i = 0;
+    int j;
+    StringBuilder sbuf = new StringBuilder(messagePattern.length() + 50);
```
purplefox: Couldn't you use the JDK Pattern and Matcher classes to do this for you, rather than coding it all yourself?

purplefox:
I've cleaned up and manually merged this into a new PR here:

https://github.com/eclipse/vert.x/pull/1062

But I'm not yet 100% convinced about it

pom.xml
PatrickSauts: That's right logback dependency is not needed to compile as it uses slf4j.
I'm removing that.

src/main/java/io/vertx/core/logging/impl/Log4jLogDelegate.java
PatrickSauts: no it was a mistake

src/main/java/io/vertx/core/logging/impl/LogBackLogDelegateFactory.java
PatrickSauts: Well if someone is not aware that LogBack uses Slf4j. 
The point was mainly for users not to ask themselves too many questions.
And as it is a very small Class.
Let me know if you want me to remove it.

PatrickSauts:
What is the part you're not convince about ? Is it the code ? I can try and come up with a more elegant solution in the future.

src/main/java/io/vertx/core/logging/helper/LogFormatter.java
PatrickSauts: Well handling escapes and double escapes make things a little more tricky. And I actually took the same approach than the Slf4j team.
I actually have some code to do that with patterns:

```
private static final String PARAM_PREFIX = '{'
  private static final String PARAM_SUFFIX = '}'
  private static final String PARAM_REGEX = ???

  private def replaceParams(String template, List params){
    int index = 0;
    int paramIndex = 0;
    Matcher matcher = Pattern.compile(PARAM_REGEX).matcher(template);
    StringBuilder sb = new StringBuilder();
    while (matcher.find()) {
      int start = matcher.start();
      int end = matcher.end();
      sb.append(template.substring(index, start));
      index = end;
      String tagParam = template.substring(start, index);
      String value = params.size() > paramIndex?params.get(paramIndex):"{}";
      sb.append(value);
    }
    sb.append(template.substring(index))
    return sb.toString()
  }
```

The right PARAM_REGEX is to be  written though.

PR #1061, opened 2015-06-07
Updated libraries versions

fmatar:


fmatar:
@vietj I can update the version to 6 the build runs smoothly with those libraries, however I can revert the netty version to 4.0.25. Would that work?

purplefox:
Hmm, we should really be using the latest Netty for the 3.0 release!

purplefox:
The versions we submit to Eclipse should be the latest ones. Netty 4.0.25 is about 6 months old

vietj:
ok, but we haven't tested the 4.0.28 version, and we have been running fine with 4.0.25. We can update to 4.0.28 but that will impact the process. Let me know.

purplefox:
I've just updated it and the test suite seems ok. We can't release V3 with a 6 month old Netty version

normanmaurer:
@purplefox @vietj FYI I will release 4.0.29.Final most likely end of the week.

vietj:
@normanmaurer good to know
@purplefox no problem, I will file an Eclipse CQ with this update

purplefox:
@normanmaurer I bet you're glad you don't have to deal with Eclipse process rules any more ;)

normanmaurer:
@purplefox I have other rules to follow ;)

PR #1062, opened 2015-06-07
Logging params

purplefox:


PatrickSauts:
Hi,

Does this mean you're not going to merge it to the master ?
I'm not the only one who worked on a similar thing 
https://github.com/eclipse/vert.x/pull/991/commits

So I guess that is something the community wants.
Can you let me know why it was closed ?

Thank you.

Patrick.

purplefox:
Hi Patrick,

I have thought about this and seems to me it does not make a lot of sense to do the log formatting in Vert.x but better to delegate it to the logging framework, e.g. SLF4J, which has a highly optimised parameterized logging implementation.

This is the approach taken in https://github.com/eclipse/vert.x/pull/991

Many thanks for your efforts though!

PatrickSauts:
That's the best choice ! Good that your moving away from delegation but use SLF4J directly instead !!! 

Thank you. I actually made that code thinking that the point was not to always use SLF4J before some logger.

Very Good news !!! Thank you !

PR #1063, opened 2015-06-08
Provide the client HTTP Error handling documentation

cescoffier:
Was mostly there already, just add an example to handle non 2XX responses.

### DIFF
--- a/src/main/asciidoc/java/http.adoc
+++ b/src/main/asciidoc/java/http.adoc
@@ -741,7 +741,7 @@ request.end(buffer);
 When you're writing to a request, the first call to `write` will result in the request headers being written
 out to the wire.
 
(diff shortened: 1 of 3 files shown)


src/main/asciidoc/java/http.adoc
```diff
@@ -850,9 +850,25 @@ request.exceptionHandler(e -> {
...
+  }
+  if (response.statusCode() == 500) {
+    System.out.println("Unexpected behavior on the server side");
+    return;
+  }
+});
+request.end();
+----
+
+IMPORTANT: `XXXNow` methods cannot receive an exception handler.
```
purplefox: This is a good point, and is probably a bug in Vert.x

PR #1067, opened 2015-06-09
Use connection as lock for synchronizing access to request/response t…

purplefox:
…o avoid potential deadlock conditions

This is a fix for https://groups.google.com/forum/?fromgroups#!topic/vertx/ck7FVOgZoe4

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerRequestImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerRequestImpl.java
@@ -58,8 +58,10 @@ import java.util.Map;
  * This class is optimised for performance when used on the same event loop that is was passed to the handler with.
  * However it can be used safely from other threads.
  *
- * The internal state is protected using the synchronized keyword. If always used on the same event loop, then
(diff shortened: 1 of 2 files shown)


purplefox:
Hi @vietj @cescoffier @pmlopes

could someone review this please?

cescoffier:
So, you are always using the Server Connection as a lock, so yes it would remove the deadlock. The only impact I see is that some code that was able to run in parallel (_i.e._ be executed by two threads) before won't anymore. Not sure it has a big impact.

purplefox:
WIth all Vert.x core classes we optimise for the normal case of everything being accessed on the correct event loop, in which case biased locking kicks in and there is very little overhead of synchronized.

However it is possible in some cases for the objects to be accessed by other threads (e.g. a worker thread in the case of execute blocking) which is why we synchronize. But we don't optimise for this case.

Using serverconnection lock for both request and response means different threads can't access request and response concurrently, but this is a very unusual use case anyway.

cescoffier:
Very unusual and probably not what we want, so it's fine for me.

PR #1072, opened 2015-06-11
Documentation improvements

cescoffier:
Move the non translatable code and language specific parts in the `docoverride` so it can be customized per languages.

This PR will be followed by PR on the different languages.


purplefox:
Hi Clement - for Eclipse core unfortunately commits need to be signed unless you're a comitter to the core project.

cescoffier:
Missed one, but it's actually better. I've made progress on the last missing section of the core documentation (parse tools). So, I close that one, and will re-open another PR with everything signed off and completing the core documentation.

PR #1073, opened 2015-06-13
Fix behaviour of absoluteURI

purplefox:
@cescoffier @pmlopes @vietj please review

Fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=470125

### DIFF
--- a/src/test/java/io/vertx/test/core/HttpTest.java
+++ b/src/test/java/io/vertx/test/core/HttpTest.java
@@ -4082,6 +4082,30 @@ public class HttpTest extends HttpTestBase {
     await();
   }
 
+  @Test
+  public void testAbsoluteURIServer() {
+    server.close();
+    // Listen on all addresses 
+    server = vertx.createHttpServer(new HttpServerOptions().setPort(DEFAULT_HTTP_PORT).setHost("0.0.0.0"));
+    server.requestHandler(req -> {
+      String absURI = req.absoluteURI();
+      assertEquals("http://localhost:8080/path", absURI);
+      req.response().end();
+    });
+    server.listen(onSuccess(s -> {
+      String host = "localhost";
+      String path = "/path";
+      int port = 8080;
+      System.out.println("Client: GET Request-URI = http://" + host + ":" + port + path);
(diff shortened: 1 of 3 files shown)


src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -4082,6 +4082,30 @@ public void start() throws Exception {
...
+    server.requestHandler(req -> {
+      String absURI = req.absoluteURI();
+      assertEquals("http://localhost:8080/path", absURI);
+      req.response().end();
+    });
+    server.listen(onSuccess(s -> {
+      String host = "localhost";
+      String path = "/path";
+      int port = 8080;
+      System.out.println("Client: GET Request-URI = http://" + host + ":" + port + path);
```
vietj: println should be removed

purplefox:
If someone could take a quick look that would be great, but if not I'm probably going to have to merge anyway as the next release is coming up fast.

cescoffier:
I gonna have a look this morning.

Sent from my iPhone

On 18 juin 2015, at 08:31, Tim Fox <notifications@github.com> wrote:

If someone could take a quick look that would be great, but if not I'm
probably going to have to merge anyway as the next release is coming up
fast.

—
Reply to this email directly or view it on GitHub
<https://github.com/eclipse/vert.x/pull/1073#issuecomment-113051653>.

leolux:
Thanks! I updated my reproducer because the HTTP Server Metrics still contains 0.0.0.0
https://github.com/leolux/vertx-absolute-uri

purplefox:
I think the metrics naming is a separate issue, and is due to the way the metrics name is identified is using host and port - but of course this is not unique.

In Vert.x you can have many servers listening on port 0.0.0.0, and the port value it uses is not the actual port.

@vietj

purplefox:
I've made a small change in core so different http server metrics with same host/port are now distinguishable.

But the server metric will still say 0.0.0.0 if you told your server to listen on port 0.0.0.0. At metrics creation time it's not possible to know the value of Server header sent to the server from the client.

PR #1077, opened 2015-06-17
Log the deploy redeploy error

vietj:
@purplefox please review


purplefox:
Really no need for a PR for this, it's a small change that you can do direct in master :)

vietj:
well, since it was not logging the exception before, I thought there might be a good reason for not doing it so I preferred you ACK it :-)

PR #1078, opened 2015-06-17
Add Vertx Start Hooks Methods in io.vertx.core.Starter

catandgrep:
This change enables subclasses of io.vertx.core.Starter to implement custom behavior before
and after the vertx instance has been started by overriding these methods:

* beforeStartingVertx(VertxOptions options)
* afterStartingVertx()

### DIFF
--- a/src/main/java/io/vertx/core/Starter.java
+++ b/src/main/java/io/vertx/core/Starter.java
@@ -171,6 +171,20 @@ public class Starter {
     stopLatch.countDown();
   }
 
+  /**
+   * Hook for sub classes of {@link Starter} before the vertx instance is started.
+   */
+  protected void beforeStartingVertx(VertxOptions options) {
+    
+  }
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/Starter.java
```diff
@@ -171,6 +171,20 @@ public void unblock() {
     stopLatch.countDown();
   }
 
+  /**
+   * Hook for sub classes of {@link Starter} before the vertx instance is started.
+   */
+  public void beforeStartingVertx(VertxOptions options) {
```
purplefox: Shouldn't they be protected not public?

catandgrep: Indeed. Visibility is now protected on both methods. PR updated.

purplefox:
Thanks!

PR #1080, opened 2015-06-17
Various performance improvements and optimisations

purplefox:


purplefox:
@vietj @cescoffier @pmlopes please review

src/main/java/io/vertx/core/datagram/impl/DatagramServerHandler.java
```diff
@@ -22,28 +22,35 @@
...
-        super(vertx, new HashMap<>());
-    this.server = server;
+  private final DatagramSocketImpl socket;
+
+  DatagramServerHandler(DatagramSocketImpl socket) {
+    this.socket = socket;
   }
 
   @Override
   public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
```
cescoffier: Is this method still useful ? Looks it just call `super`.

purplefox: You're right. I've removed it.

purplefox:
Hi Clement - did you finish the review of this, is it ok to merge?

Cheers

cescoffier:
I've made a quick review (just have my phone right now). So maybe my comment are completely wrong.

I've a couple of questions:

You detect whether or not the websockets need to be handled based on the type of handler. Would this let registering a web socket handler afterwards ? 

You set the TCCL without storing the previous one and setting it back. This can be tricky on environment manipulating the TCCL no ?

Anything else looks good to me.

purplefox:
Thanks Clement.

To answer your questions:

>> You detect whether or not the websockets need to be handled based on the type of handler. Would this let registering a web socket handler afterwards ?

If you register a websocket handler the expectWebsockets flag gets set to true thus enabling websockets.

>> You set the TCCL without storing the previous one and setting it back. This can be tricky on environment manipulating the TCCL no ?

I think you are referring to the setting in ContextImpl? The threads here are always either our own event loop threads or worker threads, so there should never be any meaningful TCCL that we need to restore

cescoffier:
About the TCCL, the code executed by the worker thread (hopefully not by
the event loop) needs to do the TCCL switch itself and restore the value
afterwards. Maybe something we need to document somewhere.

So, for me, everything else is ok can can be merged.

Sent from my iPhone

On 18 juin 2015, at 14:24, Tim Fox <notifications@github.com> wrote:

Thanks Clement.

To answer your questions:

 You detect whether or not the websockets need to be handled based on the
type of handler. Would this let registering a web socket handler afterwards
?

 If you register a websocket handler the expectWebsockets flag gets set to
true thus enabling websockets.

 You set the TCCL without storing the previous one and setting it back.
This can be tricky on environment manipulating the TCCL no ?

 I think you are referring to the setting in ContextImpl? The threads here
are always either our own event loop threads or worker threads, so there
should never be any meaningful TCCL that we need to restore

—
Reply to this email directly or view it on GitHub
<https://github.com/eclipse/vert.x/pull/1080#issuecomment-113136600>.

purplefox:
Clement,

I'm not sure I understand your point about the TCCL, could you elaborate a bit? Thanks

purplefox:
I'm going to merge now but we can continue discussing the TCCL issue and make any necessary changes afterwards.

cescoffier:
Yes I will do it as soon as I've my computer (not easy to write code from
my phone ;-)).


Sent from my iPhone

On 18 juin 2015, at 15:21, Tim Fox <notifications@github.com> wrote:

Clement,

I'm not sure I understand your point about the TCCL, could you elaborate a
bit? Thanks

—
Reply to this email directly or view it on GitHub
<https://github.com/eclipse/vert.x/pull/1080#issuecomment-113153977>.

PR #1081, opened 2015-06-18
Ws frame methods

purplefox:


purplefox:
Add some helper methods for sending websocket final frames.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=469870

purplefox:
@cescoffier @vietj @pmlopes please review

src/main/asciidoc/java/http.adoc
```diff
@@ -1251,24 +1251,24 @@ client.websocket("/some-uri", websocket -> {
 
 ==== Writing messages to WebSockets
 
-If you wish to write a single binary WebSocket message containing a single WebSocket frame to the WebSocket (a
-common case) the simplest way to do this is to use `link:../../apidocs/io/vertx/core/http/WebSocket.html#writeMessage-io.vertx.core.buffer.Buffer-[writeMessage]`:
+If you wish to write a single binary WebSocket message to the WebSocket you can do this with
+`link:../../apidocs/io/vertx/core/http/WebSocket.html#writeBinaryMessage-io.vertx.core.buffer.Buffer-[writeBinaryMessage]`:
```
cescoffier: I've found this a bit misleading in comparison to https://github.com/eclipse/vert.x/pull/1081/files#diff-1ee4c083f53d5cb366fbb5aa4adb042aR1296.

Maybe it needs some clarifications.

purplefox: Which part do you find misleading?

cescoffier: Well it should answer to the question "which method should I used when I want to send a single message to the websocket ?"

In this section you mention 'writeBinaryMessage', but later you mention 'writeFinalBinaryFrame'. You this second method be mentioned here too ?

purplefox: Sorry I don't follow which part of this comment is misleading:
"If you wish to write a single binary WebSocket message to the WebSocket you can do this with"
It seems correct to me.
Binary message and binary frame are different things.

cescoffier: you are right, it's different and it is much better when reading the whole documentation (not only the commits / diff). Anyway, if we have questions on this we can still improve it later.

PR #1086, opened 2015-06-23
Remove broken redeploy functionality, and change isolation group beha…

purplefox:
…viour to only isolate explicitly declared classes and packages, this results in a much simpler (usually) parent first delegation model which generally helps us all to retain our sanity.


purplefox:
@vietj @cescoffier @pmlopes please review

PR #1092, opened 2015-07-01
Shutdown Netty eventloopgroup gracefully to avoid annoying java.util.…

purplefox:
…concurrent.RejectedExecutionException exceptions in logs when Vert.x instance is shutdown

@pmlopes please review - this should remove those annoying exceptions in the logs


pmlopes:
code looks fine, built locally and re-run redis tests and i can report that the stacktrace does not show anymore and tests still pass :)

Good to merge

PR #1093, opened 2015-07-01
make possible to extend Starter and use it in fatjar

pwielgolaski:
Small change of reading manifest file, instead of looking for Starter class it should try to look for itself in this way it can be used to extend implementation of Starter.

On top of that logger is created as instance variable to make it possible to be shared by extending classes.

Signed-off-by: Piotr Wielgolaski <pwielgolaski@gmail.com>


purplefox:
I'm going to close this as it is superseded by Clement's recent work on Starter/CLI refactoring that we're going to merge very soon.

pwielgolaski:
New refactoring work look really promising, I'm looking forward to get it in official release.

PR #1094, opened 2015-07-02
Starter now stops vert.x instance of deploy unsuccessful, also remova…

purplefox:
…l of unnecessary block methdods


cescoffier:
Just a small cosmetic comment. Looks good to me otherwise.

PR #1095, opened 2015-07-02
File resolver improvements

purplefox:


cescoffier:
everything fine.

I wonder if we should document the new property (setting the cache base directory) in the manual.

purplefox:
Yeah, we should probably document that. There are a few other sys props we should also document too. Perhaps a new section in the docs?

cescoffier:
Yes, we probably need a new section and collect all the system properties we use.

PR #1097, opened 2015-07-06
Synchronize on connection not websocket to remove deadlock

purplefox:
This PR should fix the deadlock as reported here:

https://groups.google.com/forum/?fromgroups#!topic/vertx/LIVo9qrZrTg


cescoffier:
Everything fine for me.

PR #1104, opened 2015-07-10
Make sure currentResponse is set to null after response has ended to …

purplefox:
…prevent exception being fired when conn is closed


cescoffier:
Ok for me.

PR #1107, opened 2015-07-15
Do lookups on worker thread to avoid blocking event loop

purplefox:


src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
```diff
@@ -210,16 +212,37 @@ public DatagramSocket resume() {
   @Override
   @SuppressWarnings("unchecked")
   public DatagramSocket send(Buffer packet, int port, String host, Handler<AsyncResult<DatagramSocket>> handler) {
+    Objects.requireNonNull(packet, "no null packet accepted");
     Objects.requireNonNull(host, "no null host accepted");
-    ChannelFuture future = channel().writeAndFlush(new DatagramPacket(packet.getByteBuf(), new InetSocketAddress(host, port)));
-    addListener(future, handler);
+    InetSocketAddress addr = InetSocketAddress.createUnresolved(host, port);
+    if (addr.isUnresolved()) {
```
pabad: This will always be true. Creating an unresolved address will not even attempt to lookup in the local cache to see if it was previously resolved.

src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
```diff
@@ -210,16 +212,37 @@ public DatagramSocket resume() {
...
-
     return this;
   }
 
+  private void doSend(Buffer packet, InetSocketAddress addr, Handler<AsyncResult<DatagramSocket>> handler) {
+    ChannelFuture future = channel().writeAndFlush(new DatagramPacket(packet.getByteBuf(), addr));
+    addListener(future, handler);
+  }
+
+  public static void main(String[] args) {
```
pabad: Leftovers from debug?

pabad:
Looks good.
When Netty 4.1 is published we will problably need to change all this to make use of the connect/bind variants that use a host/port pair  as parameters.

src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
purplefox: Ok. I assumed (wrongly) that a "host" which is actually just an IP address will always be "resolved" .

Need some easy way to check if the "host" needs resolution or not as don't want to do an async resolve on each send if the user has specified IP.

src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
purplefox: yep ;)

testn:
Any chance that this will get in 3.1?

purplefox:
Probably not as I'd like to wait for netty 4.1.0 which won't be out until after 3.1.

As a workaround, if you're having problems with slow blocking lookups in Vert,x, you can resolve hosts asynchronously using the Vert.x DNS client to obtain the IP address and use that with the Vert.x HttpServer/HttpClient.

purplefox:
This is on hold until Netty 4.1.0 is released.

jtruelove:
I pulled down the http2 branch, does more work need to be done on this change or does this change just need testing to get it to completion?

vietj:
@jtruelove what do you mean ?

vietj:
if you want a branch with upgrade to Netty 4.1.0-CR1 you should fetch the http2 branch and create a new branch from this commit https://github.com/eclipse/vert.x/commit/79f39c63b488f379d54829b4d535b8f2fdbf0ad1 . Otherwise you will have all the HTTP2 commits.

vietj:
note : I pursued this work and made a new rebased branch on vertx core + integrated Netty's name resolver: https://github.com/eclipse/vert.x/tree/hostname-resolver . The integration is almost done, there is some minor work to do on hostname resolver options, like cache settings.

cescoffier:
I guess this one can be closed now.

leungchao:
I use vertx4.x, I have found the same problem with other async HttpClient, does the vertx HttpClient solve this problem? or still needed resolve hosts asynchronously using the Vert.x DNS client ?

vietj:
the vertx HTTP client has a non blocking resolver @leungchao

vietj:
so in short you don't need to make lookup before

leungchao:
Does Vertx HttpClient use worker threads to do lookup? Or use jdk InetAdress? 
Refer to https://vertx.io/docs/vertx-core/java/#_dns_client 
image

Is this already in version 4.x
link: https://github.com/eclipse-vertx/vert.x/pull/1107#issuecomment-208047707

leungchao:
> the vertx HTTP client has a non blocking resolver @leungchao

Does Vertx HttpClient use worker threads to do lookup? Or use jdk InetAddress?

Is this already in version 4.x?
link: https://github.com/eclipse-vertx/vert.x/pull/1107#issuecomment-208047707

PR #1110, opened 2015-07-27
Starter refactoring step1

cescoffier:
This is the first part in the Starter refactoring. It contains:

* an option / argument framework
* the reimplementation of the `run`, `bare`, and `version` command
* a new main runner called `Launcher` (to not break compatibility with `Starter`).

It fixes https://github.com/vert-x3/issues/issues/35 and is related to https://github.com/vert-x3/issues/issues/32.


cescoffier:
Step 1 ready for review - @vietj @pmlopes @purplefox

vietj:
you should:

- move the content of the io.vertx.core.cli package to the io.vertx.core.spi.cli as something in io.vertx.core is more targeted for developing applications
- the io.vertx.spi.Command should be moved in the same package
- we should have a split between io.vertx.core.spi.cli what someone uses to extend Vert.x and the implementation details that would be in io.vertx.core.spi.cli.impl. For instance CommandLineParser is an implementation detail used by the Starter to parse the Command developed with the SPI, etc...
- add missing headers to the file

after that I'll look at a more in details review.

vietj:
- the io.vertx.core.cli.commands package should probably be in io.vertx.core.impl.commands as this is a set of commands close to the Starter but an implementation detail of the Starter.

vietj:
there is a typo : VertxLifeycleHooks

cescoffier:
I've made the modification except one.

The ` io.vertx.core.cli.commands` was not moved to `io.vertx.core.impl.commands` but too `io.vertx.core.impl.cli.commands` because it has nothing to do with the `Launcher` class as the dispatching is made by `io.vertx.core.impl.cli.VertxCommandLineInterface` (extended by the `Launcher` class).

cescoffier:
Close this PR after refactoring (see https://github.com/vert-x3/issues/issues/43). I will open another (set of) PR.

PR #1115, opened 2015-08-06
Feature/range send file

pmlopes:
Implement range support for streaming files.
Once this is in we can support:
* range aware static file handling on web
* resumable downloads
* send chunks of files using netclients
* etc...


src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
```diff
@@ -439,7 +439,7 @@ private void doSendFile(String filename, Handler<AsyncResult<Void>> resultHandle
         try {
           raf = new RandomAccessFile(file, "r");
           conn.queueForWrite(response);
-          conn.sendFile(raf, fileLength);
+          conn.sendFile(raf, Math.min(offset, fileLength), Math.min(length, fileLength - offset));
```
purplefox: duplicate calculation. can be done once higher up in method.

src/main/java/io/vertx/core/json/Json.java
```diff
@@ -111,6 +116,8 @@ static Object checkAndCopy(Object val, boolean copy) {
       }
     } else if (val instanceof byte[]) {
       val = Base64.getEncoder().encodeToString((byte[])val);
+    } else if (val instanceof Instant) {
+      val = OffsetDateTime.ofInstant((Instant) val, ZoneOffset.UTC).format(ISO_OFFSET_DATE_TIME);
```
purplefox: How is this related to the range PR? Maybe have some other commits mixed up in here?

pmlopes:
i've ammended the branch...

src/test/java/io/vertx/test/core/FileResolverTestBase.java
```diff
@@ -230,6 +230,36 @@ public void testSendFileFromClasspath() {
     await();
   }
 
+  @Test
+  public void testSendOpenRangeFileFromClasspath() {
```
purplefox: I think it would be more appropriate to have these tests in HttpTest with the rest of the send file tests.

We do have some send file tests in FileResolverTest but these are just to make sure the offsetting is working ok.

src/main/java/io/vertx/core/http/HttpServerResponse.java
```diff
@@ -262,7 +262,40 @@
...
+   * @param offset offset to start serving from
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  default HttpServerResponse sendFile(String filename, long offset) {
+    return sendFile(filename, offset, Long.MAX_VALUE);
+  }
+
+  /**
+   * Ask the OS to stream a file as specified by {@code filename} directly
```
purplefox: When creating an overloaded method, instead of duplicating the javadoc, it's preferable to use {@link} to refer to the version with the full javadoc and just explain the differences.

src/main/java/io/vertx/core/http/HttpServerResponse.java
```diff
@@ -262,7 +262,40 @@
...
+   * @param filename  path to the file to serve
+   * @param offset offset to start serving from
+   * @param length length to serve to
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  HttpServerResponse sendFile(String filename, long offset, long length);
 
   /**
    * Like {@link #sendFile(String)} but providing a handler which will be notified once the file has been completely
```
purplefox: e.g. here's an example of using {@link}

purplefox:
Also, we should probably have something in the docs (package-info.java in the http package) about this.

PR #1118, opened 2015-08-18
Cluster public props

purplefox:
@cescoffier @pmlopes please review.


cescoffier:
I wonder if we should we add a parameter to `Starter` to configure the public host / port ? We can use the system property mapping for this, so it's not required. 

Except that, +1.

pmlopes:
:+1:

PR #1119, opened 2015-08-19
Bug fix: HttpClientImpl::requestAbs does not respect default port option

simonpai:
At this section, 
https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java#L252
```Java
  public HttpClientRequest requestAbs(HttpMethod method, String absoluteURI) {
    URL url = parseUrl(absoluteURI);
    return doRequest(method, url.getHost(), url.getPort(), url.getFile(), null);
  }
```
If `absoluteURI` does not include a port, `url.getPort()` gives -1, which is not handled before it's used to set up a connection and throws port out of range error.

The patch itself does not include a test case as I am not sure how you would like to maintain the test cases, but a sample can be found here:
https://gist.github.com/simonpai/a2d7aa0832854df4634a

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
@@ -250,7 +250,8 @@ public class HttpClientImpl implements HttpClient, MetricsProvider {
   @Override
   public HttpClientRequest requestAbs(HttpMethod method, String absoluteURI) {
     URL url = parseUrl(absoluteURI);
-    return doRequest(method, url.getHost(), url.getPort(), url.getFile(), null);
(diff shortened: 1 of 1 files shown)


purplefox:
Thanks.

Can you add a test in HttpTest for this?

purplefox:
Ok, don't worry I'll add the test :)

purplefox:
Thanks for the report. I fixed this directly.

simonpai:
Thanks, actually I am not very certain it should fallback to the default port in `HttpClient` options or the default port of the protocol. (The latter probably makes more sense.) I think it's the best to leave the official team to decide. :)

PR #1120, opened 2015-08-19
Implement Temporal Support for JSON

pmlopes:

This completes the implementation of I-JSON

### DIFF
--- a/src/main/asciidoc/java/override/json.adoc
+++ b/src/main/asciidoc/java/override/json.adoc
@@ -13,6 +13,14 @@ A JSON object is basically just a map which has string keys and values can be of
 
 JSON objects also support null values.
 
+The simplicity behind JSON is one of its biggest strengths however in many occasions developers will want some extra
+data types than the ones provided by the spec, namely binary and temporal data. For this reason there is
+`link:http://rfc-editor.org/rfc/rfc7493.txt[RFC-7493]`. The RFC defined I-JSON as a interop data type which extends
+the base spec with knowledge about temporal and binary data.
+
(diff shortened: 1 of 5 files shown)


purplefox:
Would it be a good idea to allow instants to be added/retrieved as longs too?

In many cases, Java programs handle date/times as longs and forcing a conversion to Instant just so it can be converted to String seems a bit long-winded.

purplefox:
Maybe that's just a a suggestion for the future. What you have here looks good.

pmlopes:
long timestamps are not part of the RFC, the RFC defines temporal -> String only, the String -> Instant is added for convenience.

The issue with timestamps is that languages like javascript have no long int type which may cause longs to be truncated, although the json spec does allow any precision.

Acording to ES6 (EcmaScript2015):

```
Number.MAX_SAFE_INTEGER // 9007199254740991
Math.pow(2, 53) - 1     // 9007199254740991
```

It will for sure handle all the current dates but if someone is making a forecast app that uses only far future dates we are still missing the last 10 bits of the long range...

For users only caring about longs then the current API should be enough, for example:

I can put a timestamp as long:
```
obj.put("myTimestamp", System.currentTimeMillis());
```
and later read it to an instant like:
```
Instant instant = Instant.ofEpochMilli(obj.getLong("myTimestamp"));
```

purplefox:
I didn't mean storing the long as a value in the json object but just an ability to set it with a long.

long now = System.currentTimeMillis();
obj.putTemporal("somekey", now);

such that the long is stored as a temporal internally.

with your changes I have to convert it to an Instant first which is an extra step.

PR #1122, opened 2015-08-21
Add license and javadoc to Converters

cescoffier:
Add license header to the converter template
Compute the Converter class javadoc

### DIFF
--- a/src/main/generated/io/vertx/core/http/HttpClientOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/http/HttpClientOptionsConverter.java
@@ -1,8 +1,32 @@
+/*
+ * Copyright 2014 Red Hat, Inc.
+ *
+ * Red Hat licenses this file to you under the Apache License, version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License.  You may obtain a copy of the License at:
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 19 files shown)


src/main/generated/io/vertx/core/http/HttpClientOptionsConverter.java
```diff
@@ -1,8 +1,32 @@
...
 
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.json.JsonArray;
 
+/**
+ * Converter for {@link io.vertx.core.http.HttpClientOptions}.
+ *
+ * NOTE: This class has been automatically generated from the {@link io.vertx.core.http.HttpClientOptions} original class using Vert.x codegen.
+ *
+ * @author <a href="http://tfox.org">Tim Fox</a>
```
purplefox: The author shouldn't be me.

src/main/generated/io/vertx/core/metrics/MetricsOptionsConverter.java
```diff
@@ -1,8 +1,32 @@
...
 
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.json.JsonArray;
 
+/**
+ * Converter for {@link io.vertx.core.metrics.MetricsOptions}.
+ *
+ * NOTE: This class has been automatically generated from the {@link io.vertx.core.metrics.MetricsOptions} original class using Vert.x codegen.
+ *
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
```
purplefox: Or julien

cescoffier: It gets the author from the original class.
We can remove the `@author` tag.

cescoffier:
I've removed the tag copy form the original class, so @author won't be inherited anymore.

PR #1123, opened 2015-08-27
Fix race where failure can occur if verticle is undeployed concurrent…

purplefox:
…ly from self and parent


cescoffier:
Ok for me.

pmlopes:
:+1:

PR #1126, opened 2015-09-04
Keepalive pipelining

purplefox:
1. Fixes bug in pipelining which could result in requests waiting indefinitely in waiters queue
2. Changes behaviour of keep alive so keep alive connections are never automatically closed.


pmlopes:
:+1: fine for me.

purplefox:
I'll add the documentation changes in a further commit...

Issue #48, opened 2012-03-08
Introduce pingers on cluster connections


purplefox:
Pingers so cluster connections are properly closed when nodes disappear from network

PR #1128, opened 2015-09-04
#48 escape special chars on status message

pmlopes:

The previous commit was imcomplete


pmlopes:
This part does the reverse, "revives" the \r and \n

src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -69,7 +69,7 @@
   HttpClientResponseImpl(Vertx vertx, HttpClientRequestImpl request, ClientConnection conn, HttpResponse response) {
     this.vertx = vertx;
     this.statusCode = response.getStatus().code();
-    this.statusMessage = response.getStatus().reasonPhrase();
+    this.statusMessage = response.getStatus().reasonPhrase().replaceAll("\\\\r", "\r").replaceAll("\\\\n", "\n");
```
purplefox: I'd caution against this, as it will cause replaceAll to executed twice for each client response which could kill performance.

pmlopes:
See new pull request this is not supposed to be merged and the previous issue is to be reverted. see https://github.com/eclipse/vert.x/pull/1129

PR #1131, opened 2015-09-08
added cause of SSLHandshakeException

zdanek:
I've added cause to SSLHandshakeException. Otherwise the real cause is lost.


purplefox:
Thanks!

zdanek:
Thank you for accepting.

PR #1132, opened 2015-09-09
#51: check if file exists before send

pmlopes:
Solves: https://github.com/vert-x3/issues/issues/51


src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -2295,6 +2295,29 @@ private void sendFile(String fileName, String contentExpected, boolean handler)
   }
 
   @Test
+  public void testSendNonExistingFile() throws Exception {
+    server.requestHandler(req -> {
+      req.response().sendFile("/not/existing/path", event -> {
```
vietj: check that this is the correct Context in the callback

purplefox:
I'm not sure why this fix was necessary. In the previous implementation the error is reported fine in the sendFile handler...

pmlopes:
using the included test, without the fix:

```java
req.response().sendFile("/not/existing/path", event -> {
         if (event.failed()) {
           req.response().end("failed");
         }
```

The event state will be `success` and the final response is 200 with content length of 0. With the fix, the event state will be `failed`.

purplefox:
I don't get those results.

If I comment out the code you added in HttpServerResponseImpl and set a breakpoint in the test you added to inspect the result (event) I can see it is failed and has a cause of FileNotFoundException.

PR #1134, opened 2015-09-10
Add API to copy a Buffer into a byte array more efficiently

testn:
Add API to copy a Buffer into a byte array more efficiently without array allocation.


purplefox:
Thanks!

PR #1135, opened 2015-09-14
Provide the CLI system

cescoffier:
Implement the new CLI system

This PR contains:

* the CLI model
* the parser
* annotations usable from java
* tests


src/main/java/io/vertx/core/cli/UsageMessageFormatter.java
```diff
@@ -0,0 +1,666 @@
...
+    }
+
+    return s.substring(0, pos);
+  }
+
+  /**
+   * This class implements the <code>Comparator</code> interface
+   * for comparing Options.
+   */
+  private static class OptionComparator implements Comparator<OptionModel> {
```
vietj: why not use a lambda ?

src/main/java/io/vertx/core/cli/UsageMessageFormatter.java
```diff
@@ -0,0 +1,666 @@
...
+    return new String(padding);
+  }
+
+  /**
+   * Remove the trailing whitespace from the specified String.
+   *
+   * @param s The String to remove the trailing padding from.
+   * @return The String of without the trailing padding
+   */
+  protected String rtrim(String s) {
```
vietj: could be made static

src/main/java/io/vertx/core/cli/CommandLine.java
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  /**
+   * Sets the various command line arguments. Method called by {@link CommandLineParser}.
+   *
+   * @param args the arguments
+   * @return the current {@link CommandLine}
+   */
+  @Fluent
+  CommandLine setAllArguments(List<String> args);
```
vietj: I think this could be removed from the interface as the user does not care about it and the CommandLineParser could make a cast to the impl

src/main/java/io/vertx/core/cli/CommandLineParser.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+
+/**
+ * Interface implemented by classes responsible for parsing the user command line argument and creating the
+ * {@link CommandLine}.
+ *
+ * @author Clement Escoffier <clement@apache.org>
+ */
+@VertxGen
+public interface CommandLineParser {
```
vietj: not sure we need to expose this as API. Creating a command line from a CLI would be just as simple as doing `CommandLine cli.parse(List<String> args)`

src/main/java/io/vertx/core/cli/OptionModel.java
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  /**
+   * Checks whether or not the option is valid. This implementation check that it has a short name or a long name.
+   * This method is intended to be extended by sub-class. {@link CommandLineParser} should check that the set of
+   * option of a {@link CLI} is valid before starting the parsing.
+   *
+   * @return whether or not the configuration of this option is valid.
+   */
+  public boolean isValid() {
```
vietj: weird method return says "whether or not the configuration of this option is valid." however it always returns true or throw an exception. one of the two statements if false.

src/main/java/io/vertx/core/cli/ArgumentModel.java
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * Defines a command line argument. Unlike options, argument don't have names and are identified using an index. The
+ * first index is 0 (because we are in the computer world).
+ *
+ * @author Clement Escoffier <clement@apache.org>
+ * @see OptionModel
+ */
+@DataObject(generateConverter = true)
+public class ArgumentModel {
```
vietj: Why not name this "Argument" instead of "ArgumentModel" ? besides TypedArgument that extends it is not named "TypeArgumentModel".

src/main/java/io/vertx/core/cli/OptionModel.java
```diff
@@ -0,0 +1,371 @@
...
+/**
+ * Models command line options. Options are values passed to a command line interface using -x or --x. Supported
+ * syntaxes depend on the parser.
+ * <p/>
+ * Short name is generally used with a single dash, while long name requires a double-dash.
+ *
+ * @author Clement Escoffier <clement@apache.org>
+ */
+@DataObject(generateConverter = true)
+public class OptionModel {
```
vietj: Same comment than for ArgumentModel/Argument.

src/main/java/io/vertx/core/cli/OptionModel.java
```diff
@@ -0,0 +1,371 @@
...
+   *
+   * @param multiValued whether or not this option is multi-valued.
+   * @return the current {@link OptionModel} instance
+   */
+  public OptionModel setMultiValued(boolean multiValued) {
+    this.multiValued = multiValued;
+    if (this.multiValued) {
+      // If we accept more than one, we also accept 1.
+      this.singleValued = true;
+    }
```
vietj: what if multivalued == false ?

cescoffier:
reviews integrated.

vietj:
ok for me, looks good.

PR #1137, opened 2015-09-15
Expose executeBlocking method on Context to make it easier to execute…

testn:
… job without reference to Vertx instance

### DIFF
--- a/src/main/java/io/vertx/core/Context.java
+++ b/src/main/java/io/vertx/core/Context.java
@@ -100,6 +100,35 @@ public interface Context {
   void runOnContext(Handler<Void> action);
 
   /**
+   * Safely execute some blocking code.
+   * <p>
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/Context.java
```diff
@@ -100,6 +101,32 @@ static boolean isOnVertxThread() {
   void runOnContext(Handler<Void> action);
 
   /**
+   * Like {@link #executeBlocking(Handler, boolean, Handler)} called with ordered = true.
+   */
+  <T> void executeBlocking(Action<T> action, Handler<AsyncResult<T>> resultHandler);
```
purplefox: Action is not a public class - this method should not be part of the public API.

purplefox:
Thanks.

Also we'd need at least one test for the new API.

testn:
Tests added

PR #1139, opened 2015-09-15
Command SPI

cescoffier:
PR providing the new "Launcher" class, the "Command" SPI and the implementation of the "run", "bare" and "version" commands


cescoffier:
Need to fix Starter formating. Closing.

PR #1147, opened 2015-09-16
Implemented parameterized logging in Vertx core api

manish-panwar:


manish-panwar:
Tim, do you have any idea what wrong I might be doing here ? 

I have same email (manishk@vmware.com) in github and Eclipse, and same user id (manish-panwar) in github and Eclipse. Also I had signed off my commit but seeing this error. I have tried various emails but not sure why this is happening.

PR #1149, opened 2015-09-16
Implemented parameterized logging in Vertx core api

manish-panwar:


manish-panwar:
Tim, do you have any idea what wrong I might be doing here ? My commit isn't approved by Eclipse CLA.

I have same email (manishk@vmware.com) in github and Eclipse CLA account, and same user id (manish-panwar) in github and Eclipse CLA account. Also I had signed off my commit but seeing this error. I have tried using emails but not sure why this is happening.

purplefox:
This PR is badly mangled and I couldn't accept it in this state even if the commits were signed and CLA signed. Please take a look how many commits and reversions there are.

I suggest you take a diff of your changes and re-apply them against a clean fork of the project.

PR #1151, opened 2015-09-18
Headers end handler no longer takes a Future as async completion is v…

purplefox:
…ery hard to implement correctly in the response


pmlopes:
shouldn't we keep the old method and mark it @deprecated to keep API compatibility with 3.x?

purplefox:
The old method didn't work properly so this is really a bug fix.

I think it's ok to make breaking changes like this for bug fixes or security reasons.

PR #1152, opened 2015-09-18
Deprecate ClusterManager.nodeListener() and replace it with add/removeNodeListener()

vascokk:
Hi all,

This pull request is a proposal to deprecate ClusterManager.nodeListener() and replace it with add/removeNodeListener().
This will allow the applicationa to listen for cluster membership events. An example use case is - fault-tolerance: verticles with persistent state, which previously resided on a crashed node, need to be recovered on the available cluster nodes, while the application logic handles the deployment of verticles.


vascokk:
I have signed the Eclipse CLA: 
image
Not sure why it is not accepted by the IP validator....

purplefox:
I'm not sure I understand this PR. Why are the changes preferable to the previous API?

purplefox:
We are actually going to deprecate ClusterManager in Vert.x 3.2 anyway and provide a more flexible clustering/event bus SPI, which would make changes here redundant anyway.

But thanks for looking at this :)

PR #1153, opened 2015-09-18
Add an ability to control the read buffer size for AsyncFile's async …

testn:
…read


cescoffier:
Looks fine to me.

PR #1154, opened 2015-09-18
Add option to handle http client redirect

testn:

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
@@ -29,6 +29,7 @@ import io.netty.handler.codec.http.HttpVersion;
 import io.netty.handler.codec.http.LastHttpContent;
 import io.vertx.core.Handler;
 import io.vertx.core.MultiMap;
+import io.vertx.core.VertxException;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpClientRequest;
 import io.vertx.core.http.HttpClientResponse;
@@ -38,6 +39,8 @@ import io.vertx.core.logging.LoggerFactory;
 import io.vertx.core.net.NetSocket;
 import io.vertx.core.spi.metrics.HttpClientMetrics;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.TimeoutException;
@@ -57,12 +60,13 @@ public class HttpClientRequestImpl implements HttpClientRequest {
 
   private static final Logger log = LoggerFactory.getLogger(HttpClientRequestImpl.class);
 
-  private final String host;
-  private final int port;
+  private String host;
+  private int port;
+  private int redirectCount;
   private final HttpClientImpl client;
-  private final HttpRequest request;
+  private HttpRequest request;
   private final VertxInternal vertx;
-  private final io.vertx.core.http.HttpMethod method;
+  private io.vertx.core.http.HttpMethod method;
   private Handler<HttpClientResponse> respHandler;
   private Handler<Void> endHandler;
   private boolean chunked;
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -57,12 +60,13 @@
...
   private static final Logger log = LoggerFactory.getLogger(HttpClientRequestImpl.class);
 
-  private final String host;
-  private final int port;
+  private String host;
+  private int port;
+  private int redirectCount;
   private final HttpClientImpl client;
-  private final HttpRequest request;
+  private HttpRequest request;
```
cescoffier: if request is not final anymore, access need to be guarded by the monitor lock. It's not the case in uri().

src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -57,12 +60,13 @@
...
   private final VertxInternal vertx;
-  private final io.vertx.core.http.HttpMethod method;
+  private io.vertx.core.http.HttpMethod method;
```
cescoffier: missing synchronization (same as for request)

src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -351,6 +378,42 @@ synchronized void handleResponse(HttpClientResponseImpl resp) {
...
+    }
+
+    resp.endHandler((nothing) -> {
+      String newUri = absoluteUri.toString();
+      reset(newUri);
+      end();
+    });
+  }
+
+  private void reset(String newUri) {
```
cescoffier: unprotected access to various non-final fields.

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -4305,6 +4305,138 @@ public void testRequestEnded() {
...
+      });
+      post.end();
+    });
+    await();
+  }
+
+  @Test
   public void testRequestEndedNoEndHandler() {
     server = vertx.createHttpServer(new HttpServerOptions().setPort(HttpTestBase.DEFAULT_HTTP_PORT));
     server.requestHandler(req -> {
```
purplefox: Would need to add tests to the HttpClientOptions tests too.

src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
testn: Need to be able to change that for HTTP 303 redirect where the method can be changed from POST to GET
https://en.wikipedia.org/wiki/HTTP_303

src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
testn: To handle the redirect seamlessly without creating a new HttpRequest, we need to modify some non-final fields when the redirection happens.

cescoffier: Ok, but then all access to the field must be guarded. It's not the case. So there are race conditions.

src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
cescoffier: Ok, but then all access to the field must be guarded. It's not the case. So there are race conditions.

purplefox:
A different (and safer) way of implementing this, which does not involve changing the logic of HttpClientRequestImpl at all is simply to make a normal request and do the redirect logic in the standard response handler using the public API, this could be hidden in a helper method in HttpClientImpl.

I think I would prefer that as it wouldn't require fiddling with the internals of HttpClientImpl and reasoning about thread safety.

cescoffier:
I also agree, it would be safer.

luengnat:
It doesn't change the internal state of httpclientimpl. It only deal with the transition from one httpclientrequestimpl to the new one after the redirect. If all of these operations happen in the same event loop thread, it should be thread safe right?

purplefox:
Closing this, but I am still open to the idea of this functionality, just not implemented this way.

PR #1155, opened 2015-09-21
Support for client auth

manish-panwar:
Right now Vertx API(3.0.0) expose a boolean variable in HttpServerOptions to set the Client Auth, which in turn allows the certificate based auth. This doesn't allow to set the value to one of these NONE, REQUEST, REQUIRED, which would allow HTTP Server to serve Certificate Auth as well as Basic Auth.

I have done the necessary changes to set the enum value in HttpServerOptions. I have tested it and all works fine.

Please see more details here - https://groups.google.com/forum/#!topic/vertx/YtojWw0a0bU

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
   </parent>
 
   <artifactId>vertx-core</artifactId>
-  <version>3.1.0-Client-Auth-Support-SNAPSHOT</version>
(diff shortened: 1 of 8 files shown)


manish-panwar:
Closing this pull request and going to create a new one. I found out what was the issue with Eclipse CLA and I am going to fix it.

PR #1156, opened 2015-09-21
Added support for parametrized logging(similar to how SLF4J API).

manish-panwar:
Signed-off-by: Manish Kumar <manishk@vmware.com>

Added the ability to pass the additional argument to be printed in the logs.

### DIFF
--- a/src/main/java/io/vertx/core/logging/Log4jLogDelegate.java
+++ b/src/main/java/io/vertx/core/logging/Log4jLogDelegate.java
@@ -16,7 +16,6 @@
 
 package io.vertx.core.logging;
 
-import io.vertx.core.logging.Logger;
 import io.vertx.core.spi.logging.LogDelegate;
 import org.apache.log4j.Level;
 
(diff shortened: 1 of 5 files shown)


src/main/java/io/vertx/core/logging/Log4jLogDelegate.java
```diff
@@ -101,4 +150,8 @@ private void log(Level level, Object message) {
   private void log(Level level, Object message, Throwable t) {
     logger.log(FQCN, level, message, t);
   }
+
+  private void throwUnsupportedOperationException() {
```
cescoffier: I'm not sure that this is acceptable, as it would avoid using the new methods for safety (As you don't know on which log system you are running). 

I would be in favor of computing the formatted message. This should check whether or not the log level is enabled before.

purplefox: We actually started off down the route of computing our own formatted messages but decided to delegate to the logging library as most probably the logging library implementation is much better optimised.

I think it's probably ok for the new methods on the Log4JLogDelegate to throw UnsupportedOperationException as this won't break existing code as the methods are new.

PR #1157, opened 2015-09-21
Added support to set the "Client Auth" value as enum. It allows us to…

manish-panwar:
Signed-off-by: Manish Kumar <manishk@vmware.com>

Right now Vertx API(3.0.0) expose a boolean variable in HttpServerOptions to set the Client Auth, which in turn allows the certificate based auth. This doesn't allow to set the value to one of these NONE, REQUEST, REQUIRED, which would allow HTTP Server to make Client Auth as Optional.

I have done the necessary changes to set the enum value in HttpServerOptions. I have tested it and all works fine.

Please see more details here - https://groups.google.com/forum/#!topic/vertx/YtojWw0a0bU

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServerOptions.java
+++ b/src/main/java/io/vertx/core/http/HttpServerOptions.java
@@ -20,6 +20,7 @@ import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.net.*;
+import io.vertx.core.net.impl.SSLHelper;
 
 /**
  * Represents options used by an {@link io.vertx.core.http.HttpServer} instance
@@ -272,10 +273,10 @@ public class HttpServerOptions extends NetServerOptions {
   public String getWebsocketSubProtocols() {
     return websocketSubProtocols;
   }
-  
+
   @Override
-  public HttpServerOptions setClientAuthRequired(boolean clientAuthRequired) {
-    super.setClientAuthRequired(clientAuthRequired);
+  public HttpServerOptions setClientAuth(SSLHelper.ClientAuth clientAuth) {
+    super.setClientAuth(clientAuth);
     return this;
   }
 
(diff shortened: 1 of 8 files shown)


cescoffier:
These changes are breaking the API compatibility. It should keep the previous method.

src/main/java/io/vertx/core/http/HttpServerOptions.java
```diff
@@ -20,6 +20,7 @@
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.net.*;
+import io.vertx.core.net.impl.SSLHelper;
```
purplefox: SSLHelper is not part of the public API (notice the "impl" package) -we can't expose private implementation classes in the API.

purplefox:
I think what this PR needs:

1. Create a new enum that's part of the public API (don't use the one in SSLHelper)
2. Add a _new_ method to set client auth on the options. Don't remove the old method but deprecate it.

manish-panwar:
Sure, I will make the necessary changes.

manish-panwar:
Hey guys,

As suggested, I have created enum in public API, and also retained the old API but marked them deprecated. Thanks a lot for your time to look into the pull request.

vietj:
@manish-panwar  please annotate your enum with @VertxGen and add Javadoc to it.

src/main/java/io/vertx/core/net/NetServerOptions.java
```diff
@@ -274,8 +274,26 @@ public boolean isClientAuthRequired() {
    * @param clientAuthRequired  true if client auth is required
    * @return a reference to this, so the API can be used fluently
    */
+  @Deprecated
   public NetServerOptions setClientAuthRequired(boolean clientAuthRequired) {
-    this.clientAuthRequired = clientAuthRequired;
+    this.clientAuth = clientAuthRequired ? ClientAuth.REQUIRED : ClientAuth.NONE;
```
purplefox: If you've created a public ClientAuth enum there's no need to keep the old internal one.

manish-panwar: I have got rid of internal one. Just so you know, I will still need to keep this here to keep the backward compatibility. Also, I have got rid of boolean flag "clientAuthRequired"

manish-panwar:
Changes are done as suggested by you guys.

purplefox:
This PR still had a few issues, but I resolved them in this commit:

https://github.com/eclipse/vert.x/commit/3189e48aeb6e2b44e961cf58a2d8e01007c758c8

And merged. Thanks.

PR #1159, opened 2015-09-22
Added event address when reporting failure on message reply timeout

zdanek:
When event reply timeouts there is no clue which one failed.

Now we can have very specific messages on which address event reply failed:
```
Message reply handler for address [users.projects.user] timed out as no reply was received - it will be removed 
/users/bzd/projects failed 
(TIMEOUT,-1) Timed out waiting for reply for address [users.projects.user]
	at io.vertx.core.eventbus.impl.EventBusImpl.lambda$sendOrPub$137(EventBusImpl.java:503)
```

BUT I need some English person (@purplefox maybe?) to check for messages spelling as I was not sure if it's correct :smile_cat:


purplefox:
Hi Bartek, I already fixed this in master the other day, but thanks anyway.

PR #1162, opened 2015-09-26
Added support for setting custom TLS protocol versions

maxhope:


src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -226,9 +231,13 @@ private SslHandler createHandler(SSLEngine engine, boolean client) {
       engine.setEnabledCipherSuites(toUse);
     }
     engine.setUseClientMode(client);
-    Set<String> enabledProtocols = new HashSet<>(Arrays.asList(ENABLED_PROTOCOLS));
-    enabledProtocols.retainAll(Arrays.asList(engine.getEnabledProtocols()));
-    engine.setEnabledProtocols(enabledProtocols.toArray(new String[0]));
+    if (enabledProtocols != null && !enabledProtocols.isEmpty()) {
+      engine.setEnabledProtocols(enabledProtocols.toArray(new String[0]));
```
cescoffier: the passed array should be the right size avoiding another array creation.

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -226,9 +231,13 @@ private SslHandler createHandler(SSLEngine engine, boolean client) {
...
+    } else {
+      Set<String> protocols = new HashSet<>(Arrays.asList(ENABLED_PROTOCOLS));
+      protocols.retainAll(Arrays.asList(engine.getEnabledProtocols()));
+      engine.setEnabledProtocols(protocols.toArray(new String[0]));
```
cescoffier: the passed array should be the right size avoiding another array creation.

PR #1164, opened 2015-09-28
Don't set server to no handlers unless all handlers are removed

purplefox:
Fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=477822


cescoffier:
Ok for me.

pmlopes:
:+1:

PR #1167, opened 2015-09-29
Launcher process redeploy

cescoffier:
This PR adds the _redeploy_ feature to the `run` command.

It's a process-based redeploy: the application is started in background and restarted on file change. The set of watched file is configured in the command line (`--redeploy=...`).

### DIFF
--- a/src/main/java/io/vertx/core/impl/launcher/commands/Watcher.java
+++ b/src/main/java/io/vertx/core/impl/launcher/commands/Watcher.java
@@ -1,0 +1,303 @@
+/*
+ *  Copyright (c) 2011-2013 The original author or authors
+ *  ------------------------------------------------------
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v1.0
+ *  and Apache License v2.0 which accompanies this distribution.
+ *
+ *       The Eclipse Public License is available at
+ *       http://www.eclipse.org/legal/epl-v10.html
+ *
+ *       The Apache License v2.0 is available at
+ *       http://www.opensource.org/licenses/apache2.0.php
+ *
+ *  You may elect to redistribute this code under either of these licenses.
+ */
+
+package io.vertx.core.impl.launcher.commands;
+
+import com.sun.nio.file.SensitivityWatchEventModifier;
+import io.vertx.core.Handler;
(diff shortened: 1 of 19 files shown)


src/main/java/io/vertx/core/impl/launcher/commands/Watcher.java
```diff
@@ -0,0 +1,306 @@
...
+
+  private final static Logger LOGGER = LoggerFactory.getLogger(Watcher.class);
+
+  private final List<String> includes;
+  private final File root;
+  private final Handler<Handler<Void>> deploy;
+  private final Handler<Handler<Void>> undeploy;
+  private final String cmd;
+
+  private WatchService service;
```
purplefox: Beware of WatchService - we used it in the old Redeployer originally but it didn't work properly on all OSes so I removed it and wrote our own watch code.

Maybe they've fixed those bugs - but I'm not sure...

src/test/java/io/vertx/core/impl/launcher/commands/WatcherTest.java
```diff
@@ -0,0 +1,176 @@
...
+      if (files != null) {
+        for (File f : files) {
+          ret = ret && deleteRecursive(f);
+        }
+      }
+    }
+    return ret && path.delete();
+  }
+
+}
```
purplefox: Also test addition and deletion of directory? In particular:

Create new directory
Create new file in directory.
Amend file in directory.

purplefox:
Looks good. Just a few comments.

I'm a bit concerned about Watcher though - this didn't work well for us in the past...

src/main/java/io/vertx/core/impl/launcher/commands/Watcher.java
cescoffier: From my test on MacOSX, Linux and Windows it works. The polling period is different and the events are not ordered the same way, but it is ok in our case.

cescoffier:
If we get issue because of the Watcher, we will replace it. I know that they had couple of bugs in their native integration, but they fixed them. 

However, I'm also concerns by non standard file system, or on non Oracle / OpenJDK JVM. We may want to have a home made file alteration monitor alternative in the future, with a toggle.

cescoffier:
I've extended the tests.

PR #1170, opened 2015-10-02
Replace asciidoc attributes by docgen variables

cescoffier:
It avoids the issue when the asciidoc files are rendered to html in a different project.


vietj:
sounds good, gotta merge

PR #1171, opened 2015-10-02
Handle exception on callbacks

stampy88:
Ensure exceptions from the drainHandler in HttpClientRequestImpl and dataHandler and endHandler in HttpClientResponseImpl bubble up through the exceptionHandler. Currently if an exception is thrown during that processing, it is not propagated back to the exceptionHandler.


stampy88:
Gonna close this and re-submit because of the IP validation check. Whatever I do doesn't seem to fix it.

cescoffier:
The issue should have the tag 'invalid', can you add it @purplefox ?

PR #1173, opened 2015-10-04
Some improvements to net param defaulting and use different eventloop…

purplefox:
… for acceptor for faster connection setup under load


cescoffier:
Ok for me.

vietj:
looks good

vietj:
should we expose also this acceptor event loop in Vert.x as we do expose the eventLoopGroup under nettyEventLoopGroup ?

purplefox:
This is more of an internal thing, we don't really want to expose it.

vietj:
ok

PR #1174, opened 2015-10-04
Extend the FileResolver to support OSGi bundles

cescoffier:
Add the bundle:// url scheme.

### DIFF
--- a/src/main/java/io/vertx/core/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/impl/FileResolver.java
@@ -115,6 +115,8 @@ public class FileResolver {
             return unpackFromFileURL(url, fileName, cl);
           case "jar":
             return unpackFromJarURL(url, fileName);
+          case "bundle":
+            return unpackFromBundleURL(url);
           default:
             throw new IllegalStateException("Invalid url protocol: " + prot);
         }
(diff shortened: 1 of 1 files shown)


johnsteele:
what about bundleresource?

cescoffier:
By extensibility you mean relying on url handlers ?

cescoffier:
I've created https://github.com/vert-x3/issues/issues/69 to add the protocols used by equinox.

johnsteele:
Great. Thanks.

Many components (FileSystemImpl, HttpServerResponseImpl, NetSocketImpl, etc) resolve files using VertxInternal#resolveFile(fileName), but how VertxInternal resolves the file is not configurable. 

The VertxInternal constructor hard codes:

this.fileResolver = new FileResolver(this);

What if a use supports a different protocol other than what's supported within FileResolver? 
Instead of providing a way to add, extend, configure, or replace the FileResolver,
I get this nasty IllegalStateException message  saying: Invalid url protocal.

It's not invalid. You just didn't implement it. And because one can't extend this protocol
resolution process, they are to do what in the mean time?

Create a ticket, and then wait for the next release?

cescoffier:
Well, 

I agree it would be better to provide a way to extend the file resolver. 

Actually, I've issues to get bundleentry and bundleresource working on equinox, because equinox tries to be smart about the TCCL and ignore the classloader we pass to it (ContextFinder). So, I think on equinox there is no other choice than implementing a different approach, it may mean externalizing the file resolver.

cescoffier:
BTW, as vert.x is intended to work outside of OSGi it cannot relies on the URLHandler services provided by OSGi.

johnsteele:
URLConverter you mean?

cescoffier:
https://osgi.org/javadoc/r4v43/core/org/osgi/service/url/URLStreamHandlerService.html

johnsteele:
Yes, this would be nice.

johnsteele:
I don't believe org.eclipse.osgi.service.urlconversion.URLConverter is apart of the spec. I think it's a service provided by Eclipse that one can use for bundleresource resolution. Of course I'm not suggesting to use it. I only mentioned it because you mentioned having issues with Equinox, so perhaps this could provide some insight. Although it only handles bundleresource and bundleentry.

PR #1176, opened 2015-10-09
Various cli improvements

cescoffier:
Some improvement around the CLI system:

* provide multiple argument
* provide 'help' flag to options
* provide validity of command line support
* provide options with choices

### DIFF
--- a/src/main/asciidoc/java/cli-for-java.adoc
+++ b/src/main/asciidoc/java/cli-for-java.adoc
@@ -63,6 +63,8 @@ CLI cli = CLI.create("some-name")
 
 For booleans, the boolean values are evaluated to `true`: `on`, `yes`, `1`, `true`.
 
+If one of your option as an `enum` as type, it computes the set of choices automatically.
+
 === Using annotations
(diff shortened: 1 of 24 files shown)


vietj:
looks good to me, will check with vertx-shell and merge it.

vietj:
I don't see how to declare multi valued in `@Argument` annotation, is it implicitly done via `List<String>` detection?

vietj:
Could we have `@Argument()` have `-1` as default value for `index` ? this way we don't have to declare it and the auto numeration would work if it can read the members in right order

vietj:
`List<String>` works but it requires at least one element, I think it should tolerate no values as an empty list is valid for a multivalue.

cescoffier:
When the argument is not set, the "setter" method should not be called. Do you see it called ?

vietj:
ah it's because it is "required"

vietj:
yes working fine, I was expecting that required = true + zero arg => empty list, which does make sense to me. but required = false works fine

PR #1178, opened 2015-10-16
BZ480001: HandlerRegistration cancels timerID 0 in doUnregister

alexlehm:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=480001

### DIFF
--- a/src/test/java/io/vertx/test/core/LocalEventBusTest.java
+++ b/src/test/java/io/vertx/test/core/LocalEventBusTest.java
@@ -1275,5 +1275,29 @@ public class LocalEventBusTest extends EventBusTestBase {
     closeVertx();
     await();
   }
+
+  @Test
+  public void testHandlerRegistrationDoesNotCancelTimer0() throws Exception {
+    HandlerRegistration<String> handlerRegistration = new HandlerRegistration<String>(vertx, null, null, null, false,
+        false, null, -1);
+
+    AtomicBoolean timerFinished = new AtomicBoolean(false);
+
+    long tid1 = vertx.setTimer(1000, id -> {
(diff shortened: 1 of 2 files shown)


purplefox:
Alex, could you explain the reasoning behind this change?

alexlehm:
When doing netclient.close(), the method HandlerRegistration.doUnregister is called at some point.
In the method there is the following code:

    if (timeoutID != -1) {
      vertx.cancelTimer(timeoutID);
    }

which will cancel the timer timeoutID when one is set (if its set != -1). The initialisation code of the class has long timeoutID without assignment so that the id is 0 when no timer was set.
This means that if no timer is set, it will cancel timerid 0, which may be used somewhere else (so this time is cancelled unexpectedly).

purplefox:
Thanks

src/test/java/io/vertx/test/core/LocalEventBusTest.java
```diff
@@ -1275,5 +1275,29 @@ public void testCloseCallsEndHandlerWithRegistrationContext() throws Exception {
     closeVertx();
     await();
   }
+
+  @Test
+  public void testHandlerRegistrationDoesNotCancelTimer0() throws Exception {
+    HandlerRegistration<String> handlerRegistration = new HandlerRegistration<String>(vertx, null, null, null, false,
```
purplefox: HandlerRegistration is an internal class - it would be better to trigger the functionality using the public API - i.e. create a message consumer and unregister it.

alexlehm: ok, changed it to eb.consumer(xxx).unregister()

PR #1182, opened 2015-10-20
Create a new common Socket interface 

chirino:
implemented by both the NetSocket and WebSocket classes.


purplefox:
Going to close this as it's a breaking API change.

chirino:
Is it actually a breaking API change?  This introduces a new interface, but existing interfaces don't get any new methods for example.

PR #1184, opened 2015-10-21
Race reproduced with FakeClusterManager

agura:
Reproducer for issue 479943
https://bugs.eclipse.org/bugs/show_bug.cgi?id=479943


purplefox:
I'm a bit confused - is this a reproducer or does it also contain a fix?

It appears to include changes to HAManager behaviour, I'd expect a reproducer to only include a test and no changes to Vert.x itself, otherwise you're reproducing the issue against something that's not the released Vert.x version....

agura:
This PR contains reproducer only. This race is really hard to reproduce. I use latches in order to slow down and align threads accordingly to described behaviour in ticket. All changes in HAManager contain comments for clarity.

PR #1187, opened 2015-10-26
Added a EventBusHook facility to DefaultEventBus in order to be able …

nevesmarcio:
…to listen Message delivery events; Extended SockJSSocket interface to be able to return a descriptor of the TransportType established (RawWebsocket, XhrTransport, WebSocketTransport, etc.)

Signed-off-by: nevesmarcio nevesmarcio@gmail.com

### DIFF
--- a/vertx-core/src/main/java/org/vertx/java/core/eventbus/EventBus.java
+++ b/vertx-core/src/main/java/org/vertx/java/core/eventbus/EventBus.java
@@ -552,5 +552,12 @@ public interface EventBus {
    * Return the value for default send timeout
    */
   long getDefaultReplyTimeout();
-}
 
+	/**
+	 * sets the eventbus hook for message delivery
+	 * @param eventbushook
+	 * @return
+	 */
+	EventBus setHook(EventBusHook eventbushook);
+
+}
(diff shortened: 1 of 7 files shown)


vietj:
hi, the event bus recently had an interceptor addition that one can use to build similar things, could you check that feature to see if that could be built using it ? if not, perhaps we can improve the eventbus interceptors to achieve it.

emmanuelidi:
The interceptor addition allows one to see messages _sent_ on the eventbus...but there's no way to intercept delivery. @vietj what do you think about an addition to allow for interception on the delivery size? Some context: I'm working on a distributed tracing module for a vertx project using Zipkin and not having a non-hackery approach to seeing the messages delivered is a bit of a problem.

vietj:
@emmanuelidi is your distributed tracing for Vert.x 2 ?

emmanuelidi:
@vietj no, I'm working with vert.x v3.x. I just noticed this after some googling...it looks slightly closer to what I want versus the current interceptors. The current interceptors allow one to see messages at the point of send but I would need to also see the point of delivery to propagate tracing context and understand transmission time (which should be negligible...unless there are problems). 

Technically, I guess the metrics interface has all of the connection points...but we don't get to see messages so I can't manipulate the headers.

If you're open to allowing hooks for the delivery side of message, is there a preference for implementation and I can work on a PR?

PR #1188, opened 2015-10-26
annotate api with @Nullable

vietj:
the api has been annotated with `@Nullable` where it makes sense to improve the polyglot API.

please review

### DIFF
--- a/src/main/java/io/vertx/core/cli/CLI.java
+++ b/src/main/java/io/vertx/core/cli/CLI.java
@@ -18,6 +18,7 @@ package io.vertx.core.cli;
 
 import io.vertx.codegen.annotations.Fluent;
 import io.vertx.codegen.annotations.GenIgnore;
+import io.vertx.codegen.annotations.Nullable;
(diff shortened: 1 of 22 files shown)


src/main/java/io/vertx/core/cli/CLI.java
```diff
@@ -88,6 +89,7 @@ static CLI create(Class<?> clazz) {
   /**
    * @return the CLI name.
    */
+  @Nullable
```
cescoffier: That's a good question. I don't think the name can be null.

cescoffier:
I've a doubt about the CLI name. Everything else looks good.

src/main/java/io/vertx/core/cli/CLI.java
cescoffier: The name of the CLI cannot be null.

vietj: ok :-)

PR #1189, opened 2015-10-29
fixed non-null message for codec.name()

tmont:
Just noticed this while browsing the source code :)

Changed `code.name()` to `codec.name()` in the message of the non-null check.


vietj:
please sign your commit and agree to CLA.

tmont:
ugh. no thanks. if you want to just add the two characters yourself go for it.

PR #1193, opened 2015-11-06
Changed FileResolver so it can handle nested fat jars

rworsnop:
For example, a resource we want to find in a Spring Boot fat jar might be here:
app.jar!/lib/dependency.jar!/webroot/hello.html

I modified unpackFromJarURL so that, in the case of a Spring Boot fat jar, it
opens the nested jar using getResourceAsStream.

Since we're not dealing with a File anymore, I switched to using ZipInputStream
instead of ZipFile.

I also spotted what looked like a copy/paste error in FileResolverTestBase:
testRecursivelyUnpack and testRecursivelyUnpack2 were the same.

This fixes [481623](https://bugs.eclipse.org/bugs/show_bug.cgi?id=481623)

### DIFF
--- a/src/main/java/io/vertx/core/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/impl/FileResolver.java
@@ -19,6 +19,7 @@ package io.vertx.core.impl;
 import io.vertx.core.*;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


purplefox:
Thanks.

PR #1196, opened 2015-11-12
BZ482023: Pump.pump does not check that the streams != null

alexlehm:
added Objects.requireNonNull and the unit tests for that


cescoffier:
Ok for me.

PR #1198, opened 2015-11-13
Provide response size to bodyEndHandler on HttpServerResponse

stampy88:
Currently it is very difficult to write an access log containing the response size for a chunked response. By adding the response size as the parameter to the bodyEndHandler it simplifies this tremendously. Have a subsequent PR for vertx-web when this is merged.

Signed-off-by: Dave Sinclair <stampy88@yahoo.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServerResponse.java
+++ b/src/main/java/io/vertx/core/http/HttpServerResponse.java
@@ -362,13 +362,13 @@ public interface HttpServerResponse extends WriteStream<Buffer> {
 
   /**
    * Provide a handler that will be called just before the last part of the body is written to the wire
-   * and the response is ended.<p>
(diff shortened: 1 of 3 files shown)


purplefox:
Hi Stampy,

Unfortunately we can't do it this way as it would involve an incompatible change on a public API (we must maintain backwards compatibility in minor or point releases).

A different way to do this and which would preserve backwards compatibility would be to keep the signature of the body end handler the same but to expose a new method bytesWritten() which could called from inside the bodyEndHandler.

stampy88:
Understood Tim. I was hoping with Java 8 no one was typing Handler<Void>, but need to think of the Java 7 guys too. I'll create a new PR shortly with your suggestion. 

Thanks

PR #1199, opened 2015-11-19
Provide the remote name to the TCPMetrics connected callback so clien…

vietj:
…t metrics can be reported using the host name

### DIFF
--- a/src/test/java/io/vertx/test/core/MetricsTest.java
+++ b/src/test/java/io/vertx/test/core/MetricsTest.java
@@ -19,6 +19,7 @@ package io.vertx.test.core;
 import io.vertx.core.Vertx;
 import io.vertx.core.VertxOptions;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.datagram.DatagramSocket;
 import io.vertx.core.eventbus.MessageConsumer;
 import io.vertx.core.http.HttpClient;
 import io.vertx.core.http.HttpClientRequest;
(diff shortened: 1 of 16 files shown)


tsegismont:
I haven't checked, do you know if the issue applies to DatagramSocketMetrics as well?

vietj:
the is no connection notion in datagrams.

vietj:
you mean void bytesRead(S socketMetric, SocketAddress remoteAddress, long numberOfBytes); ?

vietj:
we can try to address these as well

vietj:
WDYT ?

tsegismont:
Right, no connection, but I was talking about the ability to get the host string provided by the user when he's sending UDP packets to a remote host

vietj:
I can do these as well today I think.

tsegismont:
I believe it also would be better to get the host string provided by the user when he's setting up a UDP listener

src/test/java/io/vertx/test/core/MetricsTest.java
```diff
@@ -476,7 +478,16 @@ public void testMulti() {
   }
 
   @Test
-  public void testHttpConnect() {
+  public void testHttpConnect1() throws Exception {
+    testHttpConnect("localhost", socketMetric -> assertEquals("localhost", socketMetric.remoteName));
```
tsegismont: So here the lambda is the check for the client metric, right?

src/test/java/io/vertx/test/core/MetricsTest.java
```diff
@@ -493,15 +504,17 @@ public void testHttpConnect() {
         assertFalse(serverMetric.socket.connected.get());
         assertEquals(5, serverMetric.socket.bytesRead.get());
         assertEquals(5, serverMetric.socket.bytesWritten.get());
+        assertEquals(serverMetric.socket.remoteAddress.host(), serverMetric.socket.remoteName);
```
tsegismont: I guess this is ok to test the server metric side, provided Netty does not resolve the address internally.

tsegismont:
Looks good to me.

PR #1200, opened 2015-11-19
Add a callback invoked when the event bus is initialized. 

cescoffier:
It avoids NPE in metric providers using the event bus.

They still can't use it in their constructor but are notified when the event bus is initialized.
This commit is related to https://github.com/vert-x3/vertx-hawkular-metrics/issues/4


vietj:
there is not test for this feature.

cescoffier:
I'm going to add a test.

tsegismont:
Looks good to me.

PR #1202, opened 2015-11-19
Fix default classloader in command lookup

cescoffier:
This allows the launcher to find commands when launched using the maven exec plugin.


vietj:
looks good

PR #1203, opened 2015-11-20
Added bytesWritten to HttpServerResponse

stampy88:
Currently it is very difficult to write an access log containing the response size for a chunked response. By adding the bytesWritten method to HttpServerResponse it simplifies this tremendously.

Signed-off-by: Dave Sinclair <stampy88@yahoo.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServerResponse.java
+++ b/src/main/java/io/vertx/core/http/HttpServerResponse.java
@@ -372,7 +372,7 @@ public interface HttpServerResponse extends WriteStream<Buffer> {
   HttpServerResponse bodyEndHandler(@Nullable Handler<Void> handler);
 
   /**
-   * @return the total number of bytes written for this response
+   * @return the total number of bytes written for the body of the response.
    */
   long bytesWritten();
 }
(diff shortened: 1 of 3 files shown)


stampy88:
@purplefox any chance you'll be able to look over this soon? Right now I have my version in our maven repo until this is merged. Thanks!

cescoffier:
looks good.

stampy88:
@cescoffier if this looks good, can you merge it? Thanks!

cescoffier:
@stampy88 : we are preparing vert.x 3.2, it will be merged before the release (very soon).

vietj:
I will merge it soon.

stampy88:
Great, thanks guys!

pmlopes:
:+1:

src/main/java/io/vertx/core/http/HttpServerResponse.java
```diff
@@ -371,4 +371,8 @@ default HttpServerResponse sendFile(String filename, long offset, Handler<AsyncR
   @Fluent
   HttpServerResponse bodyEndHandler(@Nullable Handler<Void> handler);
 
+  /**
+   * @return the total number of bytes written for this response
```
vietj: shouldn't we change that to `bodyBytesWritten` to be more precise ?

stampy88: Are you talking about the comment, which I agree is misleading, or the name of the method too? The method name was originally suggested by @purplefox, so may want his 2 cents.

vietj:
I think actually bytesWritten is fine too, but at least we should precise that it is for the response body.

stampy88:
Does this work?
```@return the total number of bytes written for the body of the response.```

vietj:
perfect :-)

stampy88:
Pushed that change.

PR #1210, opened 2015-11-29
Cache "Connection was closed" as a singleton

testn:
Eliminate the stacktrace generation to improve performance in high connection/disconnection rate case. 

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/ClientConnection.java
@@ -56,6 +56,8 @@ class ClientConnection extends ConnectionBase {
 
   private static final Logger log = LoggerFactory.getLogger(ClientConnection.class);
 
+  private static final VertxException CONNECTION_CLOSED_EXCEPTION;
+
   private final HttpClientImpl client;
   private final String hostHeader;
   private final boolean ssl;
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -74,6 +76,11 @@
   private HttpClientRequestImpl requestForResponse;
   private WebSocketImpl ws;
 
+  static {
+    CONNECTION_CLOSED_EXCEPTION = new VertxException("Connection was closed");
+    CONNECTION_CLOSED_EXCEPTION.setStackTrace(new StackTraceElement[0]);
```
cescoffier: I'm a bit worried by the absence of stack trace in the exception. Could you give a reason for this ?
(I understand that it is required to build the singleton)

cescoffier:
Any particular reason why you want to have a singleton ? Did you see performance hogs ?

gmagniez:
I've opened a bugzilla issue few days ago for something similar:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=483098

vietj:
we should use a real exception like now but instead check we really need to instantiate it (i.e check requests.size() > 0 etc...)

testn:
I saw the performance hog in a short request with a lot of connection/disconnection from a lot of clients. We don't really need real exception either as "Connection was closed" is only thrown by handleEnded() request so the stacktrace is not very important. @gmagniez may be able to attest to some of that

vietj:
It is now actually the case as it was reported to be an issue with JIT in some cases.

PR #1214, opened 2015-12-02
Fix thread safety issue for writePos variable

testn:
writePos variable was touched outside the synchronization block. This may cause a file corruption if the writes are done from various threads.


vietj:
shouldn't you reduce the scope of synchronization to only += statement ?

PR #1231, opened 2015-12-08
Add max queue size in HttpClient

meshuga:
https://github.com/eclipse/vert.x/issues/1212


meshuga:
see https://github.com/eclipse/vert.x/pull/1232

Issue #1212, opened 2015-11-30
Add max queue size in HttpClient


meshuga:
In Vert.x 2 there was an option to limit the size of HttpClient waiting connection queue made in this PR - https://github.com/eclipse/vert.x/pull/849 (discussion in https://github.com/eclipse/vert.x/pull/845). Apparently, in Vert.x 3, due to the refactoring, this functionality wasn't included.
See https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/ConnectionManager.java#L95

vietj:
thanks for pointing this out, that's something to consider for the next version.

ddimensia:
I guess @meshuga beat me to it by a few hours.  I included the generated source also, can't get a clear understanding if those changes are intended to be committed or not based on past commits.

meshuga:
You've got documentation, this time you win @ddimensia :wink:

ddimensia:
You're optimistic, I wouldn't close the issue until one of the PRs is merged. :smile:

meshuga:
Sorry, I wanted to close my PR.

PR #1232, opened 2015-12-08
Ported 2.x support for setting maxWaitQueueSize on HttpClient to master.

ddimensia:
See #1212 and #849

### DIFF
--- a/src/test/java/io/vertx/test/core/HttpTest.java
+++ b/src/test/java/io/vertx/test/core/HttpTest.java
@@ -186,6 +186,10 @@ public class HttpTest extends HttpTestBase {
     assertEquals(HttpVersion.HTTP_1_0, options.getProtocolVersion());
     assertIllegalArgumentException(() -> options.setProtocolVersion(null));
 
+    assertEquals(HttpClientOptions.DEFAULT_MAX_WAIT_QUEUE_SIZE, options.getMaxWaitQueueSize());
+    assertEquals(options, options.setMaxWaitQueueSize(100));
+    assertEquals(100, options.getMaxWaitQueueSize());
+
     testComplete();
   }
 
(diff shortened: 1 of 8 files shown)


src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -2660,6 +2671,57 @@ private void testPooling(boolean keepAlive, boolean pipelining) {
...
+      });
+
+      req1.end();
+      req2.end();
+      req3.end();
+    }));
+
+    awaitLatch(latch);
+
+    assertTrue("Didn't receive ConnectionPoolTooBusyException", tooBusyExceptionOccurred.get());
```
vietj: use await() here

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -2660,6 +2671,57 @@ private void testPooling(boolean keepAlive, boolean pipelining) {
...
+      });
+      req2.exceptionHandler(t -> fail("Should not be called."));
+
+      HttpClientRequest req3 = client.get(DEFAULT_TEST_URI, resp -> {
+        fail("Should not be called.");
+      });
+      req3.exceptionHandler(t -> {
+        assertTrue("Incorrect exception time.", t instanceof ConnectionPoolTooBusyException);
+        tooBusyExceptionOccurred.set(true);
+        latch.countDown();
```
vietj: use testComplete() here

ddimensia: I can try that but technically the second request marks the end of the test because we expect the third to receive an exception immediately upon calling end() since the pool size is 2 and maxWaitQueueSize is 0.

ddimensia:
Addressed comments and resolved conflicts with updated master.

PR #1233, opened 2015-12-08
Optionally specifying whether to fail on unknown properies

mwhitlaw:
Added overloaded version of Json.decode that supports specifying whether to fail on unknown properties in the source json string, defaulting to true.

Signed-off-by: Mike Whitlaw <mike.whitlaw@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -18,8 +18,10 @@ package io.vertx.core.json;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.module.SimpleModule;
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/json/Json.java
```diff
@@ -72,12 +74,26 @@ public static String encodePrettily(Object obj) throws EncodeException {
...
+
+  public static <T> T decodeValue(String str, Class<T> clazz, boolean failOnUnknownProperties) throws DecodeException {
+    ObjectReader reader = mapper.readerFor(clazz);
+
+    if (failOnUnknownProperties) {
+      reader = reader.with(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+    } else {
+      reader = reader.without(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+    }
+
```
cescoffier: Just a thought, wouldn't it create a new reader everytime ? 
Shouldn't we try to use the object mapper if `failOnUnknownProperties` is true (there is a rish that the mapper has been configured otherwise).

mwhitlaw: Okay. I modified so that the non-default reader is only created when calling for non-default behavior. Yes, this does rely on the default mapper being configured to fail on unknown properties, but that was the presumption before as well.

mwhitlaw:
Just curious, do I need to do anything else for this before it is accepted or rejected?

cescoffier:
Hi @mwhitlaw,

I would like to have the opinion of @pmlopes and @vietj about it. My concerns is about instantiating many mappers, and the compatibility with other PR (such as the one allowing registering modules). 

As we are going to cut a 3.2.1 release (bug fix), it would not be merged until we start working on 3.3.0.

mwhitlaw:
Okay. Thanks. Although, no new mappers are being instantiated. Only the class instance of ObjectMapper "mapper" is being used. As far as other object (Readers) instantiation goes, the original/current (pre pull request) implementation of Json.decode uses:
mapper.readValue(str, clazz) 
which is still the default implementation in this pull request. The added optional behavior uses:
mapper.readerFor(clazz).without(...).readValue(str)
I don't believe this second call results in any more object instantiations than the first. i.e. The mapper.readValue method call is just a convenient way of calling mapper.readFor(clazz).readValue(str)

cescoffier:
I just check the jackson code, and from what I see it creates another reader:

https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/ObjectReader.java#L450

and 

https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/ObjectReader.java#L295

mwhitlaw:
Ah. Interesting. I would have thought that a new Reader was required by ObjectMapper.readValue, as well. But maybe not. I suppose I could add another class level ObjectMapper for doing this, similar to the way "prettyMapper" is used? Would that be more acceptable?

cescoffier:
About the prettyMapper, I don't know. We need to have a look, and plug a profiler to check that we don't create too many objects. Jackson API is not that easy to tame.

About creating another ObjectMapper, this would break module registration. As, it would need to be registered on both.

mwhitlaw:
Agreed. Adding a third ObjectMapper (for instance "unknownPropertyMapper") would required the addition of that mapper to the new "registerModule" method in the other PR, but then the addition Reader  instances issue on this PR goes away. Seems there are pro's and con's to each approach.

cescoffier:
That's why I would like to have the opinion of @vietj and @pmlopes. Some external point of view would be good.

PR #1238, opened 2015-12-10
updates to make max http url length and the max header size configura…

sameer7:
…ble. Also exposed isSSL in http request interface

Signed-off-by: Patel, Sameer <sameer.us@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServerRequest.java
+++ b/src/main/java/io/vertx/core/http/HttpServerRequest.java
@@ -73,7 +73,7 @@ public interface HttpServerRequest extends ReadStream<Buffer> {
   HttpMethod method();
   
   /**
-   * @return whether the HTTP connection is secure
+   * @return true if this {@link io.vertx.core.net.NetSocket} is encrypted via SSL/TLS
    */
   boolean isSSL();
 
--- a/src/main/java/io/vertx/core/http/HttpServerRequest.java
+++ b/src/main/java/io/vertx/core/http/HttpServerRequest.java
@@ -71,6 +71,11 @@ public interface HttpServerRequest extends ReadStream<Buffer> {
    * @return the HTTP method for the request.
    */
   HttpMethod method();
+  
+  /**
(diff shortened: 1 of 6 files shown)


src/main/java/io/vertx/core/http/HttpServerRequest.java
```diff
@@ -71,6 +71,11 @@
    * @return the HTTP method for the request.
    */
   HttpMethod method();
+  
+  /**
+   * @return whether the HTTP connection is secure
```
vietj: please use the same wording that in NetSocket : `true if this {@link io.vertx.core.net.NetSocket} is encrypted via SSL/TLS.`

vietj:
please add tests:

-  that checks the line is truncated at 4000 with default option and one test that shows it is not truncated with a different options
- same for max headers
- add isSsl == true in existing Https tests

sameer7:
I could not find the Https tests for isSSL, can you point me to that. I got the rest.

vietj:
I mean a check on isSSL in HttpTest#testTLS server request handler.

Perhaps we should have the same on websocket too.

vietj:
I see no tests for the maxInitialLineLength and maxHeadersLength feature.

sameer7:
It's there. Line #630 & 726 in my second commit.

vietj:
these tests are for option. we need to have tests with an HttpClient that submit requests with such lines and headers and a verification on the server that the line and headers are correct.

sameer7:
Added tests with HttpClient with server side verification.

sameer7:
Any updates on this?

vietj:
it was not included in 3.2 by lack of time and review. It should be merged in master after review.

cescoffier:
The branch is not mergeable anymore. Could you update ?

sameer7:
Updated.

sameer7:
Did that work for you?

vietj:
thanks for the contribution!

sameer7:
thank you all

PR #1259, opened 2016-01-04
Make sure that request.handleException is called only once

testn:
If the connection is broken before the entire request is written (POST/PUT), `requests` may contain `currentRequest`. This, sometimes, causes `request.handleException` to be called twice on the same request.


vietj:
handleException is called for each request of the connection, this is necessary as there could be several requests made in the current connection (keepalive or pipeline).

testn:
@vietj, read it carefully. The change is still allowed all pipeline requests to receive a callback when the connection is closed. The change is to ensure that if the connection is closed in the middle of a put, the current request will be callbacked ony once.

vietj:
we would need at least one test to show this behavior that this change fixes

PR #1260, opened 2016-01-04
Make sure that request.handleException is called only once

testn:
Test added 


testn:
@vietj test added

vietj:
Actually I don't understand why only a single request should have this exception and not the others in the pipeline

testn:
If the exception happens when the request has not been fully written yet, the error handler will be invoked twice.

vietj:
I've added the test in vertx to ensure the behavior - thanks - I think this was fixed a while ago

Issue #1258, opened 2015-12-29
Fat-JAR-Error: on unpackFromJarURL with whitespaces in the path of windows filepath


nixoso:
This happens in Vert.x 3.2. 

On start-up a vertx-web-fat-JAR on a path with whitespace i get a internall server error.  

Example -> ok: 
Working dir: C:\\Users\\**Max**\\git\\Project\\target\\
Start-CMD: java -jar webapp.jar

Example -> failed: 
Working dir: C:\\Users\\**Max Mustermann**\\git\\Project\\target\\
Start-CMD: java -jar webapp.jar

```
500 C:\Users\Max%20Mustermann\git\Project\target\webapp.jar (Das System kann den angegebenen Pfad nicht finden)

    java.io.FileInputStream.open0(Native Method)
    java.io.FileInputStream.open(Unknown Source)
    java.io.FileInputStream.(Unknown Source)
    java.io.FileInputStream.(Unknown Source)
    io.vertx.core.impl.FileResolver.toZipInputStream(FileResolver.java:171)
    io.vertx.core.impl.FileResolver.unpackFromJarURL(FileResolver.java:180)
    io.vertx.core.impl.FileResolver.resolveFile(FileResolver.java:121)
    io.vertx.core.impl.VertxImpl.resolveFile(VertxImpl.java:657)
    io.vertx.core.file.impl.FileSystemImpl$6.perform(FileSystemImpl.java:517)
    io.vertx.core.file.impl.FileSystemImpl$6.perform(FileSystemImpl.java:514)
    io.vertx.core.file.impl.FileSystemImpl.propsBlocking(FileSystemImpl.java:150)
    io.vertx.ext.web.handler.impl.StaticHandlerImpl.lambda$getFileProps$4(StaticHandlerImpl.java:264)
```

PR #1263, opened 2016-01-06
Fix issue #1258 - resolve file from jar when the path contains spaces

cescoffier:
Fix issue #1258

* It decodes the url before opening the file input stream as it may contains %20 characters for spaces.
* Add a test suite configuration to load files from a jar files that is located in a directory containing spaces (uses a TCCL switch to append it to the class path)


cescoffier:
@vietj, @pmlopes  please review.

vietj:
:+1:

PR #1267, opened 2016-01-07
HttpClient sharing clarifications and warns - fixes #1266

### LINKED ISSUE #1266: HttpClient sharing clarifications and warns
An HttpClient should not be shared between different Verticles or Context. It should be documented and when the client detects a different context, it should print a warn on the console to make the user aware of this situation.

vietj:
please someone review.


pmlopes:
+1

PR #1269, opened 2016-01-08
Implement toString() method

Junbong:
Signed-off-by: Junbong koliie9039@gmail.com

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HeadersAdaptor.java
+++ b/src/main/java/io/vertx/core/http/impl/HeadersAdaptor.java
@@ -23,6 +23,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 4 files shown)


Junbong:
I know the platform is very JSON oriented but I think this is not that same. MultiMap obtain used as the "request header" which style is like KEY=VALUE;. Otherwise, I just followed familiar Java toString() style of Collection.

alexlehm:
I have requested and implemented the same thing for CaseInsensitiveHeaders a while back https://bugs.eclipse.org/bugs/show_bug.cgi?id=430965.
I missed the other implementation of MultiMap apparently. I am not sure which format is better, but I think it should be consistent at least. I have used that for logging mostly until now.

Currently the output of CaseInsesitiveHeaders.toString() looks like this:

````
System.out.println(new CaseInsensitiveHeaders().add("Header1", "value1").add("header1", "value2").add("header1", "value1").add("Header2", "value3").toString());
````

````
Header1: value1
header1: value2
header1: value1
Header2: value3
````

PR #1271, opened 2016-01-08
PROXY protocol support to HttpServer

jkeys089:
This PR adds support for the PROXY protocol (http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt) to the HttpServer.

Signed-off-by: Jon Keys jon.keys@gmail.com

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,6 +106,11 @@
       <version>${netty.version}</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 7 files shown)


vietj:
hi, this seems to be interesting but we didn't have time to include it in 3.3.0 - will look for 3.3.1 instead

PR #1274, opened 2016-01-09
Added possibility to register custom jackson modules in Json utility class

pjozsef:
Adds the registerModule method to the Json class, so decoding/encoding can be extended with custom serializers/deserializers.

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -39,6 +39,10 @@ public class Json {
   public static ObjectMapper prettyMapper = new ObjectMapper();
 
   static {
+    initialize();
+  }
+
+  static void initialize(){
     // Non-standard JSON but we allow C style comments in our JSON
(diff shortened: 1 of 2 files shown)


cescoffier:
If we add registration support, it would be good to have the unregister no ?

We would need unit tests, before integrating this feature.

pjozsef:
@cescoffier as far as I know there is no way in Jackson to unregister a module.
Added tests to make sure the registered module is used for serialization/deserialization.

pmlopes:
This is a common requested feature, jackson does not support unregister of modules and recommends creating a new object mapper. We could however add a method that clear the modules e.g.:

```java
void clearModules() {
  mapper = new ObjectMapper();
  // so we get back to the original state
  initialize();
}
```

vietj:
it would be great to "approve" via github only when all remarks have been done

vietj:
thanks

vietj:
the original repo from which the PR originated is not anymore available, besides the mappers are public and can be modified actually without this patch - closing

PR #1276, opened 2016-01-11
Support for SSL/TLS protocol restrictions on SSL/TLS enabled servers

davidafsilva:
Hey!

I haven't found a way to restrict the SSL/TLS protocol on SSL/TLS enabled servers, so i've added the ability to restrict them (did i miss something in either my implementation or any existent code to achieve what i needed?).

To complement a bit, this is one of the requirements for us, at my company, to be able to use vert.x, along with secure channels (eventbus traffic - i see that an issue was recently opened, great!!) and some other security related ones.

Sorry if i messed up, dunno if docs were intended to be part of the commit, i've run `mvn clean package` prior to the commit. 
I's my first contribution though, looking forward to review/comments! :)

Cheers!

### DIFF
--- a/src/test/java/io/vertx/test/core/HttpTest.java
+++ b/src/test/java/io/vertx/test/core/HttpTest.java
@@ -16,31 +16,32 @@
 
 package io.vertx.test.core;
 
-import io.netty.handler.codec.http.DefaultHttpHeaders;
-import io.netty.handler.codec.http.HttpResponseStatus;
-import io.vertx.core.*;
-import io.vertx.core.buffer.Buffer;
-import io.vertx.core.eventbus.Message;
-import io.vertx.core.eventbus.MessageConsumer;
-import io.vertx.core.http.*;
-import io.vertx.core.http.impl.HeadersAdaptor;
-import io.vertx.core.impl.*;
-import io.vertx.core.json.JsonArray;
-import io.vertx.core.json.JsonObject;
-import io.vertx.core.net.*;
-import io.vertx.core.parsetools.RecordParser;
-import io.vertx.core.streams.Pump;
 import org.junit.Assume;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
-import java.io.*;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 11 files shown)


src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -3179,6 +3255,7 @@ public void testTLSClientRevokedServerCert() throws Exception {
 
   @Test
   //Client specifies cert that the server does not trust via a revoked certificate of the CA
+  @Ignore // this test must be reviewed - it was "passing" when "fail" was expected
```
cescoffier: Can you tell us more about this ?

davidafsilva: Sure!

I've added an [assert](https://github.com/eclipse/vert.x/pull/1276/files#diff-19c82d5fdde982f60afb99476e34f478R3354) to the `testTLS()` method to ensure that the `shouldPass` value is indeed true when it reaches that work-flow.
Looking at the definition of that particular test case, it was supposed to "fail" since:
1. With a revoked client certificate, the channel will not be established
2. `shouldPass = false`
But, the test was following a "success" work-flow, which lead to that new assert to fail.

dan-lind:
@cescoffier Will this make it into 3.3 or are there some changes needed?

cescoffier:
Gonna have a look. It seems eligible (understand on time ;-)) for 3.3.

cescoffier:
We would need to cleanup the ignored test (and see why it was wrong at the first place). The documentation would need a paragraph on this too.

dan-lind:
Is this something that I can contribute, given that I didn't create the original PR?

cescoffier:
You can reuse the same branch on your own repo, and create a new PR. I'm pretty sure @davidafsilva would be happy to see enhancements. Just be sure to signed off all the commits and have signed the Eclipse CLA.

vietj:
could you rebase this PR agains the http2-openssl branch ? some reorg have been made in TLS tests to support OpenSSL testing

src/test/java/io/vertx/test/core/HttpTest.java
purplefox: We should never have ignored tests in the test suite. If the test is broken or wrong it should either be fixed or removed.

davidafsilva: I do agree with that, but i wanted the test to be reviewed by one of you, preferably the one o designed the test, so he/she can decide what to do with it. That's why i left it there (ignored though).

purplefox: If you think there is an issue with the test, a good way to deal with this is to create a new GitHub issue and explain it in there.

Adding @Ignore to tests is dangerous as, if Clement hadn't spotted it, it might have ended up in the main codebase and we'd have lost a potentially important test :)

davidafsilva: I see, my bad then! :)
Thank you.

dan-lind: So where does this leave us? :-) Will you fix this @davidafsilva or do you want me to take a look?

vietj:
closing this one

PR #1279, opened 2016-01-14
add byte order to Buffer

stream1984:
I can not found a way to set byte order from vert.x, so i think may i expose this method from interface of Buffer ?
Signed-off-by: stream <stream1984@gmail.com>


vietj:
I'm wondering how this works as ByteOrder is not an enum and not code generated.

vietj:
@stream1984 have you checked this compiles ?

stream1984:
Hi vietj
The method of buffer.order() will be deprecated in netty 4.1.0 and we can get Little Endian bytes with new method `getShortLE()` etc, so i think this PR could be closed, maybe we can also provide such method in 3.3.0? should i make another PR ?

vietj:
yes please

PR #1281, opened 2016-01-15
Composite future implementation

vietj:
See https://github.com/vert-x3/issues/issues/84

### DIFF
--- a/src/main/asciidoc/java/index.adoc
+++ b/src/main/asciidoc/java/index.adoc
@@ -331,6 +331,8 @@ CompositeFuture.any(future1, future2).setHandler(ar -> {
 });
 ----
 
+`link:../../apidocs/io/vertx/core/Future.html#compose-io.vertx.core.Handler-io.vertx.core.Future-[compose]` chains futures:
+
 == Verticles
 
 Vert.x comes with a simple, scalable, _actor-like_ deployment and concurrency model out of the box that
--- a/src/main/asciidoc/java/index.adoc
+++ b/src/main/asciidoc/java/index.adoc
@@ -316,7 +316,7 @@ CompositeFuture.all(httpServerFuture, netServerFuture).setHandler(ar -> {
 The handler returned by `link:../../apidocs/io/vertx/core/Future.html#completer--[completer]` completes the future.
 
 `link:../../apidocs/io/vertx/core/CompositeFuture.html#any-io.vertx.core.Future-io.vertx.core.Future-[CompositeFuture.any]` takes several futures arguments (up to 6) and returns a future that is
-_failed_ when all the futures are and _succeeded_ otherwise:
+_succeeded_ if one of the futures is, and _failed_ otherwise:
 
 [source,java]
 ----
@@ -341,10 +341,10 @@ Future<Void> fut1 = Future.future();
 Future<Void> fut2 = Future.future();
 
 fs.createFile("/foo", fut1.completer());
-fut1.compose(string -> {
+fut1.compose(v -> {
   fs.writeFile("/foo", Buffer.buffer(), fut2.completer());
 }, fut2);
-fut2.compose(integer -> {
+fut2.compose(v -> {
   fs.move("/foo", "/bar", startFuture.completer());
 }, startFuture);
 ----
--- a/src/main/asciidoc/java/index.adoc
+++ b/src/main/asciidoc/java/index.adoc
@@ -331,7 +331,23 @@ CompositeFuture.any(future1, future2).setHandler(ar -> {
 });
 ----
 
-`link:../../apidocs/io/vertx/core/Future.html#compose-io.vertx.core.Handler-io.vertx.core.Future-[compose]` chains futures:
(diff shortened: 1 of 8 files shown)


src/main/asciidoc/java/index.adoc
```diff
@@ -289,6 +289,66 @@ An alternative way to run blocking code is to use a <<worker_verticles, worker v
...
+  } else {
+    // At least one server failed
+  }
+});
+----
+
+The handler returned by `link:../../apidocs/io/vertx/core/Future.html#completer--[completer]` completes the future.
+
+`link:../../apidocs/io/vertx/core/CompositeFuture.html#any-io.vertx.core.Future-io.vertx.core.Future-[CompositeFuture.any]` takes several futures arguments (up to 6) and returns a future that is
+_failed_ when all the futures are and _succeeded_ otherwise:
```
tsegismont: Minor, I wonder if this would be more clear: "and returns a future that is _succeeded_ if one of the futures is, and _failed_ otherwise"

src/main/asciidoc/java/index.adoc
```diff
@@ -289,6 +289,66 @@ An alternative way to run blocking code is to use a <<worker_verticles, worker v
...
+
+[source,java]
+----
+FileSystem fs = vertx.fileSystem();
+
+Future<Void> fut1 = Future.future();
+Future<Void> fut2 = Future.future();
+
+fs.createFile("/foo", fut1.completer());
+fut1.compose(string -> {
```
tsegismont: Shouldn't the lambda argument be of the same type as the fut1 generic type?

tsegismont: Forget about this.

src/main/java/examples/CoreExamples.java
```diff
@@ -83,6 +86,50 @@ String blockingMethod(String str) {
...
+
+  public void exampleFuture3(Vertx vertx, Future<Void> startFuture) {
+
+    FileSystem fs = vertx.fileSystem();
+
+    Future<Void> fut1 = Future.future();
+    Future<Void> fut2 = Future.future();
+
+    fs.createFile("/foo", fut1.completer());
+    fut1.compose(string -> {
```
tsegismont: The parameter name is confusing, I think.

src/main/java/io/vertx/core/CompositeFuture.java
```diff
@@ -0,0 +1,152 @@
...
+    return CompositeFutureImpl.all(f1, f2, f3, f4, f5);
+  }
+
+  /**
+   * Like {@link #all(Future, Future)} but with 6 futures.
+   */
+  static <T1, T2, T3, T4, T5, T6> CompositeFuture all(Future<T1> f1, Future<T2> f2, Future<T3> f3, Future<T4> f4, Future<T5> f5, Future<T6> f6) {
+    return CompositeFutureImpl.all(f1, f2, f3, f4, f5, f6);
+  }
+
```
tsegismont: Out of curiosity, why not adding another method with a Future array as single parameter? In case you have more than 6 futures.

vietj: arrays are not supported in codegen, only List. Someone can use two all if it needs more. and I think that most will be happy with 6

src/main/java/io/vertx/core/Future.java
```diff
@@ -133,6 +137,38 @@
...
+   * @param handler the handler
+   * @param next the next future
+   */
+  default <U> void compose(Handler<T> handler, Future<U> next) {
+    setHandler(ar -> {
+      if (ar.succeeded()) {
+        try {
+          handler.handle(ar.result());
+        } catch (Throwable err) {
+          if (next.isComplete()) {
```
tsegismont: Sorry, we talked about this last week, but I'm still not sure why we should not call `next.fail(err)` if next is complete.

src/main/java/io/vertx/core/CompositeFuture.java
tsegismont: ok

2016-01-18 10:31 GMT+01:00 Julien Viet <notifications@github.com>:

src/main/java/io/vertx/core/Future.java
vietj: a Future should be completed once only.

tsegismont: ok

2016-01-18 10:41 GMT+01:00 Julien Viet <notifications@github.com>:

src/main/java/examples/CoreExamples.java
vietj: what should it be called ?

tsegismont: The name gives the impression that it is a String, but it is a Void. Maybe void1?

tsegismont:
Other than the documentation wording and parameter naming questions, LGTM

src/main/java/examples/CoreExamples.java
vietj: agreed

cescoffier:
:+1:

PR #1239, opened 2015-12-10
Fix missing pass logging arguments on slf4j delegate

zcx2001:
#1208 new fix

PR #1282, opened 2016-01-15
Log delegates fixes

### LINKED ISSUE #1243: SLF4JLogDelegate does not pass down parameters when it is location aware
LocationAwareLogger l = (LocationAwareLogger) logger;
      l.log(null, FQCN, level, msg, null, t);

Should pass params as the second last parameter

### LINKED ISSUE #1280: SLF4j logging arguments ignored
Logging with SLF4J enabled (vertx.logger-delegate-factory-class-name = io.vertx.core.logging.SLF4JLogDelegateFactory) does not use all the log arguments.

for example:
```java
LoggerFactory.getLogger(Foo.class).info("test {}", "param")
````
outputs:
```
"test {}"
```
instead of expected output:
```
"test param"
```

This is easily fixable in io.vertx.core.logging.SLF4jLogDelegate.java#log:
```
-      l.log(null, FQCN, level, msg, null, t);
+      l.log(null, FQCN, level, msg, params, t);
```

cescoffier:
Fix two issues in log delegates:

* SLF4J parameters were not passed correctly
* JUL did not set the _throwable_ when the exception was passed as last parameter

It fixes #1280 and fixes #1243. It also supersedes #1239 and #1280.


cescoffier:
@vietj @pmlopes please review.

plombardi89:
Any ETA when we might see this fix released? Not having parameters in log messages is really really unfortunate with 3.2.1.

cescoffier:
@plombardi89 this has been integrated in 3.2.1. So it should be there.

PR #1285, opened 2016-01-18
Use netty epoll support if available

tcataldo:
Use a NettySupport helper to use Epoll based implementations if available.

Signed-off-by: Thomas Cataldo thomas.cataldo@blue-mind.net

### DIFF
--- a/vertx-core/src/main/java/org/vertx/java/core/impl/NettySupport.java
+++ b/vertx-core/src/main/java/org/vertx/java/core/impl/NettySupport.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2011-2016 The original author or authors
+ * ------------------------------------------------------
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Apache License v2.0 which accompanies this distribution.
+ *
+ *     The Eclipse Public License is available at
+ *     http://www.eclipse.org/legal/epl-v10.html
+ *
+ *     The Apache License v2.0 is available at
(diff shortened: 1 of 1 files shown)


purplefox:
I think it's good to support epoll transport, but it shouldn't be the default.

We had a similar PR some time back and I did a performance comparison between native and standard nio and the native was actually slower!

I think a better way of doing this would be a config setting to choose the transport in the options class.

tcataldo:
I would be happy to do that, but I don't see where I could stick the option.

purplefox:
Probably the NetworkOptions class is the right place.

vertx-core/src/main/java/org/vertx/java/core/impl/NettySupport.java
```diff
@@ -0,0 +1,89 @@
...
+import io.netty.channel.socket.nio.NioServerSocketChannel;
+import io.netty.channel.socket.nio.NioSocketChannel;
+
+/**
+ * This class enables netty epoll support if it is available on your system
+ *
+ */
+public class NettySupport {
+
+	private static final boolean epollSupported = "true".equals(System.getProperty("vertx.epoll.try", "false")) && Epoll.isAvailable();
```
Narigo: Would it make sense to call this `epollEnabled` ? As it's not really "support" only if you have an option for this.

Narigo: The class should probably also be intended with 2 spaces.

vietj:
supported in 3.x

PR #1287, opened 2016-01-18
JsonArray.remove(int pos) should return JsonArray or JsonObject

### LINKED ISSUE #1286: JsonArray.remove(int pos) should return JsonArray or JsonObject instead of their underlying implementations
The current implementation returns / leaks a `java.util.ArrayList<?>` in case of a `JsonArray` or a Map in case of a `JsonObject`. It should check for these implementations and return a JsonArray or JsonObject wrapper.

See https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/json/JsonArray.java#L482

cescoffier:
Fix #1286.
This is a **breaking change** - do not integrate before the 3.2.1 release.

Fix the type inconsistency of the JsonArray remove method. When the object is a Map, it returns a JsonObject. When it's a List, it returns a JsonArray.

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -20,12 +20,7 @@ import io.vertx.core.buffer.Buffer;
 import io.vertx.core.shareddata.impl.ClusterSerializable;
 
 import java.time.Instant;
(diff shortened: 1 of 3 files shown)


vietj:
should be fixed but not in 3.2.1

Issue #1255, opened 2015-12-23
Add an ability to name Vert.x instance


testn:
It is useful when dealing with unit tests that involve multiple vert.x especially if the name is used as part of the threads created by vert.x

PR #1304, opened 2016-02-11
Adding support for naming the vert.x instance

ionelMihai:


ionelMihai:
related to #1255

testn:
It should use the vert.x name to name the event loop and worker thread associated to the vert.x instance.

testn:
@vietj can we get this merged in?

vietj:
not under this form - your use case is about naming threads ?

testn:
Yes. That's right.

PR #1308, opened 2016-02-16
Enable configuration of LoggerFactory through properties file

### LINKED ISSUE #1295: Configure LogDelegateFactory through file
`JULLogDelegateFactory` can be configured through a system property and a file `vertx-default-jul-logging.properties`. I wonder if it would make sense to also make `LoggerFactory` configurable through the system property `vertx.logger-delegate-factory-class-name` as well as a file (probably called `vertx-logger-factory.properties` with a single property `vertx.logger-delegate-factory-class-name`) or through the Java service mechanism maybe. 

This would make it easier to pre-configure the required logger by placing a file in the classpath. This would even work in a fat jar where system properties can only be set by the caller.

What do you think?

michel-kraemer:
This should be very useful for testing and deployment.

Fixes #1295

### DIFF
--- a/src/main/asciidoc/java/index.adoc
+++ b/src/main/asciidoc/java/index.adoc
@@ -1198,6 +1198,10 @@ Log4J and SLF4J with the class names `io.vertx.core.logging.Log4jLogDelegateFact
 respectively. If you want to use these implementations you should also make sure the relevant Log4J or SLF4J jars
 are on your classpath.
 
+As an alternative to setting the system property you may provide a file with the name `vertx-logger-factory.properties` on your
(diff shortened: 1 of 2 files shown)


michel-kraemer:
@vietj @cescoffier @pmlopes Maybe something for the spring cleanup? :-) :-) :-)

cescoffier:
I've started looking at it, and I need to give another spin.

During the last Vert.x user group in Paris, I saw something quite similar but letting you configure almost all system properties (can't remember who did this, I need to check).

michel-kraemer:
> letting you configure almost all system properties

Yeah, that would be a great alternative indeed.

PR #1315, opened 2016-02-24
expose the config after parsing to the Launcher

jtruelove:

### DIFF
--- a/src/main/asciidoc/java/override/dependencies.adoc
+++ b/src/main/asciidoc/java/override/dependencies.adoc
@@ -8,7 +8,7 @@ project descriptor to access the Vert.x Core API:
 <dependency>
   <groupId>io.vertx</groupId>
   <artifactId>vertx-core</artifactId>
-  <version>3.2.1</version>
+  <version>3.3.0-SNAPSHOT</version>
 </dependency>
 ----
 
@@ -16,5 +16,5 @@ project descriptor to access the Vert.x Core API:
 
 [source,groovy,subs="+attributes"]
 ----
-compile io.vertx:vertx-core:3.2.1
+compile io.vertx:vertx-core:3.3.0-SNAPSHOT
 ----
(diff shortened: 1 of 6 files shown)


vietj:
can you review @cescoffier ?

vietj:
can you  explain why you need that @jtruelove ?

cescoffier:
The PR looks good, but I wonder why the callback passing the deployment option cannot be used (`beforeDeployingVerticle`) instead of adding a new hook.

Are you passing in the configuration object Vert.x configuration (and not verticle-centric config) ?

jtruelove:
So here is my use case, I want to use vertx-opentsdb's SPI support and in order to do so I need to make this call happen:

```
// where config here is a value I grab out of the JsonObject that gets created/parsed from the conf.json     //   file passed via the **-conf **flag on the command line 
options.setMetricsOptions(new OpenTsDbOptions(config).setEnabled(true));
```

inside of Launcher's

`void beforeStartingVertx(Vertx vertx);`

So to answer your question, as far as I can see I can't get access to the VertxOptions once it's at the hook `beforeDeployingVerticle` which is too late and VertxOptions have already been used. I could double parse the config file as I have access to the file location via the args coming into Launcher's constructor but that is hacky and dirty which I don't love. So the logical thing seemed to me to provide the config as early in the process as possible in case you need it for doing things in any of the other hooks, and I know you can see the config on deploymentOptions via `getConfig`. I could maybe pass it via another parameter and parse that but I use **-conf** to specify all server config needed for anything which I'm guessing most others do as well, so why split config across files?

cescoffier:
OH I see, thanks for the explanation.

There is a mapping between metrics options and system properties, but I'm not sure it would be convenient for you. 

So, I think adding a callback would make sense in this case. Let me think about it, I will tell you more tomorrow.

cescoffier:
So, I guess the best way is to add the callback and parse the config "early" enough. So, it can be used to configure the `VertxOptions`.

Here is my official +1.

Thanks, and sorry for the delay.

jtruelove:
great! so I guess this will just get merged for the next release?

cescoffier:
Exactly.

Issue #1316, opened 2016-02-26
Error loading from classpath within .zip files


ernestojpg:
### Issue description

I have found an issue with the `vertx.fileSystem().readFile()` function.

The function will fail if you try to load a file from the classpath and that file is stored inside a .zip file.

The problem is in the class `io.vertx.core.impl.FileResolver`. The method `unpackFromJarURL()` is assuming that the file will be always stored inside a .jar file:

```
  private synchronized File unpackFromJarURL(URL url, String fileName, ClassLoader cl) {
    try {
      ZipFile zip;
      String path = url.getPath();
      int idx1 = path.lastIndexOf(".jar!");
      ...
```

The method will throw an `ArrayIndexOutOfBoundsException` if the file is stored inside a .zip file.

### How to reproduce it

I have attached two files:
- **[test.zip](https://github.com/eclipse/vert.x/files/147962/test.zip)**: Compressed file that implements a Vert.x service. It contains two files:
  - mainVerticle.js: Main Service Verticle, written in JavaScript, that try to load a file.
  - myMod.json: Service descriptor, that indicates the location of the main service verticle.
- **[vertx-service-factory-3.2.1.zip](https://github.com/eclipse/vert.x/files/147964/vertx-service-factory-3.2.1.zip)**: Official Vert.x Service Factories implementation, used for loading the above service. Note that the .zip extension is not relevant here. Renamed from .jar to .zip just for being able to upload the file here.

For reproducing the issue just download both files and introduced the following commands (Linux/Mac):
- `export CLASSPATH=${CLASSPTH}:.:vertx-service-factory-3.2.1.zip:test.zip`
- `vertx run service:myMod`

You should see this log:
```
Hello from the mainVerticle.js
Succeeded in deploying verticle
Unhandled exception
<eval>#88:17<eval>:7:7 Error loading myMod.json file: java.lang.ArrayIndexOutOfBoundsException: 1
	at jdk.nashorn.internal.runtime.ECMAException.create(ECMAException.java:113)
	at jdk.nashorn.internal.scripts.Script$Recompilation$138$274AA$\^eval\_#88\!17\^eval\_.L:1$L:5(mainVerticle.js:7)
	at jdk.nashorn.internal.scripts.Script$Recompilation$137$31550A$\^eval\_#88\!17\^eval\_.L:1$FileSystem$readFile$L:831(vertx-js/file_system.js:835)
[...]

PR #1317, opened 2016-02-26
#1316 Fixed bug loading from classpath within .zip files

ernestojpg:
Quick fix for the issue #1316. The method `unpackFromJarURL()` will now take into consideration .zip files in addition to standard .jar files.

I don't know if we should make it even for generic and take also more file extensions into consideration.


vietj:
Hi, can you add a test for this please ?

ernestojpg:
Sure! Can I update this pull request, or should I close it and create a new one?

vietj:
you can still push a commit on this PR

src/main/java/io/vertx/core/impl/FileResolver.java
```diff
@@ -23,7 +23,6 @@
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
-import java.net.URLClassLoader;
```
ernestojpg: No used import

src/main/java/io/vertx/core/impl/FileResolver.java
```diff
@@ -57,7 +56,7 @@
 
   private static final String DEFAULT_CACHE_DIR_BASE = ".vertx";
   private static final String FILE_SEP = System.getProperty("file.separator");
-  private static boolean NON_UNIX_FILE_SEP = !FILE_SEP.equals("/");
+  private static final boolean NON_UNIX_FILE_SEP = !FILE_SEP.equals("/");
```
ernestojpg: This is really a constant

ernestojpg:
Added some tests

vietj:
thanks!

PR #1335, opened 2016-03-10
Add number of NIO loops created by Vert.x in the documentation.

poiuytrez:
I was very weird to noticed in production to have a lot of eventloop threads. I did not noticed that in development because I have less cores. It should be written in documentation to avoid WTF moments :).

Signed-off-by: Guillaume Charhon <guillaume@databerries.com>

### DIFF
--- a/src/main/asciidoc/java/index.adoc
+++ b/src/main/asciidoc/java/index.adoc
@@ -362,8 +362,9 @@ with respect to concurrency, scaling and deployment.
 
 To use this model, you write your code as set of *verticles*.
 
-Verticles are chunks of code that get deployed and run by Vert.x. A Vert.x instance maintains N event loop threads
-(where N by default is core*2) by default. Verticles can be written in any of the languages that Vert.x supports
+Verticles are chunks of code that get deployed and run by Vert.x. Vert.x will create N event loops threads
+(where N by default is core*2) by default. When you deploy a verticle,
+it will be deployed on one of theses event loops. Verticles can be written in any of the languages that Vert.x supports
 and a single application can include verticles written in multiple languages.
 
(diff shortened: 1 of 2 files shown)


src/main/asciidoc/java/index.adoc
```diff
@@ -362,9 +362,10 @@ with respect to concurrency, scaling and deployment.
 
 To use this model, you write your code as set of *verticles*.
 
-Verticles are chunks of code that get deployed and
-run by Vert.x. Verticles can be written in any of the languages that Vert.x supports and a single application
-can include verticles written in multiple languages.
+Verticles are chunks of code that get deployed and run by Vert.x. Vert.x will create N event loops threads
```
purplefox: I would be clearer to say "A Vert.x instance maintains N event loop threads (where N by default is core*2) by default"

src/main/asciidoc/java/index.adoc
```diff
@@ -362,9 +362,10 @@ with respect to concurrency, scaling and deployment.
...
+(where N by default is core*2) by default. When you deploy a verticle,
+it will be deployed on one of theses event loops. Verticles can be written in any of the languages that Vert.x supports
```
purplefox: The verticle is not deployed _on_ an event loop. It is _associated_ to an event loop when the verticle is created.

Also, this is only true for standard (i.e.non worker) verticles.

poiuytrez:
Thank you @purplefox for the comments. I have corrected the documentation.

cescoffier:
@poiuytrez you are not editing the right file. This file is generated. Can you update the package-info.java file ?

poiuytrez:
@cescoffier I have updated the package-info.java file.

PR #1340, opened 2016-03-15
Handle Netty Http Decoder failures

gmagniez:
Vert.X does not handle the Netty Decoder result, do for exemple when an HttpClientResponse headers contains more than 8192 bytes of data the response is handled normally but no data is sent to the attached handler and the request is "frozen" (endHandler not called, exception not thrown).

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/ClientConnection.java
@@ -276,7 +276,11 @@ class ClientConnection extends ConnectionBase {
     }
     HttpClientResponseImpl nResp = new HttpClientResponseImpl(requestForResponse, this, resp);
     currentResponse = nResp;
-    requestForResponse.handleResponse(nResp);
+    if(resp.getDecoderResult().isFailure()) {
+      handleException(resp.getDecoderResult().cause());
+    } else {
+      requestForResponse.handleResponse(nResp);
+    }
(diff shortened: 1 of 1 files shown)


vietj:
can you sign the Eclipse CLA please ?

gmagniez:
Sorry, I did it but I have signed my commit with the wrong email, do I have to resubmit the pull-request?

vietj:
I think you can make a push --force in your own fork and it should be ok

gmagniez:
done

vietj:
all green!

vietj:
one question : can you describe when this case happens precisely ?

gmagniez:
Yes, it is quite simple, it happens when using HttpClient and the response sent from the server contains a total headers higher than 8192 bytes.
The ideal solution would be to handle the error nicely (like I did whithin the merge request), and it would be also great to be able to configure the limit of 8192 bytes. (It is done for HttpServer but not for the HttpClient)

vietj:
we need a test for this case

perhaps also you can provide limit configuration for http client as you suggested ?

vietj:
I made a similar fix here that close the connection for potential pipelined requests on the same connection https://github.com/eclipse/vert.x/commit/a33523add3f74a6edbdf6abcf912893b61f562fc

vietj:
thanks for contribution

PR #1344, opened 2016-03-21
Use LinkedHashMap in copy() to preserve key ordering.

LarsDenBakker:
Right now we're losing key ordering when the JsonObject is copied (for instance when communicating on the event bus). LinkedHashMap should be used to maintain key ordering. It's a tiny change. :)

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/src/main/java/io/vertx/core/json/JsonObject.java
@@ -694,7 +694,7 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
    * @return a copy of the object
    */
   public JsonObject copy() {
-    Map<String, Object> copiedMap = new HashMap<>(map.size());
+    Map<String, Object> copiedMap = new LinkedHashMap<>(map.size());
     for (Map.Entry<String, Object> entry: map.entrySet()) {
       Object val = entry.getValue();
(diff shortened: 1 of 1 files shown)


pmlopes:
Just as a note, there is no requirement that json keys are ordered and it is probably a bad assumption specially once you start consuming json from different sources or languages.

pmlopes:
see: https://github.com/eclipse/vert.x/issues/1292

vietj:
won't fix

sodomizer:
Your "json dict is not ordered wont fix" approach is narrow-minded. What will you lose if you implement this? Nothing. Yes, JSON does not **require** that keys are ordered, but it also does not **restrict** it...and we are talking about internal object representation rather than JSON text format particularly. JsonObject is merely a wrapper (a remarkably convenient one!) for LinkedHashMap which preserves order, so nothing will change for those developers who don't care about key order but that will be great improvement for those who do. For example, for those who also deal with XML where order does matter. You are right about (de)serialization in different languages, it will always be an issue that must be taken into account, but with this change every message **inside** an application will be consistent. If I send _{"x":1,"y":2}_ message I expect to receive exactly same message.
Actually I don't get why have you put HashMap instead of LinkedHashMap in the first place as the latter if simply faster and don't get even more why do you refuse to fix this so badly despite having only pros and no cons or side effects!

PR #1350, opened 2016-03-24
Deserialize JsonObject in a JsonArray consistently when clustered

shaunrader:
This appears to solve [issue 1349](https://github.com/eclipse/vert.x/issues/1349) but I am not sure if this is the most elegant solution to the problem.

Running the test with the first commit shows the issue of the `JsonObject` turning into a `LinkedHashMap` after deserialization.

The fix explicitly creates a JsonObject when we get an untyped object on deserialization. All other paths fallback to the existing default behavior using the `UntypedObjectDeserializer.deserialize`.

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -18,10 +18,15 @@ package io.vertx.core.json;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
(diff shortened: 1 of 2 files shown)


vietj:
is it possible to achieve the same when nested in JsonObject ?

tsegismont:
With this change adds an intermediate string for decoding and all json objects would be built with double parsing:
- when the JsonNode is built
- when the intermediate string is used to create a JsonObject

Besides, if a user type holds an object field, it will be filed with a JsonObject, instead of a Map.

Issue #1345, opened 2016-03-21
Nested zip tests fails on Windows


vietj:
```
Failed tests:
  NestedZipFileResolverTest>AsyncTestBase.lambda$onSuccess$1:622->FileResolverTestBase.lambda$testCacheDirDeletedOnVertxClose$3:197->AsyncTestBase.assertFalse:208 null
  NestedZipFileResolverTest>AsyncTestBase.lambda$onSuccess$1:620->AsyncTestBase.fail:262 java.nio.file.FileSystemException: C:\Users\FTM\Documents\NetBeansProjects\vert.x\.vertx\file-cache-bd9e09f4-0e7a-4400-b8cf-a97a8735ea7c\lib\nested.zip: The process cannot access the file because it is being used by another process.

  NestedZipFileResolverTest.after:60->AsyncTestBase.after:81->FileResolverTestBase.tearDown:52->AsyncTestBase.awaitLatch:592->AsyncTestBase.assertTrue:372 null
  NestedZipFileResolverTest.after:60->AsyncTestBase.after:81->FileResolverTestBase.tearDown:52->AsyncTestBase.awaitLatch:592->AsyncTestBase.assertTrue:372 null
  NestedZipFileResolverTest.after:60->AsyncTestBase.after:81->FileResolverTestBase.tearDown:52->AsyncTestBase.awaitLatch:592->AsyncTestBase.assertTrue:372 null
  NestedZipFileResolverTest.after:60->AsyncTestBase.after:81->FileResolverTestBase.tearDown:52->AsyncTestBase.awaitLatch:592->AsyncTestBase.assertTrue:372 null
  NestedZipFileResolverTest.after:60->AsyncTestBase.after:81->FileResolverTestBase.tearDown:52->AsyncTestBase.awaitLatch:592->AsyncTestBase.assertTrue:372 null
  NestedZipFileResolverTest.after:60->AsyncTestBase.after:81->FileResolverTestBase.tearDown:52->AsyncTestBase.awaitLatch:592->AsyncTestBase.assertTrue:372 null

Tests run: 17, Failures: 8, Errors: 0, Skipped: 0
```

reported by @FainTMako on Window 7

vietj:
same for nested jar:

```
Results :

Failed tests:
[...]

PR #1351, opened 2016-03-27
Fixed SLF4JLogDelegateTest on windows platform due to CR/LF inconsistency

Mart-Bogdan:
Logger tests was failing on windows platform. Error with line end characters.


vietj:
Can you capture the system.lineSeparator in a constant in the test class and reuse it ?

Mart-Bogdan:
ok, I'll update PR

Mart-Bogdan:
Updated the code

Mart-Bogdan:
This issue was fixed with #1345 patch, by @cescoffier's commit 67c7a007222adc2f4919719c161cfaf3dcaf46f5

So, Julien, what do you think, removing **\n** at all, and using contains is better solution for problem?

PR #1352, opened 2016-03-28
Extend Launcher documentation when `CLASSPATH` is set

cescoffier:
Fix issue reported in https://github.com/vert-x3/issues/issues/104

Extend Launcher documentation to indicate that the `CLASSPATH` env var (if set) must contain vert.x core.


vietj:
looks good to me

PR #1353, opened 2016-03-28
Http2 support

vietj:
please review


pom.xml
```diff
@@ -162,9 +170,32 @@
...
+    </dependency>
   </dependencies>
 
   <build>
+
+    <extensions>
+      <extension>
+        <groupId>kr.motd.maven</groupId>
+        <artifactId>os-maven-plugin</artifactId>
+        <version>1.4.1.Final</version>
```
cescoffier: we need to ensure that the resulting jar is OS agnostic. (that's more a note for myself)

src/main/asciidoc/java/http.adoc
```diff
@@ -23,6 +28,34 @@ HttpServerOptions options = new HttpServerOptions().setMaxWebsocketFrameSize(100
 HttpServer server = vertx.createHttpServer(options);
 ----
 
+=== Configuring an HTTP/2 server
+
+Vert.x supports HTTP/2 over TLS `h2` and over TCP `h2c`.
```
cescoffier: Add links or explain

src/main/asciidoc/java/http.adoc
```diff
@@ -23,6 +28,34 @@ HttpServerOptions options = new HttpServerOptions().setMaxWebsocketFrameSize(100
...
+----
+HttpServerOptions options = new HttpServerOptions()
+    .setUseAlpn(true)
+    .setSsl(true)
+    .setKeyStoreOptions(new JksOptions().setPath("/path/to/my/keystore"));
+
+HttpServer server = vertx.createHttpServer(options);
+----
+
+ALPN is a TLS extension that negotiates the protocol before the client and the server start to exchange data.
```
cescoffier: Add link to https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation

src/main/asciidoc/java/http.adoc
```diff
@@ -421,6 +483,8 @@ Non keep-alive connections will be automatically closed by Vert.x when the respo
 Keep-alive connections are not automatically closed by Vert.x by default. If you want keep-alive connections to be
 closed after an idle time, then you configure `link:../../apidocs/io/vertx/core/http/HttpServerOptions.html#setIdleTimeout-int-[setIdleTimeout]`.
 
+HTTP/2 connections will GOAWAY frame before closing the response.
```
cescoffier: `GOAWAY`

src/main/asciidoc/java/http.adoc
```diff
@@ -590,6 +656,93 @@ vertx.createHttpServer().requestHandler(request -> {
...
+----
+int frameType = 40;
+int frameStatus = 10;
+Buffer payload = Buffer.buffer("some data");
+
+// Sending a frame to the client
+response.writeFrame(frameType, frameStatus, payload);
+----
+
+These frames are sent immediatly and are not subject to flow control - when such frame is sent there it may be done
```
cescoffier: immediately

src/main/asciidoc/java/http.adoc
```diff
@@ -590,6 +656,93 @@ vertx.createHttpServer().requestHandler(request -> {
...
+
+==== Server push
+
+Server push is a new feature of HTTP/2 that enables sending multiple responses in parallel for a single client request.
+
+When a server process a request, it can push a request/response to the client:
+
+[source,java]
+----
+HttpServerResponse response = request.response();
```
cescoffier: I don't get the API here.

Why `response.push` async ? Is this because it take time to establish the connection (handshake, negotiation ...) ?

There is a `response.sendFile`. Does that mean that the response must write something before being able to establish a push link with the client ? Is so, what the goal of the `/main.js` ?

src/main/asciidoc/java/http.adoc
```diff
@@ -1150,6 +1389,77 @@ httpServer.requestHandler(request -> {
...
+
+// End the request
+request.end();
+----
+
+If the client does not want to receive a pushed request, it can reset the stream:
+
+[source,java]
+----
+HttpClientRequest request = client.get("/index.html", response -> {
```
cescoffier: Wrong snippet ?

src/main/asciidoc/java/http.adoc
```diff
@@ -1150,6 +1389,77 @@ httpServer.requestHandler(request -> {
...
+  pushedRequest.handler(pushedResponse -> {
+    System.out.println("The response for the pushed request");
+  });
+});
+
+// End the request
+request.end();
+----
+
+When no handler is set, any stream pushed by the server will be automatically reset by the client.
```
cescoffier: What is the error code sent in this case ?

src/main/asciidoc/java/http.adoc
```diff
@@ -1214,6 +1524,185 @@ By default pipe-lining is disabled.
...
+
+The settings can be changed at any time after the connection is established:
+
+[source,java]
+----
+connection.updateSettings(new Http2Settings().setMaxConcurrentStreams(100));
+----
+
+As the remote side should acknowledge on reception of the settings update, it's possible to give a callback
+to be notified of the acknowledgent:
```
cescoffier: acknowledgment

src/main/asciidoc/java/http.adoc
```diff
@@ -1214,6 +1524,185 @@ By default pipe-lining is disabled.
...
+  data.appendByte(i);
+}
+connection.ping(data, pong -> {
+  System.out.println("Remote side replied");
+});
+----
+
+Vert.x will send automatically an acknowledgement when a `PING` frame is received,
+an handler can be set to be called when the connection is pinged:
+
```
cescoffier: The the `ACK` still sent when the handler is set ?

src/main/asciidoc/java/net.adoc
```diff
@@ -797,4 +797,54 @@ NetServerOptions options = new NetServerOptions().
...
+==== OpenSSL engine
+
+The default SSL/TLS engine implementation is provided by the JDK.
+
+The engine implementation can be configured to use https://www.openssl.org[OpenSSL] instead. OpenSSL provides
+better performances and CPU usage than the JDK engine, as well as JDK version independence.
+
+OpenSSL requires to configure `link:../../apidocs/io/vertx/core/net/TCPSSLOptions.html#setSslEngine-io.vertx.core.net.SSLEngine-[setSslEngine]` to `link:../../apidocs/io/vertx/core/net/SSLEngine.html#OPENSSL[OPENSSL]`
+and use http://netty.io/wiki/forked-tomcat-native.html[netty-tcnative] jar on the classpath, as well as the correct
+version of OpenSSL installed on your OS.
```
cescoffier: correct or consistent ? 

Is the tc-native version aligned with the openssl version ?

src/main/asciidoc/java/net.adoc
```diff
@@ -797,4 +797,54 @@ NetServerOptions options = new NetServerOptions().
...
+===== Jetty-ALPN support
+
+Jetty-ALPN is a small jar that overrides a few classes of Java 8 distribution to support ALPN.
+
+The JVM must be started with the _alpn-boot-${version}.jar_ in its `bootclasspath`:
+
+----
+-Xbootclasspath/p:/path/to/alpn-boot${version}.jar
+----
+
```
cescoffier: If I understand correctly, `lib/ext` would not be high enough in the classloader hierarchy. Maybe it needs to be mentioned.

src/main/asciidoc/java/net.adoc
```diff
@@ -797,4 +797,54 @@ NetServerOptions options = new NetServerOptions().
...
+list is available on the http://www.eclipse.org/jetty/documentation/current/alpn-chapter.html[Jetty-ALPN page].
+
+The main drawback is that the version depends on the JVM.
+
+To solve this problem the _Jetty-ALPN_ agent can be use instead. The agent is a JVM agent that will chose the correct
+ALPN version for the JVM running it:
+
+----
+-javaagent:/path/to/alpn/agent
+----
```
cescoffier: Add a link from where we can download the agent.

src/main/java/io/vertx/core/http/Http2Settings.java
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * The settings expose the parameters defined by the HTTP/2 specification, as well as extra settings for
+ * protocol extensions.
+ *
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+@DataObject(generateConverter = true)
+public class Http2Settings {
+
+  public static final long DEFAULT_HEADER_TABLE_SIZE = 4096;
```
cescoffier: Should we have javadoc on these public constants ?

src/main/java/io/vertx/core/http/Http2Settings.java
```diff
@@ -0,0 +1,340 @@
...
+        "maxHeaderListSize must be >= " + Http2CodecUtil.MIN_HEADER_LIST_SIZE);
+    this.maxHeaderListSize = maxHeaderListSize;
+    return this;
+  }
+
+  /**
+   * @return the extra settings used for extending HTTP/2
+   */
+  @GenIgnore
+  public Map<Integer, Long> getExtraSettings() {
```
cescoffier: Should it be converted to JsonObject to be polyglot ?

src/main/java/io/vertx/core/http/HttpVersion.java
```diff
@@ -25,5 +25,5 @@
  */
 @VertxGen
 public enum HttpVersion {
-  HTTP_1_0, HTTP_1_1
+  HTTP_1_0, HTTP_1_1, HTTP_2
```
cescoffier: Should it be 2_0 ?

src/main/java/io/vertx/core/http/StreamResetException.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2011-2013 The original author or authors
```
cescoffier: Outdated copyright year.

src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -274,37 +289,77 @@ void handleResponse(HttpResponse resp) {
...
       throw new IllegalStateException("No response handler");
     }
-    HttpClientResponseImpl nResp = new HttpClientResponseImpl(requestForResponse, this, resp);
+    io.netty.handler.codec.http.HttpVersion nettyVersion = resp.protocolVersion();
+    HttpVersion vertxVersion;
+    if (nettyVersion == io.netty.handler.codec.http.HttpVersion.HTTP_1_0) {
+      vertxVersion = HttpVersion.HTTP_1_0;
+    } else if (nettyVersion == io.netty.handler.codec.http.HttpVersion.HTTP_1_1) {
+      vertxVersion = HttpVersion.HTTP_1_1;
+    } else {
```
cescoffier: Should HTTP2 have a different error here ?

cescoffier: Oh this class is only used for HTTP 1

src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -420,4 +559,19 @@ public void channelRead(ChannelHandlerContext chctx, Object msg) throws Exceptio
...
   }
+
+  @Override
+  public HttpClientConnection connection() {
+    return this;
+  }
+
+  @Override
+  public HttpVersion version() {
+    return null;
```
cescoffier: not `vertxVersion` ?

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
```diff
@@ -16,139 +16,535 @@
...
+      connectionFailed(context, ch, waiter::handleFailure, sslException);
+    }
+
+    private void fallbackToHttp1x(Channel ch, ContextImpl context, HttpVersion fallbackVersion, int port, String host, Waiter waiter) {
+      // Fallback
+      // change the pool to Http1xPool
+      ConnQueue<ClientConnection> http1Queue = qm1x.getConnQueue(address);
+      applyHttp1xConnectionOptions(ch.pipeline(), context);
+      http1Queue.http1xConnected(fallbackVersion, context, port, host, ch, waiter);
+      // Should remove this queue as it may be empty ????
```
cescoffier: any answer to this existential question ?

src/main/asciidoc/java/http.adoc
vietj: it is async because the client may have limited the number of streams a server can send in the settings when establishing the connection, when it happens the push are queued.

the server must push streams before it ends the current response.

it the example:
- handler calls push : this sends a _push promise_ frame to the client
- handler calls send file : it sends the _headers_ frame + the _data_ frames to the client
- the push handler is called, it will send the headers + data frames of `main.js` to the client

probably that data frames of the original response and `main.js` are interleaved and multiplexed.

src/main/asciidoc/java/http.adoc
vietj: at the moment it sends `0` aka _ NO_ERROR_ but perhaps it should send _8_ CANCEL_

src/main/java/io/vertx/core/http/impl/Http2ServerConnection.java
```diff
@@ -0,0 +1,251 @@
...
+      @Override
+      public void handle(AsyncResult<Integer> ar) {
+        if (ar.succeeded()) {
+          synchronized (Http2ServerConnection.this) {
+            int promisedStreamId = ar.result();
+            String contentEncoding = HttpUtils.determineContentEncoding(headers);
+            Http2Stream promisedStream = handler.connection().stream(promisedStreamId);
+            Push push = new Push(promisedStream, contentEncoding, completionHandler);
+            streams.put(promisedStreamId, push);
+            if (maxConcurrentStreams == null || concurrentStreams < maxConcurrentStreams) {
```
cescoffier: maxConcurretnStreams accessed without lock.

src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
```diff
@@ -0,0 +1,629 @@
...
+      if (chunk != null && !chunked && !headers.contains(HttpHeaderNames.CONTENT_LENGTH)) {
+        headers().set(HttpHeaderNames.CONTENT_LENGTH, String.valueOf(chunk.readableBytes()));
+      }
+      write(chunk, true);
+    }
+  }
+
+  private boolean checkSendHeaders(boolean end) {
+    if (!headWritten) {
+/*
```
cescoffier: to be removed ?

src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
```diff
@@ -0,0 +1,629 @@
...
+      checkEnded();
+
+      Context resultCtx = resultHandler != null ? stream.vertx.getOrCreateContext() : null;
+
+      File file = stream.vertx.resolveFile(filename);
+      if (!file.exists()) {
+        if (resultHandler != null) {
+          resultCtx.runOnContext((v) -> resultHandler.handle(Future.failedFuture(new FileNotFoundException())));
+        } else {
+          // log.error("File not found: " + filename);
```
cescoffier: should we do something here ? If not maybe a comment explaining why.

src/main/asciidoc/java/http.adoc
vietj: yes, the ACK is actually sent by Netty HTTP/2 automatically and not by Vert.x . This is just a mean to be aware of the ping, a protocol like gRPC uses the HTTP/2 ping frames and need this callback.

src/main/asciidoc/java/net.adoc
vietj: tc-native is independant, however there is a minimum version to use.

for example the OOTB version on OSX is too old.

src/main/asciidoc/java/net.adoc
vietj: that’s unclear to me :-) if you determine, I will trust you

src/main/java/io/vertx/core/http/Http2Settings.java
vietj: good catch.

most of them are the default spec values.

so I think that’s what we should mention indeed.

cescoffier:
That was a big one.

There are several issue in the copyright year. For the rest it looks good to me.

src/main/java/io/vertx/core/http/Http2Settings.java
vietj: json object cannot use int as key.

that’s the main problem.

src/main/java/io/vertx/core/http/HttpVersion.java
vietj: it is HTTP/2 : http://httpwg.org/specs/rfc7540.html

src/main/java/io/vertx/core/http/impl/ClientConnection.java
vietj: good catch, thanks and I will add tests for these.

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
vietj: soon yes :-)

src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
vietj: I need to figure out.

src/main/asciidoc/java/http.adoc
cescoffier: It should be written in the doc.

vietj: there is a response.sendFile because a response can only be pushed in reaction to an existing request. a server cannot push to the client just like that.

src/main/asciidoc/java/http.adoc
vietj: yes wrong one :-)

src/main/asciidoc/java/net.adoc
vietj: this depends if the tcnative jar is statically or dynamically linked, so I would rather remove the part about OpenSSL installation as it is simply wrong

vietj: I will change to something like  "Using tcnative may require OpenSSL to be installed on your OS depending on the tcnative implementation”

PR #1356, opened 2016-03-29
Support command set in the manifest without main verticle.

### LINKED ISSUE #1346: Default command specified in Main-Command MANIFEST entry not run if no Main-Verticle entry
Vext.x version 3.2.1

In VerxCommandLauncher line 378 there is check for verticle != null only then getDefaultCommand() executed. But may be we not needed any vehicles in manifest at all.

cescoffier:
Fix #1346

Execute the command specified in the manifest if set and not overridden in the command line.


vietj:
looks good to me

PR #1357, opened 2016-03-29
Do not dump the configuration in the log

### LINKED ISSUE #1328: Config should not be printed to log when JSON cannot be parsed
Currently the entire config file (passed to -conf) will be printed to the log if it cannot be correctly parsed as a JSON object on vertx startup. Since the config potentially can contain sensitive information that you don't want printed in a log this is not a good default. 

I suggest vertx don't print the content of the config, rather just an error message telling that the config cannot be parsed.

### LINKED ISSUE #1346: Default command specified in Main-Command MANIFEST entry not run if no Main-Verticle entry

cescoffier:
Fix #1328

Do not display the configuration in the log as it can contain sensitive data.

### DIFF
--- a/src/test/java/io/vertx/core/impl/launcher/commands/RunCommandTest.java
+++ b/src/test/java/io/vertx/core/impl/launcher/commands/RunCommandTest.java
@@ -29,6 +29,7 @@ import java.net.URL;
 import java.nio.file.CopyOption;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
+import java.util.Random;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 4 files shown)


src/test/java/io/vertx/core/impl/launcher/commands/RunCommandTest.java
```diff
@@ -183,6 +183,20 @@ public void testWithConfProvidedInline() throws IOException {
   }
 
   @Test
+  public void testWithBrokenConfProvidedInline() throws IOException {
+    setManifest("MANIFEST-Launcher-Http-Verticle.MF");
+    cli.dispatch(new Launcher(), new String[] {"--conf={\"name\":\"vertx\""});
```
vietj: you could generate a random alpha string for config and check that this particular string is not in the output.

PR #1363, opened 2016-04-03
fixing some formatting, spelling and unicode characters

ahus1:
I found the spelling mistake of -> off. When looking closer in my IDE I also found some unicode long-hyphens where there should have been double ascii hypens on a command line.


cescoffier:
:+1:

PR #1364, opened 2016-04-04
Add address and replyAddress to message on reply timeout to aid in pr…

diabolicallabs:
Add address and replyAddress to message on reply timeout to aid in problem determination.

Signed-off-by: David Bush <david.bush@process-intellect.com>


diabolicallabs:
It's difficult to know what handler is timing out. This expanded message should help.

diabolicallabs:
Oops. Didn't mean to close

cescoffier:
As the reply address is a generated one (UUID), I'm not sure it is useful to write it. However, the initial address, that's a great idea !

Issue #1218, opened 2015-12-04
ClientConnection should be closed upon http RequestTimeout


ddimensia:
If an HttpClientRequest get's in a state where it is no longer receiving data from a server and a request timeout fires then Vertx should force the connection to close so that it can be freed up so another request can be processed.   Most likely since the connection is in an unknown state it shouldn't be reused if keepalive is enabled.

This issue combined with issue #1212 can easily cause an OOM or stalled HTTPClient situation on high throughput servers.

testn:
If `HttpClientOptions.setIdleTimeout` is set, it should provide exactly what you want I assume?

ddimensia:
```HttpClientOptions.setIdleTimeout``` is not really the same and I believe is intended to handle unused keep-alive connections similar to Apache's [KeepAliveTimeout](https://httpd.apache.org/docs/2.4/mod/core.html#keepalivetimeout) setting.  You could try to use ```setIdleTimeout``` to handle this particular problem but it would be inadequate in some situations as it wouldn't handle slow senders and isn't a per-request setting.

Specifically, the requestTimeout/connection issue is something we saw with periodic packet loss on a high throughput api call with low timeouts.  Basically what can happen, is that if you have a short timeout, say 250ms and a high request rate you can get into a situation where a slow sender or waiting for packet retransmits can exceed the request timeout.  If the request rate is high enough this will cause requests to load up on the pool and force requests into the wait queue.  Since the ConnectionManager won't view the timed out connections as 'available' until it either sees a responseEnd or channel close the HttpClient will effectively block any new requests being handled.  This combined with bugs #1219 and #1212 and high load can cause a server to go into a death spiral where it can never catch up.  

I plan to provide a PR for this shortly, I plan on making the behavior configurable with it defaulting to the existing behavior.  The downside of closing the connection is that it has impacts on pipelined requests and the server you're connecting to so I could see how some may not want this behavior.

ddimensia:
I finally got back around to working on a PR and just updated to the latest code that includes HTTP2 support.  With the latest code I found a few problems...

[...]

PR #1370, opened 2016-04-07
Allow configuration to force close sockets on HttpClientRequest exceptions.

ddimensia:
Relates to #1218 

I'm not sure if this is the best approach as it exposes the ability to close a socket connection via a HttpClientRequest object.  In some cases this may actually be desirable but I could see the incorrect use of the method causing issues also.

I added configuration in HttpClientOptions to enable this feature, it's off by default to maintain backwards compatibility but ultimately I believe the default behavior should be to have it enabled.  See the associated issue for an explanation of the problems caused by the default behavior.

It should be noted that we're running a patched version a 2.1.x that includes effectively the same change (the HTTP2 changes caused it to be implemented differently) in our production environment (multiple apps with high request volumes) and haven't seen any issues.

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -153,6 +153,32 @@ Set the key store as a buffer
 +++
 |===
 
+[[GoAway]]
+== GoAway
+
+++++
+ A  frame.
+++++
+'''
+
+[cols=">25%,^25%,50%"]
+[frame="topbot"]
+|===
+^|Name | Type ^| Description
+|[[debugData]]`debugData`|`Buffer`|
++++
+Set the additional debug data
++++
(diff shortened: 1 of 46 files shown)


vietj:
that's something you cannot do on your own with an exception handler ?

ddimensia:
Well, currently there is no method to force close a connection, just adding a close method to the request interface would make that a possibility.  However, if you read the explanation in the associated issue you would see that the current behavior has some bad side effects and I'm unaware of any advantages.

ddimensia:
Updated with latests master to deal with merge conflicts.

vietj:
hi, I've added this PR that gives more connection access for HTTP/1 https://github.com/eclipse/vert.x/pull/1424 I think you could implement this feature using the HttpConnection

PR #1374, opened 2016-04-11
Hostname resolver

vietj:
ready for review, please review


vietj:
@pmlopes @cescoffier or any other Vert.x committer please review

src/main/java/io/vertx/core/dns/HostnameResolverOptions.java
```diff
@@ -0,0 +1,285 @@
...
+package io.vertx.core.dns;
+
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonObject;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Configuration options for Vert.x hostname resolver.
```
cescoffier: Add something like: It is used to revolve IP address only (A record only).

src/main/java/io/vertx/core/spi/metrics/DatagramSocketMetrics.java
```diff
@@ -43,8 +43,9 @@
    * Called when a socket is listening. For example, this is called when an http or net server
    * has been created and is listening on a specific host/port.
    *
+   * @param localName
    * @param localAddress the local address the net socket is listening on.
    */
-  void listening(SocketAddress localAddress);
+  void listening(String localName, SocketAddress localAddress);
```
cescoffier: Breaking change.
Must be documented.

src/main/java/io/vertx/core/metrics/impl/DummyVertxMetrics.java
```diff
@@ -319,7 +319,7 @@ public boolean isEnabled() {
   protected class DummyDatagramMetrics implements DatagramSocketMetrics {
 
     @Override
-    public void listening(SocketAddress localAddress) {
+    public void listening(String localName, SocketAddress localAddress) {
```
cescoffier: Breaking change
Must be documented.

src/main/java/io/vertx/core/dns/HostnameResolverOptions.java
vietj: it does AAAA also, but yes indeed.

src/main/java/io/vertx/core/spi/metrics/DatagramSocketMetrics.java
vietj: that's SPI, but should be documented anyway

vietj: where is the 3.3 page for this ?

cescoffier: https://github.com/vert-x3/wiki/wiki/3.3.0-Breaking-Changes

vietj: maybe we should have in the doc a difference between API and SPI ? so it is clear that SPI changes won't affect normal users.

PR #1376, opened 2016-04-13
Avoid accessor method for `OrderedExecutorFactory::log`

circlespainter:
The accessor method is not an issue per-se but it causes the Quasar Java agent in the forthcoming release, when used, to print an harmless but annoying error at vert.x boot:

```
ERROR: Unable to instrument class io/vertx/core/impl/OrderedExecutorFactory$OrderedExecutor co.paralleluniverse.fibers.instrument.UnableToInstrumentException: Unable to instrument io/vertx/core/impl/OrderedExecutorFactory$OrderedExecutor#lambda$new$261()V because of synchronization
```

(See https://github.com/puniverse/quasar/issues/160#issuecomment-205977756 and following for details)

An easy workaround for the user is to run the agent with the `=m` suffix ("allow monitors").

### DIFF
--- a/src/main/java/io/vertx/core/impl/OrderedExecutorFactory.java
+++ b/src/main/java/io/vertx/core/impl/OrderedExecutorFactory.java
@@ -31,7 +31,7 @@ import java.util.concurrent.Executor;
  * @version <tt>$Revision$</tt>
  */
 public class OrderedExecutorFactory {
-  // It must be package-private rather than `private` in order to avoid the generation of an accessor, which would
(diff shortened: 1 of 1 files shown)


vietj:
@circlespainter you need to sign the Eclipse CLA and sign-off your commit

circlespainter:
Sorry, it should be OK now.

vietj:
can you add a comment on the package protected modifier that it is important for quasar ?

circlespainter:
Done.

PR #1381, opened 2016-04-14
Add host verification to NetClient

dan-lind:
Add host verification to NetClient
Added missing tests for host verification to HttpTLSTest

Please note breaking change in NetSocket.java
public NetSocket upgradeToSsl(Handler<Void> handler);
changes to
public NetSocket upgradeToSsl(String host, int port, Handler<Void> handler);

Signed-off-by: Daniel Lindberg <daniel.lindberg@omegapoint.se>


src/main/java/io/vertx/core/net/NetSocket.java
```diff
@@ -217,7 +217,7 @@ default NetSocket sendFile(String filename, long offset, Handler<AsyncResult<Voi
    * @return a reference to this, so the API can be used fluently
    */
   @Fluent
-  NetSocket upgradeToSsl(Handler<Void> handler);
+  NetSocket upgradeToSsl(String host, int port, Handler<Void> handler);
```
vietj: breaking API change + javadoc is not updated

src/test/java/io/vertx/test/core/TLSCert.java
```diff
@@ -164,7 +164,26 @@ TrustOptions getClientTrustOptions() {
     KeyCertOptions getClientKeyCertOptions() {
       return new PemKeyCertOptions().setKeyPath("tls/client-key.pem").setCertPath("tls/client-cert-ca.pem");
     }
-  };
+  },
+
+    FAKE() {
```
vietj: the procedure for producing the fake-server-keystore.jks should be documented in _ssl.txt_

vietj:
this features provided for generic TCP clients, relies on the HTTPS rules for host verification, that is a problem.

vietj:
I think this is the main problem with this feature, is that it is only for HTTPS rules. LDAP has other rules for instance. If someone would be an LDAP client on top of NetClient he would use the LDAP rules which are different, etc...

dan-lind:
Hmm, yeah I see your point. Looking at the sun.security.ssl.X509TrustManagerImpl.checkIdentity() it looks like the only two available algoritms are HTTPS and LDAP/LDAPS. But I guess we currently have no idea of knowing if a NetClient is an LDAP client or not?

vietj:
beyond that the client might have its own way for validating the certificate.

dan-lind:
Any thoughts on how to work around that? It would be nice if we could provide an OOTB way to use the HTTPS rules for host verification if you are using a plain TCP client rather than LDAP or whatnot. As it stands now you would basically have to fork master and apply something along the lines of this PR.

dan-lind:
Perhaps a simple solution would be to disable it by default (rather than enable like the HTTP client). 
Then rename the method to something like NetClientOptions.setVerifyHostUsingHttpsRules() and clearly state in the documentation that the hostname will be verified according to HTTPS rules.

vietj:
I would say that in SSLHelper the boolean verifyHost would be changed to a String that is the identifier of the verification algorithm.

HttpClientOption would set it to HTTPS if verifyHost == true

NetClientOptions would have then a property String hostnameVerifier that could be either HTTPS or LDAPS .

dan-lind:
I can fix that. Only problem is that I don't really know how to test the LDAP verification...

vietj:
do it this way, test that "HTTPS" string works, it's a feature of the SSLEngine anyway, so we would test this feature actually. The important point is that we allow the user to configure verification in general and that it works for HTTPS that is an important use case.

dan-lind:
Besides updating the javadoc for NetSocket, should the breaking change be documented in some special way?

vietj:
there should be no breaking changes

dan-lind:
Probably need to add an overloaded upgradeToSsl then, the current method breaks if you enable host verification since it doesn't pass the hostname to the ssl engine.

vietj:
the current should reuse the host/port specified when the NetSocket is created instead

dan-lind:
I think I have made all the necessary changes now, please review

vietj:
yup looks good to me 👍

vietj:
have you signed the eclipse CLA @dan-lind ?

dan-lind:
Yep

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -410,9 +407,9 @@ private SslHandler createHandler(SSLEngine engine, boolean client) {
           break;
         }
       }
-    } else if (verifyHost) {
+    } else if (!endpointIdentificationAlgorithm.isEmpty()) {
```
vietj: endpointIdentificationAlgorithm could be null with a malformed option value in NetClientOptions : add a non null ckeck too

dan-lind: Perhaps endpointIdentificationAlgorithm should be an enum instead? Would remove both null and invalid option values.

vietj: no need to add an enum for this

dan-lind: Added Objects.requireNonNull for the setter, I think this is preferred over null checking everywhere the value might be used.

vietj:
thanks for the contribution!

PR #1384, opened 2016-04-20
Additional http methods

### LINKED ISSUE #1339: Additional HTTP methods support
Vert.x 2 had support for any extended HTTP method as methods are accepted as String. Vert.x has now limited itself to work only with standard HTTP methods and this has been a limitation for our application to migrate to Vert.x 3.

As per the [discussion](https://groups.google.com/forum/#!msg/vertx/PoJMC_wEsNo/-k3hE1N_AwAJ) with @vietj , it is expected in 3.3. Adding this issue for tracking convenience.

vietj:
fixes #1339

please review

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientRequestPushPromise.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientRequestPushPromise.java
@@ -37,6 +37,7 @@ class HttpClientRequestPushPromise extends HttpClientRequestBase {
   private final Http2ClientConnection conn;
   private final Http2ClientConnection.Http2ClientStream stream;
   private final HttpMethod method;
+  private final String rawMethod;
   private final String uri;
   private final String host;
   private final MultiMap headers;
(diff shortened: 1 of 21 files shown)


src/main/java/io/vertx/core/http/impl/HttpClientRequestPushPromise.java
```diff
@@ -116,6 +117,16 @@ public HttpMethod method() {
   }
 
   @Override
+  public String getRawMethod() {
+    throw new UnsupportedOperationException("todo");
```
cescoffier: Write todo the same "way" in both places.

cescoffier: Is it a TODO or a not supported feature.

cescoffier:
I think it should be documented somewhere.

src/main/java/io/vertx/core/http/impl/HttpClientRequestPushPromise.java
vietj: oups

vietj:
fair enough

PR #1397, opened 2016-04-29
Named worker pools

vietj:
This PR handles:

* https://github.com/eclipse/vert.x/issues/1332
* https://github.com/eclipse/vert.x/issues/1333
* https://github.com/vert-x3/issues/issues/83
* https://github.com/eclipse/vert.x/issues/1246

It also provides "named" thread pool (as discussed on https://groups.google.com/forum/#!topic/vertx-dev/-J9ISW1x8eY)

* creation of extra named thread pool for specific purposes configured with pool size and max execute time
* specify a specific thread pool in Verticle deployment options
* a new SPI for ThreadPoolMetrics
* the collection of _worker_ thread metrics (worker, executeBlocking, internal blocking, named thread pools)

### DIFF
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -481,6 +481,33 @@ public interface Vertx extends Measured {
   EventLoopGroup nettyEventLoopGroup();
 
   /**
+   * Like {@link #createWorkerExecutor(String, int)} but with the {@link VertxOptions#setWorkerPoolSize} {@code poolSize}.
+   */
+  WorkerExecutor createWorkerExecutor(String name);
+
+  /**
+   * Like {@link #createWorkerExecutor(String, int, long)} but with the {@link VertxOptions#setMaxWorkerExecuteTime} {@code maxExecuteTime}.
+   */
+  WorkerExecutor createWorkerExecutor(String name, int poolSize);
+
+  /**
+   * Create a named worker executor, the executor should be closed when it's not needed anymore to release
+   * resources.<p/>
+   *
+   * This method can be called mutiple times with the same {@code name}. Executors with the same name will share
+   * the same worker pool. The worker pool size and max execute time are set when the worker pool is created and
+   * won't change after.<p>
+   *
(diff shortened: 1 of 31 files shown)


testn:
should integrate with #1255

cescoffier:
👍

src/main/java/io/vertx/core/Vertx.java
```diff
@@ -481,6 +481,33 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
   EventLoopGroup nettyEventLoopGroup();
 
   /**
+   * Like {@link #createWorkerExecutor(String, int)} but with the {@link VertxOptions#setWorkerPoolSize} {@code poolSize}.
+   */
+  WorkerExecutor createWorkerExecutor(String name);
+
+  /**
+   * Like {@link #createWorkerExecutor(String, int, long)} but with the {@link VertxOptions#setMaxWorkerExecuteTime} {@code maxExecuteTime}.
```
purplefox: Wrong javadoc

src/main/java/io/vertx/core/Vertx.java
```diff
@@ -481,6 +481,33 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
...
+   */
+  WorkerExecutor createWorkerExecutor(String name, int poolSize);
```
purplefox: As the method doesn't always create a new executor, but returns an existing one if it already exists, then perhaps it should be called workerExecutor() (or getOrCreateWorkerExecutor) instead?

vietj: my thinking is "createSharedWorkerExecutor" like we do have in JDBCClient#createShared

purplefox: +1

testn:
No plan to make event loop to be nameable as well?

vietj:
No firm plans per se, I think somebody opened a request for naming Vert.x threads, so they can be identified when several vertx instances are created in the same VM. My opinion is that it can be useful in this case.

PR #1405, opened 2016-05-04
Support HttpClient access to https via proxy 

### LINKED ISSUE #1361: Support HttpClient access to https via proxy
(please note that this description applies to HTTP 1.0/1.1, I am not sure how proxies will work with HTTP2)

When running applications in a network that does not have direct internet access but have e.g. a squid proxy, it is currently not possible to access https urls, since that requires a CONNECT tunnel request to the proxy, upgradeToSsl and then a http request to the origin server on the NetSocket.

It is also necessary to enable Proxy authentication for the CONNECT request since the proxy might require that, which is a different account/password than authentication on the origin site.

the section explaining the CONNECT method, but that is a bit brief: https://tools.ietf.org/html/rfc7231#section-4.3.6

an example request flow is the following:

client connects to proxy.example.com:3128
`CONNECT www.example.org:443 HTTP/1.1`
`Host: www.example.org:443`

server replies with
`HTTP/1.0 200 Connection established`

client and server do ssl upgrade
client sends on ssl netsocket
`GET / HTTP/1.1`
`Host: www.example.org`
`Connection: close`

server replies with http response

alexlehm:
fixes #1361


alexlehm:
sorry, had to redo the last few commit msgs since I forgot the signed-off footer

src/main/java/io/vertx/core/http/package-info.java
```diff
@@ -1527,6 +1527,21 @@
...
+ * The proxy can be configured in the {@link io.vertx.core.http.HttpClientOptions} by setting hostname, port and optionally
+ * username and password. 
+ *
+ * Here's an example:
+ *
+ * [source,$lang]
+ * ----
+ * {@link examples.HTTPExamples#example58}
+ * ----
+ *
```
vietj: add a note about the fact that the jar io.netty:netty-handler-proxy must be on the classpath for this feature, as the jar is in <optional>true</optional> config in the pom

src/main/java/io/vertx/core/http/package-info.java
```diff
@@ -1527,6 +1527,21 @@
  * When using a WebSocket as a write stream or a read stream it can only be used with WebSockets connections that are
  * used with binary frames that are no split over multiple frames.
  *
+ * === Using a HTTP proxy for HTTPS connections
+ *
+ * The http client supports accessing https server via a HTTPS proxy (HTTP CONNECT method), e.g. Squid. This is supported
+ * for HTTP/1.x and HTTP/2.
```
vietj: clarify that only HTTP/1 is supported to the proxy but it can connect to HTTP/1.x and HTTP/2 hosts

alexlehm:
changed the address to be sent to proxy to not be resolved at all, changed unit test for that
small doc fixes

vietj:
thanks!

alexlehm:
well, thanks for clearing up the resolve issue for me

PR #1408, opened 2016-05-05
Author tag removal and reformat

jponge:
`@author` tags tend to create a sense of ownership / authority over files, which is harmful in opensource contexts.

Historical context regarding Apache Cocoon: http://bluxte.net/archives/000109.html

I also applied basic source code formatting along the way.


jponge:
Note: I've grown uncomfortable with `@author` tags over the years, I don't think it encourages participation and rather delimit territories within projects.

I think Vert.x would benefit from removing `@author` tags, hence this pull-request to open the discussion. Feel-free to use it as a basis for debates.

cescoffier:
Actually, at Apache we often use `@author the $project team`.

cescoffier:
If this is applied, we should also change all stack projects.

jponge:
Apache committers grant a license to the Foundation (unlike Eclipse), so `@author Project` is doable there.

cescoffier:
That's right. 

Do you know if the Eclipse development rules say something about this ?

jponge:
I don't think so.

What is for sure is that they prefer clean copyright statements with initial contributors and subsequent contributors claiming copyright (see https://eclipse.org/legal/copyrightandlicensenotice.php, although listing contributors at the bottom is more or less like having `@author` tags).

If you claim something like "Copyright the original authors" I think that:

1. for sure it applies to who did commits, but
2. it is not very strong from a legal POV, something like "Copyright (c) Year Red Hat" + other "Copyright (c) Year Foo Bar" lines is better I think.

Legalese apart, my point was really that these tags tend to create territories, and in general contributors don't dare adding their own `@author` tag.

PR #1413, opened 2016-05-08
Issues/#1361 add another unit test

alexlehm:
Issues/#1361 add another unit test:

The unit test previously tested if localhost is resolved or not (which shows the issue we had), however I prefer to have a test where the dns resolution really fails it is being done.
The unit test is quite simple, the one general thing I had to change in the TLS test implementation is to support clientVerifyHost(false).


vietj:
thanks Alex

PR #1417, opened 2016-05-11
Allow java.time.Instant in JsonObject/JsonArray put as java.lang.Object

pmlopes:
Fixes vert-x3/issues#107

### DIFF
--- a/src/test/java/io/vertx/test/core/JsonObjectTest.java
+++ b/src/test/java/io/vertx/test/core/JsonObjectTest.java
@@ -1626,6 +1626,8 @@ public class JsonObjectTest {
     Object instant = Instant.now();
     JsonObject jsonObject = new JsonObject();
     jsonObject.put("instant", instant);
+    // assert data is stored as String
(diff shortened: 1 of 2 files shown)


src/test/java/io/vertx/test/core/JsonObjectTest.java
```diff
@@ -1622,6 +1622,13 @@ public void testJsonObjectEquality2() {
   }
 
   @Test
+  public void testPutInstantAsObject() {
+    Object instant = Instant.now();
+    JsonObject jsonObject = new JsonObject();
+    jsonObject.put("instant", instant);
```
vietj: need to test we can retrieve "instant" and its type / expected value

pmlopes: You shouldn't since instants are converted to string according to the rfc, if you want it automatically casted then you must do getInstant(key)

vietj: we need to test then we get the expected string value from the json object

Issue #1415, opened 2016-05-10
redeploy="ant_pattern" incorrectly checks path in Windows


msanguineti:
As it is right now, passing something like `--redeploy="src/**/*.java"` has no effect on Windows machines, i.e. no file is ever matched against this pattern since the pattern tokenizer fails to break the pattern into tokens due to the fact that `/` is not a file separator on Windows.

Rather than modifying either the tokenizer or the matching algorithm, a possible solution would be to store the patterns (called includes, see `io/vertx/core/impl/launcher/commands/Watcher.java
` line `77`) after converting between *nix and Windows file separator char, if needed.

I believe a better solution would be to enforce a file separator char on the pattern, i.e. refuse to accept something like `--redeploy="src\**\*.java"`, only accept path patterns separated by `/` and then replace this with the correct OS file separator, if needed.
Ant has `/` as its default file separator character but allows you to change it. I don't think this is necessary. An "enforce and convert" should be all that is needed in this case.

PR #1419, opened 2016-05-11
Fixed issue #1415 

msanguineti:
and also avoid checking for case in Windows file/folders names

Signed-off-by: msanguineti <mirco@sanguineti.gr>


msanguineti:
Hello @cescoffier,

I see that you fixed #1415 but may I suggest to have a look at my code since I take care of allowing OS/X and *nix to pass Windows style path (it might be important in mixed environments where developers write scripts in either systems) but more importantly I do not check for case sensitivity in Windows files/folders when the Watcher tries to match a change. Also I wrote a couple of tests for those cases.

Regards,
Mirco

cescoffier:
Can you rebase your commits on mine to see the difference ?

msanguineti:
Hi @cescoffier 

I did the rebase.

cescoffier:
Looks good to me.

+1

src/test/java/io/vertx/core/impl/launcher/commands/WatcherTest.java
```diff
@@ -175,6 +171,58 @@ public void testFileAdditionAndModificationInDirectory() throws IOException, Int
...
+    // create a file to be matched against "FOO.bar" pattern
+    File file = new File(root, "fOo.BAr");
+    file.createNewFile();
+    
+    if(ExecUtils.isWindows()) {      
+      // undeployment followed by redeployment. Windows is not case sensitive
+      waitUntil(() -> undeploy.get() == 1 && deploy.get() == 2);
+    }
+    else {
+      Thread.sleep(500);
```
vietj: why do we have this pause here ?

vietj:
I see lot of thread sleep in the tests. can we do without these ? in CI (cloudbees), such tests can fail because it runs slower than on our dev machines. so I'm not really a fan of having potential extra failures in the CI testsuite.

cescoffier:
The watcher tests were / are passing on Cloudbees so, it should be fine. However, I agree we should find a different way to deal with these Thread.sleep. On way would be to use awaitility (https://github.com/jayway/awaitility) - a (test) library that relax the time pressure on the test and provide a very neat API. It's the lib I use in the other modules. However, it would require a CQ.

msanguineti:
Hi, I've practically copied and pasted the code from Clement's tests in the same class. Just changed the conditions, obviously.

On 30 May 2016 09:43:41 EEST, Clement Escoffier <notifications@github.com> wrote:
>The watcher tests were / are passing on Cloudbees so, it should be
>fine. However, I agree we should find a different way to deal with
>these Thread.sleep. On way would be to use awaitility
>(https://github.com/jayway/awaitility) - a (test) library that relax
>the time pressure on the test and provide a very neat API. It's the lib
>I use in the other modules. However, it would require a CQ.
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/eclipse/vert.x/pull/1419#issuecomment-222422456

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #1420, opened 2016-05-14
add exit codes in the launcher commands

### LINKED ISSUE #1401: Vert.x command line tools/launcher should return an error code when a failure occurs
All Vert.x command line operations currently always return a "0" exit value, even if it is known the command did not run successful.

For automated deployment tooling (e.g. puppet scripts) it is very useful to return a non-zero exit value. This will allow tooling to check the exit value, stop executing when one command fails, and provide better feedback on which step a failure occurs.

### LINKED ISSUE #1402: The Vert.x "stop" command should check if native operations complete successful
The Vert.x stop command currently does not check if the underlying native commands complete successful. E.g. when the "kill" process fails due to a "Operation not permitted" failure, the vertx stop command does not report this problem. Instead, it reports the Vert.x application has been stopped (but it would likely still be running).

Native commands executed during the stop command should be checked for the exit value. When a failure occurs, any output of native processes should be shown in the vertx command process output to get more feedback on why the command did not run successful.

cescoffier:
Fix #1401 - exit the process with a non 0 status code when an error is detected
Fix #1402 - the stop command should propagate the exit code from the stopped process

Notice that this support is not enabled when the redeploy mode is enabled.

### DIFF
--- a/src/main/asciidoc/java/index.adoc
+++ b/src/main/asciidoc/java/index.adoc
@@ -1576,3 +1576,15 @@ overriding `link:../../apidocs/io/vertx/core/Launcher.html#afterStartingVertx-io
 `link:../../apidocs/io/vertx/core/impl/launcher/VertxCommandLauncher.html#getDefaultCommand--[getDefaultCommand]`
 * add / remove commands using `link:../../apidocs/io/vertx/core/impl/launcher/VertxCommandLauncher.html#register-java.lang.Class-[register]`
 and `link:../../apidocs/io/vertx/core/impl/launcher/VertxCommandLauncher.html#unregister-java.lang.String-[unregister]`
+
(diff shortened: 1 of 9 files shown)


vietj:
this is a nice addition, however it should be added in the main documentation that exit codes are now meaningful.

cescoffier:
The only meaningful thing is: it does not return 0 when an error is caught, which should have been the default. 

The error code used right now are likely going to change. I've a spec about that, and obviously I can't find it anymore... So right now, it should only check for status not 0 (which is what puppet, chef and ansible are expecting when running a command).

Notice that, these status code are not meaningful on windows too.

cescoffier:
Added documentation and fix the status code to follow some good practices.

PR #1424, opened 2016-05-16
Implement HttpConnection for HTTP/1.1 client and server

vietj:
This provides a limited impl of HttpConnection for HTTP/1.1 : close / exceptionHandler and closeHandler . These operations are useful for dealing with the underlying connection, for instance when writing a remote proxy.

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpConnection.java
+++ b/src/main/java/io/vertx/core/http/HttpConnection.java
@@ -17,7 +17,6 @@
 package io.vertx.core.http;
 
 import io.vertx.codegen.annotations.Fluent;
-import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.Nullable;
 import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.AsyncResult;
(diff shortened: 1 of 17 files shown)


testn:
Do you think that it could be useful to have another handler when the connection is released from the request/response?

src/main/java/io/vertx/core/http/HttpConnection.java
```diff
@@ -49,13 +55,15 @@ default HttpConnection goAway(long errorCode, int lastStreamId) {
   }
 
   /**
-   * Send a go away frame to the remote endpoint of the connection.<p/>
-   *
+   * Send a go away frame to the remote endpoint of the connection.
+   * <p/>
    * <ul>
    *   <li>a {@literal GOAWAY} frame is sent to the to the remote endpoint with the {@code errorCode} and {@@code debugData}</li>
```
alexlehm: typo with @@code

alexlehm:
looks ok for me
I found one javadoc typo

michel-kraemer:
+1

vietj:
@testn what's your use case ? isn't the endHandler sufficient for this ?

testn:
The key issue is that the response can come before the request ends in case of error. When that happens, the connection cannot be reused until the request is done.

vietj:
I see what you mean @testn .

for 3.3 we will not have time to do such thing I think.

what's your use case btw ?

PR #1429, opened 2016-05-20
Add SSL options to delegate key and trust store configuration to Java Security Providers.

ellebrecht:
See also: https://groups.google.com/forum/?fromgroups#!topic/vertx/q5zLoEWMH1U

This change adds new SSL options that allow configuration of key and trust stores to be delegated to named algorithms implemented by Java Security Providers. These might be standard algorithms or custom classes plugged in via the standard Java security SPI.

My specific use case was to enable dynamic key/certificate updates in a HTTPS server using pre-existing custom providers. It might be helpful to anyone with key/trust handling requirements beyond the static file support currently offered.

### DIFF
--- a/src/main/java/io/vertx/core/net/SecurityProviderOptions.java
+++ b/src/main/java/io/vertx/core/net/SecurityProviderOptions.java
@@ -1,0 +1,105 @@
+package io.vertx.core.net;
+
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Key or trust store options delegating configuration of private key and/or certificates to previously setup
+ * <a href="https://docs.oracle.com/javase/8/docs/api/java/security/Provider.html">Java Security Providers</a>.
+ * <p>
+ * This can be used for key and trust stores. The managerAlgorithm property references the algorithm name of
+ * the chosen Provider service, whose KeyManagerFactory or TrustManagerFactory implementation is then used.<br/>
+ * Additionally, a KeyStore implementation can be used in both cases by setting the keyStoreAlgorithm property
+ * accordingly.
+ * <p>
+ * This may be useful in cases where Vert.x should use existing custom code based on the java.security API, e.g.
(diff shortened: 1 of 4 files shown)


vietj:
I was actually thinking at something simpler that allows you to come up with your specific use case and allow others to do their own if they want to : https://github.com/eclipse/vert.x/tree/keystore-load-refactor

please have a look and check that it allows you to do so.

vietj:
Hi @ellebrecht any chance you can provide feedback ?

ellebrecht:
Hi Julien, thanks for the fast change, works nicely if I subclass e.g. JksOptions. 
ATM, I cannot directly subclass KeyCertOptions/TrustOptions because the setters in TCPSSLOptions all work with internal implementations only. If there is no special reason for this, might be worth changing the setters to the interface classes. 
Cheers

vietj:
I don't understand what you mean, can you clarify ?

ellebrecht:
You added factory methods to interfaces KeyCertOptions and TrustOptions for custom code to inject their KeyManagerFactory/TrustManagerFactory. 
To work with custom implementations of those interfaces, instances would need to be set via TCPSSLOptions. However, currently there are only setters for the concrete implementations of KeyCertOptions/TrustOptions that come with vertx-core. 
Generic versions of those setters would allow custom implementations subclassing the interfaces directly, rather than subclassing one of the existing implementations.
Here's a branch to illustrate: https://github.com/ellebrecht/vert.x/tree/generic-ssl-options
Hope I could clarify.

vietj:
ah ok, yes you mean that we need to add a generic setter setKeyCertOptions(KeyCertOptions) , right ?

ellebrecht:
Yes, that'd be perfect, and also one for setTrustOptions(TrustOptions).
You can just merge the generic-ssl-options branch from above, if you like.
Cheers

vietj:
sure

vietj:
note that it would not work in polyglot

PR #1434, opened 2016-05-23
Enable adding netty logging handler to log HTTP requests/responses - …

### LINKED ISSUE #1262: Enable adding netty logging handler to log HTTP requests/responses (for debugging)
```
if (...)
   pipeline.addLast("log", new LoggingHandler())
```

Those who ever needed to debug Netty know that LoggingHandler is one of the first instruments.

vietj:
…fixes #1262

### DIFF
--- a/src/main/java/io/vertx/core/net/package-info.java
+++ b/src/main/java/io/vertx/core/net/package-info.java
@@ -290,6 +290,40 @@
  *
  * By default, multiple connection attempts are disabled.
  *
+ * [[logging_network_activity]]
+ * === Logging network activity
+ *
+ * For debugging purposes, network activity can be logged:
+ *
+ * [source,$lang]
+ * ----
+ * {@link examples.NetExamples#exampleNetworkActivityLoggingOnServer}
+ * ----
+ *
+ * for the client
+ *
+ * [source,$lang]
+ * ----
+ * {@link examples.NetExamples#exampleNetworkActivityLoggingOnClient}
+ * ----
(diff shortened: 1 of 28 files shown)


src/main/java/io/vertx/core/net/package-info.java
```diff
@@ -290,6 +290,40 @@
...
+ * ----
+ *
+ * Network activity is logged by Netty with the `DEBUG` level and with the `io.netty.handler.logging.LoggingHandler`
+ * name. Netty will try to locate the following logger implementations, in the following order:
+ *
+ * - Slf4j
+ * - Log4j
+ * - JDK
+ *
+ * The presense of the slf4j or log4j classes on the classpath is enough to pick up the logging implementation.
```
cescoffier: Add a sentence that this resolution is independent from the vert.x log delegate (that can be very confusing actually).

src/main/java/io/vertx/core/net/package-info.java
```diff
@@ -290,6 +290,40 @@
...
+ *
+ * The logger implementation can be forced to a specific implementation by setting Netty's internal logger implementation directly:
+ *
+ * [source,java]
+ * ----
+ * // Force logging to SLF4J
+ * InternalLoggerFactory.setDefaultFactory(Log4JLoggerFactory.INSTANCE);
```
cescoffier: Should we use the a factory from the vert.x log delegate ?

PR #1441, opened 2016-05-27
Metrics for endpoints + make dummy metrics singleton instances

vietj:
Provides HttpClient endpoint monitoring providing per endpoint queing / connections / usage

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/ClientConnection.java
@@ -74,6 +74,7 @@ class ClientConnection extends ConnectionBase implements HttpClientConnection, H
   private final String host;
   private final int port;
   private final Http1xPool pool;
+  private final Object endpointMetric;
   // Requests can be pipelined so we need a queue to keep track of requests
   private final Queue<HttpClientRequestImpl> requests = new ArrayDeque<>();
   private final Object metric;
@@ -89,7 +90,7 @@ class ClientConnection extends ConnectionBase implements HttpClientConnection, H
   private boolean paused;
   private Buffer pausedChunk;
 
-  ClientConnection(HttpVersion version, HttpClientImpl client, Channel channel, boolean ssl, String host,
+  ClientConnection(HttpVersion version, HttpClientImpl client, Object endpointMetric, Channel channel, boolean ssl, String host,
                    int port, ContextImpl context, Http1xPool pool, HttpClientMetrics metrics) {
     super(client.getVertx(), channel, context, metrics);
     this.client = client;
@@ -100,6 +101,7 @@ class ClientConnection extends ConnectionBase implements HttpClientConnection, H
     this.metrics = metrics;
     this.metric = metrics.connected(remoteAddress(), remoteName());
     this.version = version;
+    this.endpointMetric = endpointMetric;
   }
 
   @Override
@@ -247,7 +249,7 @@ class ClientConnection extends ConnectionBase implements HttpClientConnection, H
       handshaker.finishHandshake(channel, response);
       context.executeFromIO(() -> {
         log.debug("WebSocket handshake complete");
-        webSocket.setMetric(metrics().connected(metric(), webSocket));
+        webSocket.setMetric(metrics().connected(endpointMetric, metric(), webSocket));
         wsConnect.handle(webSocket);
       });
     }
@@ -391,8 +393,8 @@ class ClientConnection extends ConnectionBase implements HttpClientConnection, H
       for (HttpClientRequestImpl req: requests) {
         metrics.requestReset(req.metric());
       }
-      if (requestForResponse != null) {
-        metrics.requestReset(requestForResponse.metric());
+      if (currentResponse != null) {
+        metrics.requestReset(currentResponse.request().metric());
       }
     }
 
@@ -504,7 +506,7 @@ class ClientConnection extends ConnectionBase implements HttpClientConnection, H
       throw new IllegalStateException("Connection is already writing a request");
     }
     if (metrics.isEnabled()) {
-      Object reqMetric = client.httpClientMetrics().requestBegin(metric, localAddress(), remoteAddress(), req);
+      Object reqMetric = client.httpClientMetrics().requestBegin(endpointMetric, metric, localAddress(), remoteAddress(), req);
       req.metric(reqMetric);
     }
     this.currentRequest = req;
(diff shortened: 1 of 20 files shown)


vietj:
there is an implementation for dropwizard after that will follow

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -55,7 +55,6 @@
   private final ConnectionManager connectionManager;
   private final Closeable closeHook;
   private final SSLHelper sslHelper;
-  final HttpClientMetrics metrics;
```
tsegismont: Out of curiosity, was there a particular reason for moving the HttpClientMetrics field to the connection manager? Wondering why the move since the lifecycle was already managed here.

src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -391,8 +393,8 @@ protected synchronized void handleClosed() {
       for (HttpClientRequestImpl req: requests) {
         metrics.requestReset(req.metric());
       }
-      if (requestForResponse != null) {
-        metrics.requestReset(requestForResponse.metric());
+      if (currentResponse != null) {
```
tsegismont: This is not strictly related to the goal of the PR, is it? It looks like a bug fix.

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
```diff
@@ -281,8 +294,14 @@ private void createNewConnection(Waiter waiter) {
...
     Waiter getNextWaiter() {
       Waiter waiter = waiters.poll();
+      if (waiter != null && ConnectionManager.this.metrics.isEnabled()) {
+        ConnectionManager.this.metrics.dequeueRequest(metric, waiter.metric);
+      }
       while (waiter != null && waiter.isCancelled()) {
         waiter = waiters.poll();
+        if (waiter != null && ConnectionManager.this.metrics.isEnabled()) {
+          ConnectionManager.this.metrics.dequeueRequest(metric, waiter.metric);
+        }
```
tsegismont:
```java
Waiter waiter;
do {
  waiter = waiters.poll();
  if (waiter == null) break;
  if (ConnectionManager.this.metrics.isEnabled()) {
    ConnectionManager.this.metrics.dequeueRequest(metric, waiter.metric);
  }  
} while (waiter.isCancelled());
```

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
```diff
@@ -281,8 +294,14 @@ private void createNewConnection(Waiter waiter) {
      */
     Waiter getNextWaiter() {
       Waiter waiter = waiters.poll();
+      if (waiter != null && ConnectionManager.this.metrics.isEnabled()) {
+        ConnectionManager.this.metrics.dequeueRequest(metric, waiter.metric);
```
tsegismont: Thinking out loud.

The SPI implementer has no way to determine if `dequeue` is called for a canceled metric? Maybe we don't care because we want to measure waiting time in the queue for both waiters who passed and who timed out.

src/main/java/io/vertx/core/metrics/impl/DummyVertxMetrics.java
```diff
@@ -109,7 +111,9 @@ public boolean isMetricsEnabled() {
     return false;
   }
 
-  protected class DummyEventBusMetrics implements EventBusMetrics<Void> {
+  public static class DummyEventBusMetrics implements EventBusMetrics<Void> {
```
tsegismont: Nice move on making all those subclasses public here. Dummy impl is useful if you want to partially implement the SPI without having to recompile your code when a new method is added.

src/test/java/io/vertx/test/core/MetricsTest.java
```diff
@@ -546,6 +546,118 @@ public void testHttpClientName() throws Exception {
...
+    }
+    assertEquals(Collections.singleton("localhost:8080"), metrics.endpoints());
+    assertEquals(8, (int)metrics.queueSize("localhost:8080"));
+    assertEquals(5, (int)metrics.connectionCount("localhost:8080"));
+    ArrayList<Runnable> copy = new ArrayList<>(requests);
+    requests.clear();
+    copy.forEach(Runnable::run);
+    awaitLatch(responsesLatch);
+    waitUntil(() -> requests.size() == 5);
+    copy = new ArrayList<>(requests);
```
tsegismont: The test looks good to me. I'm just curious how we know that a maximum of 5 requests can be handled at a time. Did we set that in config somewhere? Maybe I'm missing the obvious...

src/main/java/io/vertx/core/http/impl/ClientConnection.java
vietj: bug fix yes - was calling endRequest in metrics twice when connection was closed after response sent.

vietj: there is a test for this in this PR

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
vietj: yes we don't care we dont keep track of queue/deque like we do for pool

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
vietj: I prefer to keep it our of the HttpClientImpl that does not care much about

src/main/java/io/vertx/core/metrics/impl/DummyVertxMetrics.java
vietj: yes that's the point : hack something quickly

src/test/java/io/vertx/test/core/MetricsTest.java
vietj: that's the default client config : 5 max

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
vietj: we could but it means more complexity, I don't think it's worth to have it

src/test/java/io/vertx/test/core/MetricsTest.java
tsegismont: OK, makes sense then.

PR #1443, opened 2016-05-27
update tcnative-boringssl to Fork17

alexlehm:
this build fixes a Windows dependency issue

I tracked the ssl issues when running the tests in Windows to a dll dependency issue, which is already fixed in new build of tcnative boringssl


vietj:
can't use it right now, need to make an Eclipse CQ for this - it's a works-with dependency so should be quick

alexlehm:
I think it is based on the same source code as before, the change is only in the build-specific command https://github.com/netty/netty-tcnative/commit/629334009a372bb6b76c44f9acbf99c9f5638c78

vietj:
ok, I filled the CQ though

PR #1444, opened 2016-05-31
Use Service Helper to avoid classloading issue

cescoffier:
Fix issue #1438 by generalizing the usage of the `ServiceHelper` class.

Also add tests to check the behavior of the `ServiceHelper` when using different classloaders.


vietj:
looks good to me

PR #1445, opened 2016-05-31
Pipe lined limit

vietj:
This feature sets an upper limit for the number of pipe-lined requests for which no response have been received yet on an HTTP/1.1 persistent connection. 

When a client creates a very large number HTTP client requests at the same time, the distribution of the requests on the pool is not fair (i.e a few connections can get more connections than others). This feature limits the number of requests sent on an connection, the other pending requests remain in the endpoint queue.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xPool.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xPool.java
@@ -41,6 +41,7 @@ public class Http1xPool implements ConnectionManager.Pool<ClientConnection> {
   private final Map<Channel, HttpClientConnection> connectionMap;
   private final boolean pipelining;
   private final boolean keepAlive;
+  private final int pipeliningLimit;
   private final boolean ssl;
   private final HttpVersion version;
   private final Set<ClientConnection> allConnections = new HashSet<>();
@@ -55,6 +56,7 @@ public class Http1xPool implements ConnectionManager.Pool<ClientConnection> {
     this.metrics = metrics;
     this.pipelining = options.isPipelining();
     this.keepAlive = options.isKeepAlive();
+    this.pipeliningLimit = options.getPipeliningLimit();
     this.ssl = options.isSsl();
     this.connectionMap = connectionMap;
     this.maxSockets = maxSockets;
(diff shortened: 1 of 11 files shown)


vietj:
please review this.

src/main/java/io/vertx/core/http/impl/Http1xPool.java
```diff
@@ -81,39 +83,40 @@ public HttpClientStream createStream(ClientConnection conn) {
...
-      // Close it now
-      conn.close();
-    }
-  }
-
-  private void doRecycle(ClientConnection conn) {
     synchronized (queue) {
       Waiter waiter = queue.getNextWaiter();
       if (waiter != null) {
-        Context context = waiter.context;
```
tsegismont: Why is the context check no longer needed? Maybe missing the obvious...

vietj: when we recycle, we should really use the Context of the connection + now the runOnContext() is done before calling recycle (because it should not be done in some cases) rather than in recycle.

tsegismont: Got you

vietj: it's done in responseEnded and requestEnded now

vietj: the important thing is that there should be no runOnContext between the time a connection was check to be reusable because pipelining limit is not reached and the time it is dispatched (where the getOutStandingRequestCount() will incrase). Otherwise two checks be valid and a number > pipeliningLimit of requests can be done (race condition)

src/test/java/io/vertx/test/core/Http1xTest.java
```diff
@@ -1579,30 +1675,50 @@ private void testClientContext() throws Exception {
...
     AtomicReference<Context> c = new AtomicReference<>();
-    client.getNow(DEFAULT_HTTP_PORT, DEFAULT_HTTP_HOST, "/1", res -> {
+    HttpClientRequest req1 = client.get(DEFAULT_HTTP_PORT, DEFAULT_HTTP_HOST, "/1");
+    req1.handler(res -> {
       c.set(Vertx.currentContext());
       res.endHandler(v -> req1Latch.countDown());
     });
+    req1.end();
+    Consumer<HttpClientRequest> checker = req -> {
+      if (req1.connection() == req.connection()) {
```
tsegismont: OK, this is where you check the context is that of the connection.

src/test/java/io/vertx/test/core/Http1xTest.java
```diff
@@ -1579,30 +1675,50 @@ private void testClientContext() throws Exception {
...
+        assertSame(Vertx.currentContext(), c.get());
+      } else {
+        // This may happen sometimes if the connection
```
tsegismont: if  the connection...?

tsegismont:
Looks good to me. There's just an incomplete comment in tests.

src/test/java/io/vertx/test/core/Http1xTest.java
vietj: I think now it should not happen anymore, so I will rather remove the if

tsegismont: Right

PR #1454, opened 2016-06-03
catch NoClassDefFoundError and give a warning

alexlehm:
fix for vert-x3/issues#129

catch NoClassDefFoundError exception when getting the ProxyChannel instance, give a warning and pass the exception to the correct handler

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
+++ b/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
@@ -417,7 +417,13 @@ public class ConnectionManager {
       if (options.getProxyOptions() == null) {
         channelProvider = ChannelProvider.INSTANCE;
       } else {
-        channelProvider = ProxyChannelProvider.INSTANCE;
+        try {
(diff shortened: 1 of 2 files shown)


cescoffier:
👍

cescoffier:
If would be great to have test case about this, but it might be a bit hard (need a custom classloader not delegating to the parent one when loading the class, load vert.x using this classloader and check).

alexlehm:
I have considered that as well, I assume the easiest way would be to do a an exec task in maven, but that is not a general unit test that can be run in java. I will take a look with the classloader

vietj:
maybe we should make it simpler and just add this particular `io.netty:netty-handler-proxy` dependency instead ?

alexlehm:
If that is ok, I would prefer that. The jar adds about 20 kb to the fatjar I assume, which is not bad compared to the rest of netty about 3 MB

vietj:
fine for me

alexlehm:
closed in favour of PR#1455

PR #1457, opened 2016-06-05
Fix how is passed inlined configuration in redeploy mode.

cescoffier:
The inlined configuration was not quoted correctly when passed to the spawn process, and so could not be interpreted as valid json.

Initial discussion here: https://groups.google.com/d/msgid/vertx/a28558c2-345c-4427-9aab-21853ba0ff43%40googlegroups.com?utm_medium=email&utm_source=footer


cescoffier:
@vietj - ready for review

vietj:
looks good

PR #1459, opened 2016-06-06
Log the error when client isn't able to connect to remote host.

manish-panwar:
https://groups.google.com/forum/#!searchin/vertx/log$20the$20error/vertx/fwZNJDR6IFI/Va_GTtZ0BAAJ 

We are using a cluster of 2 nodes and intentionally blocked the port 41232 (we are using a dedicated port for Vertx communication). Since port is blocked, we get connection timeout error and "else" condition in given below code is executed.


manish-panwar:
Being handled part of https://github.com/eclipse/vert.x/pull/1562, so closing this PR.

PR #1464, opened 2016-06-13
DNS resolver issues

vietj:
- resolve localhost to 127.0.0.1 on Windows as Windows 10
- fix bug with resolver with a DNS answer containing CNAME+A


pom.xml
```diff
@@ -420,6 +420,18 @@
               </systemProperties>
             </configuration>
           </execution>
+          <execution>
+            <id>window-resolve-localhost</id>
```
tsegismont: window**s** ?

src/test/java/io/vertx/test/core/HostnameResolutionTest.java
```diff
@@ -309,4 +310,22 @@ public void testResolveFromBuffer() {
     }));
     await();
   }
+
+/*
```
tsegismont: Remove this ?

alexlehm:
this is probably dependent on the hosts file contains an entry for localhost 127.0.0.1 since the hosts file is evaluated by Netty resolver.
I didn't notice that before, since my windows at home apparently contains an entry for that (windows7), but the machine I am using at work doesn't (also windows7)

vietj:
FTR:

https://github.com/netty/netty/issues/5386
https://github.com/netty/netty/issues/5391

src/main/java/io/vertx/core/impl/AddressResolver.java
```diff
@@ -131,14 +133,18 @@ public void close() {
...
-    io.netty.util.concurrent.Future<InetSocketAddress> fut = resolver.resolve(InetSocketAddress.createUnresolved(hostname, 80));
+    io.netty.util.concurrent.Future<InetSocketAddress> fut = resolver.resolve(InetSocketAddress.createUnresolved(hostname, 0));
     fut.addListener(a -> {
       callback.runOnContext(v -> {
         if (a.isSuccess()) {
           InetSocketAddress address = fut.getNow();
           resultHandler.handle(Future.succeededFuture(address.getAddress()));
         } else {
-          resultHandler.handle(Future.failedFuture(a.cause()));
+          if (Utils.isWindows() && hostname.equals("localhost")) {
```
alexlehm: hostname resolution is case-insensitive so LOCALHOST is a valid hostname for 127.0.0.1 (had the same issue in netty hosts file resolver)

hostname.toLowerCase(Locale.EN).equals("localhost");

vietj: in the more recent version there is not comparison anymore, what is done instead is to add "alias=127.0.0.1" in the "hosts" Map, so I think this is handled by the underlying resolver, can you check ?

alexlehm:
this works correctly, the entry is added when localhost is not listed in the hosts file and the comparison in correct case-insensitive in the netty code.

alexlehm:
For some reason I cannot reproduce the failing resolution with LOCALHOST anymore when I take out the code to add the localhost alias

PR #1490, opened 2016-07-04
Fix #1320 - improve shared data documentation

### LINKED ISSUE #1320: Improve documentation of getClusterWideMap
There are many aspects when going distributed which are missing from the documentation

"Get the cluster wide map with the specified name. The map is accessible to all nodes in the cluster and data put into the map from any node is visible to any other node."

More specific topics:
 - how much HA is in there (is the data replicated?)
 - are there situations when data can be lost?
 - how about performance?
 - how about split brain? How about concurrent updates in a split brain situation? How about merging of the split brain?

------

Incorrect documentation for put with a TTL.

http://vertx.io/docs/apidocs/io/vertx/core/shareddata/AsyncMap.html#put-K-V-long-io.vertx.core.Handler-

It seems that the parameter is not a timeout but a TTL which does not match the comment below. 

"Like put(K, V, io.vertx.core.Handler<io.vertx.core.AsyncResult<java.lang.Void>>) but specifying a timeout. If the value cannot be put within the timeout a failure will be passed to the handler"

I assume the word timeout does not refer to the TTL in which case i assume there is some internal timeout to consider although not mentioned for other methods which take the completionHandler as parameter.

cescoffier:
Addresses comments from #1320:

* add a note about the behavior of distributed data structure
* fix mistake in the AsyncMap javadoc

### DIFF
--- a/src/main/java/io/vertx/core/shareddata/package-info.java
+++ b/src/main/java/io/vertx/core/shareddata/package-info.java
@@ -23,6 +23,10 @@
  * Shared data includes local shared maps, distributed, cluster-wide maps, asynchronous cluster-wide locks and
  * asynchronous cluster-wide counters.
  *
+ * IMPORTANT: The behavior of the distributed data structure depends of the cluster manager you use. Backup
+ * (replication) and behavior when a network partition is faced are defined by the cluster manager and its
+ * configuration. Refer to the cluster manager documentation as well as to the underlying framework manual.
+ *
  * === Local shared maps
  *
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/shareddata/package-info.java
```diff
@@ -23,6 +23,10 @@
  * Shared data includes local shared maps, distributed, cluster-wide maps, asynchronous cluster-wide locks and
  * asynchronous cluster-wide counters.
  *
+ * IMPORTANT: The behavior of the distributed data structure depends of the cluster manager you use. Backup
```
vietj: _depends on_

PR #1492, opened 2016-07-04
Preserve order of enabled cipher suites and protocols for TCPSSLOptio…

### LINKED ISSUE #1491: Preserve order of enabled cipher suites and protocols for TCPSSLOptions
Currently we use HashSet for storing the suites and protocols.

Ordering set when building the options should be respected in the options and when creating the ssl engine.

vietj:
…ns - fixes #1491

please review.


tsegismont:
LGTM

PR #1493, opened 2016-07-04
Eventbusmetrics schedule message

vietj:
please review

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
@@ -246,7 +246,7 @@ public class ClusteredMessage<U, V> extends MessageImpl<U, V> {
     return fromWire;
   }
 
-  boolean isLocal() {
-    return isFromWire();
+  protected boolean isLocal() {
+    return !isFromWire();
   }
 }
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
@@ -245,4 +245,8 @@ public class ClusteredMessage<U, V> extends MessageImpl<U, V> {
   public boolean isFromWire() {
     return fromWire;
   }
+
+  boolean isLocal() {
+    return isFromWire();
+  }
 }
--- a/src/main/java/io/vertx/core/spi/metrics/EventBusMetrics.java
+++ b/src/main/java/io/vertx/core/spi/metrics/EventBusMetrics.java
@@ -50,8 +50,9 @@ public interface EventBusMetrics<H> extends Metrics {
    * No specific thread and context can be expected when this method is called.
    *
    * @param handler the handler processing the message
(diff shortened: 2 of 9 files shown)


src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
```diff
@@ -245,4 +245,8 @@ ServerID getSender() {
   public boolean isFromWire() {
     return fromWire;
   }
+
+  boolean isLocal() {
+    return isFromWire();
```
tsegismont: Shouldn't it be `return !isFromWire();`

src/main/java/io/vertx/core/spi/metrics/EventBusMetrics.java
```diff
@@ -45,6 +45,16 @@
...
 
   /**
+   * Schedule a message for processing.<p/>
+   *
+   * No specific thread and context can be expected when this method is called.
+   *
+   * @param handler the handler processing the message
+   * @param local when the scheduled message is local
+   */
+  void scheduleMessage(H handler, boolean local);
```
tsegismont: Have you updated the breaking changes page?

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
vietj: good catch, it would not have introduced a bug in eventbus, rather reporting incorrectly

src/test/java/io/vertx/test/core/MetricsTest.java
```diff
@@ -270,28 +270,51 @@ private HandlerMetric assertRegistration(FakeEventBusMetrics metrics) {
...
         assertEquals(expectedLocalCount, registration.localCount.get());
-        testComplete();
-      });
-    }));
+        msg.reply("pong");
+      }).completionHandler(onSuccess(v2 -> {
+        to.runOnContext(v3 -> {
+          latch1.countDown();
+          try {
+            awaitLatch(latch2);
```
tsegismont: I don't understand why we need to wait here. Could you explain?

tsegismont: I mean, I don't understand the need for latch2 altogether.

vietj: we need to wait for Latch2 because it guarantees that this particular event loop is busy when we send the message, so `from.eventBus().send` will schedule the message to send but it will not be delivered.

tsegismont: Ha right. Got it, thanks!

tsegismont:
LGTM, no other pending question/comment

src/main/java/io/vertx/core/spi/metrics/EventBusMetrics.java
vietj: will do

vietj:
I fixed the `isLocal` can you check it is correct now ?

tsegismont:
Yes, it looks good to me.

PR #1495, opened 2016-07-05
 Fixes #1494 Let the user access the logger delegate and logger implementation

### LINKED ISSUE #1494: Let the user access the logger delegate and logger implementation
This has been discussed on the mailing list. A user wants to be able to reconfigure some loggers programatically after the logging SPI has been initialized.

tsegismont:
Added a getter in Logger to get the log delegate.
Added and an unwrap method to the log delegate interface (with default impl returning null for backwards compatibility)

### DIFF
--- a/src/main/java/io/vertx/core/spi/logging/LogDelegate.java
+++ b/src/main/java/io/vertx/core/spi/logging/LogDelegate.java
@@ -72,4 +72,15 @@ public interface LogDelegate {
   void trace(Object message, Throwable t);
 
   void trace(Object message, Throwable t, Object... params);
+
+  /**
+   * Give access to a typed instance of the underlying framework logger object.
+   *
+   * @param cls the expected type, or an interface the logger object implements
+   * @return a typed logger instance, null in the default implementation
+   * @throws ClassCastException if the logger object is not assignable to {@code cls}
+   */
+  default <T> T unwrap(Class<T> cls) {
+    return null;
+  }
 }
--- a/src/main/java/io/vertx/core/spi/logging/LogDelegate.java
+++ b/src/main/java/io/vertx/core/spi/logging/LogDelegate.java
@@ -74,13 +74,9 @@ public interface LogDelegate {
   void trace(Object message, Throwable t, Object... params);
 
   /**
-   * Give access to a typed instance of the underlying framework logger object.
-   *
-   * @param cls the expected type, or an interface the logger object implements
-   * @return a typed logger instance, null in the default implementation
-   * @throws ClassCastException if the logger object is not assignable to {@code cls}
+   * @return the underlying framework logger object, null in the default implementation
    */
-  default <T> T unwrap(Class<T> cls) {
+  default Object unwrap() {
     return null;
   }
 }
--- a/src/test/java/io/vertx/core/logging/JULLogDelegateTest.java
+++ b/src/test/java/io/vertx/core/logging/JULLogDelegateTest.java
@@ -16,13 +16,14 @@
 
 package io.vertx.core.logging;
 
+import io.vertx.core.spi.logging.LogDelegate;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.IOException;
 
-import static junit.framework.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 /**
  * These tests check the JUL log delegate. It analyses the output, so any change in the configuration may break the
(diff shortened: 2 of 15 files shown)


tsegismont:
@vietj @cescoffier this was asked in the user group recently. It's not too complicated but I'll defer to you on whether it can go in 3.3.1.

src/test/java/io/vertx/core/logging/JULLogDelegateTest.java
```diff
@@ -40,10 +41,24 @@
   public static void initialize() throws IOException {
     // Clear value.
     System.clearProperty("vertx.logger-delegate-factory-class-name");
+    LoggerFactory.initialise();
```
tsegismont: @cescoffier this was missing, compared to other logging tests. Maybe that would solve the CI issue you were referring to in the class Javadoc? If so then we can remove the ignore annotations.

src/main/java/io/vertx/core/spi/logging/LogDelegate.java
```diff
@@ -72,4 +72,15 @@
...
   void trace(Object message, Throwable t, Object... params);
+
+  /**
+   * Give access to a typed instance of the underlying framework logger object.
+   *
+   * @param cls the expected type, or an interface the logger object implements
+   * @return a typed logger instance, null in the default implementation
+   * @throws ClassCastException if the logger object is not assignable to {@code cls}
+   */
+  default <T> T unwrap(Class<T> cls) {
```
vietj: just return Object, users will do:

```
Foo foo = (Foo) delegate.unwrap();
```

instead of:

```
Foo foo = delegate.unwrap(Foo.class);
```

the later looks too smart for a cast, let just use the cast operator Java has been providing for years.

tsegismont: Done

src/test/java/io/vertx/core/logging/JULLogDelegateTest.java
tsegismont: @cescoffier with re-initialization of the logger factory and removal of ignore annotations, the tests pass on my box, even as part of the suite. They didn't pass when the call to initialize was missing. I miss some context to make sure you had the same issue on CI, but if you can confirm it was something like:

```
Running io.vertx.core.logging.JULLogDelegateTest
[main] ERROR my-jul-logger - hello
[main] INFO my-jul-logger - hello
[main] WARN my-jul-logger - hello
Tests run: 4, Failures: 4, Errors: 0, Skipped: 0, Time elapsed: 0.015 sec <<< FAILURE! - in io.vertx.core.logging.JULLogDelegateTest
testError(io.vertx.core.logging.JULLogDelegateTest)  Time elapsed: 0.009 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at io.vertx.core.logging.JULLogDelegateTest.testError(JULLogDelegateTest.java:119)

testInfo(io.vertx.core.logging.JULLogDelegateTest)  Time elapsed: 0.004 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at io.vertx.core.logging.JULLogDelegateTest.testInfo(JULLogDelegateTest.java:65)

testDelegateUnwrap(io.vertx.core.logging.JULLogDelegateTest)  Time elapsed: 0 sec  <<< FAILURE!
java.lang.AssertionError: Unexpected unwrapped type: org.apache.logging.log4j.core.Logger cannot be cast to java.util.logging.Logger
	at org.junit.Assert.fail(Assert.java:88)
	at io.vertx.core.logging.JULLogDelegateTest.testDelegateUnwrap(JULLogDelegateTest.java:55)

testWarning(io.vertx.core.logging.JULLogDelegateTest)  Time elapsed: 0.002 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at io.vertx.core.logging.JULLogDelegateTest.testWarning(JULLogDelegateTest.java:172)


Results :

Failed tests: 
  JULLogDelegateTest.testDelegateUnwrap:55 Unexpected unwrapped type: org.apache.logging.log4j.core.Logger cannot be cast to java.util.logging.Logger
  JULLogDelegateTest.testError:119 null
  JULLogDelegateTest.testInfo:65 null
  JULLogDelegateTest.testWarning:172 null
```

then I'll push a last commit.

cescoffier: Yes it's this issue.

Test orders depends on the system, and on the CI we have one (or more) test touching the logging system before the logger tests. As JUL is handled using a JVM singleton (somehow), we can reset it.

vietj: such tests we should consider moving them as "it" to run in their own VM

tsegismont: @vietj done

PR #1515, opened 2016-07-13
Removed static initialiser from FileResolver for reading system properties

### LINKED ISSUE #1514: FileResolver does not use vertx.disableFileCaching when using Eclipse
Problem:
Setting the environment vertx.disableFileCaching to true does not disable file caching.

Cause:
FileResolver statically initialises ENABLE_CACHING

Environment:
Eclipse with dynamic reload:
run main.WebVerticle --redeploy="src/**/*.java" --launcher-class=io.vertx.core.Launcher -conf src/resources/conf/config.json -Dvertx.disableFileCaching=true

andrewwh:
Removed static initialiser from FileResolver for reading system properties. Appeared to prevent correct initialisation when using vertx Launcher in eclipse. Closes #1514

Signed-off-by: andrewwh <andrewwh@datacom.co.nz>

### DIFF
--- a/src/main/java/io/vertx/core/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/impl/FileResolver.java
@@ -52,12 +52,12 @@ public class FileResolver {
   public static final String DISABLE_CP_RESOLVING_PROP_NAME = "vertx.disableFileCPResolving";
   public static final String CACHE_DIR_BASE_PROP_NAME = "vertx.cacheDirBase";
 
-  private static final String DEFAULT_CACHE_DIR_BASE = ".vertx";
-  private static final String FILE_SEP = System.getProperty("file.separator");
-  private static final boolean NON_UNIX_FILE_SEP = !FILE_SEP.equals("/");
-  private static final boolean ENABLE_CACHING = !Boolean.getBoolean(DISABLE_FILE_CACHING_PROP_NAME);
-  private static final boolean ENABLE_CP_RESOLVING = !Boolean.getBoolean(DISABLE_CP_RESOLVING_PROP_NAME);
-  private static final String CACHE_DIR_BASE = System.getProperty(CACHE_DIR_BASE_PROP_NAME, DEFAULT_CACHE_DIR_BASE);
+  private final String DEFAULT_CACHE_DIR_BASE = ".vertx";
+  private final String FILE_SEP = System.getProperty("file.separator");
+  private final boolean NON_UNIX_FILE_SEP = !FILE_SEP.equals("/");
+  private final boolean ENABLE_CACHING = !Boolean.getBoolean(DISABLE_FILE_CACHING_PROP_NAME);
+  private final boolean ENABLE_CP_RESOLVING = !Boolean.getBoolean(DISABLE_CP_RESOLVING_PROP_NAME);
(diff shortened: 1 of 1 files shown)


vietj:
@cescoffier @pmlopes are you OK with this ?

src/main/java/io/vertx/core/impl/FileResolver.java
```diff
@@ -52,12 +52,12 @@
...
   public static final String CACHE_DIR_BASE_PROP_NAME = "vertx.cacheDirBase";
 
-  private static final String DEFAULT_CACHE_DIR_BASE = ".vertx";
-  private static final String FILE_SEP = System.getProperty("file.separator");
-  private static final boolean NON_UNIX_FILE_SEP = !FILE_SEP.equals("/");
-  private static final boolean ENABLE_CACHING = !Boolean.getBoolean(DISABLE_FILE_CACHING_PROP_NAME);
-  private static final boolean ENABLE_CP_RESOLVING = !Boolean.getBoolean(DISABLE_CP_RESOLVING_PROP_NAME);
-  private static final String CACHE_DIR_BASE = System.getProperty(CACHE_DIR_BASE_PROP_NAME, DEFAULT_CACHE_DIR_BASE);
+  private final String DEFAULT_CACHE_DIR_BASE = ".vertx";
+  private final String FILE_SEP = System.getProperty("file.separator");
```
cescoffier: This one can be static, it's an immutable value.

src/main/java/io/vertx/core/impl/FileResolver.java
```diff
@@ -52,12 +52,12 @@
...
+  private final boolean NON_UNIX_FILE_SEP = !FILE_SEP.equals("/");
```
cescoffier: can be kept static for the same reason

src/main/java/io/vertx/core/impl/FileResolver.java
```diff
@@ -52,12 +52,12 @@
   public static final String DISABLE_CP_RESOLVING_PROP_NAME = "vertx.disableFileCPResolving";
   public static final String CACHE_DIR_BASE_PROP_NAME = "vertx.cacheDirBase";
 
-  private static final String DEFAULT_CACHE_DIR_BASE = ".vertx";
-  private static final String FILE_SEP = System.getProperty("file.separator");
-  private static final boolean NON_UNIX_FILE_SEP = !FILE_SEP.equals("/");
-  private static final boolean ENABLE_CACHING = !Boolean.getBoolean(DISABLE_FILE_CACHING_PROP_NAME);
-  private static final boolean ENABLE_CP_RESOLVING = !Boolean.getBoolean(DISABLE_CP_RESOLVING_PROP_NAME);
-  private static final String CACHE_DIR_BASE = System.getProperty(CACHE_DIR_BASE_PROP_NAME, DEFAULT_CACHE_DIR_BASE);
+  private final String DEFAULT_CACHE_DIR_BASE = ".vertx";
```
cescoffier: just a constant, can be kept static

cescoffier:
@andrewwh I'm not sure to understand correctly. You say it prevent the initialization issue of #1514. So that would mean that we are loading the `FileResolver` class (static field are computed when the class is loaded) before having the launcher setting the system properties. Did you look where this happen ?

With this, it means that the value are not loaded when the class is loaded, but when the instance is created (so in VertxImpl).

PR #1516, opened 2016-07-13
fix issue #1512 certcheck hostname error in upgradeToSsl

### LINKED ISSUE #1512: ns.upgradeToSsl fails hostname cert check
When connecting with NetClient using upgradeToSsl (which is for example used in mail STARTTLS), the cert name is matched against the resolved ip address, which fails.

https://gist.github.com/alexlehm/8b609f70c848453ab7da5b360b740db0

alexlehm:
fixes #1512


src/test/java/io/vertx/test/core/NetClientTLSConnectTest.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class NetClientTLSConnectTest extends VertxTestBase {
+
+  private HttpServer server;
+
+  @Before
+  public void startServer() throws InterruptedException {
+    CountDownLatch latch = new CountDownLatch(1);
+
+    HttpServerOptions serverOptions = new HttpServerOptions();
```
vietj: why does it matter to test with an HttpServer and not a NetServer ?

src/test/java/io/vertx/test/core/NetClientTLSConnectTest.java
```diff
@@ -0,0 +1,72 @@
...
+    latch.await();
+  }
+
+  @After
+  public void stopServer() {
+    server.close();
+  }
+
+  @Test
+  public void testHostnameCheck() {
```
vietj: rather have this test in NetTest instead of having a new class just for this test

src/test/java/io/vertx/test/core/NetClientTLSConnectTest.java
```diff
@@ -0,0 +1,72 @@
...
+    serverOptions.setSsl(true);
+    serverOptions.setKeyCertOptions(new JksOptions().setPath("tls/server-keystore-root-ca.jks").setPassword("wibble"));
+    server = vertx.createHttpServer(serverOptions.setPort(4043));
+    server.requestHandler(req -> {
+      req.bodyHandler(buffer -> {
+        req.response().end();
+      });
+    });
+    server.listen(ar -> latch.countDown());
+    latch.await();
```
vietj: await(latch)

src/test/java/io/vertx/test/core/NetClientTLSConnectTest.java
alexlehm: it doesn't, I started testing the issue by connecting to https://www.yahoo.com/ and then wrote the local server as https, it should work with a NetServer as well

alexlehm:
ok, that should be all for now

vietj:
looks good to me, thanks

PR #1526, opened 2016-07-20
custom launcher - set clustered and avoid NPE on afterConfigParsed

cescoffier:
Allow custom launcher to:

* set whether or not the vert.x instance is clustered (this was not possible before, clearly a bug)
* improve the afterConfigParsed callback to not call the method with a `null` object


vietj:
👍

PR #1530, opened 2016-07-22
Add support for redeploy patterns using absolute paths

cescoffier:
Fix https://github.com/vert-x3/issues/issues/156

* Add support for absolute path pattern in the redeploy commands
* Fix a race condition in the modification test


vietj:
👍

Issue #1532, opened 2016-07-24
Overloads for CompositeFuture.all() and any() that collect results prior to completion 


aschrijver:
As discussed in this thread on the [Vert.x forum](https://groups.google.com/forum/#!topic/vertx/3V44vLG_LjI) there is a need for an additional set of overloaded static members of both`CompositeFuture.all()` and `CompositeFuture.any()` that have the same behaviour, except that the composite feature completes when _all_ of its constituent futures have completed (regardless of their outcome). The current behaviour has the composite future complete on the first success or failure.

The new method overloads have an additional parameter `collectResults` with a value of `true`.
The existing members will not be changed so as not to break backwards-compatibility and will defer to the composite future implementation with a default value of `false` for `collectResults`.

I will create a PR shortly that will implement this design and provide unit tests.

vietj:
can you review https://github.com/eclipse/vert.x/issues/1555 ?

aschrijver:
Looks good, @vietj ! Overall a better solution than the boolean flag from before..

PR #1534, opened 2016-07-24
Overloads of CompositeFuture that collect results before completion (#1532)

aschrijver:
This PR extends `CompositeFuture` with static method overloads of the `any()` and `all()` methods that operate in the same way, but complete only after all individual futures have been completed.

See also issue: #1532

(Note: This request replaces [a previous one](https://github.com/eclipse/vert.x/pull/1533), where I had trouble amending commits with Eclipse CLA)

### DIFF
--- a/src/main/asciidoc/java/override/dependencies.adoc
+++ b/src/main/asciidoc/java/override/dependencies.adoc
@@ -8,7 +8,7 @@ project descriptor to access the Vert.x Core API:
 <dependency>
   <groupId>io.vertx</groupId>
   <artifactId>vertx-core</artifactId>
-  <version>3.3.2</version>
+  <version>3.4.0-SNAPSHOT</version>
 </dependency>
 ----
 
@@ -16,5 +16,5 @@ project descriptor to access the Vert.x Core API:
 
 [source,groovy,subs="+attributes"]
 ----
-compile io.vertx:vertx-core:3.3.2
+compile io.vertx:vertx-core:3.4.0-SNAPSHOT
 ----
--- a/src/main/java/io/vertx/core/CompositeFuture.java
+++ b/src/main/java/io/vertx/core/CompositeFuture.java
@@ -22,13 +22,13 @@ import io.vertx.core.impl.CompositeFutureImpl;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.function.Function;
 
 /**
  * The composite future wraps a list of {@link Future futures}, it is useful when several futures
  * needs to be coordinated.
  *
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ * @author <a href="https://github.com/aschrijver/">Arnold Schrijver</a>
  */
 @VertxGen
 public interface CompositeFuture extends Future<CompositeFuture> {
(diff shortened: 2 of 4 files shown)


vietj:
After reading some similar cases for it (in when.js) that provides similar behavior for all/any, I'm leaning toward using rather a new `join` method that seems to describe more the intent we have in this use case. It would wait for the components to be completed, the returned CompositeFuture would never fail, always succeeds when all components have been completed.

aschrijver:
Oeff, that's a bit late after inviting on a PR with a boolean parameter and me creating it on a free Sunday :(

I am not sure that makes it all that much clearer. And having it always succeed means that you always have to write a loop to check individual results, and you lose the `any` and `all` semantics.

One of the use cases I have is for [vertx-graphql-service-discovery](https://github.com/engagingspaces/vertx-graphql-service-discovery) in `SchemaPublisher.pubishAll` where you publish the schema's that define the API's and data model of your microservice.
If one publication fails other publications should still be published, otherwise the one schema brings the whole microservice down. But there is a serious error, so `publishAll` should return failure for me to handle directly.

Also I think it might be better to have things like `join` together with other reactive goodness in a library class, just like `when.js`, or maybe even a separate vertx module altogether. I think there could be a place for more reactive stuff in the supported vertx modules list.

vietj:
in your case, do you have the usage for an "any" that waits for all results ?

aschrijver:
Yes, in the case of the example I gave, I could just as well use the `any()` variant (in the `publishAll` implementation) e.g. return success by default and send information on failures to some monitoring service. The choice is to the implementer, depending on the seriousness of a constituent future failing.

aschrijver:
I am getting more use cases for this. What do you think of my arguments above?

To sum up, I am in favour of having `CompositeFuture` only having any / all semantics with the option to collect results and keep it lean.
I think there is a place for a more elaborate async library class (or set of classes). I [created an issue](https://github.com/vert-x3/issues/issues/164) on the Vert.x wish-list for a `vertx-common` module (compare to Apache Commons and Guava, and like ) that would be a proper location (they are extras, not core).

BTW Are all the additional `f1, f2, ...` overloads needed for polyglot language support? If yes, a nice feature of a future Vert.x would be to offload that to code generation, and clean up Java class signatures in Vert.x 4.x (or at least obviate the need for such overloads in future code if this is not feasible)

vietj:
I'm thinking that either we add a new "join" method that is like "all" but with the behavior you describe and we drop to wait for the any.

Or we do like you propose and we somehow rename the collectResults to "join".

Overall, I think the word "join" that describes very well the intent.

vietj:
my point is that I would like to not rush on this and have the best api possible because after it is too late to fix.

aschrijver:
Fair enough. If there would be `any`, `all` and `join` and the latter one would include a flag in the result to check if failures occurred (without having to iterate the constituent futures) then it would be better than the boolean parameters. I just created it like this to comply with your PR request.

vietj:
so if I understand correctly there are two things here:

- having the ability to have a method that waits for all futures to be completed with a predicate that completes to successful or failed (for example _all_ behavior is a predicate and _any_ is another predicate)
- have more methods on CompositeFuture to query about the status of the nested futures, like counting the number of failures, determine if there is at least one failure or all are success.

is it correct ?

aschrijver:
Yes, that is correct. But note that the second point is not required, but could be nice for convenience.
I don't mind to process the list of results on failure, but the success path should be as smooth as possible.

With bullet point 1 if there is failure with either `any` or `all` predicates, I know the constituent results to be present and can iterate + process the `cause`s of failed ones. This is already facilitated for in existing class.

vietj:
@aschrijver I think it's possible to come up now with a join method where you provide a min and max (a range) of the success count, so with a List of 5 Future:

join(5, 5, list) <=> All are success
join(1, 5, list) <=> At least one is success
join(0, 0, list) <=> All are failed
join(0, 5, list) <=> Always success

aschrijver:
Yeah that looks nicd. So you can specify your success criteria in more detail? Maybe there could also be an overload with a predicate param where you can add your custom (even more specific) criteria (e.g. 'if logging service fails return success, but fail when either auth service or auth service db' fails.

vietj:
@aschrijver this can be possible but it needs to wait for 3.4 as it means minor improvements in codegen

aschrijver:
That is no problem. Thanks!

vietj:
so for 3.3.3 we provide only join(List<Future>) and in 3.4 we allow to use a Function<CompositeFurure, Boolean> predicate with predefined predicates for ALL, ANY, NONE.

aschrijver:
That would be awesome :+1:

aschrijver:
Replaced with #1567 Closing.

PR #1537, opened 2016-07-27
Add the ability to chain bodyEndHandlers

fzakaria:
It is useful sometimes to register more than 1 bodyEndHandler at various parts of the
Router (in the case of Vert.x Web).

This change will allow users to compose multiple bodyEndHandlers as they see fit.
There is an example provided in the test case that demonstrates such a scenario.

If this change is accepted I think similar changes should occur for all methods that register a Handler.

I'm not too attached to the name of the `FunctionHandler` class and open to change.

### DIFF
--- a/src/main/java/io/vertx/core/FunctionHandler.java
+++ b/src/main/java/io/vertx/core/FunctionHandler.java
@@ -1,0 +1,40 @@
+package io.vertx.core;
+
+import io.vertx.codegen.annotations.VertxGen;
+import java.util.Objects;
+
+/**
+ *  A generic FunctionHandler interface
+ *  <p>
+ *  This interface is copy of the JDK FunctionHandler interface to support VertxGen and
+ *  composing {@link Handler}
+ *  <p>
+ *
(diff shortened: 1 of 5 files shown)


fzakaria:
Signed the request.. not sure why its still failing.

vietj:
I would rather not merge this as it can be done easily:

```
Handler<Buffer> h1 = ...
Handler<Buffer> h2 = ...
req.bodyHandler(body -> h1.handle(body); h2.handle(body));
```

fzakaria:
The problem is all the bodyHandlers have to be done in the same spot for your example to work. 
I'm looking at my usage of Very.x web and I've setup multiple route handlers on the `RoutingContext`. 

Separate route handlers may want to know register a bodyHandler (separate classes at this point). Your example would work also if the bodyHandler had a getter (another solution perhaps also simpler).

fzakaria:
Would you be open to a change where the handlers are all exposed via a getter ?
That would solve my problem. Not being able to compose the handlers across different classes really kills making larger applications with Vert.x

fzakaria:
Closing.

PR #1540, opened 2016-07-28
Issues/#1498 http proxy request

### LINKED ISSUE #1498: Implement HTTP Proxy for non SSL HttpClient
At the moment the HttpClient proxy configuration uses CONNECT for non SSL HttpClient and instead it should do a straight connection with a target requestURI, i.e

```
client.get(80, "www.google.com", resp -> {});
```

should be equivalent to:

```
client.get(80, "the-proxy", "http://www.google.com", resp -> {}).host("www.google.com:80");
```

so we do have request like:

```
GET http://www.google.com HTTP/1.1
Host: www.google.com
```

while the client connects to `the-proxy` host.

Note: this does not affect `NetClient` that will always continue to create a tunnel using CONNECT

alexlehm:
implement proxy support for http requests
fixes #1498

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
+++ b/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
@@ -406,7 +406,7 @@ public class ConnectionManager {
       applyConnectionOptions(options, bootstrap);
 
       ChannelProvider channelProvider;
-      if (options.getProxyOptions() == null) {
+      if (!options.isSsl() || options.getProxyOptions() == null) {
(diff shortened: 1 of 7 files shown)


vietj:
@alexlehm one commit does not pass the CLA check

alexlehm:
I will rewrite the commits or squash then, not sure why the last commit fails

vietj:
sometimes it happens and we don't know really why...

alexlehm:
ok, changing the last commit didn't work, if will create a new commit

alexlehm:
replaced by #1548

PR #1544, opened 2016-07-29
Improve watcher performances

cescoffier:
* change how the watching is made to not watch the full root, but watch individual roots. It avoids issues when watching huge directories
* add performance tests

This PR is related to https://github.com/vert-x3/vertx-examples/issues/150


src/main/java/io/vertx/core/impl/launcher/commands/Watcher.java
```diff
@@ -92,6 +112,10 @@ public Watcher(File root, List<String> includes, Handler<Handler<Void>> deploy,
     }).collect(Collectors.toList());
   }
 
+  private void addFileToWatch(List<File> roots) {
```
vietj: should be named `addFilesToWatch`

cescoffier:
@vietj I've integrated the reviews.

vietj:
I will merge it this morning

PR #1545, opened 2016-07-29
unit test for #1524 localhost resolve issue with netty in Windows

### LINKED ISSUE #1524: localhost resolve issue in Windows when using netty resolver
when using the netty resolver code (https://github.com/eclipse/vert.x/tree/netty-4.1.3.Final)
the resolution of "localhost" does not have the hostname set in the InetAddr object so that getHostName() returns the ip address. This breaks some unit tests using localhost and expecting "localhost" as hostname (e.g. tls cert check).
I will create a netty issue for that, since their resolver code creates the object, but it would be good to implement a test and a fix inside vert.x as well, when the netty code is fixed, we can take out the fix in vert.x again (that mostly depends on what comes first, the next vert.x version or the next netty version and whether the netty people decide the fix is valid)

alexlehm:
fixes #1524 
add test to resolve localhost with different case spellings and check that the returned address
object has the hostname localhost set.
the issue itself didn't require changes in vert.x as it is fixed by the update to netty 4.1.4

### DIFF
--- a/src/test/java/io/vertx/test/core/HostnameResolutionTest.java
+++ b/src/test/java/io/vertx/test/core/HostnameResolutionTest.java
@@ -45,6 +45,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


alexlehm:
thanks

PR #1548, opened 2016-07-30
pr issue#1498 Implement HTTP Proxy for non SSL HttpClient 

### LINKED ISSUE #1498: Implement HTTP Proxy for non SSL HttpClient

alexlehm:
fixes #1498

Squash commits into new branch:

Implement http proxy requests, add unit test, add support for GET proxy requests
to ConnectHttpProxy

fix ProxyErrorTest which was missing setSsl(true)

add tests for proxy errors for http, refactor the ProxyError tests a bit
handle DNS errors in Proxy for GET properly

add ProxyAuthentication to http proxy proxy requests

change test support methods to private, remove one unused method, clarify comments a bit

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
@@ -27,6 +27,8 @@ import io.vertx.core.impl.ContextImpl;
 import io.vertx.core.impl.VertxInternal;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
+import io.vertx.core.net.ProxyOptions;
+import io.vertx.core.net.ProxyType;
 import io.vertx.core.net.impl.SSLHelper;
 import io.vertx.core.spi.metrics.HttpClientMetrics;
 import io.vertx.core.spi.metrics.Metrics;
@@ -35,6 +37,7 @@ import io.vertx.core.spi.metrics.MetricsProvider;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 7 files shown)


alexlehm:
new pr replacing #1548

src/test/java/io/vertx/test/core/ConnectHttpProxy.java
```diff
@@ -56,49 +60,91 @@ public void start(Vertx vertx, Handler<Void> finishedHandler) {
     options.setHost("localhost").setPort(PORT);
     server = vertx.createHttpServer(options);
     server.requestHandler(request -> {
+      log.debug("request "+request.absoluteURI());
```
vietj: remove test log

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -710,6 +713,24 @@ private HttpClientRequest doRequest(HttpMethod method, String host, int port, St
...
     checkClosed();
+    ProxyOptions proxyOptions = options.getProxyOptions();
+    if (!options.isSsl() && proxyOptions != null && proxyOptions.getType() == ProxyType.HTTP) {
+      if (headers == null) {
+        headers = MultiMap.caseInsensitiveMultiMap();
+      }
+      relativeURI = "http://" + host + (port != 80 ? ":" + port : "") + relativeURI;
+      host = proxyOptions.getHost();
+      port = proxyOptions.getPort();
+      log.debug("changing request to proxy request " + relativeURI);
```
vietj: remove log

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -710,6 +713,24 @@ private HttpClientRequest doRequest(HttpMethod method, String host, int port, St
...
+      log.debug("username "+proxyOptions.getUsername()+" password "+proxyOptions.getPassword());
+      if (proxyOptions.getUsername() != null && proxyOptions.getPassword() != null) {
+        log.debug("adding authorization header");
+        headers.add("Proxy-Authorization", "Basic " + Base64.getEncoder()
+            .encodeToString((proxyOptions.getUsername() + ":" + proxyOptions.getPassword()).getBytes()));
+      }
+      // TODO: setting the host here does not work yet
+      headers.add("Host", host + (port != 80 ? ":" + port : ""));
```
vietj: what is the purpose of this line ? normally the host header is set at connect time because it's not the same for HTTP and HTTP/2 using HttpClientRequestImpl hostHeader() method

alexlehm: I added that initially to set the host header but it didn't have an effect, you are right, the header should be set in the request impl

alexlehm:
replaced by https://github.com/eclipse/vert.x/pull/1550

Issue #539, opened 2013-03-21
Json.decodeValue should not return Object


matzew:
Whould be nice improvement if the `Json.decodeValue` (see https://github.com/vert-x/vert.x/blob/master/vertx-core/src/main/java/org/vertx/java/core/json/impl/Json.java#L53) would use generics, e.g. like:

```
 public static <T> T decodeValue(String str, Class<?> clazz) throws DecodeException;
```

PR #1556, opened 2016-08-02
Added Json#decodeValue with generic types support

grossws:
See also #539

Usage:

``` java
Map<String, List<String>> map = Json.decodeValue(str, new TypeReference<Map<String, List<String>>(){});
```

Signed-off-by: Konstantin Gribov grossws@gmail.com

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -18,6 +18,7 @@ package io.vertx.core.json;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
@@ -77,8 +78,15 @@ public class Json {
   public static <T> T decodeValue(String str, Class<T> clazz) throws DecodeException {
     try {
       return mapper.readValue(str, clazz);
+    } catch (Exception e) {
+      throw new DecodeException("Failed to decode:" + e.getMessage());
     }
-    catch (Exception e) {
+  }
+
+  public static <T> T decodeValue(String str, TypeReference<T> type) throws DecodeException {
+    try {
+      return mapper.readValue(str, type);
+    } catch (Exception e) {
       throw new DecodeException("Failed to decode:" + e.getMessage());
     }
   }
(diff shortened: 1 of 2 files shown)


vietj:
@grossws can you add a small test for this feature ?

aschrijver:
@vietj When honouring this PR `vertx-core` will directly expose `jackson` types in its public API, and thus be tightly bound, just as is the case with https://github.com/eclipse/vert.x/pull/1274 where I made this comment https://github.com/vert-x3/issues/issues/169#issuecomment-237160014 on it.

For both PR's I would argue it to be better to have https://github.com/vert-x3/issues/issues/166 and/or https://github.com/vert-x3/issues/issues/164 first, then expose these features on a jackson-specific json implementation, a module that can be used as a plug-in replacement to default Vert.x json implementation (which exposes no particular external json library in its API).

vietj:
that's a good point @aschrijver , @grossws can you tell us what are the primary use cases for this ?

grossws:
@vietj primary usecase is deserializing collections of POJOs. I'll add simple test to PR to demonstrate this issue.

If we have simple POJO `public class SimplePojo { public String value; }` and json string `[{"value":"test"}]` and can't deserialize it as `List<SimplePojo>`: current `Json.decodeValue("[{\"value\":\"test\"}]", List.class)` will produce `ArrayList` which holds `LinkedHashMap` and there's no way to pass info that we need `List<SimplePojo>`. If I use jackson api I can get what I need:
```java
List<SimplePojo> pojos = Json.mapper.readValue("[{\"value\":\"test\"}]", new TypeReference<List<SimplePojo>>(){});
Assert.assertTrue(pojos.get(0) instanceof SimplePojo);
```

@aschrijver, currently Json exposes *public* fields `mapper` and `prettyMapper` of type `com.fasterxml.jackson.databind.ObjectMapper`. It binds `Json` object to jackson as well. It's better to move this part to another module and add dependency to `vertx-core`. Anyway, it *is* bound to jackson now and I see no way to decouple it and keep this part lightweight.

aschrijver:
Yes, I see it now. Overlooked it assuming them to be private. Damage already done, then, I assume..

vietj:
Discussable... We also expose netty types and that s not a bad thing

grossws:
Added test to pull request.

aschrijver:
Hmm yes, although I would consider Netty much more as a core system component of Vert.x than the Json library that is used.
I guess, given that currently Jackson is already exposed in `Json` class, these PR's can just be merged. This can be independent of a choice for vert-x3/issues#166 and vert-x3/issues#164 in future release.

grossws:
I prefer to have such method *if* `Json` class is provided. If it will be in `vertx-common`/`vertx-util`/whatever -- so be it.

Issue #1558, opened 2016-08-02
Safe jackson module registration


grossws:
Currently `mapper` and `prettyMapper` can be customized because they are `public static` fields but such customization can be a bit fragile (because of visibility issues) even if no concurrent modification happens. 

I propose to add method `Json.customize(Consumer<? super ObjectMapper> customizer)` which will pass `mapper` and `prettyMapper` to customizer to allow user safe customization (like registering modules for e.g. guava serialization/deserialization) and either 
- make these field `public volatile` (to preserve backward compatibility for ones who modified mappers in these fields);
- make them `private volatile` and provide accessors;
- make them `private`, provide accessors and use them under syncronized/rwlock.

Also I propose register modules from `com.fasterxml.jackson.databind.Module` SPI by default for `mapper` and `prettyMapper` using `ObjectMapper#findAndRegisterModules` method.

If this proposal makes sence for project maintainers I can make pull requests.

vietj:
Hi,

I would be happy to have such request for visiblity concerns

however I think we should first try to understand how Jackson encourages to json mapper safe publication between threads.

is that something you can investigate @grossws ?

grossws:
IIRC, `ObjectMapper` is effectively thread-safe after all configuration done. In CDI apps it's usually is application scoped which allows safe publication through CDI's synchronization. I'll look into it to recheck that I recall true and share in this ticket.

PR #1559, opened 2016-08-02
Find Jackson modules from SPI and register in Json

grossws:
Use jackson SPI for modules to find and register them in `Json#mapper` and `Json#prettyMapper`. This allows user to add things like `com.fasterxml.jackson.datatype:jackson-datatype-guava` or `com.fasterxml.jackson.module:jackson-module-jaxb-annotations` to classpath and get serializers/deserializers/introspectors/etc available from `Json`. 

Currently `Json#decodeValue` fails on things like guava types and `mapper`/`prettyMapper` customization is quite inconvenient.

Related to #1558

Signed-off-by: Konstantin Gribov grossws@gmail.com

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -19,6 +19,7 @@ package io.vertx.core.json;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.Module;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.SerializerProvider;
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/json/Json.java
```diff
@@ -51,6 +52,10 @@
     prettyMapper.configure(JsonParser.Feature.ALLOW_COMMENTS, true);
     prettyMapper.configure(SerializationFeature.INDENT_OUTPUT, true);
 
+    List<Module> modules = ObjectMapper.findModules(Thread.currentThread().getContextClassLoader());
```
cescoffier: What happen if the TCCL is `null` ?

cescoffier: What happens if the TCCL is `null` ?

grossws: Fixed with using `Json.class.getClassLoader` if TCCL is null. Json's CL can't be null since it's not primitive or void type and isn't loaded by bootstrap classloader.

vietj:
not a good default for a library behavior, users should register them manually or have this snippet of code to achieve it in an automated manner.

PR #1561, opened 2016-08-03
DatagramSocket should use unpooled buffers - fixes #1554

### LINKED ISSUE #1554: DatagramSocket should use unpooled buffers
DatagramSocket copies the direct buffer to an pooled heap buffer - it should use an unpooled heap buffer instead.

Workaround : use `-Dio.netty.allocator.type=unpooled`

vietj:


tsegismont:
LGTM but I can't merge.

Issue #1525, opened 2016-07-19
Load the ndots value on linux from /etc/resolv.conf and allow a value of 0


poiuytrez:
I have issues to run a webserver connected to a redis server with Vert.x 3.3.2 in a docker environment. I get the following error: 
```
webserver               | java.net.UnknownHostException: failed to resolve 'my-redis.home'. Exceeded max queries per resolve 3
webserver               | 	at io.vertx.core.dns.impl.fix.DnsNameResolverContext.finishResolve(DnsNameResolverContext.java:435)
webserver               | 	at io.vertx.core.dns.impl.fix.DnsNameResolverContext.tryToFinishResolve(DnsNameResolverContext.java:362)
webserver               | 	at io.vertx.core.dns.impl.fix.DnsNameResolverContext$2.operationComplete(DnsNameResolverContext.java:160)
webserver               | 	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:518)
webserver               | 	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:492)
webserver               | 	at io.netty.util.concurrent.DefaultPromise.notifyListenersWithStackOverFlowProtection(DefaultPromise.java:431)
webserver               | 	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:420)
webserver               | 	at io.netty.util.concurrent.DefaultPromise.setSuccess(DefaultPromise.java:99)
webserver               | 	at io.vertx.core.dns.impl.fix.DnsQueryContext.setSuccess(DnsQueryContext.java:197)
webserver               | 	at io.vertx.core.dns.impl.fix.DnsQueryContext.finish(DnsQueryContext.java:180)
webserver               | 	at io.vertx.core.dns.impl.fix.DnsNameResolver$DnsResponseHandler.channelRead(DnsNameResolver.java:680)
webserver               | 	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:348)
webserver               | 	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:334)
webserver               | 	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:326)
webserver               | 	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)
[...]

PR #1460, opened 2016-06-06
Log the error when client isn't able to connect to remote host.

manish-panwar:
https://groups.google.com/forum/#!searchin/vertx/log$20the$20error/vertx/fwZNJDR6IFI/Va_GTtZ0BAAJ 

We are using a cluster of 2 nodes and intentionally blocked the port 41232 (we are using a dedicated port for Vertx communication). Since port is blocked, we get connection timeout error and "else" condition in given below code is executed.

PR #1562, opened 2016-08-04
better logging when eventbus failed

jan-zajic:
Currently exception is "eaten" when event bus connecting to another node in cluster failed and is no way to find error reason without remote debuggging of application. This makes it very difficult to find bugs like #1525 or configuration and network problems. It's very annoying, so i suggest to log it. if you're not interested, you can disable login for class ConnectionHolder by configuration (although I do not know why).

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
@@ -57,6 +57,7 @@ class ConnectionHolder {
       if (res.succeeded()) {
         connected(res.result());
       } else {
+    	log.warn("Cannot connect to "+serverID, res.cause());
         close();
       }
     });


manish-panwar:
@jan-zajic  Is it possible to use parameterized logging  
logger.warn("Some error {}", someArgument); ?

Also, can you please also print the port in log, not just the host. If you don't define the cluster-port then Vertx will pick a default port, so it will be good to print the port as well.

jan-zajic:
I think parameterized logging is problematic because vert.x logger use different params placeholder format based on used delegate - see documentation at http://vertx.io/docs/vertx-core/java/#_logging : 
> 
Notice that, the provided delegate for Log4J 1 does not support parameterized message. The delegate for Log4J 2 uses the {} syntax like the SLF4J delegate. JUL delegate uses the {x} syntax.

I explore through vert.x core sources and parameterized loggers are not used, there are similar messages like:
> 
"Failed to propagate registration for handler " + handler + " and address " + address)
 

Also, i print both host and port. I print serverID, it's object whose toString method looks like:
> 
 return host + ":" + port;

manish-panwar:
@jan-zajic Cool !!!

jan-zajic:
@vietj what do you think about this PR, any plans to merge it? If there is anything else we can do for passing please let me know.

manish-panwar:
@vietj Can we please review/merge this PR ?

manish-panwar:
@vietj @purplefox Can you guys please merge this PR? We are waiting on this change to be merged for more than 6 months.

jan-zajic:
@manish-panwar i try to mediate that, but please close that duplicate PRs (#1459, #1460), so we are more clear

manish-panwar:
@jan-zajic - Done. Can you please take a look at it and fix the merge conflict?

PR #1576, opened 2016-08-23
Allow tuning for gzip/deflate compression level (for performance)

creativeprogramming:
Allowing Vert.Xers to cofigure their webserver deflater compression level.

Inspired by the nginx equivalent: http://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_comp_level

Usage example:

``` java
HttpServerOptions()
                .setCompressionSupported(true)
                .setCompressionLevel(1) /* valid range 1-9 (1: fastest, less compression - 9 slower, best (pedantic) compression )*/
```

**Note:**
Before this PR the .setCompressionSupported(true) was enabling the netty's HttpCompressor default compression level that is 6.

In my proposal i  set the **default compression level to 1** (fastest, not the best size saver but anyway very good compression the modern web usage case (in my experience), less cpu cylce wasting, less http requests stucking in 'waiting state')

**Why this PR?**
1. Making vert.x flexible in this configuration too cause in order to demonstrate we have nothing to envy to other webservers like nginx ;)
2. Saving server resource usage (less cpu time for compression = more cpu time for the event loop, less overwarming ;) )
3. Reducing latency TTFB for static assets when compression is enabled

**A bit more  rationale:**

A best practice suggested by any moder site 'speed test' is to enable compression, but - as a drawback - enabling it we consume more cpu cycles at server end, and if we tune up it uncarefully we can end up - paradoxically - to get a worst time-to-first-Byte (TTFB) and even a worst total request time with compression enabled compared to no compression at all.

A best pratice to avoid such issue and achieve the better performance compromise, is to tune  compressor to do the 'bare minimal' work and don't overkillig, in other words, remembering that our target is **reducing latency** at the very minimum level and not really reducing 'file size'.

From my personal experience with nginx and apache too, and confirmed by many online guides, the best value for gzip compression level is 1 or 2 for a web server. Going higher that than just saves some bits, but pay back it hardly in server resource usages and processing latency (when compression is not cached but done on-the-fly) 

Bonus: a further performance improvement can be to implement a compression cache layer in vertx but this is uncommon and complex also in other webservers so not a priority.

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServerOptions.java
+++ b/src/main/java/io/vertx/core/http/HttpServerOptions.java
@@ -43,6 +43,11 @@ public class HttpServerOptions extends NetServerOptions {
    * Default value of whether compression is supported = false
    */
   public static final boolean DEFAULT_COMPRESSION_SUPPORTED = false;
+  
+  /**
+   * Default gzip compression level = 1, fastest
+   */
+  public static final int DEFAULT_COMPRESSION_LEVEL = 1;
 
   /**
    * Default max websocket framesize = 65536
@@ -85,6 +90,7 @@ public class HttpServerOptions extends NetServerOptions {
   public static final int DEFAULT_HTTP2_CONNECTION_WINDOW_SIZE = -1;
 
   private boolean compressionSupported;
+  private int compressionLevel;
   private int maxWebsocketFrameSize;
   private String websocketSubProtocols;
   private boolean handle100ContinueAutomatically;
@@ -112,6 +118,7 @@ public class HttpServerOptions extends NetServerOptions {
   public HttpServerOptions(HttpServerOptions other) {
     super(other);
     this.compressionSupported = other.isCompressionSupported();
+    this.compressionLevel = other.getCompressionLevel();
     this.maxWebsocketFrameSize = other.getMaxWebsocketFrameSize();
     this.websocketSubProtocols = other.getWebsocketSubProtocols();
     this.handle100ContinueAutomatically = other.handle100ContinueAutomatically;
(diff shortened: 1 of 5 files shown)


creativeprogramming:
I'm running this in production since 7 days, all ok, anyway needs testing.

vietj:
indeed we need a test and also this is only implemented for HTTP/1 and not HTTP/2 so we need also HTTP/2 to use this setting.

creativeprogramming:
Right, I completley missed http/2 here, thank you for pointing out I need it.

I'll update this when ready for http/2 too and maybe with some benchmarks.

Cheers.

jeveloper:
hey @vietj  Julien , curious if this is going to be merged , essp for http/2

thanks

vietj:
@jeveloper the feature is still missing tests and I think also support on the client side - beside that it looks good indeed - @creativeprogramming ?

creativeprogramming:
@vietj support on client side is native by browsers, it's missing tests and support for http2. If you want to help me in completing this PR please give some hint to check where in the codebase http2 gzip/deflate is currently implemented.

vietj:
@creativeprogramming 

 I actually was thinking if we would ever support `HttpClient` sending a compressed body but for now it is not implemented so no need for client slide support.

However indeed there is not test and HTTP/2 is not supported : look at `VertxHttp2ConnectionHandlerBuilder`

creativeprogramming:
Thanks, I'm pushing http/2 support

vietj:
can you make a new PR with a single commit (squashed) sign-off ?

vietj:
I mean that because the CLA checker seems confused

creativeprogramming:
Yes thanks better if i open again a PR from scratch, I'm going crazy trying to fix it with git

vietj:
I can see that ^^ !!

creativeprogramming:
Anyway I'll do it later when i finish all, now I just pushed http/2 implementation and http1.x tests, I  still miss http/2 tests.

Anyway all works fine, I'm testing in production too with http/2.

I'll open an issue just to keep track of this discussion too

PR #1582, opened 2016-08-28

LoneRifle:
* Adjust LocalMapImpl ctor and SharedDataImpl.localMaps such that the key parameter of the latter is a String

* Use computeIfAbsent for all getLocalXXX implementations. This avoids the overhead associated with creating a speculative object for use as an arg in putIfAbsent, while retaining operation atomicity. This implementation is also more concise


cescoffier:
@LoneRifle unfortunately we can't re-run the check. You would need to close the PR (mark it as invalid) and re-open the "same" one. Sorry for this inconvenient process.

PR #1584, opened 2016-08-29

LoneRifle:
* Adjust LocalMapImpl ctor and SharedDataImpl.localMaps such that the key parameter of the latter is a String

* Use computeIfAbsent for all getLocalXXX implementations. This avoids the overhead associated with creating a speculative object for use as an arg in putIfAbsent, while retaining operation atomicity. This implementation is also more concise


LoneRifle:
I'm _really_ confused as to what's going on. I've signed the CLA and tagged my account with my GitHub id but your CI job insists on identifying me by my github mail address, which doesn't really exist.

PR #1589, opened 2016-08-29

LoneRifle:
* Adjust LocalMapImpl ctor and SharedDataImpl.localMaps such that the
key parameter of the latter is a String

* Use computeIfAbsent for all getLocalXXX implementations. This is
avoids the overhead associated with creating a speculative object for
use as an arg in putIfAbsent, while retaining operation atomicity. This
implementation is also more concise


LoneRifle:
Something is

PR #1591, opened 2016-08-29
SharedDataImpl: Use ConcurrentMap#computeIfAbsent to atomically get-or-create objects

LoneRifle:
- Adjust LocalMapImpl ctor and SharedDataImpl.localMaps such that the
  key parameter of the latter is a String
- Use [ConcurrentMap#computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html#computeIfAbsent-K-java.util.function.Function-) for all getLocalXXX implementations. This avoids
  the overhead associated with creating a speculative object for use as an
  arg in putIfAbsent, while retaining operation atomicity. This
  implementation is also more concise


LoneRifle:
Could I humbly seek this PR to be reviewed possibly for inclusion into the 3.4.x series of releases?

vietj:
@LoneRifle yes I think it definitely can!

PR #1593, opened 2016-08-30
always remove subs when removing node in clusterMap

pontushellgren:
When killing multiple nodes with kill -9, all registered message consumers for the killed nodes aren't removed from the distributes subs map.

Further information is in this issue:
https://github.com/vert-x3/vertx-hazelcast/issues/13

Signed-off-by: pontushellgren <phellgren@gmail.com>


cescoffier:
looks good to me.

cescoffier:
(we will need to run all cluster manager test suites once this is merged)

PR #1594, opened 2016-08-30
Implement High Fault Tolerance: Vertx Cluster is fault tolerant when several nodes are killed

polipodi:
Hi all,

We are big fans of VertX! We are working on setuping a cluster of at least 20 Vertx JVM in a single cluster. When starting playing killing nodes, more than one node in few seconds, we found that the vertx event bus cluster becomes unstable. After drilling down into the code, we found that the clusterMap of HAManager, and subs multimap of the ClusteredEventBus are often corrupted:
- sometimes, references on nodes killed are still present in clusterMap and subs multimap
- sometimes, references on nodes still running are missing in clusterMap and subs multimap

So we suggest to make sure to clean the maps from each node and not only from an elligible node. And we also suggest that each node is sending every 10 seconds their key/value into the clusterMap and subs multimap. 

The main benefits are that when the nodes are killed, after few seconds, the map are correctly cleaned-up, and running nodes are populating again the maps. Even under load (we tested it), the cluster becomes available again and 100% operational. We tested up to 20 running nodes and 80% of nodes loss, while the system were under load.
The main drawback is that each node is responsible for sending (every 10s) their key/value constantly in the clusterMap and the subs multimap.

So we propose to enable this behavior when the HFT is enabled from VertxOptions (High Fault Tolerance), since it will add additional traffic on the cluster which is not always necessary depending the deployment. (by default set to false). We have also the capability to change the delay of sending data to maps (HFTInterval) which is 10s by default.

Please comment, accept, deny or anything else about this pull request. We are strongly considering Vertx at the heart of our new infra for deploying micro-services, and the fault tolerance is only thing that we see as a showstopper.

Same issue already raised by other users: 
- https://github.com/eclipse/vert.x/pull/1593
- https://github.com/vert-x3/vertx-hazelcast/issues/13

### DIFF
--- a/src/main/java/io/vertx/core/impl/launcher/commands/RunCommand.java
+++ b/src/main/java/io/vertx/core/impl/launcher/commands/RunCommand.java
@@ -46,6 +46,7 @@ public class RunCommand extends BareCommand {
 
   protected boolean cluster;
   protected boolean ha;
+  protected boolean hft;
 
   protected int instances;
   protected String config;
@@ -72,6 +73,18 @@ public class RunCommand extends BareCommand {
       "fail over to any other nodes in the cluster started with the same HA group.")
   public void setHighAvailability(boolean ha) {
     this.ha = ha;
+  }
+  
+  /**
+   * Enables / disables the high fault tolerance
+   *
+   * @param hft whether or not to enable the HFT.
+   */
+  @Option(longName = "hft", acceptValue = false, flag = true)
+  @Description("If specified the verticle will be deployed as a high availability (HA) deployment. This means it can " +
+      "fail over to any other nodes in the cluster started with the same HA group.")
+  public void setHighFaultTolerance(boolean hft) {
+    this.hft = hft;
   }
 
   /**
@@ -237,6 +250,11 @@ public class RunCommand extends BareCommand {
   public boolean getHA() {
     return ha;
   }
+  
+  @Override
+  public boolean getHft() {
+	return hft;
+}
 
   /**
    * Starts vert.x and deploy the verticle.
@@ -357,6 +375,9 @@ public class RunCommand extends BareCommand {
     }
     if (quorum != -1) {
       args.add("--quorum=" + quorum);
+    }
+    if (hft) {
+        args.add("--hft");
     }
     if (classpath != null && !classpath.isEmpty()) {
       args.add("--classpath=" + classpath.stream().collect(Collectors.joining(File.pathSeparator)));
--- a/src/main/java/io/vertx/core/VertxOptions.java
+++ b/src/main/java/io/vertx/core/VertxOptions.java
@@ -114,6 +114,16 @@ public class VertxOptions {
   public static final boolean DEFAULT_HA_ENABLED = false;
 
   /**
+   * The default value of HFT enabled = false
+   */
+  public static final boolean DEFAULT_HFT_ENABLED = false;
+  
+  /**
+   * The default value of the HFT messages when HFT is enabled, in millis
+   */
+  public static final long DEFAULT_HFT_INTERVAL = 10000;
+
+  /**
    * The default value of warning exception time 5000000000 ns (5 seconds)
    * If a thread is blocked longer than this threshold, the warning log
    * contains a stack trace
@@ -523,6 +533,48 @@ public class VertxOptions {
    */
   public VertxOptions setHAEnabled(boolean haEnabled) {
     this.haEnabled = haEnabled;
+    return this;
+  }
+  
+  /**
+   * Will HFT (High Fault Tolerance) be enabled on the Vert.x instance?
+   *
+   * @return true if HFT enabled, false otherwise
+   */
+  public boolean isHFTEnabled() {
+    return eventBusOptions.isHFTEnabled();
+  }
+  
+  /**
+   * Set whether HFT (High Fault Tolerance) will be enabled on the Vert.x instance.
+   *
+   * @param hftEnabled true if enabled, false if not.
+   * @return a reference to this, so the API can be used fluently
+   */
+  public VertxOptions setHFTEnabled(boolean hftEnabled) {
+	  eventBusOptions.setHFTEnabled(hftEnabled);
+	  return this;
+  }
+  
+  /**
+   * Get the value of cluster HFT (High Fault Tolerance) messages interval, in ms.
+   * <p>
+   * In order to guarantee consistency of vertx internal maps (HFT enabled), cluster map and subscribers map need to be constantly resynchronized, at each interval
+   *
+   * @return the value of cluster ping reply interval
+   */
+  public long getHFTInterval() {
+	  return eventBusOptions.getHFTInterval();
+  }
+  
+  /**
+   * When HFT (High Fault Tolerance) is enabled, set the interval between internal vertx synchronization
+   *
+   * @param hftInterval interval in miliis
+   * @return a reference to this, so the API can be used fluently
+   */
+  public VertxOptions setHFTInterval(long hftInterval) {
+    eventBusOptions.setHFTInterval(hftInterval);
     return this;
   }
 
--- a/src/main/java/io/vertx/core/eventbus/EventBusOptions.java
+++ b/src/main/java/io/vertx/core/eventbus/EventBusOptions.java
@@ -37,6 +37,8 @@ public class EventBusOptions extends TCPSSLOptions {
   private int clusterPublicPort = VertxOptions.DEFAULT_CLUSTER_PUBLIC_PORT;
   private long clusterPingInterval = VertxOptions.DEFAULT_CLUSTER_PING_INTERVAL;
   private long clusterPingReplyInterval = VertxOptions.DEFAULT_CLUSTER_PING_REPLY_INTERVAL;
+  private boolean hftEnabled = VertxOptions.DEFAULT_HFT_ENABLED;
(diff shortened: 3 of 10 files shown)


cescoffier:
Hi,

Can't it be done by configuring the underlying component (for example Hazelcast in the HZ cluster manager). It has configuration variables to specify the ping time (http://docs.hazelcast.org/docs/3.5/manual/html/systemproperties.html). For example `hazelcast.heartbeat.interval.seconds` may make it, no ?

polipodi:
Hi,

For HZ, what I've seen is that when a node is killed, backup partitions are migrated to other nodes. 
But if the node containing the master partition and the node containing the backup partition are killed before migration of the partition, partition data is lost.
HZ 3.7 tries to solve the issue, check this: https://github.com/hazelcast/hazelcast/issues/5444
But even with HZ 3.7, in our tests, the clusterMap and subs multimap are still inconsistent in some situation. 
So the only way to keep them consistent that I see is to guarantee that running nodes are cleaning and populating again the maps. Else do you have another idea to solve this issue ?

We also give a try with Apache Ignite Cluster Manager, but with no luck, same result. So that's why we believe that since Vertx is storing "infrastructure" data in the Cluster Manager, this is the responsibility of Vertx itself to keep them consistent.

Thanks for your comments

melke:
👍  We are currently running our environment using this PR branch. Without it, vert.x is more or less unusable due to instability and corruption of the vert.x subs map when we autoscale our services. So I vote for merging this.

jmlong94:
I am using this solution as well, and it is working great so far. I have referenced this pull request in the Google Groups thread I started. https://groups.google.com/d/msg/vertx/7FOO1BZG-DA/OGrJsryQBwAJ

src/main/java/io/vertx/core/impl/launcher/commands/RunCommand.java
```diff
@@ -73,6 +74,18 @@
...
     this.ha = ha;
   }
+  
+  /**
+   * Enables / disables the high fault tolerance
+   *
+   * @param hft whether or not to enable the HFT.
+   */
+  @Option(longName = "hft", acceptValue = false, flag = true)
+  @Description("If specified the verticle will be deployed as a high availability (HA) deployment. This means it can " +
```
jmlong94: This description should be updated?

vietj:
we will make a thorough review of this after F2F meeting - @tsegismont will take care of it

cazacugmihai:
+1 for this pull request (or for any other way that will fix the issue)

ashertarno:
@vietj is there any eta for this to be merged? thanks.

vietj:
we are waiting for a general review and brainstorming from @tsegismont

ashertarno:
is there any update on this? sorry for nudging, we just have to make a decision if we implement it by ourselves or wait for it to be fixed in the next version. thanks.

vietj:
@ashertarno we are investigating - it will be fixed for sure

vietj:
@ashertarno what does "we implement it by ourselves" mean by the way ?

ashertarno:
@vietj i mean we have to either merge polipodi's solution and wait until the official fix or don't waste our time on merging as the official fix is on its way. this issue is a pretty much showstopper for us, as we see cluster nodes getting lost after restarts on our staging servers too often, something we can't live with in production environment.
thanks.

vietj:
can you give details on how you restart your nodes ?

ashertarno:
we tried both regular kill coming from our script that stops the service and kill -9. 
there is not something we can reproduce again and again, but sometimes the killed node seems to be not cleaned from the cluster and messages still targeting it. this is for sure, as failed sendings correspond to round robin order.

jmlong94:
@ashertarno, we have been using a locally compiled version vertx 3.3.3 with this pull request merged in, and we are still experiencing this behavior in production even though it appeared to be fixed at first. @vietj, we are running our cluster as one node per AWS EC2 instance, and the nodes are added removed by starting/stopping instances. I believe Amazon sends a regular kill signal to all processor running on the instance when shutting it down.

ashertarno:
@jmlong94: thanks for the update. if it's important, we are running on simple multi server cluster, no ec2 involved. tried both broadcast and TCP, happens in both cases as far as i remember.

pochadri:
can somebody confirm if this issue would be fixed in next version of vertx?

this is a show stopper for us as we frequently upgrade the services so shutdown and starting is very common for us.

thank u

tsegismont:
Hi,

I am working on the issue, and the goal is to get a fix ready for the next
version.

Regards,
Thomas

2016-12-07 5:07 GMT+01:00 pochadri <notifications@github.com>:

mhstnsc:
@polipodi 
Hazelcast can notify if partitions are lost in which case isn't it enough to reconstruct the maps only after this notification?
http://docs.hazelcast.org/docs/3.5/manual/html/distributedevents.html

bdarwin:
@mhstnsc this happens on ignite cluster as well. In my opinion fix should be agnostic to cluster manager.

vietj:
my opinion is that the FakeClusterManager should have an option to be inconsistent like other cluster manager can be and tests should be written testing the clustered EventBus using inconsistent behavior of the CM (like for example not evicting a dead node).

nakshathri:
Hello all, we are also facing this issue in our stage and prod env. Restarting every time for clean state is really a big pain. Is this fix pushed in the latest vertx ?

vietj:
@nakshathri not yet , we are working on it and it should be released early March

nakshathri:
That's great. Eagerly waiting for this fix.

tsegismont:
@polipodi thanks for your contribution. I'm closing this one as the issue is fixed via #1848

nakshathri:
@tsegismont, can you please confirm the fix version ?

tsegismont:
3.4.0

2017-03-09 10:30 GMT+01:00 nakshathri <notifications@github.com>:

SerLongfellow:
Thanks so much for all the hard work, guys!

tsegismont:
An important part of the credit goes to Luc and Rikard who did great with
their investigations and patches.

bdarwin:
I can still see this issue on 3.4.0 with Hazelcast. Anyone tried?

tsegismont:
I can't reproduce anymore with the original test. Besides there's a test
for data loss now in the cluster test suite. If you still experience with
Hazelcast, could you open an issue (with a reproducer) in the Hazelcast
module? Thanks

2017-03-14 9:20 GMT+01:00 bdarwin <notifications@github.com>:

SerLongfellow:
The behavior I'm seeing with 3.4.1 is as follows: 

- Several nodes are killed at once
- The timeout behavior is seen as was before at first, but after a few minutes it appears to be corrected and functionality is restored without having to entirely restart the cluster

Is this the desired behavior? I'm assuming either some sort of health check was implemented or it simply takes a bit for the cluster manager to catch up?

Either way it's great that the cluster doesn't have to be restarted now!

tsegismont:
Thanks for your feedback Jeff!

What you're seeing is indeed the desired behavior. The time between the
nodes die and the cluster is fully functional is due to:
- the cluster manager engine (Hazelcast/Infinispan) making sure the node
actually died
- the cluster manager engine rebalancing data
- Vert.x ClusteredEventBus removing the stale subscriptions / re-adding
valid ones in case of data loss

The more nodes you lose/add at the same time, the longer the process.

And yes, the main goal was that the whole cluster shouldn't have to be
restarted. There could still be some odd cases where the cluster does not
self heal but restarting one node should fix it.


2017-03-22 17:38 GMT+01:00 Jeff Long <notifications@github.com>:

tvb:
@tsegismont we face major problems with vertx/hazelcast on AWS when multiple nodes died. How and where can I submit an issue for this topic?

tsegismont:
Have you sent an email to the user group explaining your problem with some
details on the setup?

2017-11-22 16:52 GMT+01:00 Tristan <notifications@github.com>:

tsegismont:
@nakshathri Hazelcast Client instances are not supported (see https://vertx.io/docs/vertx-hazelcast/java/#_using_an_existing_hazelcast_cluster)

If you want to avoid having data on Vert.x nodes, make the Vert.x node a [lite member](http://docs.hazelcast.org/docs/latest/manual/html-single/#enabling-lite-members) (member that does not store data).

nakshathri:
@tsegismont : Thank you for that info. But i had gone through doc ( https://vertx.io/docs/vertx-hazelcast/java/#_using_an_existing_hazelcast_cluster). As per this doc, data lose can happen only when we have enabled high availability. But in our case, we have not enabled HA. Do you still think that we can use hazelcast client?  

> IMPORTANT Do not use Hazelcast clients or smart clients when using high-availability (HA, or fail-over) in your cluster as they do not notify when they leave the cluster and you may loose data, or leave the cluster in an inconsistent state

Will try with having hazelcast lite members which are getting registered to an existing hazelcast cluster and will update you the result.

tsegismont:
The doc should be updated, Hazelcast clients or smart clients are not
supported even when HA is not used.

2018-07-27 8:11 GMT+02:00 nakshathri <notifications@github.com>:

nakshathri:
Making the instances a lite member works. Thanks a lot @tsegismont .

siddshadab:
Still facing data loss issue in 3.12.2, making two app as lite members with 3 replicas running and one app as data member with 2 replicas .

Vertx : 3.8.2
Hazelcast : 3.12.2
Hazelcast-Kubernates : 1.5.5

Restarted Master node which was data member and found one of lite member lost with some of discoveries data

PR #1611, opened 2016-09-06
JsonObject copy() preserves LinkedHashMap

giovibal:
JsonObject.copy() method, create a new instance of JsonObject with HashMap and not with LinkedHashMap, changing the order of keys.
If a verticle send a JsonObject throught EventBus, now others  receive the same JsonObject with the same order of keys that was sent.


pmlopes:
just a small comment, this is not a bug, according to the spec http://json.org:

> An object is an unordered set of name/value pairs.

Users should not make assumptions on the order of the keys for several reasons:

* vertx nodes in other languages than java will use language specific encode/decode which will not guarantee the order
* bridge members from e.g.: sockjs will receive json encoded by browsers which will not preserve order
* nodejs, C++, python, C# bridges will also not guarantee the order

It could become a false assumption to say that the order is preserved. However swapping from HashMap to LinkedHashMap could be considered for performance issues but **not** for the order.

sodomizer:
At least the order will be persistent _inside_ a pure java environment (and also in JS using Map object). _External_ components are still to be considered anyway. 
This is neither the first nor the last such issue/pull request as this feature is really demanded by developers. It will be better to make ordering persistent just for java and emphasize ordering issues in docs than leave it broken completely.
Add six characters and get thousands of happy developers or let it be and get bad carma and dozens of same issues and useless forks. Not much of a difficult decision.

tsegismont:
It's not completely broken. But anyway I agree that for consistency we
should try to preserve order in the copy method.

2016-09-13 14:25 GMT+02:00 sodomizer <notifications@github.com>:

vietj:
one thing we could do is test if the current map is a LinkedHashMap and if that's a LinkedHashMap then we use a LinkedHashMap + provide an option to created a JsonObject with a LinkedHashMap (boolean ordered).

giovibal:
We use vert.x in production into some rest/json services.
One of this is consumed from a client that use a 3rd party SDK that - wrongly - use ordered of json keys...  we've discovered this bug after upgrading the rest service to new vert.x.
I'd like to mantains order for "consistency" if possibile... 
Thanks.

vietj:
@giovibal are you having issue with copy() operation as well ?

vietj:
have a look at:

https://github.com/eclipse/vert.x/tree/json-object-preserves-linked-hashmap

tsegismont:
LGTM

2016-09-27 21:56 GMT+02:00 Julien Viet <notifications@github.com>:

PR #1617, opened 2016-09-07
Documentation fix for running bare instances

jponge:
The title pretty much says it I guess 😉

### DIFF
--- a/src/main/asciidoc/java/index.adoc
+++ b/src/main/asciidoc/java/index.adoc
@@ -1118,7 +1118,7 @@ dies. To create a _bare_ instance, launch:
 
 [source]
 ----
-vertx -ha
+vertx bare
 ----
 
 Depending on your cluster configuration, you may have to append the `cluster-host` and `cluster-port` parameters.


src/main/asciidoc/java/index.adoc
```diff
@@ -1118,7 +1118,7 @@ dies. To create a _bare_ instance, launch:
 
 [source]
 ----
-vertx -ha
+vertx bare
```
cescoffier: both should work no ? 
`vertx -ha` is the old way (but still supported)

cescoffier:
@jponge you need to update the package-info.java file. The asciidoc is generated from this file (at compile time)
https://github.com/vert-x3/wiki/wiki/How-is-managed-vert.x-documentation

src/main/asciidoc/java/index.adoc
jponge: It didn't work in my trials.

jponge:
I'll fix it, but I always get confused by generated code under VCS 😉

PR #1620, opened 2016-09-09
Fixes #1619 Too many log generation - LoadBalancer Heartbeat

### LINKED ISSUE #1619: Too many log generation - LoadBalancer Heartbeat
See [[vertx:36914] Too many log generation - LoadBalancer Heartbeat](https://groups.google.com/forum/#!topic/vertx/CF1qTX3Ak1E)

Connection is reset by client before we could determine if it's HTTP 1 or 2.

tsegismont:
Override exceptionCaught, log at trace level and close channel

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -284,7 +284,6 @@ public class HttpServerImpl implements HttpServer, Closeable, MetricsProvider {
                   pipeline.addLast(new Http1xOrHttp2Handler());
                 }
               }
-              HttpServerImpl.this.onChannelInitialized(ch);
             }
         });
 
@@ -342,7 +341,10 @@ public class HttpServerImpl implements HttpServer, Closeable, MetricsProvider {
   }
 
   // Visible for testing
-  protected void onChannelInitialized(Channel ch) {
+  protected void handleConnectionException(ChannelHandlerContext ctx, Throwable cause) {
+    Channel channel = ctx.channel();
+    log.trace("Connection failure", cause);
+    channel.close();
   }
 
   private VertxHttp2ConnectionHandler<Http2ServerConnection> createHttp2Handler(HandlerHolder<HttpHandler> holder, Channel ch) {
@@ -1062,10 +1064,8 @@ public class HttpServerImpl implements HttpServer, Closeable, MetricsProvider {
     }
 
     @Override
-    public void exceptionCaught(ChannelHandlerContext chctx, Throwable cause) throws Exception {
-      Channel channel = chctx.channel();
-      log.trace("Connection failure", cause);
-      channel.close();
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+      HttpServerImpl.this.handleConnectionException(ctx, cause);
     }
   }
 }
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -145,6 +145,7 @@ public class HttpServerImpl implements HttpServer, Closeable, MetricsProvider {
   private ContextImpl listenContext;
   private HttpServerMetrics metrics;
   private boolean logEnabled;
+  private Handler<Throwable> connectionExceptionHandler;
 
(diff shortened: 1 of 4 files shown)


src/test/java/io/vertx/test/core/HttpConnectionEarlyResetTest.java
```diff
@@ -0,0 +1,108 @@
...
+  private CountDownLatch resetLatch = new CountDownLatch(1);
+
+  @Override
+  public void setUp() throws Exception {
+    super.setUp();
+    CountDownLatch listenLatch = new CountDownLatch(1);
+    httpServer = new HttpServerImpl((VertxInternal) vertx, new HttpServerOptions()) {
+      @Override
+      protected void onChannelInitialized(Channel ch) {
+        ch.pipeline().addFirst(new ResetLatchCountDown()).addLast(new ThrowableRecorder());
```
vietj: can we test without having to mess with Netty pipeline ?

vietj: I would rather be in favor of adding a connectionExceptionHandler in HttpServerImpl (not exposed) that would be called by the pipeline, the default implementation closes the channel.

The test can then change this handler to be aware of the failure and test properly.

tsegismont: Just pushed another version. Tell me if this is what you had in mind.

2016-09-13 14:03 GMT+02:00 Julien Viet <notifications@github.com>:

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -304,6 +340,13 @@ protected void configurePipeline(ChannelHandlerContext ctx, String protocol) thr
     return this;
   }
 
+  // Visible for testing
+  protected void handleConnectionException(ChannelHandlerContext ctx, Throwable cause) {
```
vietj: ok so more like:

1. the close is done in Netty's handler
2. not a method to override but instead an `Handler<Throwable>` called `connectionExceptionHandler` that by default logs the cause and can be `set` in the test

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -1019,5 +1063,11 @@ private void http1(ChannelHandlerContext ctx, ByteBuf buf) {
       ctx.fireChannelRead(buf);
       ctx.pipeline().remove(this);
     }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+      ctx.channel().close();
+      HttpServerImpl.this.connectionExceptionHandler.handle(cause);
```
vietj: I think it should use Context#executeFromIO and also test handler is not null.

tsegismont: PTAL

2016-09-13 18:24 GMT+02:00 Julien Viet <notifications@github.com>:

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -1019,5 +1065,13 @@ private void http1(ChannelHandlerContext ctx, ByteBuf buf) {
       ctx.fireChannelRead(buf);
       ctx.pipeline().remove(this);
     }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+      Channel channel = ctx.channel();
+      HandlerHolder<HttpHandler> reqHandler = reqHandlerManager.chooseHandler(channel.eventLoop());
+      reqHandler.context.executeFromIO(channel::close);
```
vietj: actually channel.close should be called normally but it's the callback to connectionExceptionHandler that should be executed with executeFromIO.

tsegismont: Done

2016-09-14 18:25 GMT+02:00 Julien Viet <notifications@github.com>:

vietj:
don't we have the same problem with the ApplicationProtocolNegotiationHandler ?

tsegismont:
No, ApplicationProtocolNegotiationHandler has an exceptionCaught method
(which closes the channel).

2016-09-15 10:41 GMT+02:00 Julien Viet <notifications@github.com>:

PR #1626, opened 2016-09-14
Improve timeout message

marcust:
Give all available information about what was actually trying to be requested
when communicating a timeout.


jponge:
LGTM @vietj

PR #1639, opened 2016-09-22
Use Netty DefaultHttpHeaders by default for performance

testn:
Use DefaultHttpHeaders so that headers in HttpHeaders can be used directly. This helps Netty to serialize header name without having to encode. In addition to that, it can also use AsciiString.hashCode() for hash code for faster hash table lookup.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
@@ -18,6 +18,7 @@ package io.vertx.core.http.impl;
 
 import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.util.AsciiString;
 import io.vertx.core.MultiMap;
 
 import java.util.AbstractList;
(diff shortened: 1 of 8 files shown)


vietj:
@testn have you measured a performance improvement ?

testn:
It offers a few % improvement in request rate by reducing the CPU and memory consumption

testn:
I updated it to contain the similar change to HTTP/2 as well

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -113,29 +98,27 @@ public boolean isEmpty() {
...
+          }
+          @Override
+          public String next() {
+            return it.next().toString();
+          }
+        };
+      }
+      @Override
+      public int size() {
+        return nameSet.size();
```
testn: need to change this to ensure that the size() returns the correct value as you can see from testEntries() test

vietj: good catch, I thought there was a copy like the previous issue

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -35,26 +36,10 @@
...
-        }
-        buffer.setCharAt(index, (char)(c + ('a' - 'A')));
-      }
-    }
-    if (buffer != null) {
-      return buffer.toString();
-    } else {
-      return s;
-    }
+    return AsciiString.of(s).toLowerCase();
```
vietj: this creates an extra AsciiString object.

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
vietj: add a test

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
testn: fixed

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
testn: the test is already on "testMultipleValues"

PR #1644, opened 2016-09-25
Json sugar

aesteve:
Add utility functions to collect results of `java.util.Collection` into a `JsonArray` and a `Map.Entry<K, V>` into a JsonObject.


src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -648,6 +653,17 @@ public JsonObject put(String key, Object value) {
   }
 
   /**
+   * Put an Entry into the JSON object (puts the entry value under the entry key)
+   *
+   * @param entry  the entry
+   * @return  a reference to this, so the API can be used fluently
+   */
+  public<T> JsonObject put(Map.Entry<String, T> entry) {
```
vietj: should not use <T> and use ? instead

vietj: not tested

aesteve:
closing this and making another proposal : `new JsonObjectCollector()` (and with Eclipse CLA OK)

PR #1645, opened 2016-09-25
Json collectors

aesteve:
Add a collector for JsonArray and JsonObject so that users can collect results from streams directly into a JsonArray or JsonObject


src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -634,6 +639,36 @@ public int readFromBuffer(int pos, Buffer buffer) {
     return pos + length + 4;
   }
 
+  /**
+   * Reduce / collects a stream's values into a JsonArray
+   *
+   * @param <T> the type of the elements in the collected stream
+   * @return a Collector, ready to use in Stream.collect(...)
+   */
+  public static<T> Collector<T, JsonArray, JsonArray> collector() {
```
vietj: it <T> really necessary ?

pmlopes:
Fixes vert-x3/issues#184

PR #1646, opened 2016-09-26
Make all MultiMap works the same way

testn:
CaseInsensitiveHeaders/HeadersAdapter/Http2HeadersAdapter seem to work differently as the contract of MultiMap interface is not quite clear. I took that CaseInsensitiveHeaders is supposed to be the most reliable one.

BTW, I actually prefer the behavior of MultiMap interface to match DefaultHeaders in Netty especially when there are multiple entries per one header name.


vietj:
can you explain what you mean by what you prefer ?

testn:
i.e. size() returns entries.size() rather than names.size()

vietj:
MultiMap#size() returns the number of keys per javadoc.

That would anyway break compatiblity.

If you need such value, then you can use `entries().size()` instead.

src/main/java/io/vertx/core/http/impl/HeadersAdaptor.java
```diff
@@ -127,7 +136,44 @@ public MultiMap clear() {
 
   @Override
   public Iterator<Map.Entry<String, String>> iterator() {
-    return headers.iterator();
+    return new Iterator<Map.Entry<String, String>>() {
```
vietj: what's the point of wrapping ?

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -95,10 +96,14 @@ public int size() {
...
   @Override
   public List<Map.Entry<String, String>> entries() {
-    return headers.names()
-        .stream()
-        .map(name -> new AbstractMap.SimpleEntry<>(name.toString(), headers.get(name).toString()))
-        .collect(Collectors.toList());
+    if (isEmpty()) {
+      return Collections.emptyList();
+    }
+    List<Map.Entry<String, String>> list = new ArrayList<Map.Entry<String, String>>(headers.size());
```
vietj: I would prefer to return a list wrapper instead of creating a new ArrayList, beside the returned object must be immutable.

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -95,10 +96,14 @@ public int size() {
...
+    for (Map.Entry<String, String> entry : this) {
```
vietj: this changes the behavior I think, the previous was I think wrong, i.e it was not returning duplicate entries.

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -113,29 +118,27 @@ public boolean isEmpty() {
...
-          };
-        }
-        @Override
-        public int size() {
-          return headers.size();
-        }
-      };
-    }
-    return names;
+    Set<CharSequence> nameSet = headers.names();
```
vietj: why did you remove the caching ?

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -201,7 +204,44 @@ public MultiMap clear() {
 
   @Override
   public Iterator<Map.Entry<String, String>> iterator() {
-    return entries().iterator();
+    return new Iterator<Map.Entry<String, String>>() {
```
vietj: what's the point of wrapping this ?

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -113,29 +118,27 @@ public boolean isEmpty() {
...
+          }
+          @Override
+          public String next() {
+            return it.next().toString();
+          }
+        };
+      }
+      @Override
+      public int size() {
+        return nameSet.size();
```
vietj: I think indeed this is more correct to return the `nameSet.size()`

vietj:
I've done more review

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -95,10 +96,14 @@ public int size() {
...
-        .map(name -> new AbstractMap.SimpleEntry<>(name.toString(), headers.get(name).toString()))
-        .collect(Collectors.toList());
+    if (isEmpty()) {
+      return Collections.emptyList();
+    }
+    List<Map.Entry<String, String>> list = new ArrayList<Map.Entry<String, String>>(headers.size());
+    for (Map.Entry<String, String> entry : this) {
+      list.add(entry);
+    }
+    return list;
```
testn: since we do not cache the result, we don't need to make it immutable.

src/main/java/io/vertx/core/http/impl/HeadersAdaptor.java
vietj: let's stick to simplicity and remove it

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
vietj: the javadoc says it is immutable

vietj: and the `Collections.empty()` is immutable as well

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
testn: I think dup entries are required. Otherwise, ClientConnection.createRequest will not work correctly.

vietj: yes it should list all entries - just thinking we should cache the instance

PR #1648, opened 2016-09-27
Support for decompression request data on HTTP server

clonyara:
In some real-life cases, clients can send gzipped POST/PUT content.

Decompression on request's body handler can be costly for high-load systems.

This pull request implementing it by adding Netty's standard HttpContentDecompressor decoder to HttpServerImpl channel pipeline.

In our performance tests, this approach reduced memory costs in compare to naive approach.

Our real-life case was requirement to support gzipped bid requests on our DSP from one of the SSPs.


vietj:
can you also add support for HTTP/2 ?

clonyara:
Pushed

src/test/java/io/vertx/test/core/HttpDecompressionTest.java
```diff
@@ -0,0 +1,61 @@
...
+public class HttpDecompressionTest extends HttpTestBase {
+
+  public void setUp() throws Exception {
+    super.setUp();
+    client = vertx.createHttpClient(new HttpClientOptions());
+    server = vertx.createHttpServer(new HttpServerOptions().setPort(DEFAULT_HTTP_PORT).setDecompressionSupported(true));
+  }
+
+  @Test
+  public void testDefaultRequestHeaders() throws Exception {
```
vietj: can you move this into Http1xTest ?

vietj:
@clonyara 

- I believe that we should not use a new property value _tryUseDecompression_ as `HttpClientOptions#setTryUseCompression()` will achieve the same effect for the `HttpClientResponse`. Son on the same settings on `HttpClient` and `HttpServer` would be inconsistent. So it should reuse the _tryUseCompression_ property of the `HttpServerOptions`
- conversely should we also perform compression on the `HttpClientRequest` ?

src/test/java/io/vertx/test/core/HttpDecompressionTest.java
clonyara: I put it in separate file just like `io.vertx.test.core.HttpCompressionTest`.
I can move to `Http1xTest`, of course

vietj:
@clonyara any news about these remarks ?

clonyara:
@vietj 
- I've created new property named `decompressionSupported` on `HttpServerOptions` which on's/off's this specific feature on server side, similar to `compressionSupported` semantic. The reason is to support both compressed and plain traffic coming from clients, according to `Content-Encoding` header value. So you prefer to rename `decompressionSupported` to `tryUseCompression` on `HttpServerOptions` class?
- Hm... It's simple to implement, but I will be global behavior, meaning once HttpClinet configured with _send compressed content_ (we should find good name for config property) it will always send POST/PUT/... data gzipped, because it will be additional channel handler on client connection's channel pipeline.

src/test/java/io/vertx/test/core/HttpDecompressionTest.java
vietj: you can close the server and recreate a new one - some test do that please have a look

vietj:
@clonyara 

- actually I think you are right, we should use `decompressionSupported` - so no I prefer you keep the `decompressionSupported` original name.

- concerning the client, I'm thinking we should not bother with that now, as the usage is not widespread and it still can be done manually.

so only moving the test to Http1xTest should be done.

vietj:
@clonyara any news ? I'd like to merge this soon

clonyara:
I can finallize everything on Sunday

vietj:
ok

clonyara:
Test moved

PR #1651, opened 2016-09-27
Support local address for TCP clients - fixes #1569

### LINKED ISSUE #1569: Support local address for TCP clients
HttpClient should allow an option to specify the local ip address. This is useful for a machine with multiple ip addresses / network interfaces. It might want to communication to happen on a specific network interface. Another use case is for a HttpClient with a really high outbound fanout and slow response (i.e. web crawler). Without an ability to force the outgoing connection to share the same local ip address and port, it will be difficult to scale the number of requests on one machine.

vietj:
@pmlopes @tsegismont @cescoffier please review

### DIFF
--- a/src/main/java/io/vertx/core/net/ClientOptionsBase.java
+++ b/src/main/java/io/vertx/core/net/ClientOptionsBase.java
@@ -48,6 +48,7 @@ public abstract class ClientOptionsBase extends TCPSSLOptions {
   private boolean trustAll;
   private String metricsName;
   private ProxyOptions proxyOptions;
+  private String localAddress;
 
   /**
    * Default constructor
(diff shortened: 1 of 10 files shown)


src/main/java/io/vertx/core/net/ClientOptionsBase.java
```diff
@@ -167,6 +170,25 @@ public ProxyOptions getProxyOptions() {
...
+  /**
+   * @return the local interface to bind for network connections.
+   */
+  public String getLocalAddress() {
+    return localAddress;
+  }
+
+  /**
+   * Set the local interface to bind for network connections. When the local address is null,
+   * it will pick any local address, the default local address is null.
```
tsegismont: I can't think of a better wording, but strictly speaking it's not "any" address, it an address which has a route to the destination. Maybe it's not important to get into such details, still wanted to let you know.

PR #1652, opened 2016-09-28
Move to use lowercase header name instead

testn:
With HTTP/2, all the headers should be in lowercase form. By making all the headers to be lowercase, it will reduce the need for Netty/Vert.x to perform any conversion

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpHeaders.java
+++ b/src/main/java/io/vertx/core/http/HttpHeaders.java
@@ -15,6 +15,10 @@
  */
 package io.vertx.core.http;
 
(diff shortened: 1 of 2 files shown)


testn:
@vietj  got time to review it?

vietj:
have you done study concerning the conversion ? with HTTP/1 and HTTP/2 ?

testn:
Using lowercase is much more efficient for HTTP/2 AFAIK. For HTTP/1.1, it does not matter that much.

vietj:
thanks, this was actually done differently but very similar, closing this

PR #1653, opened 2016-09-28
Add the map operation to AsyncResult to easily transform an async res…

vietj:
…ult into another one

This aims to replace boiler plate code like:

```
public void start(Handler<AsyncResult<MyType>> handler) {
   ...
   startSomeServer(ar -> {
     if (ar.succeeded()) {
        handler.handle(Future.succeededFuture(this));
     } else {
        handler.handle(Future.failedFuture(ar.cause()));
     }
   });
   ...
}
```

by

```
public void start(Handler<AsyncResult<MyType>> handler) {
   ...
   startSomeServer(ar -> {
      handler.handle(ar.map(this));
   });
   ...
}
```

### DIFF
--- a/src/main/java/io/vertx/core/AsyncResult.java
+++ b/src/main/java/io/vertx/core/AsyncResult.java
@@ -16,6 +16,8 @@
 
 package io.vertx.core;
 
+import java.util.function.Function;
+
 /**
  * Encapsulates the result of an asynchronous operation.
  * <p>
@@ -58,5 +60,32 @@ public interface AsyncResult<T> {
    * @return true if it failed or false otherwise
    */
   boolean failed();
+
+  /**
+   * Apply a {@code mapper} function on this async result.<p>
+   *
+   * The {@code mapper} is called with the completed value and this mapper returns a value. This value will complete the result returned by this method call.<p>
+   *
+   * If the {@code mapper} throws an exception, the returned future will be failed with this exception.<p>
+   *
+   * When this async result is failed, the failure will be propagated to the returned future and the {@code mapper}
+   * will not be called.
+   *
+   * @param mapper the mapper function
+   * @return the mapped async result
+   */
+  <U> AsyncResult<U> map(Function<T, U> mapper);
+
+  /**
+   * Map the result of this async result to a specific {@code value}.<p>
+   *
+   * When this async result succeeds, this {@code value} will succeeeds the async result returned by this method call.<p>
+   *
+   * When this future fails, the failure will be propagated to the returned async result.
+   *
+   * @param value the value that eventually completes the mapped async result
+   * @return the mapped async result
+   */
+  <V> AsyncResult<V> map(V value);
 
 }
--- a/src/test/java/io/vertx/test/core/FutureTest.java
+++ b/src/test/java/io/vertx/test/core/FutureTest.java
@@ -31,6 +31,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 /**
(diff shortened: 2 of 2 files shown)


vietj:
@tsegismont @cescoffier @pmlopes please review

src/main/java/io/vertx/core/AsyncResult.java
```diff
@@ -59,4 +61,44 @@
...
+   */
+  <V> AsyncResult<V> map(V value);
+
+  /**
+   * Return the value if this async result is succeeded, otherwise the {@code other} value.
+   *
+   * @param other the other value to return
+   * @return the result when succeeded, otherwise the {@code other} value
+   */
+  default T orElse(T other) {
```
tsegismont: I fear this is going to be a source of user problems. If the result is not known yet (AR neither failed nor succeeded), it will immediately return `other`.
But when I look at the Javadoc, I am wondering if this is the behavior you really wanted?

src/main/java/io/vertx/core/AsyncResult.java
```diff
@@ -59,4 +61,44 @@
...
+   *
+   * If the {@code mapper} throws an exception, the returned future will be failed with this exception.<p>
+   *
+   * When this async result is failed, the failure will be propagated to the returned future and the {@code mapper}
+   * will not be called.
+   *
+   * @param mapper the mapper function
+   * @return the mapped async result
+   */
+  <U> AsyncResult<U> map(Function<T, U> mapper);
```
tsegismont: Good to expose those methods already present in Future in the AR interface (since it's what a vast majority of our APIs use).

src/test/java/io/vertx/test/core/FutureTest.java
```diff
@@ -205,6 +206,13 @@ public void testFailurFutureWithNullFailure() {
   }
 
   @Test
+  public void testOrElse() {
+    assertEquals("bye", Future.future().orElse("bye"));
```
tsegismont: This is not consistent with the behavior described in the Javadoc

src/main/java/io/vertx/core/AsyncResult.java
vietj: I see what you mean, the problem is that it is inherited by Future...

vietj: yes so perhaps it should be removed

vietj: the problem indeed is when it is not completed for Future, or it should throw an IllegalStateException if taht's not completed ?

src/test/java/io/vertx/test/core/FutureTest.java
vietj: in this case maybe it should throw an IllegalStateException ?

cescoffier: Throwing a ISE would make it almost impossible to use. I think there are 2 case:

* The user want to retrieve this default result if not completed
* The user want this default result only on failure

As it's 2 different use cases, I would rather use 2 different methods explicitly named (to avoid the `orElse` confusion).

vietj: I think we should drop the `orElse`, anyway the point of this PR was about the `map` that is more useful and works in all cases.

tsegismont:
How about having an "else" method which takes an handler of Throwable? It
would be invoked if the result completes with failure.

It  would work like setting a handler on future in a way, but it would make
it possible to use asyncResult like a promise (which was your original
intention I guess)

2016-09-28 13:52 GMT+02:00 Julien Viet <notifications@github.com>:

vietj:
@tsegismont I think then it becomes quite complex

the real issue is that people will make the amalgam between Future and AsyncResult because the former extends the later

tsegismont:
Could be. Anyway I'm fine with removing altogether.

2016-09-28 14:20 GMT+02:00 Julien Viet <notifications@github.com>:

PR #1655, opened 2016-09-28
Improve HttpClientRequest reset support - fixes #1642

### LINKED ISSUE #1642: Improve HttpClientRequest reset support
Currently we support `HttpClientRequest#reset()` for HTTP/2 connections.

The support could be extended to:

- HTTP/1.x with keep-alive : close the current connection
- HTTP/1.x with pipe-lining : wait until all pending responses have been received and then close connection
- request not yet connected : the request has been queued waiting for a connection or perhaps it is triggered the creation of the connection, however the request should not be made anymore : simply put the connection back in the pool and let the pool find another request for this connection

It is useful for supporting reverse proxies implementations with Vert.x : when the client of the proxy sends an entity and something wrong happens (like an exception signed by the `HttpServerRequest#exceptionHandler`) the corresponding `HttpClientRequest` can be reset. For HTTP/2 it is natively supported, for HTTP/1.x this means closing the connection (and waiting for responses in case of pipelining). Having this behavior built-in the reset method for HTTP/1.1 makes easy to implement this for the proxy implementation.

vietj:
@cescoffier @pmlopes @tsegismont please review

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/ClientConnection.java
@@ -44,6 +44,7 @@ import io.vertx.core.spi.metrics.HttpClientMetrics;
 
 import java.net.URI;
 import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Queue;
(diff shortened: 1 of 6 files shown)


src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -377,7 +379,18 @@ void handleResponseEnd(LastHttpContent trailer) {
...
+      } else {
+        if (reset) {
+          if (requests.isEmpty()) {
+            pool.responseEnded(this, true);
+          }
+        } else {
+          pool.responseEnded(this, false);
+        }
+      }
     }
```
vietj: cryptic! (and harder to debug)

tsegismont:
A mater of taste then, because to me it's much more easier to read when we
want to close with the oneliner than with the if/else tree.

Not a blocker for merging anyway.

2016-09-29 15:43 GMT+02:00 Julien Viet <notifications@github.com>:

PR #1662, opened 2016-10-06
Use inflight DNS client instead

testn:
I noticed a similar problem as @luengnat with the DNS vert.x client which did not use inflight dns resolver and caused too much load on our DNS server. So I cook up a patch that should address the issue

### DIFF
--- a/src/main/java/io/vertx/core/impl/AddressResolver.java
+++ b/src/main/java/io/vertx/core/impl/AddressResolver.java
@@ -16,13 +16,15 @@
 
 package io.vertx.core.impl;
 
+import io.netty.channel.ChannelFactory;
 import io.netty.channel.EventLoop;
+import io.netty.channel.socket.DatagramChannel;
 import io.netty.channel.socket.nio.NioDatagramChannel;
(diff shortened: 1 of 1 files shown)


vietj:
can you give a thorough explanation of the issues you encountered and how this change solves it ?

testn:
When there is a request to resolve a host, it will consult with the local DNS cache first. If the entry does not exist in the cache, it will make a request. Everything seems well. However, when another request is made while the other DNS request is in flight, it will send out another DNS request. This means that if there are a lot of requests being made to a server while the DNS entry is not cached yet, it will send multiple requests to the DNS server. A lot of DNS server implementation, in our case, DNS BIND, have a rate limiting feature where it refuses to serve some requests when there are too many identical requests coming from the same ip address.

vietj:
ok great, thanks!

luengnat:
+1. Thank you very much for providing the PR

vietj:
did you run the testsuite with this change ?

testn:
yes but let me re-run it against the rebased code

vietj:
any news from taht ?

testn:
all green!

PR #1668, opened 2016-10-10
Fixes vert-x3/vertx-infinispan#3 InfinispanHATest>HATest.testFailureInFailover fails intermittently

tsegismont:
Actually it does not fail in Infinispan CM only, it fails in Hazelcast CM as well.

The reason is that simulateKill returns just after calling "leave" on the CM. So calling tests assumes rebalance and stuff happened already but actually the information may even have not been handed over to the cluster manager.


tsegismont:
See https://github.com/vert-x3/vertx-infinispan/issues/3

tsegismont:
@vietj can you have a look? Thx

PR #1669, opened 2016-10-10
In test code, show throwable caught from non main thread

tsegismont:
Otherwise in the build logs we only see the IllegalStateException without the actual problem


tsegismont:
@vietj can you have a look at this one too? Thx

PR #1675, opened 2016-10-13
Future convenience success and failure handlers

Xargos:
Signed-off-by: Jakub Pierzchlewicz jakubpierzchlewicz@hotmail.com

Success and failure convenience methods for Futures. Intended to make code that uses Futures more readable. In addition some additional null checks and spelling fixes.

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -120,7 +120,13 @@ public interface Future<T> extends AsyncResult<T> {
    *
    */
   @Fluent
-  Future<T> setSuccessHandler(Handler<T> successHandler);
+  default Future<T> setSuccessHandler(Handler<T> successHandler) {
+    return setHandler(ar -> {
+      if (ar.succeeded()) {
+        successHandler.handle(ar.result());
+      }
+    });
+  }
 
   /**
    * Set a handler for future that has failed.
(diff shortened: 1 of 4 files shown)


vietj:
I'm not against having shortcut method like you added to the interface, however I find confusing the fact to have more than one handler in the implementation, i.e

```
default void setSuccessHandler(Handler<T> handler) {
  setHandler(ar -> {
    if (ar.succeeded()) {
      handler.handle(ar.result());
    }
  });
}
```

Xargos:
True, might be confusing, especially since it is not clear in what order will they be called. Standard handler will be called first if it was set before futures completion and called in order of assignment if assigned after completion (if that makes sense). Upon reviewing the code I have not stated that in the method documentation. It should be added. 

With your approach you will not be able to handle situation where the future can either fail or succeed. Although it might be a good middle ground.

The idea behind this addition was to make the code that uses Futures more readable. We have recently started to implement our microservices using returned Futures instead of callbacks as arguments and it quickly became apparent that there will be a lot of if-succeeded-failed blocks. The approach with success-failure handlers seemed nice to have, however it might be just my opinion. I think with proper documentation of the functions it should be quite clear how the methods work, however in the end it's a design decision.

vietj:
when you need to handlers then just doing setHandler is fine I think

Xargos:
OK, I understand. I will modify the pull request and implement your suggestion.

Xargos:
I am not sure as to the contributing process. Should I squash the changes?

PR #1676, opened 2016-10-13
Fixed command line parameter name to be "hagroup" instead of "ha-group"

riperez:
Signed-off-by: Richard Perez <riperez@gmail.com>


riperez:
Wrong source file to edit documentation. Will submit a new PR with the fix on the correct source file.

PR #1684, opened 2016-10-21
Fix a few broken tls tests

vietj:
@alexlehm can you review the change to the proxy tls test ?


src/test/java/io/vertx/test/core/HttpTLSTest.java
```diff
@@ -783,7 +783,7 @@ public void testHttpsProxy() throws Exception {
   // Check that proxy auth fails if it is missing
   public void testHttpsProxyAuthFail() throws Exception {
     startProxy("username", ProxyType.HTTP);
-    testTLS(Cert.NONE, Trust.SERVER_JKS, Cert.SERVER_JKS, Trust.NONE).useProxy().useProxyAuth().fail();
+    testTLS(Cert.NONE, Trust.SERVER_JKS, Cert.SERVER_JKS, Trust.NONE).useProxy().fail();
```
vietj: @alexlehm I removed the `useProxyAuth()` to make the test work correctly / before it was incorrect but test check was incorrect so it was passing

PR #1690, opened 2016-10-26
Addition of Future.discardingCompleter()

jponge:
This method is similar to `Future.completer()` except that it works for any type of `AsyncResult` and discards the result value to set null on success, and the throwable in case of error.

The motivation for this addition is to simplify code such as:

``` java
@Override
  // (...)
  .listen(8080, ar -> {
    if (ar.succeeded()) {
      startFuture.complete();
    } else {
      startFuture.fail(ar.cause());
    }
  });
}
```

to:

``` java
@Override
  // (...)
  .listen(8080, startFuture.discardingCompleter());
}
```

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -293,13 +293,29 @@ public interface Future<T> extends AsyncResult<T> {
   }
 
   /**
-   * @return an handler completing this future
+   * @return a handler completing this future
+   * @see #complete(Object)
    */
   @CacheReturn
   default Handler<AsyncResult<T>> completer() {
     return ar -> {
       if (ar.succeeded()) {
         complete(ar.result());
+      } else {
+        fail(ar.cause());
+      }
+    };
+  }
+
+  @CacheReturn
+  /**
+   * @return a handler completing this future with {@code null} on success
+   * @see #complete()
+   */
+  default <U> Handler<AsyncResult<U>> discardingCompleter() {
+    return ar -> {
+      if (ar.succeeded()) {
+        complete();
       } else {
         fail(ar.cause());
       }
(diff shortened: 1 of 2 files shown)


vietj:
it seems it only fit well with the `Future<Void>` for which `complete()` is for.

it would not fit well for a `Future<String>` for which the provider of the future would expect a non value.

aesteve:
:+1:  implements https://github.com/eclipse/vert.x/issues/1654

jponge:
It indeed fits the case of `Future<Void` for which you want to complete no matter what the asynchronous result type is. This can reduces boilerplate like in verticle lifecycle methods where regular completers have type issues.

jponge:
It indeed fits the case of Future<Void> for which you want to complete no
matter what the asynchronous result type is. This can reduces boilerplate
like in verticle lifecycle methods where regular completers have type
issues.
On Wed, 26 Oct 2016 at 23:55, Julien Viet <notifications@github.com> wrote:

it seems it only fit well with the Future<Void> for which complete() is for.

it would not fit well for a Future<String> for which the provider of the
future would expect a non value.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/eclipse/vert.x/pull/1690#issuecomment-256488918>, or mute
the thread

vietj:
maybe we can overload the completer method with a function:

```
<U> Handler<U> completer(Function<U, T> f);
```

and

```
<U> Handler<U> completer(U u);
```

so in your case you could do `completer(null)` to complete `Future<Void>`

jponge:
That could be an option, but I'm not much of a fan, just like `map(value)` is a peculiar design (if you don't want a function but a value then a provider is a better design since it allows for lazy evaluation).

cescoffier:
I would go for the `discardingCompleter` as the name if straightforward.

vietj:
as I said this can be work arounded and with the plan we have to use more future as return types for 3.4, this make this it much less useful.

PR #1698, opened 2016-11-03
Add support for setting SNI hostname and set for client SSL handler

lulf:


lulf:
I wasn't sure if this should be put in ClientOptionsBase instead, since I think it only makes sense for clients. But I came to the conclusion that I liked it better along with other ssl options.

vietj:
I just checked and it supports only a single host, we want instead to support a mapping of domain->key/cert that is supported by Netty's `io.netty.handler.ssl.SniHandler`.

so it seems (as far as I understand after a quick look), we should use the `SniHandler` with a mapping of domain->SslContext to fully implement this feature.

the TCPSSLOptions would have a Map<String, KeyCertOptions> somehow.

lulf:
@vietj I've updated the PR to use SniHandler for server side mapping of host -> SslContext instead of SNIMatcher directly. The end2end interaction is tested in the SSLEngineTest integration test, where I made it use a different certificate for the sni host. 

I haven't paid much attention to how this is exposed in the options, so please comment on how you would like that.

vietj:
@lulf thanks going to review it next week

src/main/java/io/vertx/core/http/HttpServerOptions.java
```diff
@@ -280,6 +281,13 @@ public HttpServerOptions addEnabledCipherSuite(String suite) {
   }
 
   @Override
+  @GenIgnore
```
vietj: I believe we should try to handle Map<String, Options> in codegen to not have GenIgnore

vietj:
@lulf 

- instructions on how to create server-virt-cert.pem and the key are missing in src/test/resources/tls/ssl.txt
- there should be test in HttpTLSTest with various formats keystore/pkcs12/pem

src/main/java/io/vertx/core/http/HttpServerOptions.java
lulf: Fixed

lulf:
@vietj 

Instructions added, and I added a single test to HttpTLSTest that verifies that the cert is working. It's generated the same way as the server-cert.pem, because for SNI I needed another cert that was not trusted by the default ssl context.  I'm not sure if adding more tests to HttpTLSTest is necessary.

lulf:
@vietj 

In addition to addressing your comments, I had to add some more functionality to SSLHelper to be able to support listening for handshakes and setting SSLEngine properties as several tests were failing due to that. Now all the unit tests pass here.

vietj:
@lulf has it been pushed ? can you update the issue and list all the changes ?

lulf:
Ah, that issue :)

vietj:
@lulf there are a few things to change but before continuing I would like to focus on the most important part:

The SSL test in `HttpTLSTest` is incorrect because it only tests the SNI certificate validity as a certificate and does not configure the SNI server. Also the `HttpServer` does not implement SNI, otherwise the added test in `HttpTLSTest` would fail and it passes.

Such test can only be added if the *Client supports SNI as well or if we use directly Netty to write the client part of the tests.

So right now we need to have at least the `HttpServer` implement SNI and to have this tested effectively.

lulf:
@vietj 

This is tested in src/test/java/io/vertx/test/it/SSLEngineTest.java  . It uses HttpServer  and tests that a client sets SNI and that it works and doesn't work.

vietj:
@lulf my bad I did not have the correct version of files I see now createSNIHandler called by HttpServer and client

still the test `testTLSCLientTrustVirtServerCertPEM` does not seem correct as the caller does not configure SNI in server/client

lulf:
@vietj 

That is why i was a bit puzzled why I should have to add that test, since I already tested it in SSLEngineTest. I thought it was just to test certificate validity. I can extend that test to use SNI also, although it would still test the same thing.

vietj:
@lulf yes, I beleive the Trust / Cert classes should configure the TCPSSLOptions rather with somehting configure(TCPSSLOPtions) which means a bit of refactoring.

lulf:
@vietj 

I modified the TLSTest to take a SNI parameter for client and for server, and extended with a few testcases.

I'll just create separate commits during the review and squash them at the end to make it easier to see what has changed.

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -441,4 +476,49 @@ public SslHandler createSslHandler(VertxInternal vertx) {
...
+    for (String sniServername : sniKeyCertOptions.keySet()) {
+      builder.add(sniServername, getContext(vertx, sniServername));
+    }
+    return new SNIHandler(builder.build());
+  }
+
+  /**
+   * Custom SNI handler that allows for listening to handshakes and configuring the SSLEngine.
+   */
+  public class SNIHandler extends SniHandler {
```
vietj: Make it a normal class (not inner) and rename `VertxSniHandler`

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -292,6 +292,12 @@ public HttpClientOptions setSsl(boolean ssl) {
   }
 
   @Override
+  public HttpClientOptions setSNIServerName(String serverName) {
```
vietj: rename `setSniServerName` as the rest of the options tends to use non capital acronyms

src/main/java/io/vertx/core/net/TCPSSLOptions.java
```diff
@@ -405,6 +410,50 @@ public TCPSSLOptions addCrlPath(String crlPath) throws NullPointerException {
   }
 
   /**
+   * Add key options for SNI host.
+   *
+   * @param domain The domain name to match. May contain wildcards.
+   * @param keyCertOptions The key options to use when matching this host.
+   * @return a reference to this, so the API can be used fluently
+   */
+  public TCPSSLOptions addSNIKeyCertOptionsForDomain(String domain, KeyCertOptions keyCertOptions) throws NullPointerException {
```
vietj: - rename `addSniKeyCertOption`
- annotate it with @ GenIgnore
- add new adder / getter `addSniKeyStoreOption/getSniKeyStoreOptions`, `addSniPfxKeyCertOption/getSniPfxKeyCertOptions`, `addSniPemKeyCertOption/getSniPemKeyCertOptions` , not annotated with @ GenIgnore - follow the pattern used by keyCertOptions / setKeyStoreOption / setPfxKeyCertOption / setPemKeyCertOption

src/test/java/io/vertx/test/core/HttpTLSTest.java
```diff
@@ -75,6 +75,30 @@ public void testTLSClientTrustServerCertPEM() throws Exception {
   }
 
   @Test
+  // Test SNI fails if SNI is not set by client
```
vietj: add tests with KeyStore format and PFX format

vietj:
I've added further comments, we also need documentation in https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/net/package-info.java and at least an example at https://github.com/vert-x3/vertx-examples/tree/master/core-examples/src/main/java/io/vertx/example/core/http

vietj:
I think also we need to change the `SniServerName` from the options to the `HttpClientRequest` and do something similar for `NetClient` as a client is designed to handle servers spanning on different hosts.

vietj:
perhaps also we need to handle it also in the `NetSocket#upgradeToSsl` method

vietj:
need also at least one HTTP/2 test

lulf:
@vietj Thanks!

Will try to have a look this or next week.

vietj:
@lulf we are planning a 3.4 beta release soon, it could fit in if we merge it this week

src/main/java/io/vertx/core/net/TCPSSLOptions.java
lulf: There are no getters for regular keyCertOptions other than getKeyCertOptions. I assume I can skip the getters for each type as well then, since it would have to be casted  back to the appropriate type.

vietj: there is notion of "adder" where getter + adder match together in data object:

public Map<String, String> getValues() { ... }
public void addValue(String key, String value) { ... }

notice the "s" difference.

so we should have this correspondance in the data object

lulf:
@vietj I've updated the PR with the requested fixes, except moving the setSniName to HttpClientRequest. I'm not sure if it's a commonly used feature to set this explicitly, and it does relate directly to the SSL context. It also feel's like a very specialized option compared to the other methods there. Same with NetClient. I think moving it to the request should be done at a later point after some good thinking first.

Regarding Http/2 tests, all tests in HttpTLSTest are run as part of Http2TLSTest since the latter inherits the first.

I'll create a separate PR for the http example.

vietj:
@lulf ok, the sniservername should really be not on the options as the only option one has is to create one client per host

src/main/java/io/vertx/core/net/TCPSSLOptions.java
lulf: That's not what I meant.i I'm talking about the type of the return value. What I mean is that there is no way that getSniPemCertKeyOptions can return the type of Map<String, PemCertKeyOptions>, because it is stored as Map <String, KeyCertOptions >. This is why there is only one getter for the non -sni options as well getKeyCertOptions

lulf:
@vietj what about other ssl parameters then?  Why shouldn't they be per request?  The sniservername is part of an ssl session,  not a single request.

vietj:
@lulf it would be used when the request first establish a connection and re-used subsequently (pipelining / HTTP/2) , pretty much like setting the HTTP version on a request.

vietj:
@lulf we should call it `setVirtualHost()` on `HttpClientRequest`.

vietj:
well for now let's only call it setSniServerName then we will figure out how to do that properly.

vietj:
@lulf so here is what we can do : add a "setVirtualHost(String)" on `HttpClientRequest`, when this field is set then the value is used as SNI server name and it overrides the HTTP _host_ value sent to the server, i.e it would call setHost() with the value automatically (but the user might override again the host value after with `setHost` if he wants to but has to do it explicitely).

vietj:
@lulf any news for `setVirtualHost` ?

lulf:
@vietj I've added the method to the interface but haven't been able to wire things in the implementation yet. A bit busy the coming days unfortunately.

vietj:
@lulf thanks for the headsup

theoweiss:
Is there any chance to get this feature into 3.4.1?

theoweiss:
Sorry, this feature is really important for our services based on vertx. Do you have a plan when this PR will be ready to merge?

lulf:
@theoweiss Unfortunately I don't have any time left to work on this these days, and there seems to be a lot of code changes in the code that the PR touches, so I don't think this will be merged unless someone else picks up the pieces or restarts the effort.

vietj:
the effort is interesting, I would really like to grap well how the SNI configuration side, both on the server and on the client.

for instance it seems possible with Java keystores to map SNI hostname using keystores aliases instead of having a Map<String, KeyCertOptions> / Map<String, TrustOptions> which would get the actual stuff simpler.

vietj:
and I would like it to be done soon, perhaps in a Vert.x 3.4.2 release. I believe the technical bits are here, it's only a matter of providing good configuration imho.

vietj:
I've started a new branch for SNI (reused a few bits of this PR, thanks) : https://github.com/eclipse/vert.x/tree/sni-support

At the moment it provides NetSocket support only, the main point is the extension of the `KeyCertOptions` interface that provides a mapper of serverName to Key+CertChain. The ssl context are cached in `SSLHelper` based on the certificate used for the handshake.

@xkr47 you are probably interested in following this branch

xkr47:
@vietj sure!
Some quick thoughts:
* It's perfectly possible that a remote client gives our server a SNI hostname that doesn't exist, therefore `mgr = keyCertOptions.keyManagerMapper(vertx).apply(serverName);` may be null. In my [DynamicCertManager](https://github.com/xkr47/vertx-acme4j/blob/master/src/main/java/io/nitor/vertx/acme4j/util/DynamicCertManager.java) solution I provide the option of choosing a default certificate, or to leave the server without a default, in which case the TLS handshake fails when client-speicifed hostname does not match any certificate. The latter is imo a good additional layer of defense agains attackers randomly bombing servers by IP; their attack surface will be a lot smaller if they don't know a valid hostname.
* As a server implementor I would like to be able to query incoming connections/requests what certificate or hostname was used in the TLS handshake.

vietj:
thanks for the feedback, will take it in account

vietj:
you can know already the certificate on NetSocket, I believe we should add the same on HttpConnection and both should have the serverName

vietj:
I've mostly finished the SNI support (not the docs yet) : you can have a look https://github.com/eclipse/vert.x/tree/sni-support

vietj:
@lulf can you have a look at it ?

xkr47:
I would propose that DEFAULT_SNI is changed to `true` as soon as the feature is considered stable; unsupporting servers will simply ignore it.

vietj:
the `sni-support` branch has now a PR opened for review, you might want to have a look and provide feedback https://github.com/eclipse/vert.x/pull/1954

vietj:
@xkr47 you mean for `HttpClientOptions` ?

xkr47:
> @xkr47 you mean for HttpClientOptions ?
Yes

vietj:
@xkr47 I propose to leave it as false for now for the release 3.4.x and switch it to true when we'll be on 3.5.0-SNAPSHOT

vietj:
closing this PR, thanks @lulf for the contribution

xkr47:
Thanks @lulf!

PR #1710, opened 2016-11-10
generic return type for get json value

jRoadie:
Signed-off-by: shg@dsi <akter.hossain@dsinnovators.com>

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/src/main/java/io/vertx/core/json/JsonObject.java
@@ -241,6 +241,22 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
   }
 
   /**
+   * Get the value with the specified key, as an generic retrun type
+   * @param key  the key to lookup
+   * @return the value
+   */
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -241,6 +241,22 @@ public Instant getInstant(String key) {
   }
 
   /**
+   * Get the value with the specified key, as an generic retrun type
+   * @param key  the key to lookup
+   * @return the value
+   */
+  public <R> R get(String key) {
```
vietj: that's a free class cast !!

vietj:
I think we had discussed this before and decided to not provide a version of getValue() that does a free class cast.

src/main/java/io/vertx/core/json/JsonObject.java
jRoadie: @vietj do you see any downside of free/generic class cast? Can you share with me, please. I want to know about this.

PR #1725, opened 2016-11-24
HttpConnection: expose remoteAddress() and localAddress()…

xkr47:
…already implemented by all implementing classes (in vertx-core).

I assume there are no implementations of HttpConnection in other vert.x modules,


xkr47:
I added `@CacheReturn` since I thought it would make sense (they were annotated so also in the `NetSocket` interface).

PR #1732, opened 2016-12-05
Test locking on same Vert.x instance

tsegismont:
The existing clustered lock tests always involve two different Vert.x instances.
In vert-x3/vertx-hazelcast#41, it has been reported that concurrent requests to a clustered lock on the same instance can lead to deadlocking.
It happens because locks are acquired with ordered execute blocking, the same for releasing.

The fix consists in applying the technique used in jdbc client connections: a dedicated worker executor is created for a lock instance, then the lock is acquired in order, but released freely. Consequently, we never find ourselves in a situation where a lock release task is waiting for a lock acquire task to complete.

Pull requests for the cluster managers are ready and will be pushed soon.


tsegismont:
@vietj can you have a look please? Thanks!

vietj:
LGTM

tsegismont:
I'll merge the PR in the Hazelcast module then. Thanks

2017-01-03 18:33 GMT+01:00 Julien Viet <notifications@github.com>:

PR #1735, opened 2016-12-09
Changes the subs map handling for better consistency

rikardscm:
This pull request tries to fix the problem with the subs map getting out of sync with the set of actual cluster members. The following issue has a rather long discussion about it: https://github.com/vert-x3/vertx-hazelcast/issues/13. It is an alternate solution to https://github.com/eclipse/vert.x/pull/1594.

The solution uses the ClusterManager view of members rather than HAManager's distributed clusterMap, which itself may be out of sync. It cleans up and updates the subs map every time a node is added to or removed from the cluster, instead of just when a node crash is detected.
The AsyncMultiMap interface has a new method in order to support removal in a more flexible way, which means that it breaks the API for cluster plugins. A corresponding pull request for vertx-hazelcast will be created shortly to address this; see https://github.com/vert-x3/vertx-hazelcast/pull/53.

To reproduce this issue, set up clustering with at least two nodes on different hosts, with TCP clustering for Hazelcast (not multicast). The nodes should send some messages periodically, see above mentioned issue #13 for an example producer/consumer. Then simulate a network failure by blocking traffic with iptables (or kill the machines). Hazelcast will partition the cluster. Restart the verticles that are consuming messages. Remove the iptables rule (or bring the machines back up) - if hz fails to merge the cluster properly, the subs map will still contain the old entries, and so Vertx will try to send messages to these, now dead, verticles.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusterNodeInfo.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusterNodeInfo.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2017 The original author or authors
+ * ------------------------------------------------------
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Apache License v2.0 which accompanies this distribution.
+ *
+ *     The Eclipse Public License is available at
+ *     http://www.eclipse.org/legal/epl-v10.html
+ *
+ *     The Apache License v2.0 is available at
+ *     http://www.opensource.org/licenses/apache2.0.php
+ *
+ * You may elect to redistribute this code under either of these licenses.
+ */
+
+package io.vertx.core.eventbus.impl.clustered;
+
+import java.io.Serializable;
+
+/**
+ * @author Rikard Björklind
+ */
+public class ClusterNodeInfo implements Serializable {
+  public String nodeId;
+  public String host;
+  public int port;
+
+  public ClusterNodeInfo() {
+  }
+
+  public ClusterNodeInfo(String nodeId, String host, int port) {
+    this.nodeId = nodeId;
+    this.host = host;
+    this.port = port;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+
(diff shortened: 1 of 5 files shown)


vietj:
this claims to solve a problem but does not have any tests

rikardscm:
There are already some tests (testSubsRemovedForKilledNode and testSubsRemovedForClosedNode) in ClusteredEventBusTest. The ideal situation would of course be to have a test that fails before this patch, but the problem is not that easily reproduced; it happens in some cases when there is a Hazelcast partition and subsequent merge.

tsegismont:
Hi @rikardscm , in case you miss it, I've sent two PRs to your working branches with my updates. Please have a look at them. When we're fine please merge those PRs so we can keep track of the changes here and in vert-x3/vertx-hazelcast#53.

- sbs-discovery-sweden/vert.x#1
- sbs-discovery-sweden/vertx-hazelcast#1

rikardscm:
Thanks @tsegismont, merged.

tsegismont:
@rikardscm you're welcome. Could you squash the commits and signoff please? Thanks

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusterNodeInfo.java
```diff
@@ -0,0 +1,43 @@
+package io.vertx.core.eventbus.impl.clustered;
```
tsegismont: Sorry, I missed that, could you a copyright section at the top of the file? Thanks

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusterNodeInfo.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.io.Serializable;
+
+public class ClusterNodeInfo implements Serializable {
```
tsegismont: Please add an `@author` comment as well

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusterNodeInfo.java
rikardscm: Done

vietj:
closing as this has been included in https://github.com/eclipse/vert.x/pull/1848

vietj:
thanks for the contribution @rikardscm !!!!!

PR #1744, opened 2016-12-16
Expose send() as part of the Message interface

tobias:
This allows the TCP EventBus bridge to signal to the client that the
serialized message should be treated as a send or publish.

See https://github.com/vert-x3/vertx-tcp-eventbus-bridge/issues/20 for
more details.

Are there any known implementations of the `Message` interface outside of Vert.x core? If so, this change would break them unless we added a default implementation to the interface. However, a default implementation would likely need to return a constant value (possibly `true`), which may not be desirable.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/Message.java
+++ b/src/main/java/io/vertx/core/eventbus/Message.java
@@ -67,6 +67,13 @@ public interface Message<T> {
   String replyAddress();
 
   /**
+   * Signals if this message represents a send or publish event.
+   *
+   * @return true if this is a send.
+   */
+  boolean isSend();
+
+  /**
    * Reply to this message.
    * <p>
    * If the message was sent specifying a reply handler, that handler will be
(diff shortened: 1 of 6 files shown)


tobias:
Looks like my Eclipse CA had expired. I just re-signed.

src/main/java/io/vertx/core/eventbus/Message.java
```diff
@@ -67,6 +67,13 @@
   String replyAddress();
 
   /**
+   * Signals if this message represents a send or publish event.
+   *
+   * @return true if this is a send.
+   */
+  boolean send();
```
vietj: should be renamed `isSend`

tobias:
I renamed `send` to `isSend`.

vietj:
I may have to revert this. `isSend` does not make sense in some case like the AMQP bridge that use this class.

tobias:
Why doesn't it make sense for AMQP? Does it not have the concept of a send? If not, could `isSend` just always return false in that case?

vietj:
amqp have just the meaning of sending a message.

yes it could return true/false and that's the current case

so we may leave it as is / another option is to have this info in the DeliveryOptions headers when the message is received

PR #1746, opened 2016-12-17
add static method for chain possibility

mystdeim:
JsonObject object = JsonObject.create().put("foo", "bar");


vietj:
what does it buy over new JsonObject().put("foo", "bar") ?

mystdeim:
You are right, new JsonObject() is enough

PR #1759, opened 2017-01-10
Future recovering methods

ruslansennov:
I believe these are very useful methods (at least for me)

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -305,6 +305,32 @@ public interface Future<T> extends AsyncResult<T> {
     };
   }
 
+  /**
+   * Handles a failure of this Future by returning the result of another Future.
+   * If the mapper fails, then the returned future will be failed with this failure.
+   *
+   * @param mapper A function which takes the exception of a failure and returns a new future.
+   * @return A recovered future
+   */
+  default Future<T> recover(Function<? super Throwable, Future<T>> mapper) {
+    Future<T> ret = Future.future();
+    setHandler(ar -> {
+      if (ar.succeeded()) {
+        ret.complete(result());
+      } else {
+        Future<T> mapped;
+        try {
+          mapped = mapper.apply(ar.cause());
+        } catch (Throwable e) {
+          ret.fail(e);
+          return;
+        }
+        mapped.setHandler(ret.completer());
+      }
+    });
+    return ret;
+  }
+
   FutureFactory factory = ServiceHelper.loadFactory(FutureFactory.class);
 
 }
--- a/src/test/java/io/vertx/test/core/FutureTest.java
+++ b/src/test/java/io/vertx/test/core/FutureTest.java
@@ -789,6 +789,35 @@ public class FutureTest extends VertxTestBase {
     assertSame(cause, map2.cause());
   }
 
+  @Test
+  public void testSucceededFutureRecoverWith() {
+    Future<String> f = Future.future();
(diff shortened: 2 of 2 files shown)


vietj:
that seems related a bit to circuit breaker

ruslansennov:
> that seems related a bit to circuit breaker

Just another two FP methods, additional to `compose()` and `map()`

vietj:
I would only keep `recover` with a Future and people recovering with a constant, would do 'recover(err -> Future.succeeded(foo))'

ruslansennov:
makes sense, thank you
PR updated

vietj:
nice reactivity :-)

vietj:
a test for the case the mapper function fails is missing

ruslansennov:
done

vietj:
last thing, I think we need in the documentation of recover, mention that if the mapper fails, then the returned future will be failed with this failure.

ruslansennov:
done!

vietj:
thanks for the contribution @ruslansennov

ruslansennov:
@vietj 
```
Could not generate model for mapper: Illegal type ? super java.lang.Throwable of kind WILDCARD
  default Future<T> recover(Function<? super Throwable, Future<T>> mapper) {
```

Sorry, I've tested this PR by run `FutureTest` only

vietj:
@ruslansennov actually I do have some concerns about this patch.

As the `map` method, it would be good to have a similar method in `AsyncResult<T>`. However `recover` uses a function that returns a `Future<T>` (like compose).

You propose `recover` and `recoverWith` and I think it would be good to have `recoverWith` that can go in `AsyncResult`.

However I'm not fan of `recoverWith` name. If you think of how to name such method in a better way similar to:

- compose / map
- recover / ???
- ??? / recover

for instance RxJava uses:
- onErrorResumeNext
- onErrorReturn

vietj:
maybe `orElse` ?

```
void foo(ar -> {
  ar.orElse(err -> "something");
});
```

```
fut.orElse(err -> "something");
```

that could be overloaded with a constant like `map`

```
ar.orElse("constant");
```

```
fut.orElse("constant");
```

vietj:
Actually it seems like a good fit to me.

ruslansennov:
@vietj this patch inspired by [similar methods in Javaslang](https://github.com/javaslang/javaslang/blob/44a6887e34037566fca4964ccc91d5d97aab422c/javaslang/src/main/java/javaslang/concurrent/Future.java#L791) which follow [similar methods in Scala](https://github.com/scala/scala/blob/bff11b611ae4c9a02c456847fbbb4695bc3f40ec/src/library/scala/concurrent/Future.scala#L389). Also I like `onErrorResumeNext`/`onErrorReturn` names. But `orElse` sounds a bit confusing in this case

vietj:
@ruslansennov yes, I see your point of view and this is more inspired from https://github.com/advantageous/reakt and `Optional` than from a functional library.

vietj:
I believe it's a good fit because we have compose/map and map can accept a `Function<T, Value>` or just a value, the recovering method counterpart should accept both and in this case the `orElse(Value)` is exactly what we want here

vietj:
in javaslang that would be `javaslang.control.Option` that provides an `orElse`.

Given that in Vert.x `Future<T>` extends `AsyncResult<T>` that seems ok to me.

vietj:
btw in Vert.x future, the `compose` is the classic `flatMap` and the name has been chosen on purpose to stay away from the functional feel as providing a full fledged functional type API for Vert.x future is a non goal.

PR #1761, opened 2017-01-11
HttpClient leaves connection open on websocket request failure (resource leak) - fixes #1757

### LINKED ISSUE #1757: HttpClient leaves connection open on websocket request failure (resource leak)
When running this code:
```java
        Vertx vertx = Vertx.vertx();
        vertx.createHttpServer().requestHandler(req -> {
            req.response().setStatusCode(200);
            req.response().headers().set("Connection", "close");
            req.response().end();
        }).listen(44444);
        vertx.createHttpClient().websocket(44444, "localhost", "/", ws -> {
            System.out.println("WS Unexpectedly connected");
        }, t -> {
            System.err.println("WS Connection failed");
            t.printStackTrace();
        });
```
Then end result is `WS Connection failed` (as expected) but `netstat -natp|grep :44444` (in Linux) shows the connection is still ESTABLISHED. Only closing the java process (naturally) gets rid of the connection.

For a regular GET request:
```java
        Vertx vertx = Vertx.vertx();
        vertx.createHttpServer().requestHandler(req -> {
            req.response().setStatusCode(200);
            req.response().headers().set("Connection", "close");
            req.response().end();
        }).listen(44444);
        vertx.createHttpClient().get(44444, "localhost", "/", c -> {
            System.out.println("GET connected");
        }).end();
```

then `netstat -natp|grep :44444` shows no ESTABLISHED connections, as expected.

I added the `Connection: close` header so that HttpClient wouldn't pool the connection and leave it open because of that.

xkr47:
Continuing from #1757:

Ok, I tried, but also had to add `handshaker = null;` so that it wouldn't try to send a graceful websocket frame to shut it down.

This seems to work fine for me with both versions of the code posted in #1757. However I do not know if there are further resource cleanups that need to be done, maybe related to connection pools, so please consider that.


vietj:
can you provide the corresponding tests ?

xkr47:
I'll try..

xkr47:
Ok, hope this is good enough...

src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -189,6 +189,8 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
         if (msg instanceof HttpResponse) {
           HttpResponse resp = (HttpResponse) msg;
           if (resp.getStatus().code() != 101) {
+            handshaker = null;
+            close();
```
vietj: I think close should be called after the handleException statement

src/test/java/io/vertx/test/core/Issue1757_WebsocketFailureTest.java
```diff
@@ -0,0 +1,115 @@
...
+                        ws.close();
+                    }, t -> {
+                        System.out.println("Client got exception");
+                        addResult(t);
+                    });
+                });
+            }
+        });
+        try {
+            System.out.println("Test thread waiting for resultQueue");
```
vietj: no System.out

src/main/java/io/vertx/core/http/impl/ClientConnection.java
xkr47: I just copypasted from line 215-218.. I have no clue which way is better..

src/test/java/io/vertx/test/core/Issue1757_WebsocketFailureTest.java
xkr47: All System.out removed.

src/main/java/io/vertx/core/http/impl/ClientConnection.java
vietj: ok

vietj:
can you move the test to the `WebsocketTest` class ?

src/test/java/io/vertx/test/core/WebsocketTest.java
```diff
@@ -1350,4 +1357,86 @@ public void testClientWebsocketWithHttp2Client() throws Exception {
     }));
     await();
   }
+
+  @Test
+  public void testIssue1757_1() throws Throwable {
```
vietj: can you rename this to something more meaningful ?

src/test/java/io/vertx/test/core/WebsocketTest.java
```diff
@@ -1350,4 +1357,86 @@ public void testClientWebsocketWithHttp2Client() throws Exception {
...
+    doTestIssue1757(true);
+  }
+
+  @Test
+  public void testIssue1757_2() throws Throwable {
```
vietj: can you rename this to something more meaningful ?

src/test/java/io/vertx/test/core/WebsocketTest.java
vietj: @xkr47 ?

src/test/java/io/vertx/test/core/WebsocketTest.java
vietj: @xkr47 ?

vietj:
thanks for the contrib!

PR #1762, opened 2017-01-13
Expose the HttpClientRequest in HttpClientResponse

jo5ef:
The method is already there, just not exposed in the interface. I suggest to expose it.


vietj:
+1, this does not harm

PR #1763, opened 2017-01-16
Fixes vert-x3/issues#75 Send messages instantly

tsegismont:
We must make sure messages are sent instantly, whether sent from worker verticle or executeBlocking

### DIFF
--- a/src/test/java/io/vertx/test/core/EventBusTestBase.java
+++ b/src/test/java/io/vertx/test/core/EventBusTestBase.java
@@ -393,25 +393,20 @@ public abstract class EventBusTestBase extends VertxTestBase {
       receivedLatch.countDown();
     }).completionHandler(ar -> {
       assertTrue(ar.succeeded());
-      vertices[0].deployVerticle(new AbstractVerticle() {
-        @Override
-        public void start() throws Exception {
-          vertx().executeBlocking(fut -> {
-            vertices[0].eventBus().send(ADDRESS1, expectedBody);
-            try {
-              awaitLatch(receivedLatch); // Make sure message is sent even if we're busy
-            } catch (InterruptedException e) {
-              Thread.interrupted();
-              fut.fail(e);
-            }
-            fut.complete();
-          }, ar -> {
-            if (ar.succeeded()) {
-              testComplete();
-            } else {
-              fail(ar.cause());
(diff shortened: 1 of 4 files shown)


tsegismont:
Other PRs in cluster managers coming

tsegismont:
The fix is ready here and in cluster managers (I checked them all).

Here it consists in using the current context for sending only if:
- it exists (we could be sending from a regular thread)
- it is an EL context

The sending context is not very important, because we use the context stored in the handler registration to deliver the message.

So when the sending context is a worker context, fall back to the default context.
Indeed, the "event bus connected" callback (just an example) might not be executed if worker is busy on (user) blocking code.

In cluster managers, the fix consists in assigning a dedicated worker executor for AsyncMap implementations (created with ContextInternal.createWorkerExecutor). So ordering is still guaranteed where needed, but does not interfere with the user's code.

tsegismont:
@vietj please have a look. Thx

src/test/java/io/vertx/test/core/EventBusTestBase.java
```diff
@@ -381,6 +383,41 @@ public void start() throws Exception {
...
+  public void testSendFromExecuteBlocking() throws Exception {
+    String expectedBody = TestUtils.randomAlphaString(20);
+    CountDownLatch receivedLatch = new CountDownLatch(1);
+    startNodes(2);
+    vertices[1].eventBus().<String>consumer(ADDRESS1, msg -> {
+      assertEquals(expectedBody, msg.body());
+      receivedLatch.countDown();
+    }).completionHandler(ar -> {
+      assertTrue(ar.succeeded());
+      vertices[0].deployVerticle(new AbstractVerticle() {
```
vietj: why do we need to deploy a verticle here ? to get a new context ?

tsegismont:
Yes. It could be changed to use `runOnContext`. TBH I simply reused the
structure of the `testSendFromWorker` test.

2017-01-23 0:17 GMT+01:00 Julien Viet <notifications@github.com>:

vietj:
can you update it please ?

tsegismont:
Done

2017-01-23 11:01 GMT+01:00 Julien Viet <notifications@github.com>:

PR #1766, opened 2017-01-17
Fixes #1765 NullPointer in sendToSubs

### LINKED ISSUE #1765: NullPointer in sendToSubs
Cloned from vert-x3/vertx-hazelcast#52

tsegismont:
The io.vertx.core.spi.cluster.ChoosableIterable contract does not guarantee that
after calling io.vertx.core.spi.cluster.ChoosableIterable.isEmpty,
io.vertx.core.spi.cluster.ChoosableIterable.choose returns a non null entry.

This could happen if a consumer is unregistered on another in the cluster.

Publish shouldn't be affected but I still added a test. In order to reproduce
I had to annotate the test with @Repeat(times=1000)
I can't leave the annotation as this would take much too long with real cluster managers.
The issue could be reproduced with the FakeClusterManager anyway.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
@@ -16,17 +16,35 @@
 
 package io.vertx.core.eventbus.impl.clustered;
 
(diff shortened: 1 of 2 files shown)


vietj:
👍

PR #1767, opened 2017-01-18
Recursive mergeIn for JsonObject

jo5ef:
Implementation of a recursive mergeIn method for JsonObject. The existing mergeIn method keeps its semantic of a flat merge. But the new overloads allow deep merges. e.g.:

```
JsonObject defaultConfig = new JsonObject()
   .put("mongo", new JsonObject()
      .put("db_name", "mydb")
      .put("host", "localhost")
      .put("port", 27017));

// configFile: { "mongo": { "host": "prod.mongo.dbhost" } }
JsonObject config = defaultConfig.mergeIn(configFile, true);
// config now contains all defaults and the host that was overriden in the configFile
```

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/src/main/java/io/vertx/core/json/JsonObject.java
@@ -660,12 +660,58 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
   /**
    * Merge in another JSON object.
    * <p>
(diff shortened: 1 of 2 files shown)


vietj:
looks good, can you add spaces between keywords like if/for and the following parenthesis ?

for instance: `if(depth < 1) ` -> `if (depth < 1) `

jo5ef:
absolutely, done.

PR #1771, opened 2017-01-20
NetServer rework

vietj:
Provide the `NetServerBase` that manages the state of a TCP server that is extracted from `NetServerImpl` in order to reuse it in _vertx-mqtt-server_ that will benefit of its feature. The _vertx-mqtt-server_ is based on Netty's MQTT codec.

### DIFF
--- a/src/main/java/io/vertx/core/spi/metrics/VertxMetrics.java
+++ b/src/main/java/io/vertx/core/spi/metrics/VertxMetrics.java
@@ -126,12 +126,11 @@ public interface VertxMetrics extends Metrics, Measured {
    * the provided {@code server} argument can be used to distinguish the different {@code TCPMetrics}
    * instances.
    *
-   * @param server       the Vert.x net server
    * @param localAddress localAddress the local address the net socket is listening on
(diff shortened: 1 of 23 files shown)


src/main/java/io/vertx/core/spi/metrics/VertxMetrics.java
```diff
@@ -126,12 +126,11 @@
...
    * instances.
    *
-   * @param server       the Vert.x net server
    * @param localAddress localAddress the local address the net socket is listening on
-   * @param options      the options used to create the {@link io.vertx.core.net.NetServer}
+   * @param options      the options used to create the {@link NetServer}
    * @return the net server metrics SPI
    */
-  TCPMetrics<?> createMetrics(NetServer server, SocketAddress localAddress, NetServerOptions options);
+  TCPMetrics<?> createMetrics(SocketAddress localAddress, NetServerOptions options);
```
vietj: yes, I'll record a line for that in change log

vietj: I don't think it is used because NetServer only provides useless stuff : listen / close / handler, so it should be fine.

PR #1774, opened 2017-01-22
JsonObject POJO mapping

lukehutch:


lukehutch:
Will sign the Eclipse CLA if needed, but I'm first submitting this for feedback on the approach.

lukehutch:
PS I opted to write a custom POJO-to-JsonObject mapper for speed, but I used Jackson for the JSON-to-POJO mapper (because JSON-to-POJO mapping is extremely complicated, and hard to get right).

An option would be to instead use Jackson for both, by defining:

```java
public JsonObject(Object obj) {
    this(Json.encode(obj));
}
```

This just seems really wasteful if the JsonObject is going to be serialized yet again to be sent over the wire or stored in MongoDB (the POJO is serialized, deserialized, and serialized again).

vietj:
or this : https://www.leveluplunch.com/java/examples/convert-object-bean-properties-map-key-value/ show how to encode a Pojo to a Map using jackson, then the map could be wrapped in JsonObject

lukehutch:
It would be worth benchmarking methods like that. Efficiency is my primary concern here.

vietj:
@lukehutch sure, but I was asking, wouldn't the way using jackson work ok ?

lukehutch:
I assume you were referring specifically to this example code:

```java
Map<String, Object> objectAsMap = objectMapper.convertValue(moleskineNoteBook, Map.class);
```

That looks like it does the right thing. Note the caveats in the docs:

> `<Map> Map com.fasterxml.jackson.databind.ObjectMapper.convertValue(Object fromValue, Class<Map> toValueType) throws IllegalArgumentException`
> 
> Convenience method for doing two-step conversion from given value, into instance of given value type, if (but only if!) conversion is needed. If given value is already of requested type, value is returned as is.
> 
> This method is functionally similar to first serializing given value into JSON, and then binding JSON data into value of given type, but should be more efficient since full serialization does not (need to) occur. However, same converters (serializers, deserializers) will be used as for data binding, meaning same object mapper configuration works.
> 
> Note that it is possible that in some cases behavior does differ from full serialize-then-deserialize cycle: in most case differences are unintentional (that is, flaws to fix) and should be reported. It is not guaranteed, however, that the behavior is 100% the same: the goal is just to allow efficient value conversions for structurally compatible Objects, according to standard Jackson configuration.
> 
> Further note that functianality is not designed to support "advanced" use cases, such as conversion of polymorphic values, or cases where Object Identity is used.

With this, the required methods in `JsonObject` would be something like:

```java
  /**
   * Creating a JsonObject from the fields of a Java object.
   * Faster than calling new JsonObject(Json.encode(pojo)).
   * 
   * @param obj
   *          The object to convert to a JsonObject.
   * @throws IllegalArgumentException
   *          if conversion fails due to an incompatible type.
   */
  @SuppressWarnings("unchecked")
  public JsonObject(Object obj) {
    this((Map<String, Object>) Json.mapper.convertValue(obj, Map.class));
  }

  /**
   * Method for instantiating a Java object from a JsonObject.
   * 
   * @param type
   *          The type to instantiate from the JsonObject.
   * @throws IllegalArgumentException
   *          if the type cannot be instantiated.
   */
  public <T> T toInstance(Class<T> type) {
    try {
      return Json.mapper.readValue(Json.encode(this), type);
    } catch (Throwable e) {
      throw new IllegalArgumentException(
          "Could not instantiate type " + type.getName(), e);
    }
  }
```

This looks pretty good to me. Thoughts?

lukehutch:
@vietj OK, well if you like the code above, feel free to copy and paste into JsonObject.java, then close the original PR. I don't want to have to go through the whole Eclipse CLA validation steps right now, but I give you permission to use the code above for any purpose, and with any license!

vietj:
@lukehutch sure and the tests most importantly :-)

vietj:
@lukehutch eclipse validation is not that terrible, no need to send a fax :-)

lukehutch:
I'll get around to eclipse validation eventually! But yes, you have my permission to use the test code too. Just to confirm, you'll take it from here, right?

lukehutch:
@vietj I just discovered that `convertValue()` works in both directions, so I updated the `toInstance()` code in the comment above. The result is even simpler.

vietj:
@lukehutch yes I will take care of it

lukehutch:
@vietj I signed the CLA, and updated the PR, but I still see the "All checks have failed" message.

vietj:
@lukehutch you would need to make a new PR with squashed commits and signed-off, the checker is confused with mulitple commits whose some are not signed-off

lukehutch:
@vietj OK, I needed to learn how to do this anyway, so I created a new PR (#1793). Closing this one in favor of that. Sorry that this was such a long-winded process!

PR #1776, opened 2017-01-25
Websocket unmasked frame support

stecolet:
As you can see from WebSocket RCF protocol: https://tools.ietf.org/html/rfc6455#section-5.2 masking frame is only suggested (not mandatory). Vert.x 3.3.3 supports only masked frame.  My project instead, requires unmasked frame so I'v added unmasked frame configuration in order to create client and server web socket without masking frames.

Netty library has already provided an API which supports unmasked frame. As you can see by the commits I've only configured a flag which calls correctly Netty library with un/masked flag properly setted.


vietj:
your revert commit is not signed and make the ip-validation fail.

can you provide a new clean PR ?

stecolet:
Closed due to missing signed commits

stecolet:
I've submitted a new PR. Is https://github.com/eclipse/vert.x/pull/1778
should be correct for your merge ?

thank you
Stefano Coletta.


2017-01-25 12:32 GMT+01:00 Julien Viet <notifications@github.com>:

PR #1778, opened 2017-01-25
Unmasking frame on websocket

stecolet:
Signed-off-by: stecolet@gmail.com <stecolet@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpClientOptions.java
+++ b/src/main/java/io/vertx/core/http/HttpClientOptions.java
@@ -141,7 +141,7 @@ public class HttpClientOptions extends ClientOptionsBase {
   /**
    * Default WebSocket Masked bit is true as depicted by RFC. PerformingUnMasking will be false
    */
-  public static final boolean DEFAULT_FRAME_UNMASKING = false;
+  public static final boolean DEFAULT_SEND_UNMASKED_FRAMES = false;
 
   private boolean verifyHost = true;
   private int maxPoolSize;
@@ -236,7 +236,7 @@ public class HttpClientOptions extends ClientOptionsBase {
     initialSettings = new Http2Settings();
     alpnVersions = new ArrayList<>(DEFAULT_ALPN_VERSIONS);
     http2ClearTextUpgrade = DEFAULT_HTTP2_CLEAR_TEXT_UPGRADE;
-    sendUnmaskedFrames = DEFAULT_FRAME_UNMASKING;
+    sendUnmaskedFrames = DEFAULT_SEND_UNMASKED_FRAMES;
   }
 
   @Override
--- a/src/main/java/io/vertx/core/http/HttpClientOptions.java
+++ b/src/main/java/io/vertx/core/http/HttpClientOptions.java
@@ -159,7 +159,7 @@ public class HttpClientOptions extends ClientOptionsBase {
   private Http2Settings initialSettings;
   private List<HttpVersion> alpnVersions;
   private boolean http2ClearTextUpgrade;
-  private boolean unmaskedFrame;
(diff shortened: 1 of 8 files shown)


stecolet:
As you can see from https://tools.ietf.org/html/rfc6455#section-5.2 websocket protocol supports also unmasked frame while Vetx 3.3.0 not. In this PR I've configured Netty properly to support unmasked frame when unmaskedFrame flasg is enabled in httpServer/client options

vietj:
rather name the option "performFrameMasking" or "performFrameUnmasking"

stecolet:
As you can see from last commit I've renamed option to performUnmaskingFrame.

vietj:
almost good : we need to have equals/hashcode in HttpClientOptions and HttpServerOptions and these tested in the options test in Http1xTest

stecolet:
Yes, I'm very agree. I've done.

Thank you
Stefano

vietj:
looking at the interest of the PR, I don't understand why you need this feature as the RFC6455 clearly says that the client frames must be masked and the server frames must not be masked.

stecolet:
In my project there's a WebSocket communication between two server side services. Maybe this kind of protocol shoud not be the best way, but I can't change it because I'm going to integrate an existing service (which are using WebSocket). For this reason my need is to create a WebSocket server accepting unmasked frame.

Am I Clear??
Stefano.

vietj:
yes it is clear now

vietj:
having read the RFC more carefully, we should rename the options to have something more clear:

- client option should be called : `sendUnmaskedFrames` default = false
- server option should be called : `acceptUnmaskedFrames` default = false

There should be more documentation:

- the client documentation must say that this option for sending frames only and by default masked frames and activating this option bypasses the RFC
- the server documentation must say that this option is for accepting frames only and by default unmasked frames are rejected and activating this option bypasses the RFC

stecolet:
@vietj, When you say "more documentation", it means to explain the behaviour only by JavaDoc onthe set method of the field? or more documentation on others doc file ?

vietj:
@stecolet javadoc only, as this is really specific behavior for a non general use case, it's better to avoid the general reference guide

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -138,6 +138,11 @@
    */
   public static final boolean DEFAULT_HTTP2_CLEAR_TEXT_UPGRADE = true;
 
+  /**
+   * Default WebSocket Masked bit is true as depicted by RFC. PerformingUnMasking will be false
+   */
+  public static final boolean DEFAULT_FRAME_UNMASKING = false;
```
vietj: last but not least rename `DEFAULT_FRAME_UNMASKING` to `DEFAULT_SEND_UNMASKED_FRAMES`

src/main/java/io/vertx/core/http/HttpServerOptions.java
```diff
@@ -94,6 +94,11 @@
    */
   public static final boolean DEFAULT_DECOMPRESSION_SUPPORTED = false;
 
+  /**
+   * Default WebSocket Masked bit is true as depicted by RFC. PerformingUnMasking will be false
+   */
+  public static final boolean DEFAULT_FRAME_UNMASKING = false;
```
vietj: last but not least rename `DEFAULT_FRAME_UNMASKING` to `DEFAULT_ACCEPT_UNMASKED_FRAMES`

stecolet: Sure, I'm sorry, I forgot it.
Thank you for your support.
Stefano

vietj:
thanks for the contribution

PR #1780, opened 2017-01-27
Http client redirect - fixes #1779

### LINKED ISSUE #1779: HttpClient follow redirect
HttpClient request can follow redirect on a 30x status code response.

vietj:
please review

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpClient.java
+++ b/src/main/java/io/vertx/core/http/HttpClient.java
@@ -17,10 +17,13 @@
 package io.vertx.core.http;
 
 import io.vertx.codegen.annotations.Fluent;
+import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.Handler;
 import io.vertx.core.MultiMap;
 import io.vertx.core.metrics.Measured;
+
+import java.util.function.Function;
 
 /**
  * An asynchronous HTTP client.
@@ -1363,6 +1366,31 @@ public interface HttpClient extends Measured {
    */
   WebSocketStream websocketStream(String requestURI, MultiMap headers, WebsocketVersion version,
                                   String subProtocols);
+
+  /**
+   * Set a redirect handler for the web client.
+   * <p>
+   * The redirect handler is called when a {@code 3xx} response is received and the request is configured to
+   * follow redirects with {@link HttpClientRequest#setFollowRedirects(boolean)}.
+   * <p>
+   * The redirect handler is passed the {@link HttpClientResponse}, it can return an {@link HttpClientRequest} or {@code null}.
+   * <ul>
+   *   <li>when null is returned, the original response is processed by the original request response handler</li>
+   *   <li>when a new {@code HttpClientRequest} is returned, the client will send this new request</li>
+   * </ul>
+   * The handler must return a request unsent so the client can further configure it and send it.
+   *
+   * @param handler the new redirect handler
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  HttpClient redirectHandler(Function<HttpClientResponse, HttpClientRequest> handler);
+
+  /**
+   * @return the current redirect handler.
+   */
+  @GenIgnore
+  Function<HttpClientResponse, HttpClientRequest> redirectHandler();
 
   /**
    * Close the client. Closing will close down any pooled connections.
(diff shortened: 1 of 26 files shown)


src/main/java/io/vertx/core/http/HttpClient.java
```diff
@@ -1089,6 +1369,31 @@ WebSocketStream websocketStream(String requestURI, MultiMap headers, WebsocketVe
                                   String subProtocols);
 
   /**
+   * Set a redirect handler for the web client.
```
tsegismont: for the "http client" would be better to avoid confusion with the vertx-web client

src/main/java/io/vertx/core/http/HttpClient.java
```diff
@@ -1089,6 +1369,31 @@ WebSocketStream websocketStream(String requestURI, MultiMap headers, WebsocketVe
...
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  HttpClient redirectHandler(Function<HttpClientResponse, Future<HttpClientRequest>> handler);
+
+  /**
+   * @return the current redirect handler.
+   */
+  @GenIgnore
+  Function<HttpClientResponse, Future<HttpClientRequest>> redirectHandler();
```
tsegismont: Usually we don't return handlers (because the user code knows what was set). Is there a reason for this one to be present?

src/main/java/io/vertx/core/http/impl/HttpClientRequestBase.java
```diff
@@ -35,18 +35,21 @@
...
   protected final String path;
   protected final String host;
+  protected final int port;
   protected final String query;
   private Handler<Throwable> exceptionHandler;
   private long currentTimeoutTimerId = -1;
+  private long currentTimeoutMs;
   private long lastDataReceived;
-  protected boolean exceptionOccurred;
+  protected Throwable exceptionOccurred;
```
tsegismont: Is renaming possible? I know the class could be extended but it's an internal class at the same time

src/main/java/io/vertx/core/http/impl/HttpUtils.java
```diff
@@ -52,6 +52,172 @@
...
+   * Removed dots as per <a href="http://tools.ietf.org/html/rfc3986#section-5.2.4>rfc3986</a>.
+   *
+   * There are 2 extra transformations that are not part of the spec but kept for backwards compatibility:
+   *
+   * double slash // will be converted to single slash and the path will always start with slash.
+   *
+   * @param path raw path
+   * @return normalized path
+   */
+  public static String removeDots(CharSequence path) {
```
tsegismont: I reviewed the same kind of code for @pmlopes in vertx-web recently. Can you check with him if we can keep just one?

tsegismont:
Please add a test to make sure one doesn't have to provide "RequestOptions" to send a https request if the client was not configured with ssl enabled but a https url is provided in any xxxAbs method.

src/main/java/io/vertx/core/http/HttpClient.java
vietj: indeed

src/main/java/io/vertx/core/http/HttpClient.java
vietj: yes we do sometimes, it is here in case someone wants to sets its own handler but get the previous one to do some kind of decoration or only wants to handle a specific case and delegates the rest of the behavior to the previous handler

src/main/java/io/vertx/core/http/impl/HttpClientRequestBase.java
vietj: renaming to ?

src/main/java/io/vertx/core/http/impl/HttpUtils.java
vietj: it was a copy paste and I think we said it would be reused

tsegismont: Sorry I missed that conversation.

src/main/java/io/vertx/core/http/impl/HttpClientRequestBase.java
tsegismont: `throwableCaught` or just `throwable`

src/main/java/io/vertx/core/http/HttpClient.java
tsegismont: Makes sense. Thanks for the details.

src/main/java/io/vertx/core/http/impl/HttpClientRequestBase.java
vietj: exceptionOccured seems more precise

Issue #1783, opened 2017-01-29
EventBus interceptors should be able to intercept on both send and receipt


emmanuelidi:
Currently, the eventbus interceptor can only inspect and transform messages on send. The interceptor does not have the ability to perform complementary operations on message delivery. 

A use-case would be for distributed tracing technologies where tracing of intra-cluster communication is the goal. One would want to "start" a trace on message send and "close" the trace on delivery (or, perform further trace-state management if a response is expected).

PR forthcoming.

MazizEsa:
Not yet merge to master?

vietj:
it needs some rework but it could eventually be

MazizEsa:
@vietj Another thing, is there anyway as a workaround to achieve this so far? @emmanuelidi too.

vietj:
what is your use case ?

MazizEsa:
I have a case where when the message passby to the event bus, it must be validated and checked before being published to other subscriber. After an approval process, we will publish those message to the subscribers.

vietj:
in your case I think you should create a MessageProducer that wraps the EventBus MessageProducer and performs this validation.

MazizEsa:
Just to clarify, will this work if the message comes from bridged sockjs server? We plan to intercept and stop it from publishing when the message come in.

vietj:
yes I think it should

DrJonnie:
Hi, any progress on this this? It would be really awesome. In my current project i would
like to use this feature to transparently propagate a tenant identification between event loop contexts.

emmanuelidi:
I've stopped using Vert.x so, unfortunately, the window of time for me to work on this has come and gone

vietj:
@DrJonnie here is a PR that rebases @emmanuelidi contribution with some lifting https://github.com/eclipse-vertx/vert.x/pull/2632 it shall be in 3.6

PR #1784, opened 2017-01-29
feat: Extend EventBus listener to allow for delivery interception

emmanuelidi:
resolves: issue #1783

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/EventBus.java
+++ b/src/main/java/io/vertx/core/eventbus/EventBus.java
@@ -261,8 +261,28 @@ public interface EventBus extends Measured {
    *
    * @param interceptor  the interceptor
    * @return a reference to this, so the API can be used fluently
+   * @deprecated use {@link EventBus#addInterceptor(Interceptor)} instead.
    */
+  @Deprecated
   EventBus addInterceptor(Handler<SendContext> interceptor);
+
+  /**
+   * Add an interceptor that will be called whenever a message is sent from, and delivered by, Vert.x
+   *
+   * @param interceptor  the interceptor
+   * @return a reference to this, so the API can be used fluently
+   */
+  EventBus addInterceptor(Interceptor interceptor);
+
+  /**
+   * Remove an interceptor
+   *
+   * @param interceptor  the interceptor
+   * @return a reference to this, so the API can be used fluently
+   * @deprecated use {@link EventBus#removeInterceptor(Interceptor)} instead.
+   */
+  @Deprecated
+  EventBus removeInterceptor(Handler<SendContext> interceptor);
 
   /**
    * Remove an interceptor
@@ -270,7 +290,6 @@ public interface EventBus extends Measured {
    * @param interceptor  the interceptor
    * @return a reference to this, so the API can be used fluently
    */
-  EventBus removeInterceptor(Handler<SendContext> interceptor);
-
+   EventBus removeInterceptor(Interceptor interceptor);
 }
 
--- a/src/main/java/io/vertx/core/eventbus/Interceptor.java
+++ b/src/main/java/io/vertx/core/eventbus/Interceptor.java
@@ -1,0 +1,20 @@
+package io.vertx.core.eventbus;
+
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.Handler;
+
+/**
+ * An eventbus interceptor that will be called whenever a message is sent from, and delivered by,
+ * Vert.x
+ */
+@VertxGen
+public interface Interceptor extends Handler<SendContext> {
+  /**
+   * A message delivery attempt is about to occur.
+   *
+   * @param context the message delivery context to handle
+   */
+  default void handleDelivery(DeliveryContext context) {
+    context.next();
+  }
+}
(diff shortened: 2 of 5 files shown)


src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -512,6 +522,61 @@ private void unregisterAll() {
     });
   }
 
+  // NOTE: Consider merging the delivery & send context into a general `TransmissionContext`
```
vietj: good idea, we should do that now instead instead of postponing it

src/main/java/io/vertx/core/eventbus/Interceptor.java
```diff
@@ -0,0 +1,20 @@
...
+ * Vert.x
+ */
+@VertxGen
+public interface Interceptor extends Handler<SendContext> {
+  /**
+   * A message delivery attempt is about to occur.
+   *
+   * @param context the message delivery context to handle
+   */
+  default void handleDelivery(DeliveryContext context) {
```
vietj: rather called `handleReceive` as this is the name used internally, _delivery_ is more general and care about both _send_ and/or _receive_ actions

src/main/java/io/vertx/core/eventbus/Interceptor.java
```diff
@@ -0,0 +1,20 @@
...
+
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.Handler;
+
+/**
+ * An eventbus interceptor that will be called whenever a message is sent from, and delivered by,
+ * Vert.x
+ */
+@VertxGen
+public interface Interceptor extends Handler<SendContext> {
```
vietj: instead of `Handler<SendContext>`, rather a `handleSend(DeliveryContext)`

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: DeliveryContext is good for both _send_ and _receive_ , we should remove `SendContext` instead. I know it can break, but this API was never publicly advertised or documented, so I would rather change it now and provide something we can support

src/main/java/io/vertx/core/eventbus/EventBus.java
```diff
@@ -261,16 +261,35 @@
    *
    * @param interceptor  the interceptor
    * @return a reference to this, so the API can be used fluently
+   * @deprecated use {@link EventBus#addInterceptor(Interceptor)} instead.
    */
+  @Deprecated
   EventBus addInterceptor(Handler<SendContext> interceptor);
```
vietj: remove it instead of deprecating it, this API was never really advertised and supported. So I prefer to finish it, document it and support it, instead of deprecating things.

src/main/java/io/vertx/core/eventbus/Interceptor.java
vietj: now I rethink about it, we should rather have 4 methods:

- sendHandler(Handler<DeliveryContext>)
- Handler<DeliveryContext> sendHandler() (with @ genignore)
- receiveHandler(Handler<DeliveryContext>)
- Handler<DeliveryContext> receiveHandler() (with @ genignore)

This way the interface is really @ genignore enabled

src/main/java/io/vertx/core/eventbus/EventBus.java
emmanuelidi: Ok. I'll likely have to go and update the existing interceptors, though, which shouldn't be an issue.

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
emmanuelidi: Ok.

emmanuelidi:
I'd like to edit and finalize tomorrow. Any other comments forthcoming?

vietj:
@emmanuelidi no further comments for now

vietj:
hi @emmanuelidi any news about it ?

emmanuelidi:
@vietj I'll take another look tonight. My Patriots were in the Super Bowl so I had some distractions the past few days.

src/main/java/io/vertx/core/eventbus/Interceptor.java
emmanuelidi: I'm not sure what you mean with the second comment. Are you asking to make the Interceptor stateful and allow it to store a send and receive handler?

jan-zajic:
+1 for this, @emmanuelidi any news about it ?

src/main/java/io/vertx/core/eventbus/Interceptor.java
emmanuelidi: @vietj I don't understand your second comment. Is there an example in the existing project code which I should use as an example?

emmanuelidi:
@jan-zajic I had a question for @vietj but otherwise done. If I don't hear back, I'll just rebase and push the changes

vietj:
actually implemented based on this proposal in 3.6

PR #1793, opened 2017-02-02
Add JsonObject POJO mapping convenience methods

lukehutch:
Signed-off-by: Luke Hutchison <luke.hutch@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/src/main/java/io/vertx/core/json/JsonObject.java
@@ -71,7 +71,7 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
 
   /**
    * Create a JsonObject from the fields of a Java object.
-   * Faster than calling `new JsonObject(Json.encode(pojo))`.
+   * Faster than calling `new JsonObject(Json.encode(obj))`.
    * 
    * @param obj
    *          The object to convert to a JsonObject.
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -70,6 +70,32 @@ public JsonObject(Map<String, Object> map) {
...
+   * Create a JsonObject from the fields of a Java object.
+   * Faster than calling `new JsonObject(Json.encode(pojo))`.
+   * 
+   * @param obj
+   *          The object to convert to a JsonObject.
+   * @throws IllegalArgumentException
+   *          if conversion fails due to an incompatible type.
+   */
+  @SuppressWarnings("unchecked")
+  public JsonObject(Object obj) {
```
vietj: I would prefer we move that to the Json as static method. Constructors with Object type tends to be error prone because of overloading

lukehutch: Maybe then there should be a static factory method `JsonObject.fromInstance(Object)`, to match `JsonObject#toInstance()`? It would be nice to only ever have to deal with JsonObject, and not additionally with Json.

vietj: @lukehutch agreed, however it should be renamed with a name like `mapTo` / `mapFrom` in both case that has the semantic that mappings happens.

lukehutch:
@vietj done

vietj:
looks good, thanks for the contribution

lukehutch:
You're welcome. I just made a few changes, but I'm done now :-) Thanks for reviewing, and for the suggestions!

lukehutch:
@vietj I just realized something about the `ObjectMapper#convertValue()` approach: List-typed fields (and, I assume, array-typed fields) by default will not be properly serialized to a JsonArray -- so you'll end up with a JsonObject that contains values of type List (or array), not of type JsonArray.

PR #1795, opened 2017-02-02
Add documentation for #1793 (JSON<->POJO mapping)

lukehutch:
Signed-off-by: Luke Hutchison <luke.hutch@gmail.com>


lukehutch:
@vietj There may need to be some discussion here as to whether the fields need to be public or not to be (de)serialized (or have public getters/setters), depending on how the object mapper is configured.

Also, I'm guessing that the Java class being deserialized needs to have a public default constructor so that it can be instantiated by Jackson.

vietj:
I believe that we should say explicitly that this uses Jackson mapping and it should follow  the results, without being agnostic or hiding this fact.

lukehutch:
@vietj does the most recent commit work?

vietj:
looks good, however we create documentation from the package-info javadoc (because it allows us to have a type proof documentation and embed the examples from real java classes).

I think the file to modify is https://github.com/eclipse/vert.x/blob/master/src/main/java/docoverride/json/package-info.java and if you make `mvn clean compile` it will rebuild the asciidoctor file

lukehutch:
Sorry, on the previous commit, I tried running `mvn clean compile` as you suggested, and it failed with 

No Vert.x Snapshots Allowed during a release!
Found Banned Dependency: io.vertx:vertx-docgen:jar:3.4.0-SNAPSHOT
Found Banned Dependency: io.vertx:vertx-codegen:jar:3.4.0-SNAPSHOT

I committed the changes anyway (after moving the doc contributions to the right place, as you suggested). I assumed that `mvn clean compile` would not cause any large-scale file changes to what was checked in. However, it looks like a bunch of files were deleted from `src/main/generated`. What happened here?

(Since I don't know how the doc autogeneration works, this is probably simpler for you to fix than for me, so maybe you can just take my few changes and add them to the docs :o)  Sorry this wasn't a clean PR!)

vietj:
thanks, I copied it there https://github.com/eclipse/vert.x/commit/f4c5b3919406ba36ce146ee9cf4b1ac3d4a40c5a

vietj:
Setting as invalid for the record, it does not mean it was valid :-)

Issue #1797, opened 2017-02-06
When cluster manager fails to join the cluster, VertxImpl discards the error, leaving the caller stuck


guss77:
As evident in https://github.com/eclipse/vert.x/blob/868595511d8c63a6589a260da7cdea585219f20c/src/main/java/io/vertx/core/impl/VertxImpl.java#L165

if the `clusterManager.join()` call returns a failure, the `resultHandler` submitted to the `VertxImpl` constructor is never called (neither success nor failure) resulting in the handler never completing and the caller getting stuck with an unresponsive stack.

jan-zajic:
@vietj, @guss77 please close this issue, because it's resolved with #1798 ?

vietj:
can you confirm @tsegismont ?

tsegismont:
Yes, #1798 closes this issue.

PR #1798, opened 2017-02-06
notify caller that the cluster manager they requested failed to join

guss77:
Signed-off-by: Oded Arbel <oded@geek.co.il>

Fixes issue #1797

### DIFF
--- a/src/main/java/io/vertx/core/impl/VertxImpl.java
+++ b/src/main/java/io/vertx/core/impl/VertxImpl.java
@@ -163,6 +163,7 @@ public class VertxImpl implements VertxInternal, MetricsProvider {
       this.clusterManager.join(ar -> {
         if (ar.failed()) {
           log.error("Failed to join cluster", ar.cause());
+          resultHandler.handle(Future.failedFuture(ar.cause()));
         } else {
           // Provide a memory barrier as we are setting from a different thread
           synchronized (VertxImpl.this) {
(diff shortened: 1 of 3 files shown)


tsegismont:
Could you add a test to your PR?

guss77:
I can try - setting up the Hazelcast cluster manager to fail wasn't easy for me :smiley: . I'll try to do something for tomorrow - I'm kind of out of time today.

tsegismont:
No worries :)

You shouldn't use a failing Hazelcast cluster manager here. The Vert.x core
project should not be aware of any implementation. You could write your own
mock cluster manager for the test.

2017-02-06 15:21 GMT+01:00 Oded Arbel <notifications@github.com>:

guss77:
Added unit test - I hope I put it in the right place. Its kind of simplistic, and if the bug is not fixed it doesn't "fail" - instead it errors out.

tsegismont:
The test looks good to me. As for its location, I believe it should be
added to io.vertx.test.core.CreateVertxTest, instead
of io.vertx.test.core.VertxTest

Thank you!

2017-02-06 21:35 GMT+01:00 Oded Arbel <notifications@github.com>:

guss77:
@tsegismont - thanks for your comments. I've moved the test as you requested, and I see that `CreateVertxTest` has a facility to create a clustered vertx so I used that. After the change the test still errors out without the fix and passes with the fix.

tsegismont:
Can you revert the changes on io.vertx.test.core.VertxTest ?

guss77:
Oops. Thanks for catching that - automatic tools, what can you do? ;-)

tsegismont:
LGTM, thanks!

2017-02-07 11:14 GMT+01:00 Oded Arbel <notifications@github.com>:

PR #1806, opened 2017-02-09
Avoid class method calls in examples

tsegismont:
Class method calls get translated, so the resulting snippet contains a
weird method definition throwing an error.


tsegismont:
Closing as we'll update codetrans to skip those methods.

Issue #1807, opened 2017-02-09
Feature: Support large message framing for text messages in websockets


rmelick:
We would like to send and receive large text responses over websockets from our vertx server.  Since the message may be too large for the default websocket frame, the vertx library should split it into multiple frames.

[io.vertx.core.http.WebSocketBase](http://vertx.io/docs/apidocs/io/vertx/core/http/WebSocketBase.html) already supports sending large binary messages with the writeBinaryMessage function, so I would like to add a new function writeTextMessage.

This would be required to resolve issue 521 in vertx-web (https://github.com/vert-x3/vertx-web/issues/521).  In the pull request (https://github.com/vert-x3/vertx-web/pull/520) for that issue, it was identified that making some changes to WebSocketBase in vertx-core would be the best way forward.

This is similar to Issues #932 and #964.

vietj:
contribution done

PR #1808, opened 2017-02-09
Issue #1807 - Add support for websocket text framing.

rmelick:
Following the guidance from my pull request against vertx-web (https://github.com/vert-x3/vertx-web/pull/520), I have created this pull request against vertx-core.

I also filed issue #1807 detailing the desired behavior: vertx-core can break large text messages up into websocket frames when sending, and re-assemble multiple frames into a single frame when receiving.

There are a few open questions:

1. In WebSocketImplBase, I am creating buffers to store the frames as I re-assemble them.  These should probably have a maximum size which is configurable.  I am looking for suggestions about the best way to do that.
2. In WebsocketTest, some of the tests I added fail.  Everything that uses Unicode for the text message fails, and also the protocol version 00 fails with plain ascii.  It would be great if someone with more websocket experience could look at the errors.  The unicode tests fail with `io.netty.handler.codec.TooLongFrameException`.  To me, it looks like the frames that are given to the client have an invalid size header, which means that netty cannot figure out how to decode them.

Signed-off-by: rmelick <rmelick@cs.hmc.edu>

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpClientOptions.java
+++ b/src/main/java/io/vertx/core/http/HttpClientOptions.java
@@ -84,6 +84,12 @@ public class HttpClientOptions extends ClientOptionsBase {
   public static final int DEFAULT_MAX_WEBSOCKET_FRAME_SIZE = 65536;
 
   /**
+   * The default value for maximum websocket messages (could be assembled from multiple frames) is 4 full frames
+   * worth of data
+   */
+  public static final int DEFAULT_MAX_WEBSOCKET_MESSAGE_SIZE = 65536 * 4;
+
+  /**
    * The default value for host name = "localhost"
    */
   public static final String DEFAULT_DEFAULT_HOST = "localhost";
@@ -149,6 +155,7 @@ public class HttpClientOptions extends ClientOptionsBase {
 
   private boolean tryUseCompression;
   private int maxWebsocketFrameSize;
+  private int maxWebsocketMessageSize;
   private String defaultHost;
   private int defaultPort;
   private HttpVersion protocolVersion;
@@ -187,6 +194,7 @@ public class HttpClientOptions extends ClientOptionsBase {
     this.http2ConnectionWindowSize = other.http2ConnectionWindowSize;
     this.tryUseCompression = other.isTryUseCompression();
     this.maxWebsocketFrameSize = other.maxWebsocketFrameSize;
+    this.maxWebsocketMessageSize = other.maxWebsocketMessageSize;
     this.defaultHost = other.defaultHost;
     this.defaultPort = other.defaultPort;
     this.protocolVersion = other.protocolVersion;
@@ -234,6 +242,7 @@ public class HttpClientOptions extends ClientOptionsBase {
     http2ConnectionWindowSize = DEFAULT_HTTP2_CONNECTION_WINDOW_SIZE;
     tryUseCompression = DEFAULT_TRY_USE_COMPRESSION;
     maxWebsocketFrameSize = DEFAULT_MAX_WEBSOCKET_FRAME_SIZE;
+    maxWebsocketMessageSize = DEFAULT_MAX_WEBSOCKET_MESSAGE_SIZE;
     defaultHost = DEFAULT_DEFAULT_HOST;
     defaultPort = DEFAULT_DEFAULT_PORT;
     protocolVersion = DEFAULT_PROTOCOL_VERSION;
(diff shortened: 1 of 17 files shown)


vietj:
something on top of my mind with unicode chars : string.length() != string.getByte().length

vietj:
so we may have `slice.getBytes().length > maxWebSocketFrameSize`

vietj:
for the maximum size we should have that in HTTP options : maxWebsocketMessageSize

rmelick:
@vietj I fixed the utf-8 tests by slicing the bytes instead of the string.  All of the tests now pass.  Do you see anything missing?

vietj:
@rmelick I'm going to review it now

vietj:
@rmelick I think we are missing now documentation in the package-info.java of the `io.vertx.core.http` package, cf the _Writing messages to WebSockets_ section.

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
```diff
@@ -107,29 +118,68 @@ public SocketAddress remoteAddress() {
...
-      isFinal = true;
-    } else {
-      isFinal = false;
+  /**
+   * Splits the provided buffer into chunks of byte lengths not exceeding the maximum web socket frame size.
+   *
+   * @param buffer the buffer to be split. Is assumed to not be null
+   * @return list of chunks
+   */
+  private List<Buffer> chunkMessage(Buffer buffer) {
```
vietj: I would prefer you keep the original `writePartialMessage` as it does not involve the creation of a list and instead you pass a boolean that says wether it is binary or text frame that should be written.

rmelick: Ok.  I thought this was a good chance to get rid of the recursive code, but not creating the list also makes sense.  I updated the pr.

vietj:
I think what we are missing is a max message size in HttpClientOptions and HttpServerOptions that give the max size for the message size when aggregating a full message.

And a test for this (i.e a sender that sends a message taht exhaust the buffer).

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -94,6 +94,11 @@
   public static final int DEFAULT_MAX_WEBSOCKET_FRAME_SIZE = 65536;
 
   /**
+   * The default value for maximum websocket messages (could be assembled from multiple frames) = 655360 bytes
+   */
+  public static final int DEFAULT_MAX_WEBSOCKET_MESSAGE_SIZE = 655360;
```
vietj: why this particular value ?

rmelick: I figured approximately 10 partial frames was a reasonable default.  No particular reason.

vietj: I'm wondering if we should not a lower value, what are the usual cases for this feature ?

rmelick: The particular use case we have is a large json message that is usually 2-3 partial frames.  I'm fine with whatever default you suggest (even having it set to the size of a single message)

vietj:
4 frames ?

src/test/java/io/vertx/test/core/WebsocketTest.java
```diff
@@ -1002,19 +1002,19 @@ public void testFragmentedTextMessage2Hybi13() {
 
   @Test
   public void testFragmentedUnicodeTextMessage2Hybi07() {
-    String messageToSend = TestUtils.randomUnicodeString(65536 + 65536 + 256);
+    String messageToSend = TestUtils.randomUnicodeString(65536 + 256);
```
vietj: why don't we test more than this size, at least 3 frames of length ?

rmelick: Since these are utf8 strings, they might already be 3 frames long.  I added a new test that uses the maximum message size though.

src/test/java/io/vertx/test/core/WebsocketTest.java
```diff
@@ -1020,7 +1026,7 @@ public void testFragmentedUnicodeTextMessage2Hybi13() {
 
   @Test
   public void testTooLargeMessage() {
-    String messageToSend = TestUtils.randomAlphaString(HttpClientOptions.DEFAULT_MAX_WEBSOCKET_MESSAGE_SIZE * 2);
+    String messageToSend = TestUtils.randomAlphaString(HttpClientOptions.DEFAULT_MAX_WEBSOCKET_MESSAGE_SIZE + 1);
```
vietj: :-)

vietj:
looks good to me!

vietj:
thanks for the contribution

src/test/java/io/vertx/test/core/WebsocketTest.java
vietj: it's fine as long as you use the ASCII range like you did with alpha char

rmelick:
Thanks for all your help @vietj !  What is the merge process like at this point?  I will need to update my PR against vertx-web (https://github.com/vert-x3/vertx-web/pull/520) once this is merged

vietj:
I'll update it asap

vietj:
I'll merge it asap

vietj:
there are conflicts with Vert.x upstream, you should remerge vertx/master in your branch before I can merge

rmelick:
I think I will squash all my commits down into a fresh branch based off of master

rmelick:
Ok, I think the conflicts should be resolved now

vietj:
last comment : in the handlers for receving frames, I would try to make the textHandler and binaryHandler fit as normal handlers instead of having two kinds of handlers, pretty much like we do have for bodyHandler(Handler<Buffer>) that setup an HttpServerRequest handler and endHandler.

if you don't want to do it (I understand as it is feature complete), I would try myself.

rmelick:
I think I have an idea of what you mean, but I think doing it as a separate feature would be a good idea

vietj:
alright, that was an afterthoughts

PR #1809, opened 2017-02-12
Verify pool size against the correct parameter

guss77:
Seems like a simple enough mistake.

Do you guys want a unit test with that?


guss77:
OK, Github editing doesn't play nice with Eclipse CLA checker. I'm going to trash this and try again.

Issue #1815, opened 2017-02-15
TrustManagerFactorySpi not initiated before call


domhauton:
When adding TrustOptions with a TrustManagerFactory, the factory is never instantiated by Vert.x despite the init method present.

Relevant forum thread here. Information here: https://groups.google.com/forum/#!topic/vertx/V25r0zc_wPc

domhauton:
#1816 Pull Request created. Tests added but issue is not fixed.

PR #1818, opened 2017-02-15
#1815 - Created tests for trustManagerFactoryFix

domhauton:
Creating tests to ensure engineInit called before engineGetTrustManagers

Signed-off-by: domhauton@gmail.com
Signed-off-by: Dominic H <domhauton@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/SSLHelper.java
+++ b/src/main/java/io/vertx/core/net/impl/SSLHelper.java
@@ -257,7 +257,6 @@ public class SSLHelper {
         }
       }
       if (trustMgrFactory != null) {
-        // FIXME Should init before here.
         builder.trustManager(trustMgrFactory);
       }
       if (cipherSuites != null && cipherSuites.size() > 0) {
--- a/src/main/java/io/vertx/core/net/impl/SSLHelper.java
+++ b/src/main/java/io/vertx/core/net/impl/SSLHelper.java
@@ -18,7 +18,6 @@ package io.vertx.core.net.impl;
 
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.handler.ssl.*;
-import io.netty.handler.ssl.util.SimpleTrustManagerFactory;
 import io.vertx.core.VertxException;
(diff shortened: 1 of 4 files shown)


vietj:
@domhauton so looking at it, the TrustOptions should itself initialize the `TrustManagerFactory` in the `getTrustManagerFactory` method. That's what the other implementation do. For instance `KeyStoreHelper#getTrustMgrFactory`:

```
  public TrustManagerFactory getTrustMgrFactory(VertxInternal vertx) throws Exception {
    TrustManagerFactory fact = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
    KeyStore ts = loadStore(vertx);
    fact.init(ts);
    return fact;
  }
```

They should do it, because the TrustOptions implementation is supposed to have access to the `KeyStore` and can do it. The `SSLHelper` cannot.

I will update the documentation to mention that the returned factory is assumed to be initialized.

domhauton:
I have updated the test to reflect the change - However, I do think it would be useful if you could setKeyStore in TrustOptions to make it obvious to users. If it's not set then it should load the default keystore. Thoughts?

vietj:
@domhauton that would create issues with other pre-initiliazed options like the one I mentionned.

However I've added a new tests that checks that TrustOptions can be used to validate this SPI : https://github.com/eclipse/vert.x/commit/3034cfdb10735acc48281209d79383733cabf7f5

so I think your PR should not be necessary anymore, as the test added in HttpTLSTest does the same but with a full TLS usage.

vietj:
i.e https://github.com/eclipse/vert.x/commit/3034cfdb10735acc48281209d79383733cabf7f5#diff-607c245648ec2e1e7226237a0ef45113R365

domhauton:
@vietj Any thoughts on this?

This solves the issue neatly for anyone that wants to add their own Trust Managers.

vietj:
@domhauton I prefer users come up with their own impl for TrustOptions as it's quite trivial to implement

domhauton:
No need for basic TrustManager Implementation

PR #1819, opened 2017-02-16
Patch/thread max execution time setting

### LINKED ISSUE #1812: BlockedThreadChecker treats max execution time as if it is in nanos, though Java API docs claim it to be in millis
Looking at the JavaDocs for [`Vertx.createSharedWorkerExecutor`](http://vertx.io/docs/apidocs/io/vertx/rxjava/core/Vertx.html#createSharedWorkerExecutor-java.lang.String-int-long-) , it appears that the max execution time for a task is measured in milliseconds, but [`BlockedThreadChecker`](https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/impl/BlockedThreadChecker.java#L50) compares this value to `VertxThread.startTime()` that returns a value in nanoseconds, and indeed, `BlockedThreadChecker` divides that value by `1000000` to get at a milliseconds value.

All the other entities in the chain from `Vertx.createSharedWorkerExecutor` to `BlockedThreadChecker` don't document the meaning of the max execution time value, so I'm not sure which is correct, and I've also failed to locate where that value is being acted upon other than to issue warnings to the logs.

guss77:
Fix #1812 by having src/main/java/io/vertx/core/impl/BlockedThreadChecker.java be more consistent in what timing values mean.

### DIFF
--- a/src/main/java/io/vertx/core/impl/BlockedThreadChecker.java
+++ b/src/main/java/io/vertx/core/impl/BlockedThreadChecker.java
@@ -45,10 +45,10 @@ public class BlockedThreadChecker {
           long now = System.nanoTime();
           for (VertxThread thread : threads.keySet()) {
             long execStart = thread.startTime();
-            long dur = now - execStart;
+            long dur = (now - execStart) / 1000000;
             final long timeLimit = thread.getMaxExecTime();
             if (execStart != 0 && dur > timeLimit) {
-              final String message = "Thread " + thread + " has been blocked for " + (dur / 1000000) + " ms, time limit is " + (timeLimit / 1000000);
+              final String message = "Thread " + thread + " has been blocked for " + dur + " ms, time limit is " + timeLimit + " ms";
               if (dur <= warningExceptionTime) {
                 log.warn(message);
               } else {
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -502,7 +502,7 @@ public interface Vertx extends Measured {
    *
    * @param name the name of the worker executor
    * @param poolSize the size of the pool
-   * @param maxExecuteTime the value of max worker execute time, in ns
+   * @param maxExecuteTime the value of max worker execute time, in ms
    * @return the named worker executor
    */
   WorkerExecutor createSharedWorkerExecutor(String name, int poolSize, long maxExecuteTime);


src/main/java/io/vertx/core/impl/BlockedThreadChecker.java
```diff
@@ -45,10 +45,10 @@ public void run() {
           long now = System.nanoTime();
           for (VertxThread thread : threads.keySet()) {
             long execStart = thread.startTime();
-            long dur = now - execStart;
+            long dur = (now - execStart) / 1000000;
```
vietj: the / 100000 should be done only when print occur not, I changed ns -> ms already.

guss77:
Why do the division only when printing? There's just a single op here - I don't believe it's a performance issue.

Please see my comments on #1812 - this is an alternative fix suggestion where we leave the timeout value in ms (so it's similar to other Java timing APIs that mostly take ms values), and instead, in BlockedThreadChecker's we do all calculations in ms (there's basically just one place where we need to convert).

src/main/java/io/vertx/core/impl/BlockedThreadChecker.java
vietj: time limit is in nanos, so if you convert to ms the expression `dur > timeLimit` does not have the same meaning.

vietj: Ah, so I changed ms -> ns because that's the value we use and that is used already in `VertxOptions#maxEventLoopExecuteTime` that is in nanos.

guss77: 1. please explain why do the ns->ms conversion only on print and not earlier?
2. my commit to `Vertx.java` in this PR reverts the javadoc to declare `timeLimit` in ms - which I think makes more sense when comparing to other timing APIs in Java (see the relevant commit comment), so then the change in line 48 causes `dur` to be in ms so it can be adequately compared to `timeLimit` (the original bug #1812 was about this comparison being wrong, and as I stated in the comments there, I believe that doing this comparison in ms makes more sense).
3. ah, I see. I wasn't aware of that. That makes this whole discussion moot, don't you think? I'll close this PR.

guss77:
Closing, as the changes here conflict with the definition of max execution time in `VertxOptions`

PR #1823, opened 2017-02-17
fixes mapper signature in Future.recover()

ruslansennov:
see #1759


vietj:
I should have seen it too!!!

PR #1825, opened 2017-02-17
Future fail/complete returns boolean instead of throwing an IllegalSt…

### LINKED ISSUE #1824: Future fail/complete returns boolean instead of throwing an IllegalStateException
## Motivation

Future completion methods throw an `IllegalStateException` when the future is already completed. Vert.x itself does not rely on this behavior.

When a future might be completed several times (for instance a close or a timeout), Vert.x takes the necessary steps with the idiom, in order to propagate the event correctly.

```
if (!f.isComplete()) {
  f.complete();
}
```

This change allows to simplify the existing code and use complete the future directly and keep the ability to remain aware of whether the Future was actually completed by the method call or not.

## Changes

Future completion returns a boolean indicating whether the Future was completed or not by the method call instead of throwing an exception.

vietj:
…ateException - fixes #1824

please review

### DIFF
--- a/src/main/java/io/vertx/core/CompositeFuture.java
+++ b/src/main/java/io/vertx/core/CompositeFuture.java
@@ -186,6 +186,19 @@ public interface CompositeFuture extends Future<CompositeFuture> {
   CompositeFuture setHandler(Handler<AsyncResult<CompositeFuture>> handler);
 
   /**
+   * @return false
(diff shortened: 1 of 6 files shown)


cescoffier:
👍 

But you have to rebase first.

vietj:
oh yeah

PR #1827, opened 2017-02-19
Future<T> extends Handler<AsyncResult<T>> - fixes #1826

### LINKED ISSUE #1826: Future<T> extends Handler<AsyncResult<T>>
Motivation:

It is actually possible to turn a `Future<T>` into an `Handler<AsyncResult<T>>` using `Future#completer()` method. This is actually the polyglot method to achieve it as `AsyncResult<T>` might not be mapped correctly sometimes (like in JS or Ruby).

The implementations of `Future` actually implements `Handler<AsyncResult<T>>` and return `this` in the `completer()` method.

When using java-ish languages like Java, Kotlin or Groovy, it should be possible to use directly the represented `Future` type directly as an `Handler<AsyncResult>`.

Resolution:

This patch makes `Future<T>` extend `Handler<AsyncResult<T>>` (in addition of `AsyncResult<T>`) and the `handle(AsyncResult<T>)` is annotated with `@GenIgnore`.

vietj:
Motivation:

It is actually possible to turn a `Future<T>` into an `Handler<AsyncResult<T>>` using `Future#completer()` method. This is actually the polyglot method to achieve it as `AsyncResult<T>` might not be mapped correctly sometimes (like in JS or Ruby).

The implementations of `Future` actually implements `Handler<AsyncResult<T>>` and return `this` in the `completer()` method.

When using java-ish languages like Java, Kotlin or Groovy, it should be possible to use directly the represented `Future` type directly as an `Handler<AsyncResult>`.

Resolution:

This patch makes `Future<T>` extend `Handler<AsyncResult<T>>` (in addition of `AsyncResult<T>`) and the `handle(AsyncResult<T>)` is annotated with `@GenIgnore`.

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -31,7 +31,7 @@ import java.util.function.Function;
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
 @VertxGen
-public interface Future<T> extends AsyncResult<T> {
+public interface Future<T> extends AsyncResult<T>, Handler<AsyncResult<T>> {
 
   /**
    * Create a future that hasn't completed yet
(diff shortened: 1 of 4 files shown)


vietj:
this method relies on this commit https://github.com/vert-x3/vertx-codegen/tree/tolerate-extend-handler-async-result

cescoffier:
👍

PR #1829, opened 2017-02-20
Websocket cleanup

vietj:
Motivation:

generate cleanup in websocket impl hierarchy, a recent PR added support for message aggregation and the introduced code could be factored out as an internal frame handler of the `WebSocketServerBase` in order to increase readability. Lot of duplicated code was found in `WebSocketServerBase` sub-interfaces and could be factored out in the `WebSocketServerBase ` interface as well.

Result:
- message aggregation is now handled in an internal frame handler of `WebSocketServerBase `
- common code has been factored in `WebSocketServerBase ` instead of being duplicated in `WebSocketImpl` and `ServerWebSocketImpl`

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
+++ b/src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
@@ -26,8 +26,6 @@ import io.vertx.core.http.WebSocketFrame;
 import io.vertx.core.http.impl.ws.WebSocketFrameImpl;
 import io.vertx.core.http.impl.ws.WebSocketFrameInternal;
 import io.vertx.core.impl.VertxInternal;
-import io.vertx.core.logging.Logger;
-import io.vertx.core.logging.LoggerFactory;
 import io.vertx.core.net.SocketAddress;
 import io.vertx.core.net.impl.ConnectionBase;
 
@@ -40,9 +38,9 @@ import java.util.UUID;
  * we benefit from biased locking which makes the overhead of synchronized near zero.
  *
  * @author <a href="http://tfox.org">Tim Fox</a>
+ * @param <S> self return type
  */
-public abstract class WebSocketImplBase implements WebSocketBase {
-  private static final Logger log = LoggerFactory.getLogger(WebSocketImplBase.class);
+public abstract class WebSocketImplBase<S extends WebSocketBase> implements WebSocketBase {
 
   private final boolean supportsContinuation;
   private final String textHandlerID;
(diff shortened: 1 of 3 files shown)


vietj:
@rmelick can you review this PR ?

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
```diff
@@ -195,73 +219,116 @@ void handleFrame(WebSocketFrameInternal frame) {
...
+        case BINARY:
+          handleBinaryFrame(frame);
+          break;
+        case CONTINUATION:
+          if (textMessageBuffer != null && textMessageBuffer.length() > 0) {
+            handleTextFrame(frame);
+          } else if (binaryMessageBuffer != null && binaryMessageBuffer.length() > 0) {
+            handleBinaryFrame(frame);
+          }
+          break;
```
rmelick: Do you think we want to have some sort of warning or exception if a CONTINUATION frame arrives without a TEXT or BINARY coming first?

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
```diff
@@ -195,73 +219,116 @@ void handleFrame(WebSocketFrameInternal frame) {
...
+        textMessageBuffer = frameBuffer;
+      } else {
+        textMessageBuffer.appendBuffer(frameBuffer);
+      }
+      if (textMessageBuffer.length() > maxWebSocketMessageSize) {
+        int len = textMessageBuffer.length() - frameBuffer.length();
+        textMessageBuffer = null;
+        String msg = "Cannot process text frame of size " + frameBuffer.length() + ", it would cause message buffer (size " +
+            len + ") to overflow max message size of " + maxWebSocketMessageSize;
+        handleException(new IllegalStateException(msg));
```
rmelick: I think we should reset the buffer in the error case here.  If we don't, the internal buffer will keep growing forever if they continue to pass in continuation frames.  I thought I had added a test to WebsocketTest that would now fail (because we wouldn't be able to recover from a message that was too large)

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
```diff
@@ -195,73 +219,116 @@ void handleFrame(WebSocketFrameInternal frame) {
...
       }
+    }
 
-      if (binaryMessageHandler != null && (frame.type() == FrameType.BINARY
-        || (processingPartialBinaryMessage && frame.type() == FrameType.CONTINUATION))) {
-        handleBinaryFrame(frame);
+    private void handleTextFrame(WebSocketFrameInternal frame) {
+      Buffer frameBuffer = Buffer.buffer(frame.getBinaryData());
+      if (textMessageBuffer == null) {
+        textMessageBuffer = frameBuffer;
```
rmelick: I think we actually want to reset this buffer every time we get a TEXT frame, not just if it is currently empty.

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
vietj: no it's fine

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
vietj: so it means that we can have a case where we get a continuation frame that is not final ?

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
vietj: there is `textMessageBuffer = null` statement

rmelick: Thanks.  I don't know why I completely missed that :).

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
rmelick: Any message that is split into at least 3 frames will have a continuation frame that is not final.

I see now that your change checks if buffer != null instead of having a separate processingPartialMessage flag.  That made it more difficult for me to see that you were correctly dropping continuation frames after the failure.  Now that I have looked through it again, I think everything is ok.

vietj: great

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
vietj: great, the idea is to reuse the first buffer and avoid an unnecessary copy

PR #1837, opened 2017-02-24
Optimize removeWorker in VertxEventLoopGroup

yunyu:
This change saves iterating through the workers collection twice, and also optimizes the count comparison to use a pre-decrement.


yunyul:
I just signed the ECA. Is there anything I need to do to force the checks to rerun?

vietj:
not really needed there is no performance issue here

PR #1838, opened 2017-02-25
Fixed issue with overriding handlers

Crokoking:
A bug in binaryMessageHandler() caused it to overwrite an already existing Frame-Aggregator.


Crokoking:
resubmitted as https://github.com/eclipse/vert.x/pull/1839

PR #1839, opened 2017-02-25
Fixed issue with overriding handlers

Crokoking:
A bug in binaryMessageHandler() caused it to overwrite an already existing Frame-Aggregator.


vietj:
thanks

PR #1847, opened 2017-02-28
Helper for self-signed certificates

jponge:
This is useful for testing and development purposes.

### DIFF
--- a/src/main/java/examples/NetExamples.java
+++ b/src/main/java/examples/NetExamples.java
@@ -573,7 +573,7 @@ public class NetExamples {
     NetClient client = vertx.createNetClient(clientOptions);
     client.connect(1234, "localhost", ar -> {
       if (ar.succeeded()) {
-        ar.result().handler(System.out::println);
+        ar.result().handler(buffer -> System.out.println(buffer));
       } else {
         System.err.println("Woops: " + ar.cause().getMessage());
       }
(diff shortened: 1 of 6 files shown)


src/main/java/examples/NetExamples.java
```diff
@@ -550,4 +552,31 @@ public void example47(Vertx vertx) {
...
+
+    NetClientOptions clientOptions = new NetClientOptions()
+      .setSsl(true)
+      .setKeyCertOptions(certificate.keyCertOption())
+      .setTrustOptions(certificate.trustOptions());
+
+    NetClient client = vertx.createNetClient(clientOptions);
+    client.connect(1234, "localhost", ar -> {
+      if (ar.succeeded()) {
+        ar.result().handler(System.out::println);
```
vietj: should not use a method reference and instead a lambda because of the code generation in other languages for the doc

vietj:
I think we need also an example with HTTP with client trustAll=true as this is a very common case

jponge:
@vietj done a pass

src/main/java/io/vertx/core/net/impl/SelfSignedCertificateImpl.java
```diff
@@ -0,0 +1,46 @@
...
+public class SelfSignedCertificateImpl implements SelfSignedCertificate {
+
+  private final io.netty.handler.ssl.util.SelfSignedCertificate certificate;
+
+  public SelfSignedCertificateImpl() throws CertificateException {
+    certificate = new io.netty.handler.ssl.util.SelfSignedCertificate();
+  }
+
+  @Override
+  public PemKeyCertOptions keyCertOption() {
```
vietj: typo -> keyCertOptions

jponge: Done

jponge:
@vietj Good for another (last?) review.

PR #1855, opened 2017-03-01
Hostname resolver improvements

vietj:
Please review

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -65,6 +65,10 @@ Set the query timeout in milliseconds, i.e the amount of time after a query is c
 +++
 Set the DNS queries <i>Recursion Desired</i> flag value.
 +++
+|[[roundRobin]]`roundRobin`|`Boolean`|
++++
+Set to <code>true</code> to enable round-robin selection of the dns server to use.
++++
 |[[searchDomains]]`searchDomains`|`Array of String`|
 +++
 Set the lists of DNS search domains.
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -65,9 +65,10 @@ Set the query timeout in milliseconds, i.e the amount of time after a query is c
 +++
 Set the DNS queries <i>Recursion Desired</i> flag value.
 +++
-|[[roundRobin]]`roundRobin`|`Boolean`|
+|[[rotateServers]]`rotateServers`|`Boolean`|
 +++
-Set to <code>true</code> to enable round-robin selection of the dns server to use.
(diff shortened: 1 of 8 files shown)


testn:
I think there are different cases for RoundRobin. RoundRobinDnsAddressResolverGroup actually returns different ip addresses in a round robin fashion if multiple IN A record exists. AFAIK, Your implementation seems to pick a different DNS servers in a round robin fashion instead.

vietj:
provide round robin on the server and what I've done is round robin on the dns server itself.

that being said what I've provided is still useful, right ?

vietj:
so this would be rather configured in `ClientOptionsBase` ?

vietj:
and it seems to use random and non strict round robin.

I think in our case it would be something configured in the client (HttpClient / NetClient) that is passed when the address is resolved using : VertxInternal#resolveAddress

so it gets the list the first time when the option is active in the client and then uses the list to perform round robin

testn:
I guess but I think the name should probably follow the netty convention. The feature you implemented here should probably be named something like "shuffleDns" to separate round robin DNS from DNS shuffling?

vietj:
the name is DNS shuffling ? do you have pointers about it ?

testn:
Just that it uses DnsAddresses.shuffle

vietj:
please open a more specific issue for shuffling

testn:
BTW, the feature option is called "rotate" in  /etc/resolv.conf https://linux.die.net/man/5/resolv.conf

vietj:
ok so for this config I should rename it to rotate and not use round-robin name

vietj:
and I believe that round robin selection of the name should rather be *per client*

vietj:
I've renamed the feature to `rotateServers`, I'll open another issue for round robin selection of the client

vietj:
thanks for the insights @testn that helps :-)

alexlehm:
I think the rotation of the IP address returned should be per request, but I have to look into the feature to check how it is done in the C code in Linux

vietj:
@alexlehm good idea to look at it.

vietj:
I'm wondering so if that makes sense to have this feature enabled for when keep-alive is active in HttpClient

vietj:
@alexlehm I believe rotation only applies when doing a lookup to the DNS server and is not related to the nature of the client.

The goal is not to load balance the client requests but to spread the load on the DNS servers.

We have talked of another feature which is to make round-robin of multiple answers to spread the load of the target servers.

tsegismont:
I'm +1 for merging. On Linux at least, the JVM relies on the system configuration. So when you setup `resolv.conf` to enable rotation, all listed DNS servers are used.
It's certainly not something everyone uses, but that's good to have.

PR #1858, opened 2017-03-01
Default redirect handler: do not fail on URIs without explicit port

tsegismont:
Fixes java.lang.IllegalArgumentException: port out of range:-1


vietj:
can you explain how to reproduce it so we can add tests ?

vietj:
I see, hard to have a test for it as it uses port 80 or 443 ...

vietj:
I'm going to extract the code in a static method and test this method instead

tsegismont:
Exactly

2017-03-01 23:17 GMT+01:00 Julien Viet <notifications@github.com>:

PR #1860, opened 2017-03-05
Load the rotate option from /etc/resolv.conf on Linux

tsegismont:
Fixes the AddressResolverOptions copy constructor and adds the ability to override options on a per-process basis with the RES_OPTIONS environment var.


tsegismont:
@vietj I had this in the pipe before you pushed f7c7a87

vietj:
that was blocking the release

tsegismont:
I understand. Just wanted to share in case we want this in the future.

What about the options copy constructor fix? Not a blocker I guess.

2017-03-06 0:04 GMT+01:00 Julien Viet <notifications@github.com>:

PR #1863, opened 2017-03-08
Handle HTTP/1.1 decode errors - fixes 1862

vietj:


src/main/java/io/vertx/core/http/impl/ServerConnection.java
```diff
@@ -397,46 +397,49 @@ protected ChannelFuture sendFile(RandomAccessFile file, long offset, long length
...
+          if (!paused) {
+            handleEnd();
+          } else {
+            // Requeue
+            pending.add(LastHttpContent.EMPTY_LAST_CONTENT);
+          }
         }
-      }
-    } else if (msg instanceof WebSocketFrameInternal) {
+      }    } else if (msg instanceof WebSocketFrameInternal) {
```
alexlehm: here a newline is missing between the closing braces

src/main/java/io/vertx/core/http/impl/ServerConnection.java
```diff
@@ -397,46 +397,49 @@ protected ChannelFuture sendFile(RandomAccessFile file, long offset, long length
...
+    if (msg instanceof HttpObject) {
+      HttpObject obj = (HttpObject) msg;
+      DecoderResult result = obj.decoderResult();
       if (result.isFailure()) {
         Throwable cause = result.cause();
         if (cause instanceof TooLongFrameException) {
           String causeMsg = cause.getMessage();
           HttpResponseStatus status = causeMsg.startsWith("An HTTP line is larger than") ? HttpResponseStatus.REQUEST_URI_TOO_LONG : HttpResponseStatus.BAD_REQUEST;
-          DefaultFullHttpResponse resp = new DefaultFullHttpResponse(request.protocolVersion(), status);
+          DefaultFullHttpResponse resp = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, status);
```
tsegismont: Ideally we should keep of track the protocol version number instead of forcing 1.1 here.

vietj:
I've addressed the version number

PR #1865, opened 2017-03-08
WorkerExecutor callback should run on caller context

tsegismont:
Currently it runs on the context which created the WorkerExecutor


tsegismont:
@vietj can you please review along with the sister PR vert-x3/vertx-rx#79? Thanks

src/main/java/io/vertx/core/impl/WorkerExecutorImpl.java
```diff
@@ -56,8 +57,8 @@ public boolean isMetricsEnabled() {
   }
 
   @Override
-  public ContextImpl getContext() {
-    return context;
+  public Vertx owner() {
```
vietj: rather name it "vertx"

PR #1877, opened 2017-03-15
Add json path support to JsonObject

Jotschi:
## Motivation

Loading specific nested values from a Vert.x JsonObject is very cumbersome because each level of nesting would need to be accessed individually. I propose adding JSON Path support in order to make accessing nested values easier.

## Example Usage

```java
jsonObject.put("level1", new JsonObject().put("level2", new JsonObject().put("value1", "abc")));
String value = jsonObject.getByPath("$.level1.level2.value1");
```

Used library is well maintained: https://github.com/jayway/JsonPath

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,6 +134,11 @@
       <version>${jackson.version}</version>
     </dependency>
     <dependency>
+      <groupId>com.jayway.jsonpath</groupId>
+      <artifactId>json-path</artifactId>
+      <version>2.2.0</version>
+    </dependency>
+    <dependency>
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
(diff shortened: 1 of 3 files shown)


Jotschi:
I have signed the ECA via `cla@jotschi.de` - Do I need to change the commit email? Is there a way to list multiple email addresses?

vietj:
Hi, I prefer to avoid this feature as it introduces a new dependency in Vert.x Core . We want to keep the number of dependency to a minimum.

Jotschi:
@vietj Yes - That makes sense. I may give it another try and implement path resolving without the need of an additional library. I'll close this PR

vietj:
thanks for understanding!

PR #1878, opened 2017-03-15
Removed deprecated Netty code

tsegismont:
Some tests had to be updated because of:
- HTTP with Strings over a TCPServer
- Case sensitive header map

### DIFF
--- a/src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
+++ b/src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
@@ -15,11 +15,11 @@
  */
 package io.vertx.core.datagram.impl;
 
-import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.FixedRecvByteBufAllocator;
+import io.netty.channel.MaxMessagesRecvByteBufAllocator;
 import io.netty.channel.socket.DatagramChannel;
 import io.netty.channel.socket.DatagramPacket;
 import io.netty.channel.socket.InternetProtocolFamily;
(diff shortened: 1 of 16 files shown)


vietj:
as this has an impact on performance, can you make sure that the new changes are equivalent to the existing usage of ascii strings ?

tsegismont:
1/ I haven't checked if there's a perf impact, but I don't believe there
would be: the deprecated methods simply call the new methods.

2/ We are using AsciiString already, as
io.netty.handler.codec.http.HttpHeaders.newEntity
actually creates an AsciiString.
The difference is that with this PR, we don't create our own constants but
reuse those of Netty, which are defined with lower case letters.
The potential impact is for code doing manual http parsing, like some of
our tests do (see the changes).

2017-04-10 9:37 GMT+02:00 Julien Viet <notifications@github.com>:

vietj:
I would rather keep the existing behavior concerning the constant case and not change it

tsegismont:
Let me elaborate on point 2/

We already use AsciiString, created with HttpHeaders.newEntity. The
parameter provided to newEntity is a deprecated Netty constant, defined
with a mix of upper and lower case letters (e.g. "Content-Type")
With this PR, we also use Netty constants, but they have changed the values
to all lower case (e.g. "content-type")

If we want to keep the constant case, we need to either go back to the
deprecated constants or create our own. The former makes no sense since
it's the purpose of the PR to get rid of the deprecated code.
I'm not opposed to the latter but why do this since the spec makes it clear
the HTTP header names are case-insensitive?


2017-04-11 10:07 GMT+02:00 Julien Viet <notifications@github.com>:

tsegismont:
@vietj have you seen the previous comment?

vietj:
@tsegismont my point is that I'm doing performance test at the moment, and I would like to validate it by doing actual benchmarks

tsegismont:
@vietj ping

vietj:
let me test it :-)

tsegismont:
@vietj rebased on master (had conflicts since your perf PR was merged)

vietj:
I would like to benchmark it before merging it to see if the change of constant has an impact (it might have)

tsegismont:
I will open another PR: rebasing on master will be harder than starting from scratch. Also, I'll split the work into 2 PRs as discussed with @vietj :

1. deprecated code replacement 
2. http headers change

We'll test 2/ in the perf lab

PR #1879, opened 2017-03-16
More tests fixes

tsegismont:
Details in individual commits

### DIFF
--- a/src/test/java/io/vertx/test/core/AsyncTestBase.java
+++ b/src/test/java/io/vertx/test/core/AsyncTestBase.java
@@ -596,11 +596,21 @@ public class AsyncTestBase {
     waitUntil(supplier, 10000);
   }
 
+  protected void waitUntilNoFail(BooleanSupplier supplier) {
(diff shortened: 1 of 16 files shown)


src/test/java/io/vertx/test/core/AsyncTestBase.java
```diff
@@ -596,19 +596,29 @@ protected void waitUntil(BooleanSupplier supplier) {
     waitUntil(supplier, 10000);
   }
 
+  protected void waitUntilNoFail(BooleanSupplier supplier) {
+    waitUntilNoFail(supplier, 10000);
+  }
+
   protected void waitUntil(BooleanSupplier supplier, long timeout) {
```
vietj: we should rename this method rather `assertWaitUntil` and make it fail the test instead. The `waitUntilNoFail` should be then just named `waitUntil`.

tsegismont:
Good idea

2017-03-17 10:36 GMT+01:00 Julien Viet <notifications@github.com>:

PR #1880, opened 2017-03-16
Exception cause lost if result already complete.

tsegismont:
It's impossible to fix the original issue without the cause. Found this while trying to debug a test failure:

```
Caught unexpected Throwable
java.lang.IllegalStateException: Result is already complete: failed
	at io.vertx.core.impl.FutureImpl.fail(FutureImpl.java:121)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$1(ContextImpl.java:277)
	at io.vertx.core.impl.TaskQueue.lambda$new$0(TaskQueue.java:60)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

### DIFF
--- a/src/main/java/io/vertx/core/impl/CompositeFutureImpl.java
+++ b/src/main/java/io/vertx/core/impl/CompositeFutureImpl.java
@@ -17,8 +17,8 @@
 package io.vertx.core.impl;
 
 import io.vertx.core.AsyncResult;
-import io.vertx.core.Future;
 import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
 import io.vertx.core.Handler;
 
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


vietj:
So how would you fix it if the result was a success ?

vietj:
the cause of the failure should not be failure to complete a future twice and it will half work because you will never get the stack trace in the case of a success

tsegismont:
I can't think of an easy/cheap way to do it.

2017-03-17 10:39 GMT+01:00 Julien Viet <notifications@github.com>:

tsegismont:
I agree in case of a success, we can't do much. But in case of failure, the
exception is already there, we just lose the information.

2017-03-17 10:40 GMT+01:00 Julien Viet <notifications@github.com>:

vietj:
yes but this exception has no relation about the timing completion of the Future, the cause might be an exception created by another thread or at another moment.

Issue #1750, opened 2016-12-22
Customization of Netty "initialBufferSize"


gmagniez:
Would it be possible to add a new configuration options in both HttpClient and HttpServer to allow to customize the initial size (128b default) of the parsing buffer of Netty HttpObjectDecoder.

I see that when handling an high load on my application with requests or responses with multiple headers, the CPU is mostly consumed by netty "AppendableCharSequence.append", which has to always grow and so has to copy char arrays.

```
Stack Trace	Sample Count	Percentage(%)
io.netty.util.internal.AppendableCharSequence.append(char)	2 160	6,472
   io.netty.handler.codec.http.HttpObjectDecoder$HeaderParser.process(byte)	2 160	6,472
      io.netty.buffer.AbstractByteBuf.forEachByteAsc0(int, int, ByteBufProcessor)	2 160	6,472
         io.netty.buffer.AbstractByteBuf.forEachByte(ByteBufProcessor)	2 160	6,472
            io.netty.handler.codec.http.HttpObjectDecoder$HeaderParser.parse(ByteBuf)	2 160	6,472
               io.netty.handler.codec.http.HttpObjectDecoder.readHeaders(ByteBuf)	2 141	6,415
                  io.netty.handler.codec.http.HttpObjectDecoder.decode(ChannelHandlerContext, ByteBuf, List)	2 141	6,415
```

Methods involved in vertx:
```HttpServerImpl.configureHttp1 -> HttpRequestDecoder```
```ConnectionManager$ChannelConnector.applyHttp1xConnectionOptions -> HttpClientCodec```

tsegismont:
Would you mind to send a pull request? Usually, such pull requests get
better chances of inclusion if they come with a performance test
demonstrating the benefit.

2016-12-22 14:26 GMT+01:00 gmagniez <notifications@github.com>:

vietj:
+1

gmagniez:
Ok, I'll try to do that but I think it will be after Christmas and New Year :-)

vietj:
@gmagniez any news ?

leogomes:
@gmagniez: Please check my PR [here](https://github.com/eclipse/vert.x/pull/1881), if you have some time. It would be also nice if you could share the original use-case you had for this and/or have a look at the test case I wrote to benchmark the change. All mentioned in the PR.

leogomes:
@vietj @tsegismont I guess you guys can close this one now that #1881 is merged.

PR #1881, opened 2017-03-16
Customization of Netty "initialBufferSize" #1750

leogomes:
Motivation:
Under load and with HTTP requests or responses with multiple headers, a
lot of CPU-time can be spent on "AppendableCharSequence.append", which
has to always grow and so has to copy char arrays.

Modifications:
Changed HttpClientOptions and HttpServerOptions to allow the user to
pass an initial buffer size for Netty’s HTTP decoder that fits her
needs when not happy with the default initial buffer size, i.e. 128
bytes. Also, fixed some typos in documentation.

Result:
Initial buffer size for HTTP decoder is customisable.

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -715,6 +715,10 @@ Set a client limit of the number concurrent streams for each HTTP/2 connection, 
 Set the idle timeout, in seconds. zero means don't timeout.
  This determines if a connection will timeout and be closed if no data is received within the timeout.
 +++
+|[[initialBufferSizeHttpDecoder]]`initialBufferSizeHttpDecoder`|`Number (int)`|
++++
+set to <code>initialBufferSizeHttpDecoder</code> the initial buffer of the HttpDecoder.
++++
 |[[initialSettings]]`initialSettings`|`link:dataobjects.html#Http2Settings[Http2Settings]`|
 +++
 Set the HTTP/2 connection settings immediately sent by to the server when the client connects.
@@ -921,10 +925,10 @@ This method allows to set the compression level to be used in http1.x/2 response
  
  While one can think that best value is always the maximum compression ratio, 
  there's a trade-off to consider: the most compressed level requires the most
- computatinal work to compress/decompress data, e.g. more dictionary lookups and loops.
+ computational work to compress/decompress data, e.g. more dictionary lookups and loops.
  
  E.g. you have it set fairly high on a high-volume website, you may experience performance degradation 
- and latency on resource serving due to CPU overload, and, however - as the comptational work is required also client side 
+ and latency on resource serving due to CPU overload, and, however - as the computational work is required also client side
  while decompressing - setting an higher compression level can result in an overall higher page load time
  especially nowadays when many clients are handled mobile devices with a low CPU profile.
  
@@ -975,6 +979,10 @@ Set the default HTTP/2 connection window size. It overrides the initial window
 +++
 Set the idle timeout, in seconds. zero means don't timeout.
  This determines if a connection will timeout and be closed if no data is received within the timeout.
++++
+|[[initialBufferSizeHttpDecoder]]`initialBufferSizeHttpDecoder`|`Number (int)`|
++++
+Set the initial buffer size for the HTTP decoder
 +++
 |[[initialSettings]]`initialSettings`|`link:dataobjects.html#Http2Settings[Http2Settings]`|
 +++
--- a/src/main/generated/io/vertx/core/http/HttpClientOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/http/HttpClientOptionsConverter.java
@@ -53,6 +53,9 @@ public class HttpClientOptionsConverter {
     if (json.getValue("http2MultiplexingLimit") instanceof Number) {
       obj.setHttp2MultiplexingLimit(((Number)json.getValue("http2MultiplexingLimit")).intValue());
     }
+    if (json.getValue("initialBufferSizeHttpDecoder") instanceof Number) {
+      obj.setInitialBufferSizeHttpDecoder(((Number)json.getValue("initialBufferSizeHttpDecoder")).intValue());
+    }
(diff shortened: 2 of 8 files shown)


leogomes:
This PR adds the new option suggested here: https://github.com/eclipse/vert.x/issues/1750

@tsegismont mentioned in the issue that performance tests would be a plus, but I haven't put them in place yet. I was thinking of modifying https://github.com/vert-x3/vertx-perf to add a JMH test there to validate performance gains brought by this PR. Is it a good strategy?

tsegismont:
I don't believe it's a good way to test your changes.
You'd better create a simple application which replies to requests and
measure the response time improvement (with Gatling or JMeter). The
injector and the app should run on separate machines.

2017-03-17 0:21 GMT+01:00 Leo Gomes <notifications@github.com>:

leogomes:
@tsegismont OK, I will put that in place. Should it be on a specific existing project or should I create a project on my side to demonstrate that and provide you guys a way to execute it?

tsegismont:
A project of your own on GH would be perfect.

2017-03-20 16:29 GMT+01:00 Leo Gomes <notifications@github.com>:

leogomes:
I have done some tests with Gatling. Results are available here: [https://github.com/leogomes/vertx-benchmark/tree/master/results](https://github.com/leogomes/vertx-benchmark/tree/master/results). I also profiled the app with JFR and generated flamegraphs, which you can find on the same folder, for the different initial buffer sizes.

While I couldn't notice any significant difference in response time in my particular tests, the flamegraphs show that HttpObjectDecoder$HeaderParser.parse() goes down from 69.49% of total CPU time to 51.22%, when you change the initial buffer from 128 to 4096.

If you guys have any suggestions for a better test case, please let me know. I've written a very basic app and a Gatling scenario that sends a request with a couple of quite big headers: 
https://github.com/leogomes/vertx-benchmark/blob/master/src/gatling/vertx/VertxInitialBufferSimulation.scala

gmagniez:
Seems really great to me.
I think you did not notice a big difference because you have only tested the HttpServer side.
My use case was an HTTP Proxy handling thousand of requests with 15+ headers with some quite large (at least 500 chars) in both ways. Client <-> (HttpServer) VertX (HttpClient) <-> Backend server
Ending with near 2K of headers

tsegismont:
@gmagniez could you try the change and give some feedback?

2017-03-24 16:12 GMT+01:00 gmagniez <notifications@github.com>:

gmagniez:
I already have patched VertX to implement this behavior  and as leogomes pointed it out, I have also noticed a lower cpu consumption due to request parsing.  Even with a bigger scale as my proxy was using both of HttpServer and HttpClient.
After that, as I remember AppendableCharSequence.append is not the top consumer when I do some sampling. It is now exceeded by the char to char writing of headers to the socket, but I did not find already a way to keep headers,  netty's AsciiString are always converted to String at some point of parsing. (I would prefer to have the possibility to only handle AsciiString's with my Proxy use case as my app is "just" copying headers in both ways). But it is another "problem"

leogomes:
In a couple of weeks, I should have time to improve my current test scenario (write a proxy-like app with the server and client part, put 15+ headers as described by @gmagniez) and run a better performance test. I will keep you posted.

leogomes:
So, my new scenario has:

- A proxy like application that forwards HTTP requests to an Echo Server. Both run on the same machine, while the Client (Gatling) is sitting on another one;
- My Gatling scenario now has 15 HTTP headers with ~500 bytes each;
- I now trigger 1000 users (i.e. 1000 channels between Gatling and the Proxy);
- I initialized my Proxy with 128 bytes for the initial buffer (i.e. default case) and 10240 bytes (so there should be no need to resize within the HttpDecoder) and compared both.

With that scenario, the difference in performance starts be more visible:

- HeaderParser.parse accounts for ~52% of CPU time, according to the flamegraph below, for the default case:
image

- It goes down to ~21% when the initial buffer is set to 10240 bytes:
image

Throughput and response time are better as well (as one would expect):

128 bytes
image

10240 bytes:
image

Source code and results are available here: https://github.com/leogomes/vertx-benchmark/tree/master/results

Gatling scenario: https://github.com/leogomes/vertx-benchmark/blob/master/src/gatling/vertx/VertxInitialBufferSimulation.scala

Cheers,
Leo.

vietj:
@leogomes the results are interesting : the latency improvement  at the 99th percentile + the throughput improvement looks good. Can you give some details of the actual workload for this benchmark (machines, network) ?

leogomes:
@vietj Here are the CPU details for the two machines:
Intel(R) Xeon(R) CPU E5-2640 v3 @ 2.60GHz
CPU(s): 32
On-line CPU(s) list: 0-31
Thread(s) per core: 2
Core(s) per socket: 8
Socket(s): 2
NUMA node(s): 2
L1d cache: 32K
L1i cache: 32K
L2 cache: 256K
L3 cache: 20480K

They're running Linux 64 bits 3.0.101-63-default

They're on the same rack and have a 1 Gbps network adapter.

The request that I'm sending is the one you can find on [the Gatling scenario](https://github.com/leogomes/vertx-benchmark/blob/master/src/gatling/vertx/VertxInitialBufferSimulation.scala).

The measurement itself is pretty short (< 3min), because I haven't had much time to dedicate to it and also because here we're not actually modifying a default value, but only exposing the underlying Netty setting, so that people with this sort of workload (heavy HTTP headers) can customise the HttpDecoder. So, I was kind of just trying to build an example that could easily reproduce the scenario. But for sure, a more serious performance test would need to be run for much longer. With that setting exposed, people out there may be able to run it with different values and their own use-cases to find their sweet spot.

Finally, the last results I had put were taken while profiling with JFR to be able to get the Flamegraphs. I have now run a new test without any profiling and have been able to [reproduce the same results](https://github.com/leogomes/vertx-benchmark/tree/master/results).

vietj:
Looks good and elaborated to me @leogomes

leogomes:
@tsegismont WDYT?

tsegismont:
@leogomes LGTM

PR #1883, opened 2017-03-19
Log target address when more than maxBufferedMessages buffered in consumer

jan-zajic:
This only "warning" is very important message about data being lost. We need to known which endpoint failing to be able to fix backpressure in code .. 

Signed-off-by: jzajic <jan.zajic@corpus.cz>

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
@@ -177,7 +177,7 @@ public class HandlerRegistration<T> implements MessageConsumer<T>, Handler<Messa
           if (discardHandler != null) {
             discardHandler.handle(message);
           } else {
-            log.warn("Discarding message as more than " + maxBufferedMessages + " buffered in paused consumer");
+            log.warn("Discarding message as more than " + maxBufferedMessages + " buffered in paused consumer. address: " + address);
           }
         }
         return;


src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
```diff
@@ -177,7 +177,7 @@ public void handle(Message<T> message) {
           if (discardHandler != null) {
```
vietj: should be renamed dropHandler

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
```diff
@@ -177,7 +177,7 @@ public void handle(Message<T> message) {
...
             discardHandler.handle(message);
           } else {
-            log.warn("Discarding message as more than " + maxBufferedMessages + " buffered in paused consumer");
+            log.warn("Discarding message as more than " + maxBufferedMessages + " buffered in paused consumer (address {0})", address);
```
vietj: we should not use message interpolation as this is not portable

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
vietj: that's a note to self :-)

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
jan-zajic: @vietj ok, i do it like at line 65 in same file

jan-zajic:
@vietj, fixed according to your review

PR #1884, opened 2017-03-20
#1586 Make JsonObject and JsonArray extend a common class

### LINKED ISSUE #1586: Make JsonObject and JsonArray extend a common class
Currently both JsonObject and JsonArray are 2 completely independent classes, but they do share some method signatures. It could be beneficial to make both classes extend some common interface (for example JsonData).
The main shared methods that could belong to this common class are:
- int size()
- String encode()
- String encodePrettily()
- boolean isEmpty();

Adding this hierarchy, could enable users send/receive JsonData objects in/to the EventBus without the need for casting if any of these methods needs to be called.

jan-zajic:
Fixes #1586. 

Adding this hierarchy and correcting deserialization in Json class enable users to deserialize both Json structures automatic correctlly. Proven by test.


vietj:
that's something we want to do indeed but with @pmlopes we wanted first to name the parent `JsonElement` instead (I think it is the name it used to be previously)

pmlopes:
@jan-zajic yes we had plans for this, we had choosen `JsonElement` as the name because it was the name used before on Vert.x 2.x series.

jan-zajic:
@vietj @pmlopes i named it JsonStructure, because json.org say: "JSON is built on two structures". Element is a part or aspect of something, for example in JSON doc element only means array member and array member may be any JSON value. If i see JsonElement i first think about it as JSON value, which means string, number, true or false or null, or an object or an array, but structure is only object or array. So in my opinion, if class name is JsonElement, it should have subclasses JsonBoolean (or JsonTrue and JsonFalse in fact), ..String, ..Number, ..Null, ..Array and ..Object, which is not so useful - i personally don't allow JSON value as top level JSON, even if it's questionably valid.

vietj:
it's not because you name something JsonElement that it implies that subclasses like JsonString have to be created :-) . However making something name JsonString and extends JsonStructure would be totally wrong.

jan-zajic:
Your right, but that's not the point. It's important to us to have these parent class, no matter what name you choose, please make a decision and let me know. Or some other votes for JsonStructure ? ;-)

pmlopes:
As you know naming things is hard 😀 you named after the Jason spec so in a way it makes sense and I see no real benefit on resurrecting the old name vs this one.

tsegismont:
If it comes from the Jason spec I recommend the "Golden Fleece" name :P

2017-03-26 11:11 GMT+02:00 Paulo Lopes <notifications@github.com>:

pmlopes:
@tsegismont that's what happens when you enable auto-correct ;) json -> jason

jan-zajic:
so, what about this? btw it's non breaking change, only expands usability.

vietj:
it looks ok, we need to review it though

src/main/java/io/vertx/core/json/JsonStructure.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * A collection of name/value pairs. In various languages, this is realized as
+ * an object, record, struct, dictionary, hash table, keyed list, or associative
+ * array. An ordered list of values. In most languages, this is realized as an
+ * array, vector, list, or sequence.
+ *
+ * @author Jan Zajic
+ *
+ */
+public abstract class JsonStructure implements ClusterSerializable {
```
vietj: constructor should be package private so there cannot be other subclass than `JsonObject` or `JsonArray`

src/main/java/io/vertx/core/json/JsonStructure.java
```diff
@@ -0,0 +1,93 @@
...
+
+  public boolean isArray() {
+    return this instanceof JsonArray;
+  }
+
+  public boolean isObject() {
+    return this instanceof JsonObject;
+  }
+
+  public JsonArray asArray() {
```
vietj: not a fan of this, if user need to cast he should do a real cast

src/main/java/io/vertx/core/json/JsonStructure.java
```diff
@@ -0,0 +1,93 @@
...
+    return (JsonArray) this;
+  }
+
+  public JsonObject asObject() {
```
vietj: same as asArray()

src/main/java/io/vertx/core/json/JsonStructure.java
pmlopes: we had this on v2

vietj:
note : this is going to break also codetrans so before merging we should have codetrans understand this type otherwise it's going to break docs

vietj:
also currently codegen does not understand this type, currently it we use it in something like:

```
@VertxGen
public interface SomeApi {
  JsonStructure getSomeJson();
}
```

it will not compile

src/main/java/io/vertx/core/json/JsonStructure.java
pmlopes: I can think of a case where I would like to extend it, for example on JDBC client, the result set is a JSON object however the column names returned are case insensitive, however since we encode to json we must get a column by the name the RDBMS returns. A extension could be a JSON that uses a case insensitive key for the underlying map for this case.

But anyway this would be a pure hack for this specific case and would only work as long as it wasn't encoded over the EB or copied.

vietj:
so its only usage at the moment is to allow some API to use `JsonObject` or `JsonArray` to have a common OO ancestor. as soon as we have this PR merged, people will ask to have VertxGen to support it.

Is it possible to support it in JavaScript in both return and param type ? i.e what is the type that will represent JsonStructure ?

so again before merging this PR, we need to decide what it means for codegen.

pmlopes:
in `JavaScript` this would be an `Object` since in `JS` `[]` is also an `Object`

src/main/java/io/vertx/core/json/JsonStructure.java
vietj: but in JDBC client wouldn't you instead extend `JsonObject` or `JsonArray` ?

vietj: typical pattern of use:

```
if (json.isArray()) {
  JsonArray array = (JsonArray) json;
} else {
  // It must be JsonObject !
  JsonObject obj = (JsonObject) json;
}
```

vietj:
so I'm fine with it, as long as it comes fully supported in codegen/codetrans otherwise it won't be usable

PR #1886, opened 2017-03-22
Add precision to the AsyncMultiMap javadoc

henri-tremblay:
* A typo
* The values are (conceptually) in a Set, not in a List
* Add what the boolean means for remove
* Detail the behavior of all the removes


henri-tremblay:
test this please

Issue #1851, opened 2017-03-01
KeyStoreHelper just support PKCS8,Not support RSA private key


jeho0815:
`public PrivateKey loadPrivateKey() throws Exception {
      if (keyValue == null) {
        throw new RuntimeException("Missing private key path");
      }
      byte[] value = loadPems(keyValue.get(), "**PRIVATE KEY**").get(0);
      KeyFactory rsaKeyFactory = KeyFactory.getInstance("RSA");
      return rsaKeyFactory.generatePrivate(new PKCS8EncodedKeySpec(value));
    }`

jeho0815:
public PrivateKey loadPrivateKey() throws Exception {
      if (keyValue == null) {
        throw new RuntimeException("Missing private key path");
      }
      String pem = keyValue.get().toString();
      byte[] value = null;
      KeyFactory rsaKeyFactory = KeyFactory.getInstance("RSA");
      KeySpec keySpec = null;
      if (pem.indexOf("RSA PRIVATE KEY") > 0) {
          value = loadPems(pem, "RSA PRIVATE KEY").get(0);
          List<BigInteger> rsaIntegers = new ArrayList<BigInteger>();
          ASN1Parse(value, rsaIntegers);
          if (rsaIntegers.size() < 8)  {
            throw new Exception("'"  + 
              "' does not appear to be a properly formatted RSA key");
          }
          BigInteger publicExponent = rsaIntegers.get(2);
          BigInteger privateExponent = rsaIntegers.get(3);
          BigInteger modulus = rsaIntegers.get(1);
          BigInteger primeP = rsaIntegers.get(4);
          BigInteger primeQ = rsaIntegers.get(5);
          BigInteger primeExponentP = rsaIntegers.get(6);
          BigInteger primeExponentQ = rsaIntegers.get(7);
          BigInteger crtCoefficient = rsaIntegers.get(8);
          keySpec = new RSAPrivateCrtKeySpec(modulus, publicExponent, privateExponent,
            primeP, primeQ, primeExponentP, primeExponentQ, crtCoefficient);
      }
      else {
          value = loadPems(pem, "PRIVATE KEY").get(0);
          keySpec = new PKCS8EncodedKeySpec(value);
      }
      
      return rsaKeyFactory.generatePrivate(keySpec);
    }
    
    private static void ASN1Parse(byte[] b, List<BigInteger> integers) 
            throws Exception  {
          int pos = 0;
          while (pos < b.length)  {
[...]

PR #1897, opened 2017-03-24
Fix issue #1851

jan-zajic:
@see http://stackoverflow.com/questions/20065304/what-is-the-differences-between-begin-rsa-private-key-and-begin-private-key

I added server-key-pkcs1.pem to test sources/tls, which is server-key.pem in the same place converted using command:
openssl rsa -inform PEM -outform PEM -in server-key.pem -out server-key-pkcs1.pem

Test fail before fix:
java.lang.RuntimeException: Missing -----BEGIN PRIVATE KEY----- delimiter
	at io.vertx.core.net.impl.KeyStoreHelper.loadPems(KeyStoreHelper.java:271)
	at io.vertx.core.net.impl.KeyStoreHelper.access$0(KeyStoreHelper.java:246)
	at io.vertx.core.net.impl.KeyStoreHelper$KeyCert.loadPrivateKey(KeyStoreHelper.java:212)
	at io.vertx.core.net.impl.KeyStoreHelper$KeyCert.loadStore(KeyStoreHelper.java:202)
	at io.vertx.test.core.KeyStoreTest.testKeyStore(KeyStoreTest.java:394)
	at io.vertx.test.core.KeyStoreTest.testRsaKeyCertPath(KeyStoreTest.java:342)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

Now OK, other tests in same file as well.	
Signed-off-by: jzajic <jan.zajic@corpus.cz>


vietj:
can you add a step in `ssl.txt` that describes how to generate the key with openssl CLI ?

vietj:
and update the documentation with the supported format + an example

jan-zajic:
@vietj updated. or you mean something else?

lulf:
@vietj  Any update on this? 

The OpenShift mechanisms for generating service certificates output keys in pkcs#1 format. It would be nice to be able to use those with vert.x without having to convert.

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
```diff
@@ -242,33 +312,47 @@ public KeyStore loadStore(VertxInternal vertx) throws Exception {
...
-      if (end == -1) {
-        throw new RuntimeException("Missing " + endDelimiter + " delimiter");
-      }
-      String content = pem.substring(index, end);
-      content = content.replaceAll("\\s", "");
-      if (content.length() == 0) {
-        throw new RuntimeException("Empty pem file");
+      
+      String beginDelimiter = beginMatcher.group(1);
+      index += beginMatcher.end();
```
vietj: seems to be a bug

jan-zajic: ?

vietj:
this shall be rather in Vert.x 3.5

@lulf if you need, we can give you code that converts the key from PKCS1 to PKCS8 in the meantime

vietj:
@jan-zajic I've created a rebased version of your branch here (and amended your commits) https://github.com/vietj/vert.x/tree/feature/1802 can you check ?

lulf:
@vietj The problem isn't the conversion so much as the implication of having to rerun the conversion whenever the certificate file is updated due to certificate expiry. I'm using openssl to run the conversion for now.

vietj:
I mean that the conversion could be done with java, in your app until Vert.x supports this feature.

vietj:
@jan-zajic I've reworked your patch so it could be handled by the IP process due to the fork of the existing source key parser file and now it is in master - thanks for the contribution

@lulf would you mind to give a try ?

lulf:
It is able to read the cert now @vietj , thanks !

PR #1900, opened 2017-03-27
Timers should be cancellable until they are fired - fixes #1899

### LINKED ISSUE #1899: Timers should be cancellable until they are fired
motivation: timers are scheduled by the event loop and executed on its context, during the time between a timer callback is scheduled and its actual execution, the timer cannot be cancelled although it could still be. Such situation can happen easily with worker contexts where timer tasks can be delayed due to blocking.

change: uses an atomic boolean `cancelled` to allow a timer to be cancelled when it has been scheduled for execution on the context. The atomic boolean guarantees the visibility across threads and also the timer can be marked as cancelled once.

vietj:
@pmlopes @tsegismont please review


src/main/java/io/vertx/core/Main.java
```diff
@@ -0,0 +1,120 @@
+package io.vertx.core;
```
cescoffier: Should be removed no ?

src/main/java/io/vertx/core/Main2.java
```diff
@@ -0,0 +1,47 @@
+package io.vertx.core;
```
cescoffier: Also to be removed

vietj: omg yes :-)

src/main/java/io/vertx/core/Main.java
vietj: omg yes :-)

PR #1904, opened 2017-03-27
Test case showing retrieved buffer does not match created size.

whitingjr:
The returned array misses the last element.
Merging this test will cause the test suite to fail.

### DIFF
--- a/src/test/java/io/vertx/test/core/BufferTest.java
+++ b/src/test/java/io/vertx/test/core/BufferTest.java
@@ -605,6 +605,7 @@ public class BufferTest {
     Buffer b = Buffer.buffer(bytes);
 
     assertTrue(TestUtils.byteArraysEqual(bytes, b.getBytes()));
+    assertEquals(100, b.getBytes(0, 99).length);
   }
 
   @Test


src/test/java/io/vertx/test/core/BufferTest.java
```diff
@@ -605,6 +605,7 @@ public void testGetBytes() throws Exception {
     Buffer b = Buffer.buffer(bytes);
 
     assertTrue(TestUtils.byteArraysEqual(bytes, b.getBytes()));
+    assertEquals(100, b.getBytes(0, 99).length);
```
vietj: that's actually expected, what would you expect instead ?

whitingjr: Just what the test case is asserting. I am expecting the length of the returned byte array to be 100.
When I read the parameter comments on [Buffer.getBytes(int, int)](https://github.com/eclipse/vert.x/blob/3.4.1/src/main/java/io/vertx/core/buffer/Buffer.java#L280) it indicates array positions. 0 being the first and 99 being the 100th.
Are you telling me a request for 100 elements in the array should only return 99 is expected ?

whitingjr: To make it more obvious what I am going on about here is the test failure.
```
java.lang.AssertionError: expected:<100> but was:<99>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at io.vertx.test.core.BufferTest.testGetBytes(BufferTest.java:608)
```

vietj: so following your logic, `b.getBytes(0, 0).length == 1` ?

vietj: and to have an empty array I need to write `b.getBytes(0, -1)`

whitingjr: The point I was making was the API docs for the method is a bit vague. The first parameter being inclusive and the second exclusive.

PR #1907, opened 2017-03-27
Fixes #1905 EventBus interceptor does not get message body when clustered

### LINKED ISSUE #1905: EventBus interceptor does not get message body when clustered
From [David Klotz](https://groups.google.com/d/msg/vertx/7otR0fsI6_8/E4ZdC0usCQAJ) on the forum

```
Hi, 

I'm currently a bit confused by the differing behavior of event bus interceptors in a non-clustered and a clustered vert.x environment (currently on 3.2.1, can test again in 3.4.x if something in that area has changed since then). When run in a non-clustered environment, the interceptor has access to both the message body and the message headers, but when run in a clustered environment, the message body is always null. Is this behavior by design or a bug?

Cheers, 
David
```

tsegismont:
Fixes #1905 EventBus interceptor does not get message body when clustered

ClusteredMessage#body returns null for a clustered message not yet sent.
 The fix consists in using the codec #transform method in this case

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/SendContext.java
+++ b/src/main/java/io/vertx/core/eventbus/SendContext.java
@@ -22,8 +22,12 @@ public interface SendContext<T> {
   void next();
 
   /**
-   *
    * @return true if the message is being sent (point to point) or False if the message is being published
    */
   boolean send();
+
+  /**
+   * @return the value sent or published (before being processed by the codec)
+   */
+  Object sentBody();
 }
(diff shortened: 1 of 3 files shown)


tsegismont:
@vietj can you have a look again? I've reverted the ClusteredMessage#body changes and added SendContext#sentMessage

src/main/java/io/vertx/core/eventbus/SendContext.java
```diff
@@ -22,8 +22,12 @@
...
   /**
-   *
    * @return true if the message is being sent (point to point) or False if the message is being published
    */
   boolean send();
+
+  /**
+   * @return the value sent or published (before being processed by the codec)
+   */
+  Object sentMessage();
```
vietj: rather sentBody or sentMessageBody

tsegismont:
Done

2017-04-04 10:15 GMT+02:00 Julien Viet <notifications@github.com>:

tsegismont:
@vietj ping

PR #1915, opened 2017-03-29
Meaningful Future toString()

jponge:
Having a proper toString() implementation helps while debugging.


vietj:
what format is that :-) !

jponge:
Never used the IntelliJ toString generator?

PR #1918, opened 2017-04-02
Javadoc enhancement for JsonObject and JsonArray

EmadAlblueshi:

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -217,8 +217,9 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable {
   /**
    * Get the Instant at position {@code pos} in the array.
    * <p>
-   * JSON itself has no notion of a temporal types, so this method assumes there is a String value and
-   * it contains a ISOString encoded date, which it decodes if found and returns.
+   * JSON itself has no notion of a temporal types, this extension complies to the RFC-7493, so this method assumes
+   * there is a String value and it contains an ISODATE encoded data and time such as "2017-04-03T10:25:41Z", which it
+   * decodes if found and returns.
    * <p>
    * This method should be used in conjunction with {@link #add(Instant)}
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -224,6 +225,7 @@ public JsonArray getJsonArray(int pos) {
    * @param pos  the position in the array
    * @return  the Instant, or null if a null value present
    * @throws java.lang.ClassCastException if the value cannot be converted to String
+   * @throws java.time.format.DateTimeParseException if the String value is not a legal ISODATE encoded value
```
cescoffier: Maybe we should specify the expected format (Not sure everyone knows what ISODATE is).

src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -259,7 +259,7 @@ public JsonArray getJsonArray(String key) {
    * @param key  the key to return the value for
    * @return the value or null if no value for that key
    * @throws java.lang.ClassCastException if the value is not a String
-   * @throws java.lang.IllegalArgumentException if the String value is not a legal Base64 encoded value
+   * @throws java.time.format.DateTimeParseException if the String value is not a legal ISODATE encoded value
```
cescoffier: Maybe we should specify the expected format (Not sure everyone knows what ISODATE is).

src/main/java/io/vertx/core/json/JsonArray.java
EmadAlblueshi: I think the format example should be mentioned in the paragraph above and leave the ISODATE as is

cescoffier: Up to you for the location, it just needs to be somewhere.

EmadAlblueshi:
@cescoffier Done

cescoffier:
👍

PR #1930, opened 2017-04-08
Future: Add new defaulted composeResult() method.

xkr47:
Allows mapping both success and failure results in one go.

No unit tests yet, probing for your interest in general first..

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -211,6 +211,38 @@ public interface Future<T> extends AsyncResult<T>, Handler<AsyncResult<T>> {
   boolean failed();
 
   /**
+   * Compose this future with a {@code mapper} function.<p>
+   *
+   * When this future (the one on which {@code compose} is called) completes, the {@code mapper} will be called with
+   * the result and this mapper returns another future object. This returned future completion will complete
+   * the future returned by this method call.<p>
+   *
+   * If the {@code mapper} throws an exception, the returned future will be failed with this exception.<p>
+   *
+   * This can be used to implement e.g. "finally" clauses, but can also modify the result and/or postpone it.<p>
+   *
+   * @param mapper the mapper function
+   * @return the composed future
+   */
+  default <U> Future<U> composeResult(Function<AsyncResult<T>, Future<U>> mapper) {
+    if (mapper == null) {
+      throw new NullPointerException();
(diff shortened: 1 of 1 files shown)


xkr47:
Could perhaps also have a separate `andFinally(Runnable handler)` that has no async and no result-modifying capacity for simple cases..

tsegismont:
How is this different from
io.vertx.core.Future#compose(java.util.function.Function<T,io.vertx.core.Future<U>>)
?

2017-04-08 23:47 GMT+02:00 Jonas Berlin <notifications@github.com>:

xkr47:
> How is this different from `io.vertx.core.Future#compose(java.util.function.Function<T,io.vertx.core.Future<U>>)` ?

The function you give to compose() receives a `T` object as input if the future was successful, while the one you give to composeResult() receives a `AsyncResult<T>` object as input regardless of whether the future was successful or it failed.

The `AsyncResult<T>` contains methods with which you can retrieve the `T` object in case of success and also methods with which you can retrieve the Throwable in case of failure.

Thus, you can map both the success and failure cases in one go. For example you could implement a finally block like this:

```java
Future<Result> res = createTempFile("foo.tmp").compose((file) -> {
  return doSomethingWithFile(file).composeResult((asyncResult) -> {
    removeTempFile(file)
    return Future.future((fut) -> fut.handle(asyncResult));
  });
});
```

tsegismont:
Your problem would be solved if you could write something like

Future<Result> res = createTempFile("foo.tmp").compose((file) -> {
  return doSomethingWithFile(file).whenComplete(removeTempFile(file));
});

Just like with CompletionStage/CompletableFuture

I'd recommend to use CS/CF instead of Vert.x future for this use case, or
the RxJava API.


2017-04-10 21:28 GMT+02:00 Jonas Berlin <notifications@github.com>:

xkr47:
Are there plans to deprecate the vert.x Future etc classes in favour of java8 features?

vietj:
@xkr47 not at the moment, we had a discussion about Java 8 CompletableFuture a few months ago, that you can read on vertx-dev list

vietj:
done as https://github.com/eclipse-vertx/vert.x/commit/95f82b52bd68c28fe05c4cf0d2c9aaa88f358ced

xkr47:
Ok great, works for me! :)

PR #1933, opened 2017-04-11
Change "blocking code pool" to "worker pool"

chengenzhao:
Since " it is called worker pool in other places" :)
This change is based on following enhancement suggestion:
https://github.com/vert-x3/vertx-web-site/issues/210#issuecomment-293002313


chengenzhao:
I created another pull request as following to replace this one since it is not signed off

https://github.com/eclipse/vert.x/pull/1934

PR #1935, opened 2017-04-11
Fix Http2ServerResponse.putHeader to match the Http1ServerResponse implementation

gmokki:
The Http2 putHeader implementation did not replace but instead appended.

This broke for example pac4j authentication with Safari browser when using HTTP2 connections.


vietj:
looks good, but I think the test should add two different headers value, to make sure the second value overwrites the first one

gmokki:
Now sets two unique values and checks that the latter value wins.

gmokki:
Btw. is there a test base that would run tests with both http1 and http2 - such tests would make sure the implementations stay in sync in the future.

vietj:
@gmokki yes HttpTest that contains common tests between Http1xTest and Http2Test

vietj:
probably you can move this test in HttpTest

gmokki:
Moved the test to HttpTest

vietj:
@gmokki thanks for the contribution!

PR #1936, opened 2017-04-11
HttpClient does not remove established HTTP/2 connection when the cha…

### LINKED ISSUE #1925: HttpClient does not remove established HTTP/2 connection when the channel becomes inactive
Hi,

I'm working on a project, where I have two Vert.x applications which communicate with each other using REST. In one application I'm running an HttpServer and in the other an HttpClient, which connects directly using HTTP/2. When the server crashes, there is an unhandled "an existing connection was closed by the remote host" exception in the client, but the connection does not seem to get removed from the Http2Pool. Thus when the client application tries to make another request, the request.end() method succeeds without any exception and the client application never gets a response or an exception.

I'm using Vert.x 3.4.1 and JDK 8 Update 121.

I have created two demo Maven projects that can be used to reproduce the problem:
 [VertxHttpClientProblem.zip](https://github.com/eclipse/vert.x/files/895890/VertxHttpClientProblem.zip)

Here is the procedure:
1. Run the server using: java -jar http-server-0.1-SNAPSHOT-combined.jar
2. Run the client using: java -jar http-client-0.1-SNAPSHOT-combined.jar
3. Press enter in the client to make the first request.
4. Stop the server process using Ctrl+C
5. Press enter in the client process again to make another request (for which no response is received and no exception is thrown).

The code of the demo applications is also here:
HttpServerTest.java:

```
package test.http.server;

import io.vertx.core.AbstractVerticle;
import io.vertx.core.Vertx;
import io.vertx.core.http.HttpServerOptions;
import io.vertx.ext.web.Router;
import io.vertx.ext.web.handler.BodyHandler;

public class HttpServerTest extends AbstractVerticle {
	
		Vertx vertx = Vertx.vertx();
		vertx.deployVerticle(new HttpServerTest());
	}
	
	@Override
	public void start() {
		router.route().handler(BodyHandler.create());
		router.post("/echo").handler(routingContext -> {
			String text = routingContext.getBodyAsString();
			System.out.println("Received: " + text);
			routingContext.response().end(text);
		});
		
		HttpServerOptions serverOptions = new HttpServerOptions();
		serverOptions.setCompressionSupported(true);
		serverOptions.setTcpKeepAlive(true);
		vertx.createHttpServer(serverOptions).requestHandler(router::accept).listen(8080, result -> {
			if (result.succeeded()) {
				System.out.println("HttpServer listening.");
			} else {
				System.out.println("Error starting HttpServer.");
			}
		});
		
	}
}
```

HttpClientTest.java:

```
package test.http.client;

import io.vertx.core.AbstractVerticle;
import io.vertx.core.Vertx;
import io.vertx.core.http.HttpClient;
import io.vertx.core.http.HttpClientOptions;
import io.vertx.core.http.HttpClientRequest;
import io.vertx.core.http.HttpVersion;
import java.io.BufferedReader;
import java.io.IOException;
import java.io.InputStreamReader;
import java.util.concurrent.Semaphore;

public class HttpClientTest extends AbstractVerticle {
	
	// The HttpClient for the requests.
	private HttpClient httpClient;
	
	// Used only to signal the read console loop on the main thread when to continue.
	private Semaphore semContinue = new Semaphore(0);
	
	public static void main(String[] args) throws IOException, InterruptedException {
		Vertx vertx = Vertx.vertx();
		HttpClientTest verticle = new HttpClientTest();
		vertx.deployVerticle(verticle);
		
		BufferedReader input = new BufferedReader(new InputStreamReader(System.in));
		while (true) {
			verticle.semContinue.acquire();
			System.out.println("Press enter to make a new HTTP request.");
			input.readLine();
			verticle.context.runOnContext(verticle::makeRequest);
		}
	}
	
	@Override
	public void start() {
		HttpClientOptions clientOptions = new HttpClientOptions();
		clientOptions.setTryUseCompression(true);
		// We use HTTP/2 directly.
		clientOptions.setProtocolVersion(HttpVersion.HTTP_2);
		clientOptions.setHttp2ClearTextUpgrade(false);
		this.httpClient = this.vertx.createHttpClient(clientOptions);
		this.semContinue.release();
	}
	
	private void makeRequest(Void v) {
		HttpClientRequest request = this.httpClient.post(8080, "localhost", "/echo", response -> {
			response.bodyHandler(fullBuffer ->{
				System.out.println("Response: " + fullBuffer.toString());
			});
		});
		request.exceptionHandler(error -> {
			System.out.println(error.toString());
		});
		request.end("Test");
		System.out.println("Request: Test");
		this.semContinue.release();
	}
}
```

I've tried to debug the client application stepping into the Vert.x code and found out that an infinite loop seems to be created when a new request is made after the server has crashed. ConnectionManager.deliverStream(conn, waiter) calls Http2Pool.createStream(conn), which throws an exception because DefaultHttp2Connection.isClosed() returns true. Thus deliverStream() calls getConnection(waiter), which results in ConnectionManager.deliverStream(conn, waiter) to be added to the context's queue again.

Best regards,
Thomas

vietj:
…nnel becomes inactive - fixes #1925

please review @tsegismont

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http2ConnectionBase.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ConnectionBase.java
@@ -22,6 +22,7 @@ import io.netty.buffer.CompositeByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.handler.codec.http2.Http2Connection;
 import io.netty.handler.codec.http2.Http2Exception;
 import io.netty.handler.codec.http2.Http2Flags;
@@ -117,6 +118,7 @@ abstract class Http2ConnectionBase extends ConnectionBase implements Http2FrameL
 
   @Override
   public synchronized void handleClosed() {
+    closed = true;
     super.handleClosed();
   }
 
@@ -455,6 +457,14 @@ abstract class Http2ConnectionBase extends ConnectionBase implements Http2FrameL
   @Override
   public synchronized Http2ConnectionBase exceptionHandler(Handler<Throwable> handler) {
     return (Http2ConnectionBase) super.exceptionHandler(handler);
+  }
+
+
+  /**
+   * @return the Netty channel - for internal usage only
+   */
+  public Channel channel() {
+    return channel;
   }
 
   // Private
--- a/src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
@@ -101,9 +101,15 @@ class Http2ClientConnection extends Http2ConnectionBase implements HttpClientCon
   }
 
   @Override
+  public synchronized void handleClosed() {
+    http2Pool.discard(this);
+    super.handleClosed();
+  }
+
+  @Override
   public boolean isValid() {
     Http2Connection conn = handler.connection();
-    return !conn.goAwaySent() && !conn.goAwayReceived();
+    return !isClosed() && !conn.goAwaySent() && !conn.goAwayReceived();
   }
 
   @Override
--- a/src/test/java/io/vertx/test/core/Http2Test.java
+++ b/src/test/java/io/vertx/test/core/Http2Test.java
@@ -16,14 +16,19 @@
 
 package io.vertx.test.core;
 
+import io.netty.channel.Channel;
+import io.netty.channel.socket.nio.NioSocketChannel;
 import io.vertx.core.Context;
 import io.vertx.core.buffer.Buffer;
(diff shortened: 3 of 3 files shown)


src/main/java/io/vertx/core/http/impl/Http2ConnectionBase.java
```diff
@@ -457,6 +459,14 @@ public synchronized Http2ConnectionBase exceptionHandler(Handler<Throwable> hand
     return (Http2ConnectionBase) super.exceptionHandler(handler);
   }
 
+
+  /**
+   * @return the Netty channel - for internal usage only
```
tsegismont: Well, visible for testing only actually  :)

tsegismont:
LGTM

src/main/java/io/vertx/core/http/impl/Http2ConnectionBase.java
vietj: testing == internal :-)

PR #1954, opened 2017-04-24
Sni support

vietj:
motivation : SNI is an important features for virtual hosting

changes: provides SNI support for HttpClient/HttpServer/NetClient/NetServer. On the server certificates are matched using the certificate CN or SAN DNS, the `PemKeyCertOptions` has been extended to support multiple entries for this matter. For HttpClient, the host header value is sent as SNI server name, for NetClient the `connect` method is overloaded to provide the server name in addition of the connect host.

please review @pmlopes @alexlehm @cescoffier


src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -237,21 +245,26 @@ public SSLHelper setApplicationProtocols(List<HttpVersion> applicationProtocols)
...
         builder = SslContextBuilder.forClient();
+        KeyManagerFactory keyMgrFactory = getKeyMgrFactory(vertx);
         if (keyMgrFactory != null) {
           builder.keyManager(keyMgrFactory);
         }
       } else {
-        if (keyMgrFactory == null) {
-          throw new VertxException("Key/certificate is mandatory for SSL");
+        if (mgr != null) {
+          builder = SslContextBuilder.forServer(mgr.getPrivateKey(null), "wibble", mgr.getCertificateChain(null));
```
gmokki: Is `wibble` a magic value required here? Or would it be possible to use other passwords too =)

vietj: good catch :-)

gemmellr:
I haven't looked at this in great depth, so I might be entirely off base with the below, apologies if so :)

SNI has happened implicitly for by default from at least Java7 so long as you provided the host detail for the SSLEngine. From Java8 you could additionally control the value explicitly using the SNIHostName options so as to allow configuring it to alternate values (e.g. connect to one DNS address but indicate an different server name for SNI) or explicitly making SNI happen in some case it wouldn't implicitly (not for IP's, obviously, and not for hostnames that aren't fully qualified, like localhost). The old code looks like it was providing the hostname, and if so then SNI would have been already happening previously in many cases, and looks like it could continue to even though the new client option says 'sni is off', which if so seems like it might be a confusing to some folks.

Less interesting..might having the NetClient connect method take some options instead of just a serverName param, or just adding an explicit 'sni server name' option shared by all connections from a given NetClient, rather than adding more connect methods for each new interesting param, be something to consider?

vietj:
@gemmellr I thought initially about overloading with a `ConnectOptions` having the host/port/serverName properties, but went for overloading with serverName.

what else beside host/port/serverName could we add ?

vietj:
so I'm ok having a ConnectOptions if that makes sense.

vietj:
the only interesting case I think for providing an explicit SNI host, is to connect to an IP and provide an SNI explicit value.

vietj:
@gemmellr the `NetClient` might be used to connect to distincts hosts

vietj:
@gemmellr so what would you recommend precisely for `NetClient` ?

gemmellr:
@vietj Yep, I realise, and figured that would be the main reason not to put it on the options.

I personally think some of the options that are on NetClientOptions could be equally desirable to change on per-connection basis, such as some of the other SSL related config for example, but I appreciate that it is where it is now and we should do what makes most sense in relation to that. That probably is just adding the new param to connect. I can't immediately think of anything new that will cause us to need more connect options soon, the thought was mainly that perhaps we should guard against needing more later for things that haven't been considered yet, just like sni hostnames.

As an aside, one reason to specify an explicit SNI host besides connecting using an IP, is connecting to a DNS hostname for a service that fronts various services but doesnt offer DNS names for them, or they are secured with self signed certs etc. I believe thats the reason @lulf needs it in vertx-proton.

gemmellr:
@vietj For NetClient, going from 1 to 2 connect methods isnt so bad, so I'm fine jsut adding the param. As above, just seemed worth considering if more general options would be useful and avoid adding more connect methods later if anything else was ever needed.

Vertx-proton would have been going from 4 to 6, and there I dont see a need there since it already has per-connection options. It could just use the new connect method under the covers and choose what it gives to it from its options.

vietj:
@gemmellr 

so you rather recommend having a `ConnectOptions` with host/port/serverName that could hold more properties later ?

about the DNS hostname : that would not work for HTTP because of hostname validation though.

vietj:
so for now we keep as-is and if we ever need to overload, we use a `ConnectOptions`

vietj:
I did not introduce ConnectOptions for this reason : I found reasonnable to have a new method as NetClient as a pretty simple API to read.

gemmellr:
@vietj In this case, I'm fine either way. I'd probably keep host and port on the method, you always need them, and have options for everything else optional. Yep, its mainly for non-HTTP stuff that this gets interesting. However, on that point, if you do override the SNI name the cert name matching should be against the SNI name specified, not the DNS hostname / IP.

vietj:
I think that's what we do : https://github.com/eclipse/vert.x/blob/sni-support/src/test/java/io/vertx/test/core/HttpTLSTest.java#L677

gemmellr:
Yep I assumed so, its what Java does for its own checks afaik..I only commented given your description above, and again as I was mainly meaning of the non-HTTP stuff.

vietj:
thanks for the review @gemmellr

gemmellr:
@vietj No problem. Did you see the bit in my original comment about the HTTP client option possibly being confusing? It seemed like it would still do implicit SNI by default (wont work on just "localhost" however), though the new option rather implies it wont do SNI, with it only setting the explicit SNI value if the option is set true. Did I just misinterpret what its going to do?

vietj:
@gemmellr thanks, I will add tests to see how it goes (and perhaps use wireshark) and be sure of that and adjust, thanks for the feedback

vietj:
@gemmellr we also support OPEN_SSL, so it may be different and I will check both cases

gemmellr:
@vietj -Djavax.net.debug=all can also be good, you can tell from there whether SNI is in use. E.g the JDK stuff wont do implicit SNI for "localhost" but if your box has "localhost.localdomain" then I think it does for that.

vietj:
@gemmellr thanks for the advice, I'll follow it (and I find wireshark to be more readable than javax.net.debug :-) )

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
```diff
@@ -112,34 +118,118 @@ public static KeyStoreHelper create(VertxInternal vertx, TrustOptions options) {
...
+  private final String password;
+  private final KeyStore store;
+  private final Map<String, X509KeyManager> wildcardMgrMap = new HashMap<>();
+  private final Map<String, X509KeyManager> mgrMap = new HashMap<>();
 
-  public KeyStoreHelper(String password) {
+  public KeyStoreHelper(KeyStore ks, String password) throws Exception {
+    Enumeration<String> en = ks.aliases();
+    while (en.hasMoreElements()) {
+      String alias = en.nextElement();
```
xkr47: I'm not very good at KeyStores yet; is it possible for multiple aliases to refer to the same certificate? In that case I guess we could optimize by skipping certificates for which we have already stored in `mgrMap` and `wildcardMgrMap`?

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
```diff
@@ -112,34 +118,118 @@ public static KeyStoreHelper create(VertxInternal vertx, TrustOptions options) {
...
+            @Override
+            public PrivateKey getPrivateKey(String s) {
+              return key;
+            }
+          };
+          for (String domain : domains) {
+            if (domain.startsWith("*.")) {
+              wildcardMgrMap.put(domain.substring(2), mgr);
+            } else {
+              mgrMap.put(domain, mgr);
```
xkr47: Should we perhaps check the return values of the two `put()` calls here and check they are null to detect (and report) accidentally loading multiple certificates for the same domain(s)?

vietj:
@gemmellr I've done a test and currently there is no _implicit_ SNI with 3.4.2-SNAPSHOT, I've added this test `Http1xTest#testSNIClientDisabled` that asserts it and in wireshark I don't see the server name extension in the client hello .

The current code in `SSLHelper` configures SNI:

```
    if (serverName != null) {
      SSLParameters sslParameters = engine.getSSLParameters();
      sslParameters.setServerNames(Collections.singletonList(new SNIHostName(serverName)));
      engine.setSSLParameters(sslParameters);
    }
```

which is pretty standard, however we get the `SSLEngine` using:

```
SSLEngine engine = getContext(vertx, null).newEngine(ByteBufAllocator.DEFAULT, host, port);
```

Providing the `peerHost`, as far as I remember you said that when providing the `peerHost`, the SNI extension should use implicitly this value. But the engine creation is delegated to Netty's `SslContext` and is not done by the JDK `SSLContext` (although obviously `SslContext` more or less delegates to `SSLContext`.

So this seems to disagree with your earlier comment, any thought ?

I haven't verified for older version (such as 3.3.3) though (no time tonight).

xkr47:
I guess it is not at this point too interesting to provide direct support for dynamically updating the KeyCertOptions like I did in https://github.com/xkr47/vertx-acme4j/blob/master/src/main/java/io/nitor/vertx/acme4j/util/DynamicCertOptions.java - I guess it might also be challenging to support this properly in other languages than Java (?). However it seems it will still be possible to implement this oneself in the future as well, and perhaps even with much less effort thanks to the new `KeyCertOptions.keyManagerMapper()` api. So that's good enough for me. (I use it to implement LetsEncrypt tls-sni handshakes.)

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
vietj: I think the user is responsible to know what is inside his keystore

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
```diff
@@ -112,34 +118,118 @@ public static KeyStoreHelper create(VertxInternal vertx, TrustOptions options) {
...
+            @Override
+            public String chooseClientAlias(String[] strings, Principal[] principals, Socket socket) {
+              throw new UnsupportedOperationException();
+            }
+            @Override
+            public String[] getServerAliases(String s, Principal[] principals) {
+              throw new UnsupportedOperationException();
+            }
+            @Override
+            public String chooseServerAlias(String s, Principal[] principals, Socket socket) {
```
vietj: this implementation of `X509KeyManager` is not called by JSSE itself but only by the `SSLHelper`, it's rather a reuse to avoid introducing an interface that would bundle key/certificates

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
vietj: no, it's not possible

src/main/java/io/vertx/core/net/PemKeyCertOptions.java
```diff
@@ -116,82 +139,234 @@ public JsonObject toJson() {
   }
 
   /**
-   * Get the path to the key file
+   * Get the path to the first key file
    *
    * @return the path to the key file
    */
+  @GenIgnore
   public String getKeyPath() {
```
vietj: it's not deprecated, the @GenIgnore is rather here to support properly multiple paths.

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -409,36 +429,65 @@ private SslHandler createHandler(SSLEngine engine, boolean client) {
...
+      return sslContext;
+    } else {
+      X509KeyManager mgr;
+      try {
+        mgr = keyCertOptions.keyManagerMapper(vertx).apply(serverName);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
       }
+      if (mgr == null) {
+        return sslContext;
```
vietj: yes this is deliberate, SSLHelper#getContext(null) is always called by the `validate` method when the server is starting. It could probably be improved to be more readable I think with a small refactoring.

vietj:
How does the bombing attack work ?

in the SSLHelper the Map<Certificate, SslContext> is here to avoid a Map<String, SslContext> and avoid an attack that would make an OOM assuming the server caches the SslContext by serverName

the option looks interesting though so it could be added to `NetServerOptions`

vietj:
@xkr47 the plugability API is like before the `KeyCertOptions` that has been extended with the mapper. Providing a more user oriented API should be done outside, for instance one could provide a VertxGen API for building a specific `KeyCertOption` interface that uses a `Function<String, KeyCertOptions>` I think.

vietj:
@xkr47 providing the `serverName` in the server API does not come naturally because the server name is only known in the `VertxSniHandler` and there is no obvious way to store it. Perhaps it could be stored in the generic `Channel` attributes ? I don't know if we can use Netty `Channel` attribute to store arbitrary stuff. If yes, that would make it easier.

xkr47:
> @xkr47 providing the serverName in the server API does not come naturally because the server name is only known in the VertxSniHandler and there is no obvious way to store it.

In `ConnectionBase.peerCertificateChain()` there is `return sslHandler.engine().getSession().getPeerCertificateChain();`. There also seems to be a `sslHandler.engine().getSession().getLocalCertificates()` - wonder if this would provide the actual certificate that was used? Perhaps I could test this myself..

gemmellr:
@vietj

> I've done a test and currently there is no implicit SNI with 3.4.2-SNAPSHOT, I've added this test Http1xTest#testSNIClientDisabled that asserts it and in wireshark I don't see the server name extension in the client hello .

I find that strange. Does the test use only "host1" or "localhost" for the peer host? Java wont do implicit SNI for those, you need to use something with a domain component in the name like "localhost.localdomain" / "example.com" / etc.

As example, if I modify a vertx-proton SSL test to use "localhost.localdomain" instead of "localhost" for a connect operation it starts doing SNI, without having modified vertx-proton itself in any way at all yet: "Extension server_name, server_name: [type=host_name (0), value=localhost.localdomain]"

The main reason I'm commenting about this is that I think we *should* do SNI by default, at least to the extent that Java normally does implicitly, and I would have expected it to before. Assuming it did, I think we certainly need to continue doing it. While the only time you really want/need to explicitly specify the server name for SNI is if you want/need it to be different than the peer host used, that isn't the only time SNI is needed or should be used.

If however it does do some SNI by default, then the new default 'sni = false' option false is misleading.

vietj:
@gemmellr ok, I'll add a component and check

vietj:
@xkr47 I've added support for getting the SNI server name

vietj:
@gemmellr actually you are right, I've changed the test has you indicated and the SNI hostname is implicitely sent.

Do you know the reason why it's not sent with a single component name ?

vietj:
@gemmellr so do you think we should follow the same way or always provide an SNI name ?

vietj:
or do you think it hurts to always send it ? do we have actually a case for disabling it ?

gemmellr:
@vietj This is one of those 'what it says its for, vs what you want to do' situations, I'm not sure there is an answer that pleases everyone.

In adding to such not-FQDN peer host names, the main other issue with sending a value all the time is going to be when the 'hostname' given to connect to is actually an IP address. Those don't make sense to send for SNI (and are expressly forbidden by the RFC) and thus the implicit SNI wont send them, whereas the explicit way presumably would (not tried).

For non-HTTP cases I think I'd do what it is now doing (and presumably did before) i.e doing implicit SNI, and then have a way to provide the desired value explicitly for any less usual cases where folks want to control the value. If you provide it explicitly then it gets sent explicitly, if not then the implicit SNI happens or not based on the hostname value given.

For the HTTP case, its a little more interesting, but I think the same approach might be sufficient..i.e maybe all thats needed is renaming the new 'sni' option to something like 'force sni' and clarifying that SNI will happen in most cases, but this will force it to send the associated value.

vietj:
@gemmellr thanks for the feedback

About forceSNI, it's a good tradeoff, we can leave it aside for now (i.e do implicit SNI) and add it later if there is demand for it. WDYT ?

gemmellr:
@vietj leaving the option out until someone says they need it also sounds reasonable, if you think most folks can do without it. I'm perhaps not the best person to ask though since I've never actually used the HTTP client and don't entirely know how behaviour has changed on that side of things :)

vietj:
I believe it might be interesting for writing some integration tests, perhaps that @xkr47 would need it as he's working on a gateway server that uses SNI. WDYT @xkr47 ?

lulf:
@vietj  Overriding the SNI value explicitly is very useful in an integration test setting, which is the reason for me wanting this feature at least.

vietj:
@lulf ok so let's keep it :-) . however we were talking of HttpClient and I believe you are mostly using porton based on NetClient which has it. But we'll have a forceSNI flag instead.

gemmellr:
Spending a bit more time looking at the code, for the HTTP client it seems it will use the host header value for the SSLEngine host if specified, in which case it effectively already has an SNI hostname value override, right? The thing a 'force' option would further let you do is make SNI happen even where the peer host or host header values specified were either non-FQDN or an IP address?

vietj:
The override is taken indeed from the `host` header.

The value should always been used from the host header as the host header is also used for pooling (so the request uses the right SSL connection and the correct certificate) 
and for `HTTPS` hostname validation. So using a different value than the `host` header does not make sense.

So the _force_ would just force the SSLEngine to use the `host` header when it is a single name like `localhost`.

In case of `NetClient` it does because there is no hostname validation mechanism.

vietj:
well and actually the primary purpose of SNI is to enable virtual hosts so well :-)

gemmellr:
Yep for NetClient in some cases you just might want to provoke use of different backend services fronted on the same DNS hostname (or IP address) and port, and SNI is a convenient way to do it, especially where there are multiple protocols involved too as you dont even need to inspect them then. That said, other protocols also include hostnames too though besides HTTP, such as AMQP 1.0, which is where some of the interest in this comes from for vertx-proton.

vietj:
@gemmellr @xkr47 I've updated the PR with the client forceSNI behavior and updated the tests to use FQDN instead of shortname aliases and add specific tests for shortname aliases

vietj:
@xkr47 you haven't replied to my question about bombing.

I was thinking perhaps to handle it differently:

- if the KeyCertOptions returns null : in short don't use the default (first) certificate => fail the TLS handshake
- add on JksOptions and PKCS12Options an alias to specify the fallback certificate to use in case of unknown SNI
- do the same on KeyCertOptions but with an index

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -256,28 +257,21 @@ public synchronized HttpServer listen(int port, String host, Handler<AsyncResult
...
-                    }
-                  });
+                if (options.isSni()) {
+                  VertxSniHandler sniHandler = new VertxSniHandler(sslHelper, vertx);
+                  pipeline.addLast(sniHandler);
+                  postSSLConfig(pipeline);
                 } else {
-                  configureHttp1(pipeline);
+                  pipeline.addLast("ssl", new SslHandler(sslHelper.createEngine(vertx)));
+                  postSSLConfig(pipeline);
```
vietj: ok

src/main/java/io/vertx/core/http/impl/ServerConnection.java
```diff
@@ -46,6 +46,8 @@
 import io.vertx.core.net.impl.VertxNetHandler;
 import io.vertx.core.spi.metrics.HttpServerMetrics;
 
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.security.cert.X509Certificate;
```
vietj: ok

xkr47:
> How does the bombing attack work ?
>
> in the SSLHelper the Map<Certificate, SslContext> is here to avoid a Map<String, SslContext> and avoid an attack that would make an OOM assuming the server caches the SslContext by serverName
>
> the option looks interesting though so it could be added to NetServerOptions

> @xkr47 you haven't replied to my question about bombing.

Sorry, I failed to associate that comment with me.

So "bombing" is just something I happened to call it.. I got the idea when I accidentally returned null from `KeyManager.chooseServerAlias()` and it resulted in a TLS handshake error. So I figured that I could use it for security.

So imagine you have a malicious client that keeps opening TLS connections to random IPs for scanning purposes (without SNI hostname in handshake). You therefore have no particular interest in serving the client. So if you decide to NOT have a default certificate in your server, the malicious client's attack will be cut short with a TLS handshake error. Only if the attacker knows a SNI hostname that your server actually accepts can he/she/it proceed. So this provides a little bit of additional security against these kinds of attacks. It however has the downside of rejecting any client NOT supplying a SNI hostname. There will therefore certainly be users that cannot use this option.

A variant of this would be that the attacker client DOES send a SNI hostname in the handshake, but one that doesn't match any hostname in the server's KeyStore. In this case we can also (separately) decide whether to reply with a default certificate or not. Not reporting with a default certificate in this case is less likely to cause issues but it's still possible that e.g. hostnames not listed in any certificates' SAN are used if configured in DNS.

> I was thinking perhaps to handle it differently:
>
> * if the KeyCertOptions returns null : in short don't use the default (first) certificate => fail the TLS handshake

For the two cases I listed above, the first case is handled solely by the `KeyCertOptions.getKeyManagerFactory()` instance since no SNI was provided by the client. The `KeyCertOptions.getKeyManagerFactory().engineGetKeyManagers()[0].chooseServerName()` can choose to return null in this case, failing the TLS handshake.

The second case is first handled by the `KeyCertOptions.keyManagerMapper()` instance. By your proposal (if I understood it correctly) the mapper returns null to fail the TLS handshake, or alternatively chooses & returns the default certificate. Therefore the getKeyManagerFactory() will never be consulted if the client supplies a SNI hostname and SNI is enabled on the server. I guess this would work fine, and would allow separately configuring the two cases I listed.

> * add on JksOptions and PfxOptions an alias to specify the fallback certificate to use in case of unknown SNI
> * do the same on PemKeyCertOptions but with an index

.. and allow the alias/index to be null to not have a default certificate?

I guess the `KeyStoreHelper.create()` would determine the default alias depending on the `KeyCertOptions` type given and pass it to the `KeyStoreHelper` constructor and then used in `KeyStoreHelper.getKeyMgr(String serverName)`?

xkr47:
> The override is taken indeed from the host header.

@vietj if the host header contains an IP address, will you be able to detect that in order not to use it as a SNI hostname? Or how are you planning on handling IPs? (also IPv6:s)

vietj:
this case is handled by setting an host header explicitly.

client.get(port, port, uri).setHost(“the_host);

vietj:
thanks all for the review, your contribution definitely made a difference!

xkr47:
@vietj So you didn't end up adding support for specifying fallback certificate? And no link from http/package-info to net/package-info?

vietj:
@xkr47 

I thought about it.

The issue with specifying a fallback certificate is that a client presenting a server name should get by default a certificate. However there is no way to have a good default value because we don't know the alias name the user would use in the existing keystore files. That being said SNI on the server is not enabled by default, so it means that SNI needs explicit configuration and the fallback certificate should be configured as well.

So I do have mixed feelings about it, it would be interesting to see what other thinks.

for the link : provide a PR :-)

gemmellr:
@vietj the 'forceSni' update and related test changes etc look good to me

gemmellr:
@xkr47
> if the host header contains an IP address, will you be able to detect that in order not to use it as a SNI hostname? Or how are you planning on handling IPs? (also IPv6:s)

The default behaviour is to use the implicit SNI that Java does, it doesn't send the SNI hostname if the value given is an IP or non-FQDN.

PR #1959, opened 2017-04-29
Issues/#1956 proxy ftp

alexlehm:
support ftp:// urls for http proxy requests

### DIFF
--- a/src/main/asciidoc/java/http.adoc
+++ b/src/main/asciidoc/java/http.adoc
@@ -1380,7 +1380,7 @@ The client can be configured to follow HTTP redirections: when the client receiv
 `301`, `302`, `303` or `307` status code, it follows the redirection provided by the `Location` response header
 and the response handler is passed the redirected response instead of the original response.
 
-Hereâ€™s an example:
+Here’s an example:
 
 [source,java]
 ----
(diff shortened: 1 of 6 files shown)


tsegismont:
@alexlehm the changes look good to me. However I wonder if we shouldn't update the `web-client` code as well (there is code there similar to what's in `HttpClientImpl.java` to determine the actual port)

vietj:
good point, we should also update the WebClientOptions with 'forceSni'

alexlehm:
@tsegismont i think we should, there is currently no method to create a HttpRequest with a protocol except requestAbs, so WebClient would have to go from abs url to parsed to abs url again, not quite sure what is the sensible solution for this

alexlehm:
@tsegismont I have created the same handling for web-client, I don't like they way this is working, maybe it can be improved

https://github.com/alexlehm/vertx-web/tree/web-client-protocol

vietj:
thanks Alex!

tsegismont:
I am still clearing my email backlog, so maybe you did it already, if not
can you open a PR there?

2017-05-04 23:30 GMT+02:00 Alexander Lehmann <notifications@github.com>:

alexlehm:
@tsegismont commited a small fix now

PR #1968, opened 2017-05-05
Make the logic of NetClientImpl reusable in NetClientBase

vietj:
This shall be used by the upcoming mqtt client to reuse the connection handing provided by Vert.x

### DIFF
--- a/src/main/java/io/vertx/core/metrics/impl/DummyVertxMetrics.java
+++ b/src/main/java/io/vertx/core/metrics/impl/DummyVertxMetrics.java
@@ -33,7 +33,6 @@ import io.vertx.core.http.HttpServerResponse;
 import io.vertx.core.http.ServerWebSocket;
 import io.vertx.core.http.WebSocket;
 import io.vertx.core.spi.metrics.*;
-import io.vertx.core.net.NetClient;
 import io.vertx.core.net.NetClientOptions;
 import io.vertx.core.net.NetServerOptions;
 import io.vertx.core.net.SocketAddress;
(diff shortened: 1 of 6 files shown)


src/main/java/io/vertx/core/metrics/impl/DummyVertxMetrics.java
```diff
@@ -82,7 +81,7 @@ public TCPMetrics createMetrics(SocketAddress localAddress, NetServerOptions opt
   }
 
   @Override
-  public TCPMetrics createMetrics(NetClient client, NetClientOptions options) {
```
vietj: it's not a public API and it can be changed, nobody really implements that except us in Dropwizard and Hawkular

src/main/java/io/vertx/core/net/impl/NetClientBase.java
```diff
@@ -0,0 +1,257 @@
+/*
+ * Copyright (c) 2011-2013 The original author or authors
```
vietj: it's a copy/paste from the NetClientImpl class

src/main/java/io/vertx/core/net/impl/NetClientBase.java
```diff
@@ -0,0 +1,257 @@
...
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ *
+ * This class is thread-safe
+ *
+ * @author <a href="http://tfox.org">Tim Fox</a>
```
vietj: yes, it's a copy/paste from the NetClientImpl class and Tim is the original author

src/main/java/io/vertx/core/metrics/impl/DummyVertxMetrics.java
vietj: without changing that we cannot make the feature work

jponge:
Great then.

Issue #1975, opened 2017-05-08
Faster json parsing


amannocci:
Motivation:
We can speed up json parsing provide in Buffer. 
In fact, Jackson is a bit more efficient when the parsing is based on byte[] or InputStream ([Jackson-Performance](https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance)).
In addition, this will reduce intermediate String allocation.
We can use the underlying netty ByteBuf to achieve this.

Modification:
We need to add a method in Json to be able to decodeValue from Buffer.
And then we can modify toJsonObject and toJsonArray implementation in Buffer.

Result:
Better performance for json parsing based on Buffer.

elad-yosifon:
@amannocci  what about exposing a static parser or ctor for InputStream and Buffer in addition to String and Map<String,Object> ?

amannocci:
I think that exposing a ctor for InputStream and Buffer in JsonObject and JsonArray can be useful.

vietj:
I disagree for InputStream, InputStream are blocking so it's not something we encourage to use in the context of Vert.x

vietj:
have you done actual benchmarking ?

vietj:
Actually Buffer has `toJsonArray` and `toJsonObject`, we should avoid to add new method that do the same thing.

amannocci:
I agree with you for InputStream, this can be a problem when the underlying implementation is blocking.
I prefer to replace existing `toJsonArray` and `toJsonObject` method in Buffer instead of creating new one.
I will try to do a jmh benchmark.
Actually I have only done a very simple benchmark to check the article.

vietj:
:+1:

amannocci:
I have done a little JMH to confirm performance improvement and I think we can safely switch to the new version.
```
# Run complete. Total time: 00:27:14

Benchmark                         Mode  Cnt       Score     Error  Units
JsonBenchmark.testNewJsonArray   thrpt  200   51411,287 ± 369,807  ops/s
JsonBenchmark.testNewJsonObject  thrpt  200  103836,793 ± 944,890  ops/s
JsonBenchmark.testOldJsonArray   thrpt  200   39609,818 ± 273,973  ops/s
JsonBenchmark.testOldJsonObject  thrpt  200   78209,119 ± 519,875  ops/s
```
I get a boost of **~30% for JsonObject** parsing from Buffer and **~27% for JsonArray** parsing from Buffer.

[...]

PR #1976, opened 2017-05-08
Faster json parsing #1975

amannocci:
Motivation:
We can speed up json parsing provide in Buffer. 
In fact, Jackson is a bit more efficient when the parsing is based on byte[] or InputStream ([Jackson-Performance](https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance)).
In addition, this will reduce intermediate String allocation.
We can use the underlying netty ByteBuf to achieve this.

Modification:
We need to add a method in Json to be able to decodeValue from Buffer.
And then we can modify toJsonObject and toJsonArray implementation in Buffer.

Result:
Better performance for json parsing based on Buffer.

Signed-off-by: amannocci <adrien.mannocci@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -78,15 +78,6 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable {
   }
 
   /**
-   * Create an instance from a stream of JSON.
-   *
-   * @param stream  the stream of JSON.
-   */
-  public JsonArray(InputStream stream) {
-    fromStream(stream);
-  }
-
-  /**
    * Get the String at position {@code pos} in the array,
    *
    * @param pos  the position in the array
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -19,6 +19,7 @@ package io.vertx.core.json;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.shareddata.impl.ClusterSerializable;
 
+import java.io.InputStream;
 import java.time.Instant;
 import java.util.*;
 import java.util.stream.Stream;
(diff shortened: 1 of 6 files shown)


src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -68,6 +69,24 @@ public JsonArray(List list) {
...
+  public JsonArray(Buffer buf) {
+    fromBuffer(buf);
+  }
+
+  /**
+   * Create an instance from a stream of JSON.
+   *
+   * @param stream  the stream of JSON.
+   */
+  public JsonArray(InputStream stream) {
```
pmlopes: Not totally sure about this, since it would allow the code to be abused with blocking IO as the InputStream is blocking by nature

pmlopes:
It could be convenient also to have an encode to buffer method. For example in most web code we will observe that users create a `JsonObject` then call `.encode()` to get a String representation and then convert to `Buffer` in order to send over the wire.

We could bypass this encode to string and then extract bytes and write directly to a Buffer

src/main/java/io/vertx/core/json/JsonArray.java
amannocci: You're right as @vietj.
I will remove this constructor.

amannocci:
Maybe a `toBuffer` method in both `JsonObject` and `JsonArray` could be relevant ?

pmlopes:
@amannocci I belive that a `toBuffer` method would be good idea, was per my previous comment. @vietj what do you think?

vietj:
why not

vietj:
the master merge in this branch makes the IP validation fail because the user `noreply@github.com` has not signed the Eclipse CA

pmlopes:
@amannocci I messed up the ip validation by fixing the merge conflict using the github UI, please refer to #1980 PR to a verified merge + review the `toBuffer` implementation.

PR #1979, opened 2017-05-10
Added support for doing recursion DNS queries

jasoncopeland:
- When creating a new DnsClientImpl() yourself, you can now optionally pass in a boolean that defines if the DNS query should be recursive or not.  By default, recursion is false, so you have to query the NS hosting the name in question.


vietj:
can you add a test for this ?

PR #1982, opened 2017-05-10
Avoid NPE and call handler with Failed Future

pmlopes:
Fixes: vert-x3/issues#80

Manually tested as per issue description.

### DIFF
--- a/src/test/java/io/vertx/test/core/Http1xTest.java
+++ b/src/test/java/io/vertx/test/core/Http1xTest.java
@@ -3537,17 +3537,7 @@ public class Http1xTest extends HttpTest {
     final File f = setupFile("file.pdf", TestUtils.randomUnicodeString(1000000));
 
     server.requestHandler(req -> {
-      vertx.executeBlocking(future -> {
-
-        try {
-          Thread.sleep(100);
-        } catch (InterruptedException e) {
-
-        }
-
-        future.complete();
-
-      }, false, res -> {
+      req.connection().closeHandler(v -> {
         req.response().putHeader(HttpHeaders.CONTENT_TYPE, "application/pdf");
 
         try {
--- a/src/test/java/io/vertx/test/core/Http1xTest.java
+++ b/src/test/java/io/vertx/test/core/Http1xTest.java
@@ -24,19 +24,7 @@ import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.VertxException;
 import io.vertx.core.buffer.Buffer;
-import io.vertx.core.http.ConnectionPoolTooBusyException;
-import io.vertx.core.http.Http2Settings;
-import io.vertx.core.http.HttpClient;
-import io.vertx.core.http.HttpClientOptions;
-import io.vertx.core.http.HttpClientRequest;
-import io.vertx.core.http.HttpConnection;
-import io.vertx.core.http.HttpMethod;
-import io.vertx.core.http.HttpServer;
-import io.vertx.core.http.HttpServerOptions;
-import io.vertx.core.http.HttpServerRequest;
-import io.vertx.core.http.HttpServerResponse;
-import io.vertx.core.http.HttpVersion;
-import io.vertx.core.http.RequestOptions;
+import io.vertx.core.http.*;
 import io.vertx.core.http.impl.HttpClientRequestImpl;
 import io.vertx.core.impl.ConcurrentHashSet;
 import io.vertx.core.impl.ContextImpl;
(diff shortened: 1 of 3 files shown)


vietj:
can we have a test ?

pmlopes:
@vietj I've added a test, it looks quite hacky but I couldn't find a way to kill the client right after the request started, so I ended up doing it using the netclient.

src/test/java/io/vertx/test/core/Http1xTest.java
```diff
@@ -3542,6 +3531,49 @@ public void testPerPeerPoolingWithProxy() throws Exception {
     testPerPeerPooling(i -> client.get(80, "host" + i, "/somepath"));
   }
 
+  @Test
+  public void testNoNPEWhenClientBreaksConnection() throws Exception {
+
+    final File f = setupFile("file.pdf", TestUtils.randomUnicodeString(1000000));
+
+    server.requestHandler(req -> {
+      vertx.executeBlocking(future -> {
```
vietj: why not setTimer ?

src/test/java/io/vertx/test/core/Http1xTest.java
```diff
@@ -3542,6 +3531,49 @@ public void testPerPeerPoolingWithProxy() throws Exception {
...
+
+        try {
+          req.response().sendFile(f.getAbsolutePath(), event -> {
+            if (event.failed()) {
+              testComplete();
+            } else {
+              fail("It should not reach this point");
+            }
+          });
+        } catch (NullPointerException e) {
```
vietj: that's unnecessary

src/test/java/io/vertx/test/core/Http1xTest.java
vietj: instead you can use a `closeHandler` on the associated `HttpConnection`, that will be more reliable

pmlopes: This was the reproducer code we got in the bug report.

src/test/java/io/vertx/test/core/Http1xTest.java
pmlopes: Yes it is, the NPE bubbles up from the sendFile call, it was on the original report.

src/test/java/io/vertx/test/core/Http1xTest.java
vietj: can you update it to use it ?

src/test/java/io/vertx/test/core/Http1xTest.java
vietj: can you remove it ?

pmlopes: If I remove this `catch` the test does not cover the issue.

vietj: I think it does because we check we have a async result failure. If there was an NPE, the async result failure would not happen. What you can do perhaps instead is to use `faitFor(2)` and make the test completed by a statement after the send file and by the expected async result failure

pmlopes:
If I remove the catch and test against the base code (without the fix) this is what happens:

```
Starting test: Http1xTest#testNoNPEWhenClientBreaksConnection 
Unhandled exception 
java.lang.NullPointerException
	at io.vertx.core.http.impl.HttpServerResponseImpl.doSendFile(HttpServerResponseImpl.java:512)
	at io.vertx.core.http.impl.HttpServerResponseImpl.sendFile(HttpServerResponseImpl.java:353)
	at io.vertx.core.http.HttpServerResponse.sendFile(HttpServerResponse.java:332)
	at io.vertx.core.http.HttpServerResponse.sendFile(HttpServerResponse.java:318)
	at io.vertx.test.core.Http1xTest.lambda$null$358(Http1xTest.java:3543)
	at io.vertx.core.impl.ContextImpl.lambda$wrapTask$2(ContextImpl.java:337)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:403)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:445)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
	at java.lang.Thread.run(Thread.java:745)
```

You get the NPE that I'm asserting and the test instead of being fast to execute will take a long time to complete, while waiting for the test timeout to occur.

So the `try` `catch` is asserting that in the future there won't be a regression. I can remove it but it will make the test suite to take longer to run.

vietj:
does the test pass ?

pmlopes:
yes, the exception above is running the test without the fix, if the fix is in place then it passes

PR #1989, opened 2017-05-16
AsyncFileImpl.writeInternal swallows op failures

### LINKED ISSUE #1967: Fix the CompletionHandler in AsyncFileImpl#writeInternal
```
      public void failed(Throwable exc, Object attachment) {
        if (exc instanceof Exception) {
          context.runOnContext((v) -> handler.handle(Future.succeededFuture()));
        } else {
          log.error("Error occurred", exc);
        }
      }
```

should be a failed future, a test should be added

tsegismont:
Fixes #1967


pom.xml
```diff
@@ -204,7 +204,12 @@
       <version>${tcnative.version}</version>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
```
vietj: we can't have this dependency added, otherwise it means we need to file a CQ for 3.4.2

tsegismont:
Ok. I just pushed a commit so that we only keep the fix.

2017-05-17 8:17 GMT+02:00 Julien Viet <notifications@github.com>:

vietj:
👍 let's keep things that way

PR #1994, opened 2017-05-22
Correct java generics in CompositeFuture.all

jan-zajic:
Method CompositeFuture.all accept List<Future>. Future is a raw type. References to generic type Future<T> should be parameterized. 

We can't currently call this method with any generic list without specific cast, or we are forced to use non generic list. It's ugly, produce warnings in every popular IDE, produce bugs in the final, so please merge this non breaking change.

### DIFF
--- a/src/main/java/io/vertx/core/CompositeFuture.java
+++ b/src/main/java/io/vertx/core/CompositeFuture.java
@@ -128,7 +128,7 @@ public interface CompositeFuture extends Future<CompositeFuture> {
    *
    * When the list is empty, the returned future will be already completed.
    */
-  static CompositeFuture any(List<Future> futures) {
+  static <X extends Future<?>> CompositeFuture any(List<X> futures) {
     return CompositeFutureImpl.any(futures.toArray(new Future[futures.size()]));
   }
 
@@ -178,7 +178,7 @@ public interface CompositeFuture extends Future<CompositeFuture> {
    *
    * When the list is empty, the returned future will be already completed.
    */
-  static CompositeFuture join(List<Future> futures) {
+  static <X extends Future<?>> CompositeFuture join(List<X> futures) {
     return CompositeFutureImpl.join(futures.toArray(new Future[futures.size()]));
   }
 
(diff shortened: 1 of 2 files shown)


juanavelez:
CompositeFuture.any(List<Future>) and CompositeFuture.join(List<Future>) methods could also be modified

jan-zajic:
@juanavelez, your're right, i added it :thumbsup:

vietj:
I don't understand how this can compile because wildcard extension are not supported by the code generator.

jan-zajic:
https://github.com/vert-x3/vertx-codegen/pull/123

vietj:
I would like to understand what is really necessary to make this work.

I would be fine with support generic type upper bound such as `<T extends Foo>` is that what is missing at the moment.

vietj:
In fact I told myself several times that it would be a good thing to support to express better APIs.

jan-zajic:
also found discussion here https://stackoverflow.com/questions/41467014/vert-x-java-listfutures-parameterization

vietj:
can you explain why ` <X extends Future<?>> CompositeFuture all(List<X> futures)` shouldn't be used over `CompositeFuture all(List<Future<?>>)` ?

jan-zajic:
:) because first variant of method accept both all(new ArrayList<Future>()) and all(new ArrayList<Future<?>>()) . Second variant is better, but breaks Backward compatibility. If method has form all(List<Future>) in past and everyone was forced to call it that way (ie with "half generic", generic list with non generic content), everyone must rewrite their code if we change it to all(new ArrayList<Future<?>>()). I think that only <X extends Future<?>> CompositeFuture all(List<X> futures)  method signature accept List<Future> and List<Future<?>> and List<Future<anything>>. (Or CompositeFuture all(List futures) but that one don't care if we past List with something else than Future).

vietj:
I'm afraid we could only support this form `CompositeFuture all(List<Future<?>>)` (in 3.5) for the moment for two reasons

- the changes required to support it are easier and smaller as the impact on other languages seems limited
- the version you propose seems convoluted and harder to read (usability)

So a change for this version should be possible using () and should be discussed and agreed on the dev list.

vietj:
until we revisit that later, I will close this PR

kdurnoga:
There's still a raw type in `CompositeFuture all(List<Future>)` which make this method barely usable. Any update on this issue?

PR #1995, opened 2017-05-23
Fix all places where stackTrace was discarded

jan-zajic:
log exception using correct method, no e.tostring as when you pass it to message, so user can see failure (for example having only exception message without original trace for npe, like i see in logs of our application when some developer made mistake, is very annoying)


seanf:
I was just about to make a very similar pull request, after an SSL exception with no stack trace got me. 

I chose slightly different messages of course. My version is here: https://github.com/eclipse/vert.x/compare/master...seanf:log-stack-traces (https://github.com/seanf/vert.x/commit/468dcea824b19f38d989571a6e43621f27add618)

@jan-zajic I found one extra log call which had been missing its stack trace, in `src/main/java/io/vertx/core/Starter.java`.

Issue #1996, opened 2017-05-26
Calling ServerWebSocket reject is documented to return a 404, but returns 502 instead


CraigStuntz:
`ServerWebSocket.reject()` is [documented](http://vertx.io/docs/apidocs/io/vertx/core/http/ServerWebSocket.html#reject--) to return a HTTP 404, but we observe that it returns a 502 instead, and [the source code](https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java#L829) indicates 502 as well: 

          if (!ws.isRejected()) {
            // ....
          } else {
            ch.writeAndFlush(new DefaultFullHttpResponse(HTTP_1_1, BAD_GATEWAY));
          }

404 seems better than 502 to me, but certainly the docs and code should agree, and I think ideally the `reject()` should provide an obvious means of setting the status code.

vietj:
I think reject could be overriden to set the status code indeed, I'm tried to look at the source code which one was originally there (the 502 or 404) but could not get past enough in the history to find out.

mithunsasidharan:
@vietj : Can I pick this up. I'm new to Vert.x and eager to contribute. Kindly advice.

vietj:
@mithunsasidharan sure, your first mission is to determine which is the proper code that should be used :-)

mithunsasidharan:
@vietj : I see that presently, if isRejected() is true, then we return

`ch.writeAndFlush(new DefaultFullHttpResponse(HTTP_1_1, BAD_GATEWAY));`

But HttpResponseStatus BAD_GATEWAY stands for 502 Bad Gateway. The right status to be used is NOT_FOUND to be able to return 404 as expected.

`ch.writeAndFlush(new DefaultFullHttpResponse(HTTP_1_1, NOT_FOUND));`

vietj:
@mithunsasidharan we need a test for this in the WebsocketTest class

CraigStuntz:
We should be able to specify the status code. I like the idea of adding an overload to `reject()`. Or a fluent method like [`setStatusCode`](http://vertx.io/docs/apidocs/io/vertx/core/http/HttpServerResponse.html#setStatusCode-int-) / `setStatusMessage`

vietj:
overloaded reject method looks good to me.

PR #2004, opened 2017-06-01
#1996 Fixed ServerWebSocket reject to return 404

mithunsasidharan:
Signed-off-by: Mithun Sasidharan <mithunsasidharan89@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -16,6 +16,21 @@
 
 package io.vertx.core.http.impl;
 
+import static io.netty.handler.codec.http.HttpResponseStatus.BAD_REQUEST;
+import static io.netty.handler.codec.http.HttpResponseStatus.METHOD_NOT_ALLOWED;
+import static io.netty.handler.codec.http.HttpResponseStatus.NOT_FOUND;
+import static io.netty.handler.codec.http.HttpResponseStatus.SWITCHING_PROTOCOLS;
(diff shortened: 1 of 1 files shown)


mithunsasidharan:
Also, imported the few HttpResponseStatus and single HttpVersion from netty handler instead of all (*)

mithunsasidharan:
@vietj : Can you please review/merge. Thanks.

pmlopes:
Not sure about this. During the handshake if the operation fails I think 502 is correct... 404 doesn't make sense to me since it's an connection handshake not the resource access.

vietj:
@pmlopes good point

vietj:
for HTTP/2 when a h2c upgrade fails, we return a 400 status code

vietj:
imho, I would keep 502 error. In node reject allow to specify a status code and a reject reason

vietj:
the Netty WebSocketServerProtocolHandler also uses 400 status code when it handles a rejected connection

mithunsasidharan:
@vietj : So do we need this change anymore then ?

vietj:
I think we should use 400 code, document it and test it.

it should also added in the list of breaking changes in the wiki.

vietj:
@mithunsasidharan any news for updating this PR ?

mithunsasidharan:
@vietj : will work on it over weekend !

vietj:
@mithunsasidharan there should be a test for this

vietj:
this has been superseded by https://github.com/eclipse/vert.x/pull/2066

PR #2008, opened 2017-06-04
Packet lost in Http1xOrHttp2Handler - fixes #2007

### LINKED ISSUE #2007: Packet lost in Http1xOrHttp2Handler
If we send an http request beginning with "P" (i.e. a POST, PATCH or PUT) split across two packets where the first packet has data length of 1 byte and the second packet contains the rest of the request, then the Http1xOrHttp2Handler reads the 1 byte packet, and then throws it away.

When the second packet comes in it does not aggregate the 1 byte that it previously read with the second packet, resulting in the method being received as "OST", "ATCH", or "UT".

A workaround is to disable the Http1xOrHttp2Handler with:
`System.setProperty("vertx.disableH2c", "true");`

Reproducer:
~~~~
  @Test
  public void test1BytePost() throws Exception {

    // The test will pass if we disable the http2 decoder:
    //  System.setProperty("vertx.disableH2c", "true");


    CountDownLatch latch1 = new CountDownLatch(1);
      assertEquals("POST", req.rawMethod());
      latch1.countDown();
    });

    String fullRequest = "POST /whatever HTTP/1.1\r\n\r\n";

    CountDownLatch latch2 = new CountDownLatch(1);
    server.listen(onSuccess(s -> {
      latch2.countDown();
    }));
    latch2.await();

    try (Socket sock = new Socket("localhost", 8080)) {

      byte[] rawData = fullRequest.getBytes();
      System.out.println("total Len = " + rawData.length);

      // send 1 byte packet
      sock.getOutputStream().write(rawData, 0, 1);
      sock.getOutputStream().flush();

      Thread.sleep(750);

      // send the rest
      sock.getOutputStream().write(rawData, 1, rawData.length - 1);
      sock.getOutputStream().flush();
    }

    latch1.await();

  }
~~~~

vietj:
Fixes #2007

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xOrH2CHandler.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xOrH2CHandler.java
@@ -1,0 +1,61 @@
+package io.vertx.core.http.impl;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+
+/**
+ * Handler that detects whether the HTTP/2 connection preface or just process the request
+ * with the HTTP 1.x pipeline to support H2C with prior knowledge, i.e a client that connects
+ * and uses HTTP/2 in clear text directly without an HTTP upgrade.
+ */
+public class Http1xOrH2CHandler extends ChannelInboundHandlerAdapter {
+
+  public static final String HTTP_2_PREFACE = "PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n";
+  private static final byte[] HTTP_2_PREFACE_ARRAY = HTTP_2_PREFACE.getBytes();
+  private int current = 0;
+
+  @Override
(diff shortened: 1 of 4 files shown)


vietj:
thanks for reporting !

src/main/java/io/vertx/core/http/impl/Http1xOrH2CHandler.java
```diff
@@ -0,0 +1,61 @@
...
+    if (current + i == HTTP_2_PREFACE_ARRAY.length) {
+      end(ctx, buf, true);
+    } else {
+      current += len;
+      buf.release();
+    }
+  }
+
+  private void end(ChannelHandlerContext ctx, ByteBuf buf, boolean h2c) {
+    if (current > 0) {
```
tsegismont: Is there any case when this method is called with `current == 0` ?

vietj: that's the most common case, it happens when you get a Buffer that contains enough data to make the decision the first time.

tsegismont: Ha yes, thanks.

PR #2016, opened 2017-06-09
Handle message that's throwable

bbyk:
If a throwable is passed as message with throwable param explicitly
set to `null` (the current behavior), effectively only t.getMessage() is
printed when message.toString() is called. Let's pass the message in
place of throwable if the message object is throwable.

### DIFF
--- a/src/main/java/io/vertx/core/logging/SLF4JLogDelegate.java
+++ b/src/main/java/io/vertx/core/logging/SLF4JLogDelegate.java
@@ -145,7 +145,13 @@ public class SLF4JLogDelegate implements LogDelegate {
   }
 
   private void log(int level, Object message) {
-    log(level, message, null);
+    if (message instanceof Throwable) {
+      // We often send a throwable inside the message and
+      // therefore we will lack the stacktrace in the rendered log entry
+      log(level, "", (Throwable) message);
+    } else {
+      log(level, message, null);
+    }
   }
 
   private void log(int level, Object message, Throwable t) {


vietj:
I think the user should always give a message in addition of the throwable rather, or the Logger method should be overloaded with a `Throwable` parameter.

bbyk:
That makes sense. There's about a dozen of calls like this

```
      log.error(t);
```

in the vert.x code itself (e.g. `io.vertx.core.net.impl.ConnectionBase#handleException`).

It's either a mistake or an established pattern to avoid somewhat expensive stacktrace rendering. But one better always render in the case of an exception.

I can update all the calls to be something like: `log.error("", t)` or `log.error("An exception thrown", t)` and undo the change in `SLF4JLogDelegate.java`.

vietj:
I think we should rather overload Logger to have the method `error(Throwable t)` that would do `error(t.getMessage(), t)` . Please take care to sign-off your commit and sign the Eclipse Contributor Agreement too, so your commit is valid.

bbyk:
That will break the behavior in the apps using vert.x logging after recompilation. The code `log.error(t);` will start printing full exception layout, and the maintainers will need to rewrite it to be `log.error((Object)t)` for the call to go through the previous code path.

I still maintain that we fix the problem in the vert.x codebase for those dozen instances. I can expand the call to `log.error(t.getMessage(), t)` as per your suggestion.

bbyk:
Julien, I signed. I suppose the message will disappear on the next push.

tsegismont:
2017-06-09 23:58 GMT+02:00 Boris Byk <notifications@github.com>:

> That will break the behavior in the apps using vert.x logging after
> recompilation. The code log.error(t); will start printing full exception
> layout, and the maintainers will need to rewrite it to be
> log.error((Object)t) for the call to go through the previous code path.
>

IMHO, users of the log.error(t) method expect it to do
log.error(t.getMessage(), t). They just have not noticed it doesn't.

So it should be fine.

vietj:
@tsegismont agreed with you, it looks more like correcting a programmer's mistake than changing the intended behavior or the programmer

vietj:
@bbyk can you create a new clean PR ?

bbyk:
I will. BTW, `log.error(t.getMessage(), t)` will most likly produce something like:

```
2017-06-12 [thread-0] com.company.MyLogger java.lang.UnsupportedOperationException: never gonna support

java.lang.UnsupportedOperationException: never gonna support
   at ...
   at ...
   at ...
```
Thus the message is going to be duplicated

PR #2017, opened 2017-06-11
Fix race condition with writesOutstanding when is used from non Vert.…

### LINKED ISSUE #2012: Race condition in AsyncFileImpl writesOutstanding
If an AsyncFileImpl is written to from a thread different to the context that created then the long counter writesOutstanding can get into an invalid state. This is because the counter is updated from both the caller's thread and also on the context when a write is complete.

Fix is to use an AtomicLong for writesOutstanding:

https://gist.github.com/purplefox/723fa69a1330fb4f4f74d4a5dd5e6658

vietj:
…x thread - fixes #2012


vietj:
one of you or anyone else is fine for review

tsegismont:
Looking

PR #2021, opened 2017-06-16
Expose AsyncMap keys, values and entries

### LINKED ISSUE #2020: Expose AsyncMap keys, values and entries
Currently, AsyncMap lets you get/put/remove/replace entries asynchronously.

In some cases, you also need the ability to get the keys, values or entries (see @cescoffier 's comment in https://github.com/vert-x3/vertx-infinispan/issues/35#issuecomment-308135648)

tsegismont:
Fixes #2020 

Expose AsyncMap keys, values and entries


tsegismont:
Please do not merge yet

src/main/java/io/vertx/core/shareddata/AsyncMap.java
```diff
@@ -134,4 +141,84 @@
    */
   void size(Handler<AsyncResult<Integer>> resultHandler);
 
+  /**
+   * Get the keys of the map, asynchronously.
+   * <p>
+   * Use this method with care as the map may contain a huge number of entries.
```
vietj: huge -> large, that may not fit entirely in memory.

src/main/java/io/vertx/core/shareddata/AsyncMapStream.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.Closeable;
+import io.vertx.core.Handler;
+import io.vertx.core.streams.ReadStream;
+
+/**
+ * An {@link AsyncMap} keys, values, or entries {@link ReadStream stream}.
+ * <p>
+ * Make sure to invoke {@link #close(Handler)} after usage to avoid leaking resources.
```
vietj: should `close()` be used when the stream is completed ?

tsegismont: Yes, either ended or failed.

src/main/java/io/vertx/core/shareddata/AsyncMap.java
tsegismont: Done

src/main/java/io/vertx/core/shareddata/AsyncMapStream.java
vietj: I think that close should not have to be used when the stream is ended or failed. It should rather be used when the user wants to dispose the stream prematurely (for instance there is an error during the iteration and the developer want to release the sream resources).

tsegismont: I'm not sure I understand. Do you mean close should be automatically invoked if the stream fails (before `exceptionHandler` is called) or ends (before `endHandler` is called)? And the user should have to call it only if he wants to end things prematurely (because he doesn't need the rest of the data or an error happened downstream)?

vietj: yes

src/main/java/io/vertx/core/shareddata/AsyncMapStream.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * An {@link AsyncMap} keys, values, or entries {@link ReadStream stream}.
+ * <p>
+ * The stream will be automatically closed if it fails or ends.
+ * Otherwise you must invoke {@link #close(Handler)} after usage to avoid leaking resources.
+ *
+ * @author Thomas Segismont
+ */
+@VertxGen
+public interface AsyncMapStream<T> extends ReadStream<T>, Closeable {
```
vietj: This should rather be named `CollectionStream` or something in that spirit I think as it is used for a collection of key/value or entries from the `AsyncMap`.

src/main/java/io/vertx/core/shareddata/AsyncMapStream.java
tsegismont: Done, including cluster managers updates

src/main/java/io/vertx/core/shareddata/AsyncMapStream.java
tsegismont: You're right about the usage (it's explained in the Javadoc). To be honest I also find that `AsyncMapStream` is not self-explanatory. But neither does `CollectionStream` in my opinion. If you find anything better let me know. Otherwise I'll update this one and the cluster managers PRs

vietj: my point is that this stream could be used for other purpose than `AsyncMapStream` and if we name it `AsyncMapStream` it will feel awkward if we reuse it.

vietj: `DataStream` might be a good candidate

tsegismont: Ha I see. Then how about `CloseableReadstream` ? Isn't a stream always a data stream?

vietj: Actually I've given thoughts about it and we should instead use plain `ReadStream` and use the `handler(Handler<T>)` to control this like we do in other cases:

- when the handler is set to a value : start to stream the data from the beginning
- when the handler is set to `null` : release the resource

this is how other streams work and how it will work best with RxJava

vietj: I believe we shall add a `close` operation on `ReadStream` in the future to improve this contract (after 3.5.0).

tsegismont: So IIUC, setting a null handler would close the stream? Can you elaborate on how this will work with RxJava?

tsegismont: >Setting a null handler would close the stream and setting a non null handler would start it (lazy start).

OK, thanks for clarifying.

>When a ReadStream is adapted to an Observable its handlers are set at subscription and unset at subscription cancellation. You can look at the implementation of MessageConsumer that works this way.

So that will work because the stream implementations in cluster managers don't create the resources until you set the data handler.

Alright, I'll switch back to a plain read stream, update the Javadoc and cluster manager implementations.

tsegismont: Done

src/main/java/io/vertx/core/shareddata/AsyncMap.java
```diff
@@ -134,4 +141,90 @@
...
   void size(Handler<AsyncResult<Integer>> resultHandler);
 
+  /**
+   * Get the keys of the map, asynchronously.
+   * <p>
+   * Use this method with care as the map may contain a large number of keys,
+   * which may not fit entirely in memory of a single node.
+   * In this case, the invocation will result in an {@link OutOfMemoryError}.
+   * <p>
+   * When in doubt, prefer usage of {@link #keyStream()}.
```
vietj: that's not really true since Hazelcast all the keys are loaded right ?

tsegismont: Yes but the `xxxStream` methods have a warning that not all cluster manager support true streaming.

vietj: I'm worried that Hazelcast does not provide streaming yet for this, so it would only be effective with Infinispan which is today a very minority of the users, so I would like for now to remove streaming and defer until we can support it in the general case, otherwise we are going to have people running into issues and we will have no solution for them.

tsegismont: Not only Infinispan, Ignite too. Anyway I went back to Hazelcast doc and I believe we could do with queries and a `PagingPredicate`. Let me update the HZ cluster manager PR.

tsegismont: @vietj as discussed offline, I talked with Hazelcast guys and PagingPredicates are not a good solution: they require a fullscan on every iteration.
So I went ahead and removed the streaming methods from the generic API as well as Ignite, Hazelcast and Zookeeper cluster managers.
In Infinispan I added a new interface for extra features. The streaming features are maintained there.

PR #2023, opened 2017-06-19
Remove LICENSE file with MIT license text.

jpopelka:
[io/vertx/core/http/impl/cgbystrom/](https://github.com/eclipse/vert.x/tree/master/src/main/java/io/vertx/core/http/impl/cgbystrom) contains a [LICENSE](https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/cgbystrom/LICENSE) file with MIT License text.
I can't find any file under this license, so I guess the LICENSE file can be safely removed ?


msrb:
I think the file licensed under MIT is https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/cgbystrom/FlashPolicyHandler.java#L45, borrowed from https://github.com/cgbystrom/netty-tools.

jpopelka:
👍 @msrb

jpopelka:
Opened #2043

PR #2029, opened 2017-06-21
Http2Settings.hashCode() does not conform to contract

### LINKED ISSUE #2026: Http2Settings.hashCode() does not conform to contract.
Vertx Version -- 3.4.2

The implementation of `Http2Settings.hashCode()` does not conform to the contract where if 2 instances return `true` from `equals` they must have the same hash code value. 

The primary issue is that it uses `Object.hashCode()` as its base value.  It also does not include a hash of `headerTableSize` in its value (this property is checked by `equals`), though strictly speaking the contract does not require this.

The following groovy script shows the issue (the second assert fails):

``` groovy
import io.vertx.core.http.Http2Settings

def settings1 = new Http2Settings()
def settings2 = new Http2Settings()
assert settings1.equals(settings2)
assert settings1.hashCode() == settings2.hashCode()
```

tsegismont:
Fixes #2026

### DIFF
--- a/src/main/java/io/vertx/core/http/Http2Settings.java
+++ b/src/main/java/io/vertx/core/http/Http2Settings.java
@@ -345,14 +345,13 @@ public class Http2Settings {
 
   @Override
   public int hashCode() {
-    long result = super.hashCode();
+    int result = (int) (headerTableSize ^ (headerTableSize >>> 32));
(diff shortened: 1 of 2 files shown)


vietj:
it should have a test

tsegismont:
Done

2017-06-21 18:31 GMT+02:00 Julien Viet <notifications@github.com>:

PR #2033, opened 2017-06-26
Perf improvements

vietj:
Performance improvements for `HttpServer`, the improvements have been tested in a real environment (provided by Red Hat). Here are results of the Techempower plaintext benchmark:

|                     |3.4.0    |   3.5.0-SNAPSHOT|
|---|---|---|
|plain/256     | 3489111 | 3624167 |
|plain/1024   | 3571519 | 4027825 |
|plain/4096  | 3468908  | 3955497 |
|plain/16384 | 3296358  | 3779672 |

- Simplified the `HttpServer` handlers to combine the request and websocket handlers context that helps to simplify the internal code and anyway one would not use two different event loops for these.
- Decouple the Vert.x Netty handler that handles the `HttpRequest`/`HttpResponse` events so it can be created independantly of an HttpServer instance and can be benchmarked with a Netty embedded event loop to measure the overhead of Vert.x
- set the `ConnectionBase` on `VertxHandler` when the handler is added to the pipeline avoiding to handle the null case in the read pipeline.
- now the various server handlers write through the `ChannelHandlerContext` instead of the `Channel` to bypass the `Channel`->`ChannelHandlerContext` that is not necessary (http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#29.0)
- use a `VoidPromise` in pipeline write operations when possible (http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#8.0)
- metrics SPI instances are considered as disabled when null is returned from the SPI. Previously it was checking for null / enabled boolean
- use optimized `HttpHeaders` for `HttpServerResponse` that implements `Multimap`
- in `HttpServerResponse` avoid to double check the response headers like _content-length_
- `ServerConnection` has been made JIT friendly when possible
- JMH micro benchmarks : a few JMH micro benchmarks have been added that helps to measure the impact of changes on the `HttpServer` request handler. Note that these benchmarks have been added to quickly see the impact of changes but do not replace real benchmarking done in a lab.

The JMH benchmarks results:

```
Benchmark                             Mode  Cnt       Score       Error   Units
HeadersContainsBenchmark.nettySmall  thrpt   10  211541.865 ± 15864.149  ops/ms
HeadersContainsBenchmark.vertxSmall  thrpt   10  279507.048 ± 28970.228  ops/ms
HeadersEncodeBenchmark.baseline      thrpt   10  254731.860 ±  6419.979  ops/ms
HeadersEncodeBenchmark.nettySmall    thrpt   10    8159.088 ±    72.889  ops/ms
HeadersEncodeBenchmark.vertxSmall    thrpt   10    9804.617 ±   199.194  ops/ms
HeadersSetBenchmark.nettySmall       thrpt   10    4883.103 ±    96.648  ops/ms
HeadersSetBenchmark.vertxSmall       thrpt   10   23627.110 ±   327.857  ops/ms
HttpServerHandlerBenchmark.netty     thrpt    10  916.701 ± 1338.710  ops/ms
HttpServerHandlerBenchmark.vertx     thrpt    10  613.475 ±   68.536  ops/ms
HttpServerHandlerBenchmark.vertxOpt  thrpt    10  678.481 ±  467.255  ops/ms
RunOnContextBenchmark.baseline       thrpt   10   24237.869 ±   648.594  ops/ms
RunOnContextBenchmark.noChecks       thrpt   10  193615.947 ± 15562.721  ops/ms
```

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ServerConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/ServerConnection.java
@@ -17,12 +17,23 @@
 package io.vertx.core.http.impl;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.channel.*;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.FileRegion;
 import io.netty.handler.codec.DecoderResult;
 import io.netty.handler.codec.TooLongFrameException;
-import io.netty.handler.codec.http.*;
-import io.netty.handler.codec.http.HttpHeaders;
+import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.HttpContent;
+import io.netty.handler.codec.http.HttpObject;
+import io.netty.handler.codec.http.HttpRequest;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.HttpVersion;
+import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketHandshakeException;
 import io.netty.handler.codec.http.websocketx.WebSocketServerHandshaker;
@@ -34,7 +45,12 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
-import io.vertx.core.http.*;
+import io.vertx.core.http.GoAway;
+import io.vertx.core.http.Http2Settings;
+import io.vertx.core.http.HttpConnection;
+import io.vertx.core.http.HttpServerRequest;
+import io.vertx.core.http.ServerWebSocket;
+import io.vertx.core.http.WebSocketFrame;
 import io.vertx.core.http.impl.ws.WebSocketFrameInternal;
 import io.vertx.core.impl.ContextImpl;
 import io.vertx.core.impl.VertxInternal;
@@ -69,6 +85,8 @@ class ServerConnection extends ConnectionBase implements HttpConnection {
 
   private static final Logger log = LoggerFactory.getLogger(ServerConnection.class);
 
+  private static final Handler<HttpServerRequest> NULL_REQUEST_HANDLER = req -> {};
+
   private static final int CHANNEL_PAUSE_QUEUE_SIZE = 5;
 
   private final Queue<Object> pending = new ArrayDeque<>(8);
@@ -78,7 +96,7 @@ class ServerConnection extends ConnectionBase implements HttpConnection {
   private final HttpServerMetrics metrics;
   private boolean requestFailed;
   private Object requestMetric;
-  private Handler<HttpServerRequest> requestHandler;
+  private Handler<HttpServerRequest> requestHandler = NULL_REQUEST_HANDLER;
   private Handler<ServerWebSocket> wsHandler;
   private HttpServerRequestImpl currentRequest;
   private HttpServerResponseImpl pendingResponse;
@@ -275,17 +293,6 @@ class ServerConnection extends ConnectionBase implements HttpConnection {
     return socket;
   }
 
-  private void handleRequest(HttpServerRequestImpl req, HttpServerResponseImpl resp) {
-    this.currentRequest = req;
-    pendingResponse = resp;
-    if (metrics.isEnabled()) {
-      requestMetric = metrics.requestBegin(metric(), req);
-    }
-    if (requestHandler != null) {
-      requestHandler.handle(req);
-    }
-  }
-
   private void handleChunk(Buffer chunk) {
     if (metrics.isEnabled()) {
       bytesRead += chunk.length();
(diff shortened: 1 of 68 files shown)


tsegismont:
On my laptop, the vertx-web fails on SockJS test with these changes, passes without

src/test/java/io/vertx/test/core/Http1xTest.java
```diff
@@ -1169,6 +1169,7 @@ public void testPipeliningOrder() throws Exception {
     AtomicInteger reqCount = new AtomicInteger(0);
     server.requestHandler(req -> {
       int theCount = reqCount.get();
+//      System.out.println("begin " + theCount);
```
tsegismont: Remove

src/test/java/io/vertx/test/core/Http1xTest.java
```diff
@@ -1180,6 +1181,7 @@ public void testPipeliningOrder() throws Exception {
           req.response().headers().set("count", String.valueOf(theCount));
           req.response().write(buff);
           req.response().end();
+//          System.out.println("end " + theCount);
```
tsegismont: Remove

src/main/java/io/vertx/core/http/impl/ServerConnection.java
```diff
@@ -494,6 +494,9 @@ private void handleLastHttpContent() {
     } else {
       // Requeue
       // paused = true => queueing = true
+      // todo : this should be added first if pending.size() > 0
```
tsegismont: Has this been fixed in a later commit?

src/main/java/io/vertx/core/http/impl/headers/VertxHttpHeaders.java
```diff
@@ -0,0 +1,479 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Consumer;
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+public class VertxHttpHeaders extends HttpHeaders implements MultiMap {
```
vietj: I can use the existing multimap test to run on VertxHttpHeaders

vietj: it's slower

src/main/java/io/vertx/core/spi/metrics/Metrics.java
```diff
@@ -24,11 +24,16 @@
...
+  String DISABLE_METRICS_PROPERTY_NAME = "vertx.disabledMetrics";
+
+  boolean METRICS_ENABLED = !Boolean.getBoolean(DISABLE_METRICS_PROPERTY_NAME);
+
   /**
    * Whether the metrics are enabled.
    *
    * @return true if the metrics are enabled.
    */
+  @Deprecated
```
vietj: no because it can be disabled if metrics == null

src/main/java/io/vertx/core/spi/metrics/VertxMetrics.java
```diff
@@ -101,7 +101,7 @@
    * @param server       the Vert.x http server
    * @param localAddress localAddress the local address the net socket is listening on
    * @param options      the options used to create the {@link io.vertx.core.http.HttpServer}
-   * @return the http server metrics SPI
+   * @return the http server metrics SPI or null when metrics are disabled
    */
```
vietj: yes

vietj:
I will have a look at the sock-js test

src/main/java/io/vertx/core/spi/metrics/Metrics.java
vietj: ok

vietj:
I've addressed the changes, the failing SockJS test was due to an edge case (setting a chunked response with `HTTP_1_0` that set the _content-length_ header to `0`), a test has been added to specify this behavior.

tsegismont:
I had a look at https://www.w3.org/Protocols/HTTP/1.0/spec.html and my
understanding is that when a content-length is present then it indicates
the size of the body. When the server can't compute the length, it does not
send it and the client just waits for the connection closing.

2017-06-28 23:48 GMT+02:00 Julien Viet <notifications@github.com>:

tsegismont:
Did you have a look at the other questions about metrics changes?

vietj:
@tsegismont ok for using HTTP/1.0 behavior, so it means we need to update the SockJS implementation to test if that's an HTTP/1.0 or HTTP/1.1 to behave differently.

vietj:
@tsegismont I changed the behavior to not set a _Content-Length_ header when sending a chunked HTTP/1.0 response and I updated vertx-web to set an explicit _Content-Length: 0_ header for SockJS chunked streams

vietj:
https://github.com/vert-x3/vertx-web/tree/perf-improvements

src/test/benchmarks/io/vertx/benchmarks/HeadersSetBenchmark.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Benchmark
+  public void nettySmall(Blackhole blackhole) throws Exception {
+    nettySmallHeaders.clear();
+    setBaseHeaders(nettySmallHeaders);
+    blackhole.consume(nettySmallHeaders);
+  }
+
+  @Benchmark
+  public void vertxSmall(Blackhole blackhole) throws Exception {
```
jponge: As a matter of style since you do just _1_ `Blackhole` call you should have your benchmark methods return the object to be thrown away.

jponge: Same in other benchmark classes BTW

vietj:
here is the branch for mqtt-server https://github.com/vert-x3/vertx-mqtt-server/tree/perf-improvements

src/test/benchmarks/io/vertx/benchmarks/HeadersSetBenchmark.java
vietj: ok

vietj:
@jponge I've updated benchmarks to replace Blackhole either by a return or a non inlined method to consume the buffer when the benchmark does not produce a value and instead updates a buffer. I've used a dedicated netty allocator to reuse always the same buffer during the test

PR #2040, opened 2017-06-30
SNI: tie trusted CA's to server names (WIP)

theoweiss:
This patch adds the possibility to tie trusted CA's to server names when TLS with SNI and client auth is used.

The PR is not yet ready but I open it to discuss the implementation details. Documentation, more tests and an implementation for PfxOptions are still missing.
I've added some methods to JksOptions, PEMTrustOptions and (not yet) PfxOptions which only make sense in a sni setup as server side trust options. But these types were also used in totally different contexts (client side trust, KeyCertOptions, no sni in use, ...), therefore I'm not sure if it is better to subclass these Types?


theoweiss:
Hi @vietj,

could you give me a short feedback if it's worth for me to put further work into this PR or if my approach is going into the totally wrong direction.

Thanks,
Theo

src/main/java/io/vertx/core/net/JksOptions.java
```diff
@@ -81,8 +88,59 @@ public JksOptions(JksOptions other) {
    * @param json  the JSON
    */
   public JksOptions(JsonObject json) {
-    super();
-    JksOptionsConverter.fromJson(json, this);
+    this();
```
vietj: you need to use converter for this

src/main/java/io/vertx/core/net/JksOptions.java
```diff
@@ -92,7 +150,42 @@ public JksOptions(JsonObject json) {
    */
   public JsonObject toJson() {
     JsonObject json = new JsonObject();
-    JksOptionsConverter.toJson(this, json);
+    if (path != null){
```
vietj: you need to use converter for this

vietj:
@theoweiss rather than commenting on the PR (which changes a lot the json conversion code to not use converter for which I commented), can we discuss what are the overall changes you want to do ?

theoweiss:
@vietj Yes, I'll try to sum up my changes. But I'm not sure if I have enough time today, at least I'll be back here at the beginning of the next week.

theoweiss:
# The use case

## Basic use case

- Net/Http-Server with SNI enabled and client authentication enabled
- there a two CA's available: "A-CA" and "B-CA"
- the server uses keyCert "A-KeyCert" from CA "A-CA" for server name "A-Name"
- the server uses keyCert "B-KeyCert" from CA "B-CA" for server name "B-Name"
- for indicated name "A-Name" only clients are trusted which provide a client certificate from "A-CA"
- for indicated name "B-Name" only clients are trusted which provide a client certificate from "B-CA"

## Advanced use case

- there is a third CA "B2-CA"
- for indicated name "B-Name" clients are trusted which provide a client certificate from "B-CA" or "B2-CA"

# The implementation

## TrustOptions: Mapping trusted CAs to server names

- To tie trusted CAs to server names, implementations of TrustOptions (for all the different types PEM, JKS and Pfx) are needed which can assign trusted CAs to server names. Therefore I've added PemTrustOptions.addCertPathForName(certpath, servername) (the same for addCertValues) and JksOptions.addPathForName(path, password, servername). Internally, the TrustOptions store the mapping serverName->ListOfTrustedCAs in a Map. There are also getters for the ListOfTrustedCAs for a given serverName (PemTrustOptions.getCertPathForName(serverName), PemTrustOptions.getCertValuesForName(serverName), JksOptions.getPathForName() + JksOptions.getPasswordForName()).
- TrustOptions got a new method for getting the trustManagerMapper.
- I've also added getters for the (internal) Maps, they are only needed for the copy constructor. Now I think they should be better removed and be replaced with code which iterates over the server names.
- I've removed the converter because I've been running into problems with the code generation.
- Maybe it is a better idea to add new TrustOptions types with SNI support than extending the existing types?

##  KeyStoreHelper

- I've added a TrustManagerMap to the KeyStoreHelper which holds the mapping serverName -> KeyStore after "resolving" the TrustOptions.

## SSLHelper

- Now getContext(vertx, serverName) is also choosing the right TrustManagerFactory dependent on the serverName and is handing it over to the createContext method.

vietj:
it looks valid to me, however I would like to avoid to have a Map<ServerName, Value> in the options and instead rely on alias like it is done for the keystores. Eventually someone that wants something more flexible can implement its own `TrustOptions`.

theoweiss:
> rely on alias

Do you think of using the SAN and maybe the Subjects Common Name field of the CA certificate or do you mean the alias which is only available for JKS truststores?

vietj:
I meant CN name sorry for the confusion. The actual solution for Keystore works for JKS but also uses a list of PEM support

theoweiss:
The Keystore solution for the sni (server) certKeys handles server certificates where Subject Alternative Names are a common feature. Therefore I think here it is a good approach to use the SANs as sni names.
In contrast the Truststore handles CA certificates where SANs are very unusual. I think in most cases (I've also talked to our security experts) it would be impossible to convince a CA Manager to issue such CA certificates. I fear an implementation which relies on the SANs would only serve some corner cases but not the common use cases. I think we couldn't avoid to have an explicit configuration sniName->[CaCert,...] in one way or another.
What do you think?

vietj:
how about using aliases in keystore ?

theoweiss:
Technically I think that would be possible but only for JKS truststores. PEM and Pfx TrustOptions wouldn't be available as SNI aware TrustOptions. But I worry that it would break the "principle of least surprise".

theoweiss:
@vietj Any suggestions how to proceed?

vietj:
we need to come up with a good configuration way, the current way to use Map inside the option feels confusing to me, specially that for keystore we use a different way to map the SNI aliases.

one option would be to support using aliases name.

another way would be to manage it at the TCPSSLOptions level with `Map<String, TrustOptions>`.

theoweiss:
Ok, I will try an implementation based on the alias names. The benefit would be that no changes to the TrustOption interface are needed, similar to the KeyStore solution.

vietj:
keep in mind also that alternatives implementations are possible (even dynamic) using the SPI method in the option interface, somebody has imoplemented already dynamic keystore SNI based mapping this way (using the ACME protocol).

theoweiss:
@vietj I've started to rework this PR but I run into problems with the current NetTest "framework". 

* if shouldPass is set to false (TLSTest::test.run(false)) but the run is nevertheless successful, the test does not complete but is running into a Timeout. It seems that the latch has yet a value of 1 after AsyncTestBase.fail() is called. You can reproduce this by changing shouldPass from true to false in testSniWithServerNameStartTLS test.

* I faced another problem: if the NetClient connect call fails under certain circumtances, a test will always report success regardless if shouldPass is set to true or false. You can reproduce this with this test:

```
  @Test
  public void testTLSClientCertClientNotTrustedX() throws Exception {
    Trust<PemTrustOptions> SERVER_PEM_OTHER_CA = () -> new PemTrustOptions().addCertPath("tls/other-ca/ca-cert.pem");
    testTLS(Cert.NONE, Trust.SERVER_JKS, Cert.SERVER_JKS, SERVER_PEM_OTHER_CA, true, false, true, false);
  }
```

theoweiss:
I've opened a new PR for this feature #2147

PR #2047, opened 2017-07-11
WorkerExecutor should use the right executor when the context queue i…

### LINKED ISSUE #2046: WorkerExecutor should use the right executor when the context queue is not empty

vietj:
…s not empty - fixes #2046


tsegismont:
LGTM

PR #2052, opened 2017-07-14
Upgrade Jackson to version 2.8.9

vjkoskela:


src/main/java/io/vertx/core/json/Json.java
```diff
@@ -166,7 +167,7 @@ public static String encodePrettily(Object obj) throws EncodeException {
    */
   public static <T> T decodeValue(Buffer buf, Class<T> clazz) throws DecodeException {
     try {
-      return mapper.readValue(new ByteBufInputStream(buf.getByteBuf()), clazz);
+      return mapper.readValue((InputStream) new ByteBufInputStream(buf.getByteBuf()), clazz);
```
InfoSec812: Why would you want/need to cast this to InputStream. Doesn't ByteBufInputStream IMPLEMENT InputStream?

vjkoskela: The `ByteBufInputStream` is both a `InputStream` and a `DataReader`. Jackson `2.8.x` introduces overloads into `ObjectMapper` for `DataReader`. Therefore, this call is now ambiguous and the cast is needed to indicate which method to invoke. Alternatively, you could take a look variable as type `InputStream` (not `ByteBufInputStream`) to achieve the same result.

vietj:
we will rather upgrade to 2.9 when it's released as it contains a feature we would like to have for 3.5

vjkoskela:
That's fair. We're in a similar position using 2.8.x versions 2.7.x and thus have a package conflict with Vert.x 3.4. We can't force the upgrade because of the call ambiguity identified in this issue. However, if 3.5 will move to 2.9.x we can converge on that and just cut an internal release with this fix in the meantime assuming those test failures are not serious in your opinion.

vietj:
Upgraded to 2.9.0

PR #2053, opened 2017-07-14
Fix #handleEnd to respect paused

bbyk:
`io.vertx.core.http.impl.HttpClientResponseImpl#handleEnd` calls
`handleChunk` whose handler can pause the response, thus signaling it's
not ready to process further including the end of the response. With the
change the `paused` status is now examined twice.

Signed-off-by: Boris Byk <boris.byk@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
@@ -243,15 +243,20 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
     synchronized (conn) {
       conn.reportBytesRead(bytesRead);
       bytesRead = 0;
-      if (!paused && lastChunk != null) {
-        handleChunk(lastChunk);
-      }
-      // paused can be updated in the #handleChunk call above, check it again
       if (paused) {
         pausedLastChunk = lastChunk;
         hasPausedEnd = true;
         pausedTrailers = trailers;
       } else {
+        if (lastChunk != null) {
+          handleChunk(lastChunk);
+          if (paused) {
+            pausedLastChunk = null;
+            hasPausedEnd = true;
+            pausedTrailers = trailers;
+            return;
+          }
+        }
         this.trailers = trailers;
         if (endHandler != null) {
           try {
--- a/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
@@ -243,14 +243,15 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
     synchronized (conn) {
       conn.reportBytesRead(bytesRead);
       bytesRead = 0;
+      if (!paused && lastChunk != null) {
+        handleChunk(lastChunk);
+      }
+      // paused can be updated in the #handleChunk call above, check it again
       if (paused) {
         pausedLastChunk = lastChunk;
         hasPausedEnd = true;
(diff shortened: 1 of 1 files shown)


vietj:
can you provide a test for this ?

bbyk:
@vietj Julien, what kind of test do you have in mind? The http client barely have a single test for settings. I don't know what example to follow. A test that doesn't mock is hardly possible because the existence of `lastChunk` may not be deterministic. As for mocking – I don't see any foundation for mocking in vert.x.

If you have any idea I would be happy to throw something in. I hope you agree that there was a bug there.

vietj:
I would like to have a case that reproduce the issue that this commit fixes

bbyk:
Will a mocking test be enough? I mean, I will create a mocked instance of the response, set the handler that pauses the response and then I will call the code modified in the PR, and call for a `endHandler` will throw an exception.

Does it make sense?

bbyk:
@vietj Julien, there you go: https://github.com/bbyk/verxpull2053

As I mentioned before it may not be exactly deterministic and can depend on OS. I can reproduce the exception most of the time when I run the test `bbyk.VertxHttpClientTests#endHandlerCalledOutOfOrder` from a debugger on my

```
MacOS Sierra 10.12.5 (16F73)
MacBook Pro (Retina, 15-inch, Mid 2014)
```

vietj:
I can see your patched has changed and it handles the case where the stream has been paused during the delivery of the last chunk.

In this case, why should we not call the `endHandler` at this moment ? All the chunks have been delivered and there is no good reason not to call the `endHandler`, can you elaborate ?

bbyk:
> I can see your patched has changed and it handles the case where the stream has been paused during the delivery of the last chunk.

Well, my changes handled this case from the very first commit on the PR. I am glad it is clear now, though. My last patch was to reset `pausedLastChunk = null;` after the chunk is handled because `handleEnd` is going to be called again in `#resume`.

> In this case, why should we not call the endHandler at this moment ? All the chunks have been delivered and there is no good reason not to call the endHandler, can you elaborate ?

Why does the class have `#pause` and `#resume`? I figured it's for the case when the handler can't completely munch the chunk during the delivery time due to scheduling an async op on it e.g. saving the data to DB, file etc. I hope we're on the same page here. Alright, so the last chunk is no excuse as you can see from my code here: https://github.com/bbyk/verxpull2053/blob/master/src/test/java/bbyk/VertxHttpClientTests.java#L89 . There is a good reason not to call the `endHandler` immediately after the last chunk because the processing of the chunk happens async. Therefore, if `endHandler` was called immediately the last chunk would not be yet fully processed. The `lineBuilder` in my code would have a few lines not that were not processed yet.

I will demonstrate. For example, if the last chunk was (I deliberately omitted \n on the last line and specified `iting` as partial) and `lineBuilder` in my code was `wr`:
```
iting\n
Java\n
is\n
fun
```
`handleEnd` calls `handleChunk` and that pauses the stream and extracts the first line which is going to be `writing`, right? Now, I am scheduling an async op to save it in db – well, I emulate it via `setTimer` anyway. The next line, that is `Java` won't be extracted until the completion of the async save. If I called the `endHandled` here, the `lineBuilder` would not be fully processed, would it? In real world scenario, I may be going to need to close the connection to db in `endHandler` which I can't in the case. In my example, I have yet another reason - not all lines are separated by \n. `handleChunk` doesn't know whether `fun` is the last or it's just the `\n` will come in the next hunk. Only `endHandler` knows. That's why whatever is left in `lineBuilder` needs to be processed in `endHandler` too.

PR #2056, opened 2017-07-17
Build cleanups

jponge:
A bunch of build cleanups:

* many dependencies already have their versions set in the BOM, so repeating is bad
* simplified the usage of maven-source-plugin
* fixed a few execution ids
* properly clean {code,doc}gen files in main + test sources
* removed an unused reporting plugin.


vietj:
generated converters should not be removed from SCM

PR #2064, opened 2017-07-20
HttpServerRequest exception handler is not called when conn closed

### LINKED ISSUE #2027: HttpServerRequest exception handler is not called when the connection is closed
`HttpServerRequest` extends `ReadStream<Buffer>`. When there is an ongoing file upload and the connection closes abruptly, the exception handler on the `HttpServerRequest` should be called to signal that the stream an error occured

vietj:
fixes #2027


vietj:
it's a small PR, quick to review

PR #2066, opened 2017-07-21
A few improvements in websocket rejection: the ServerWebSocket#reject…

### LINKED ISSUE #1996: Calling ServerWebSocket reject is documented to return a 404, but returns 502 instead

vietj:
…() is overloaded to provide a specific status code, the reject() method is documented that it will send 502 by default and not 404, the client can be aware of the rejection status with the new WebsocketRejectedException. fixes #1996

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ServerWebSocketImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/ServerWebSocketImpl.java
@@ -16,6 +16,7 @@
 
 package io.vertx.core.http.impl;
 
+import io.netty.handler.codec.http.HttpResponseStatus;
 import io.vertx.core.MultiMap;
 import io.vertx.core.http.ServerWebSocket;
 import io.vertx.core.http.WebSocketFrame;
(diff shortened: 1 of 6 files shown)


src/main/java/io/vertx/core/http/impl/ServerWebSocketImpl.java
```diff
@@ -135,15 +149,18 @@ private void connect() {
...
         connect();
       }
     }
   }
 
-  boolean isRejected() {
-    synchronized (conn) {
-      return rejected;
-    }
+  synchronized boolean isRejected() {
```
jponge: `synchronized(conn)` rather than `synchronized(this)` is likely what you expected here

vietj: hum right good catch thanks

vietj:
the wrong monitor synchronization has been fixed @jponge

PR #2069, opened 2017-07-23
A couple of fixes on http exception handler called on close - fixes #…

vietj:
…2068


src/main/java/io/vertx/core/http/impl/ClientConnection.java
```diff
@@ -414,7 +414,9 @@ protected synchronized void handleClosed() {
 
     // Connection was closed - call exception handlers for any requests in the pipeline or one being currently written
```
tsegismont: Shouldn't the comment be updated as well ? The request being currently written will not have its exception handler invoked

vietj: yes it will have with the next statement:

```
if (currentRequest != null) {
  currentRequest.handleException(e);
}
```

it was just called twice in this case.

tsegismont: Oh, right!

PR #2076, opened 2017-07-27
Remove ExtendedClusterManager interface

### LINKED ISSUE #1277: remove hazelcast workaround (or at least the comment, that indicate it as a workaround)
Hi!

Issue https://github.com/hazelcast/hazelcast/issues/5220 had been closed in October, we should upgrade to the latest stable build 3.5.4 (http://hazelcast.org/download/ ) .

see https://github.com/vert-x3/vertx-hazelcast/blame/master/pom.xml#L44

and then we can remove the workaround here:

https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/impl/VertxImpl.java#L431

kind regards
Gabriel

tsegismont:
Fixes #1277


vietj:
can you add a note about the deprecation, i.e what the developer should be aware of to migrate from this interface, also please specify when (version) the interface will be removed

tsegismont:
Not sure what to add. The Javadoc already as a comment from @purplefox indicating that this interface was introduced as a workaround for a Hazelcast bug. I don't believe any other cluster manager implementation than vertx-hazelcast uses it.

tsegismont:
As for when it will be removed, it should be removed in 3.6, right? I'll update the comment if that's fine with you

vietj:
if that's only implemented by Hazelcast, then we should just remove it straight

tsegismont:
Only used by Hazelcast in the stack:
https://github.com/search?q=org%3Avert-x3+ExtendedClusterManager&type=Code

I'll go ahead and remove it

2017-07-27 15:44 GMT+02:00 Julien Viet <notifications@github.com>:

tsegismont:
@vietj done

PR #2077, opened 2017-07-27
Mismatch between documentation and implementation for LocalMap

### LINKED ISSUE #1509: Missmatch between documentation and implementation for LocalMap
The documentation of LocalMap states:

The map only allows immutable keys and values in the map, OR certain mutable objects such as {@link io.vertx.core.buffer.Buffer}

https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/shareddata/LocalMap.java

However the implementation in:

https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/shareddata/impl/Checker.java

Has only a whitelist of allowed Types. Immutable objects are refused. 

Not sure whether this is an implementation or documentation problem.

tsegismont:
Fixes #1509


pmlopes:
:+1:

PR #2078, opened 2017-07-27
SSLSession accessor support to get access to SSL session information

dachaac:
Java SE 5-8 has deprecated usage of SSLSession#getPeerCertificateChain() as described in here:
https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLSession.html#getPeerCertificateChain--

This PR adds support for SSLSession#getPeerCertificates() which has new accessor peerCertificates() to follow naming convention for peerCertificateChain(). peerCertificateChain() is left there with @deprecated note to transition to peerCertificate().

TLS test cases are adapted to use peerCertificates().

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpConnection.java
+++ b/src/main/java/io/vertx/core/http/HttpConnection.java
@@ -267,10 +267,15 @@ public interface HttpConnection {
   SSLSession sslSession();
 
   /**
+   * Note: Java SE 5+ recommends to use javax.net.ssl.SSLSession#getPeerCertificates() instead of
+   * of javax.net.ssl.SSLSession#getPeerCertificateChain() which this method is based on. Use {@link #sslSession()} to
+   * access that method.
+   *
    * @return an ordered array of the peer certificates. Returns null if connection is
    *         not SSL.
    * @throws javax.net.ssl.SSLPeerUnverifiedException SSL peer's identity has not been verified.
    * @see javax.net.ssl.SSLSession#getPeerCertificateChain()
+   * @see #sslSession()
    */
   @GenIgnore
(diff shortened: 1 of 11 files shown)


sauthieg:
Should Vert.x use the `@Deprecated` annotation (in addition to the doclet) ?

vietj:
I'm wondering if instead we should rather not give access to the SSL session instead of copying all its methods.

dachaac:
@vietj Do you mean to **add** access to SSL Session? It might be better option than having multiple accessors. SSL Session also provides other methods that user might want to use.

- getLocal*() might have some use cases. 
- Part of getPeer*() is already available otherwise but shorthand getPeerPrincipal() might have uses.

Shall I replace peerCertificates() in this PR with sslSession() accessor?
Shall I remove @Deprecated or keep it there? Now instead direct user to sslSession() usage?

vietj:
@dachaac let's go for an `sslSession()` that returns it and no `deprecated` as the original methods are not deprecated on `SSLSession`.

dachaac:
@vietj PR updated for sslSession() and changed wording about deprecation to recommendation.

vietj:
thanks, I don't think it's necessary to change the use of `Certificate` over `X509Certificate` for tests

dachaac:
@vietj Point for the change was to show how to do it with the new interface. Shall the commit be removed from this PR?

vietj:
yes

dachaac:
@vietj test commit removed.

vietj:
@ppatierno we've added this to NetSocket, perhaps that's something you want to expose in a similar fashion in MQTT ?

PR #2082, opened 2017-07-31
Deploy verticles with a supplier or a class

### LINKED ISSUE #1952: Vertx deploys a Verticle class
It would be nice to have a `Vertx#deployVerticle(C<? extends Verticle, ...)` so one can deploy a class without using classloading

### LINKED ISSUE #1953: Vertx deploys a Verticle instances
It would be nice to have a `Vertx#deployVerticle(Supplier<Verticle>, ...)` pretty much like `Vertx#deployVerticle(Verticle,...)` but that is able to handle multiple instances

tsegismont:
Fixes #1953, fixes #1952

### DIFF
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -336,6 +336,13 @@ public interface Vertx extends Measured {
 
   /**
    * Like {@link #deployVerticle(Verticle, DeploymentOptions)} but {@link Verticle} instance is created by invoking the
+   * default constructor of {@code verticleClass}.
+   */
+  @GenIgnore
+  void deployVerticle(Class<? extends Verticle> verticleClass, DeploymentOptions options);
+
+  /**
+   * Like {@link #deployVerticle(Verticle, DeploymentOptions)} but {@link Verticle} instance is created by invoking the
    * {@code verticleSupplier}.
    * <p>
    * The supplier will be invoked as many times as {@link DeploymentOptions#getInstances()}.
@@ -359,6 +366,13 @@ public interface Vertx extends Measured {
    */
   @GenIgnore
   void deployVerticle(Verticle verticle, DeploymentOptions options, Handler<AsyncResult<String>> completionHandler);
+
+  /**
+   * Like {@link #deployVerticle(Verticle, DeploymentOptions, Handler)} but {@link Verticle} instance is created by
+   * invoking the default constructor of {@code verticleClass}.
+   */
+  @GenIgnore
+  void deployVerticle(Class<? extends Verticle> verticleClass, DeploymentOptions options, Handler<AsyncResult<String>> completionHandler);
 
   /**
    * Like {@link #deployVerticle(Verticle, DeploymentOptions, Handler)} but {@link Verticle} instance is created by
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -42,6 +42,7 @@ import io.vertx.core.spi.VertxFactory;
 import io.vertx.core.streams.ReadStream;
 
 import java.util.Set;
+import java.util.function.Supplier;
 
 /**
  * The entry point into the Vert.x Core API.
@@ -334,6 +335,21 @@ public interface Vertx extends Measured {
   void deployVerticle(Verticle verticle, DeploymentOptions options);
 
   /**
+   * Like {@link #deployVerticle(Verticle, DeploymentOptions)} but {@link Verticle} instance is created by invoking the
+   * {@code verticleSupplier}.
+   * <p>
+   * The supplier will be invoked as many times as {@link DeploymentOptions#getInstances()}.
+   * It must not return:
+   * <ul>
+   * <li>null</li>
+   * <li>the same instance twice</li>
+   * <li>instances of different verticle classes</li>
+   * </ul>
+   */
+  @GenIgnore
+  void deployVerticle(Supplier<? extends Verticle> verticleSupplier, DeploymentOptions options);
+
+  /**
    * Like {@link #deployVerticle(Verticle, Handler)} but {@link io.vertx.core.DeploymentOptions} are provided to configure the
    * deployment.
    *
@@ -343,6 +359,21 @@ public interface Vertx extends Measured {
    */
   @GenIgnore
   void deployVerticle(Verticle verticle, DeploymentOptions options, Handler<AsyncResult<String>> completionHandler);
+
+  /**
+   * Like {@link #deployVerticle(Verticle, DeploymentOptions, Handler)} but {@link Verticle} instance is created by
+   * invoking the {@code verticleSupplier}.
+   * <p>
+   * The supplier will be invoked as many times as {@link DeploymentOptions#getInstances()}.
+   * It must not return:
+   * <ul>
+   * <li>null</li>
+   * <li>the same instance twice</li>
+   * <li>instances of different verticle classes</li>
+   * </ul>
+   */
+  @GenIgnore
+  void deployVerticle(Supplier<? extends Verticle> verticleSupplier, DeploymentOptions options, Handler<AsyncResult<String>> completionHandler);
 
   /**
    * Deploy a verticle instance given a name.
--- a/src/main/java/io/vertx/core/impl/DeploymentManager.java
+++ b/src/main/java/io/vertx/core/impl/DeploymentManager.java
@@ -115,7 +115,7 @@ public class DeploymentManager {
         });
       }
       if (verticles.size() != 1) {
-        throw new RuntimeException("Supplied verticles are not of the same class");
+        throw new RuntimeException("Supplied verticles are not from the same class");
       }
       Set<Verticle> instances = verticles.values().iterator().next();
       if (instances.size() != nbInstances) {
--- a/src/main/java/io/vertx/core/impl/DeploymentManager.java
+++ b/src/main/java/io/vertx/core/impl/DeploymentManager.java
@@ -16,7 +16,13 @@
 
 package io.vertx.core.impl;
 
-import io.vertx.core.*;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
+import io.vertx.core.DeploymentOptions;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.ServiceHelper;
+import io.vertx.core.Verticle;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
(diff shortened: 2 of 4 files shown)


src/main/java/io/vertx/core/Vertx.java
```diff
@@ -334,6 +335,28 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
...
+  void deployVerticle(Class<? extends Verticle> verticleClass, DeploymentOptions options);
+
+  /**
+   * Like {@link #deployVerticle(Verticle, DeploymentOptions)} but {@link Verticle} instance is created by invoking the
+   * {@code verticleSupplier}.
+   * <p>
+   * The supplier will be invoked as many times as {@link DeploymentOptions#getInstances()}.
+   * It must not return:
+   * <ul>
+   * <li>null</li>
```
cescoffier: Maybe a note when `null` is returned (obviously you won't deploy null)

src/main/java/io/vertx/core/Vertx.java
```diff
@@ -345,6 +368,28 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
...
+  void deployVerticle(Class<? extends Verticle> verticleClass, DeploymentOptions options, Handler<AsyncResult<String>> completionHandler);
+
+  /**
+   * Like {@link #deployVerticle(Verticle, DeploymentOptions, Handler)} but {@link Verticle} instance is created by
+   * invoking the {@code verticleSupplier}.
+   * <p>
+   * The supplier will be invoked as many times as {@link DeploymentOptions#getInstances()}.
+   * It must not return:
+   * <ul>
+   * <li>null</li>
```
cescoffier: Same as above.

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -79,13 +99,47 @@ public void deployVerticle(Verticle verticle, DeploymentOptions options,
...
+          throw new RuntimeException("Supplied verticle is null");
+        }
+        verticles.compute(verticle.getClass(), (k, v) -> {
+          if (v == null) v = new HashSet<>();
+          v.add(verticle);
+          return v;
+        });
+      }
+      if (verticles.size() != 1) {
+        throw new RuntimeException("Supplied verticles are not of the same class");
```
cescoffier: of -> from.

Is it important that the supplier returns instances from the same class?

tsegismont: The typo is fixed.

I thought it would be good to be conservative here. Technically speaking, nothing prevents you from writing a verticle factory which returns instances of different verticle classes.

Let me know if you believe the constraint should be removed.

cescoffier: I think your approach is right. Let's start with a conservative approach and if we see that people need a bit more, let's rediscuss it. So 👍

tsegismont: @cescoffier thanks for reviewing!

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -79,13 +99,47 @@ public void deployVerticle(Verticle verticle, DeploymentOptions options,
...
     }
     ContextImpl currentContext = vertx.getOrCreateContext();
-    doDeploy("java:" + verticle.getClass().getName(), generateDeploymentID(), options, currentContext, currentContext, completionHandler,
-        getCurrentClassLoader(), verticle);
+    ClassLoader cl = getClassLoader(options, currentContext);
+    currentContext.<Set<Verticle>>executeBlocking(fut -> {
+      int nbInstances = options.getInstances();
+      IdentityHashMap<Class<? extends Verticle>, Set<Verticle>> verticles = new IdentityHashMap<>();
+      for (int i = 0; i < nbInstances; i++) {
+        Verticle verticle = verticleSupplier.get();
```
vietj: the case where `get` throws an exception is not handled

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -79,13 +99,47 @@ public void deployVerticle(Verticle verticle, DeploymentOptions options,
       throw new IllegalArgumentException("Can't specify isolatedClasses for already created verticle");
     }
     ContextImpl currentContext = vertx.getOrCreateContext();
-    doDeploy("java:" + verticle.getClass().getName(), generateDeploymentID(), options, currentContext, currentContext, completionHandler,
-        getCurrentClassLoader(), verticle);
+    ClassLoader cl = getClassLoader(options, currentContext);
+    currentContext.<Set<Verticle>>executeBlocking(fut -> {
```
vietj: do we really need `executeBlocking` here ?

if we do, we should probably say that the thread calling `Supplier#get` is not the same than the thread that will execute the verticle, specially with an event loop based verticle.

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -79,13 +99,47 @@ public void deployVerticle(Verticle verticle, DeploymentOptions options,
...
+      int nbInstances = options.getInstances();
+      IdentityHashMap<Class<? extends Verticle>, Set<Verticle>> verticles = new IdentityHashMap<>();
```
vietj: why do we need an `IdentityHashMap` here ?

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -79,13 +99,47 @@ public void deployVerticle(Verticle verticle, DeploymentOptions options,
...
+          throw new RuntimeException("Supplied verticle is null");
+        }
+        verticles.compute(verticle.getClass(), (k, v) -> {
+          if (v == null) v = new HashSet<>();
+          v.add(verticle);
+          return v;
+        });
+      }
+      if (verticles.size() != 1) {
+        throw new RuntimeException("Supplied verticles are not from the same class");
```
vietj: do we really need this restriction ? i.e the user should be responsible for using the same class but I don't think we want to be restrictive here, i.e this is the typical check that gets in the way of someone and will annoy him.

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -79,13 +99,47 @@ public void deployVerticle(Verticle verticle, DeploymentOptions options,
...
+    }, false, ar -> {
+      if (ar.succeeded()) {
+        Set<Verticle> result = ar.result();
+        Verticle[] verticles = result.toArray(new Verticle[result.size()]);
+        String verticleClass = verticles[0].getClass().getName();
+        doDeploy("java:" + verticleClass, generateDeploymentID(), options, currentContext, currentContext, completionHandler, cl, verticles);
+      } else {
+        completionHandler.handle(Future.failedFuture(ar.cause()));
+      }
+    });
```
vietj: this seems convoluted and if we drop the "not the same class check" we can get something simpler:

```
    currentContext.<List<Verticle>>executeBlocking(fut -> {
      int nbInstances = options.getInstances();
      List<Verticle> verticles = new ArrayList<>();
      for (int i = 0; i < nbInstances; i++) {
        Verticle verticle = verticleSupplier.get();
        if (verticle == null) {
          throw new RuntimeException("Supplied verticle is null");
        }
        verticles.add(verticle);
      }
      if (verticles.stream().distinct().count() < nbInstances) {
        throw new RuntimeException("Same verticle supplied more than once");
      }
      fut.complete(verticles);
    }, false, ar -> {
      if (ar.succeeded()) {
        List<Verticle> result = ar.result();
        Verticle[] verticles = result.toArray(new Verticle[result.size()]);
        String verticleClass = verticles[0].getClass().getName();
        doDeploy("java:" + verticleClass, generateDeploymentID(), options, currentContext, currentContext, completionHandler, cl, verticles);
      } else {
        completionHandler.handle(Future.failedFuture(ar.cause()));
      }
    });
```

src/main/java/io/vertx/core/Vertx.java
```diff
@@ -345,6 +368,28 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
...
+   * <li>the same instance twice</li>
+   * <li>instances of different verticle classes</li>
+   * </ul>
+   */
+  @GenIgnore
+  void deployVerticle(Supplier<? extends Verticle> verticleSupplier, DeploymentOptions options, Handler<AsyncResult<String>> completionHandler);
```
vietj: do we really need the `<? extends Verticle>` most users will use a lambda here (I'm not a fan of wildcard generics).

src/main/java/io/vertx/core/impl/DeploymentManager.java
vietj: or at least not tested :)

tsegismont: It is handled by executeBlocking but yes, I can add a test

src/main/java/io/vertx/core/impl/DeploymentManager.java
tsegismont: We don't know what users are going to do in the supplier. Could be blocking code.

We could change the deploy method to add a `blockingCreation` boolean param. Would work like in a verticle factory.

src/main/java/io/vertx/core/impl/DeploymentManager.java
tsegismont: Convenient way to check if verticles instances are from the same class

src/main/java/io/vertx/core/impl/DeploymentManager.java
tsegismont: I was not sure so went with a conservative approach but right, someone implementing a verticle factory would not have this limitation

src/main/java/io/vertx/core/impl/DeploymentManager.java
tsegismont: You can use _distinct()_ because it relies on equals, not identity. We really want different instances

src/main/java/io/vertx/core/impl/DeploymentManager.java
vietj: `java.lang.Class` does not override equals, so it was confusing with the distinct number of instances required below

src/main/java/io/vertx/core/impl/DeploymentManager.java
tsegismont: But yes code with less constraint checking should be more concise

src/main/java/io/vertx/core/impl/DeploymentManager.java
vietj: I think it should simply do nothing, most likely add a note that the supplier will be called on the current thread and so if the user is on an event loop he will know that he should not block the creation, if he has a blocking supplier he would himself wrap it with an executeBlocking block.

src/main/java/io/vertx/core/impl/DeploymentManager.java
vietj: that's not much different of the commit that uses an `HashSet`

src/main/java/io/vertx/core/Vertx.java
tsegismont: Right

src/main/java/io/vertx/core/impl/DeploymentManager.java
tsegismont: ok

PR #2087, opened 2017-08-02
Add error log output for cluster connection failures

### LINKED ISSUE #1575: ConnectionHolder should log when queuing messages
ConnectionHolder creates a queue and inserts messages bound for a remote host if it isn't connected.  It would be very useful it it logged debug messages when this happened.  It would also be exceptionally useful if it logged a warning for failed connection attempts to remote peers, including the attempted host/port.

Also, the queue appears to grow without bound.  It would be good to limit the size of the queue via a config entry, then stop queuing messages for that remote host once that limit is reached.

Jotschi:
Fixes #1575

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
@@ -56,6 +56,7 @@ class ConnectionHolder {
       if (res.succeeded()) {
         connected(res.result());
       } else {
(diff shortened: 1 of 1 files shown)


cescoffier:
LGTM, @tsegismont WDYT ?

tsegismont:
@cescoffier looks good to me as well. I'll merge it manually, I'm going to add a debug log when queing starts/stops (as discussed in #1575).

tsegismont:
@Jotschi thanks!

PR #2089, opened 2017-08-04
Improve the CPU core detection on linux when executed in cgroups (Doc…

cescoffier:
…ker container for instance)

This PR replaces the classic call to `availableProcessors` when running on Linux. This is because, when running in a docker environment or on the top of an hypervisor the JVM does not a great job to give us the right answer (returning the number of cores on the host machine instead of the container ones).

This PR proposes to read `/proc/self/status` to find the right number of core. 

On Windows and Mac, we still rely on the JVM method (which give the right answer).


vietj:
I think actually we need a CQ for this contribution

PR #2094, opened 2017-08-08
Getting corresponding sharred map

nolequen:
Sometimes it is very uncomfortable to choose an appropriate map explicitly, especialy if we don't know in advance whether the Vert.x instance clustered or not. E.g. when writing tests which may use cluster in some cases or not, choosing the "right" map leads to changes in production code.

The goal of the proposed changes is to make it easier to work with shared data without paying attention if the instance is clustered or not.


vietj:
it looks ok to me, however I think we should honour the TTL with a timer and make sure it works correctly.

nolequen:
Great! I will add TTL support at an early date.

nolequen:
@vietj I have updated the request with TTL support.
Unfortunately, some merge-commits were added... Is it ok or new request should be created?

nolequen:
Replaced with up-to date request: #2244.

Issue #2100, opened 2017-08-23
HttpClient allow dangerous header values


mguillet:
The Vert.x http client accepts header values that contains LF in it. If a program were to add an extra header based on user input without proper data sanitation this could allow an attacker to perform header injection.

Here is a working example:
```java
@RunWith(VertxUnitRunner.class)
public class TestInjection {

  @ClassRule
  static public RunTestOnContext vertxRule = new RunTestOnContext();

  @Test
  public void testHttpHeader(TestContext context) {
    Vertx vertx = vertxRule.vertx();

    vertx.createHttpServer()
      .requestHandler(request -> {
        System.out.println(request.headers().entries().stream()
          .map(e -> e.getKey() + ": <" + e.getValue() +">").collect(Collectors.joining("\n")));
        request.response().end();
      })
      .listen(0, context.asyncAssertSuccess(server -> {
        Async async = context.async();
        vertx.createHttpClient()
          .getAbs("http://localhost:" + server.actualPort(), request -> async.complete())
          .putHeader("Content-Type", "text/plain\nX-EvilHeader: horrible injection")
          .end();
      }));
  }

}
```

This will produce the following input:
```
Content-Type: <text/plain>
X-EvilHeader: <horrible injecton>
Host: <localhost:xxxxx>
```

It seems that Vert.x does't ask Netty to perform header validation:
https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/ClientConnection.java#L451
When passing `true` in the `validateHeaders` parameter to DefaultHttpRequest's constructor an exception is then raised:
```
java.lang.IllegalArgumentException: only ' ' and '\t' are allowed after '\n': [...]
```

Is there any reason why Vert.x is not asking Netty to validate the header of the client request?

If not, I'll create a PR to fix this issue that we consider to be a SSRF vulnerability.

What do think?

vietj:
@mguillet the main reason is speed. I think you could add this as a configuration in HttpServerOptions and HttpClientOptions

mguillet:
Ok I'll create a PR for this so we can discuss about naming, default value, behaviour, etc.

mguillet:
I've created the PR #2102, so I'm closing this ticket.

PR #2102, opened 2017-08-23
#2100 Add validateHeaderValues option to detect malformed HTTP header values

mguillet:
_This is a WIP._
_As of today HTTPServer support for this option is not implemented_

Because of #2100, one might need to validate HTTP headers value to be able to detect header injection or just forbidden values.

We need to allow this check in request performed by the Vert.x `HttpClient` and response made through the `HttpServer`

My suggestion is to add a `setValidateHeaderValues` method on both `HttpClientOptions` and `HttpServerOptions`. This option will ask Netty to perform header validation when constructing a `DefaultHttpRequest` instance.

For speed reason, one might want to disable this check. But I feel like this option should be activated by default for security reasons.

**Note:** I don't think HTTP/2 is impacted by this issue since headers are now binary based.

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -905,6 +905,11 @@ Set the ALPN usage.
 +++
 Set whether Netty pooled buffers are enabled
 +++
+|[[validateHeaderValues]]`validateHeaderValues`|`Boolean`|
++++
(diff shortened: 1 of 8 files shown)


vietj:
thanks @mguillet this is now done by default in vertx and there is an system property to disable validation

Issue #2099, opened 2017-08-22
Api on AsyncResult.map is wrong


kerko:
The documentation on AsyncResult.map contains the following passage:
> If the {@code mapper} throws an exception, the returned future will be failed with this exception.
>     When this async result is failed, the failure will be propagated to the returned future and the {@code mapper}     will not be called.

This is simply not true, throwing an exception in the Mapper results in an asyncresult that never completes and therefore breaks the complete flow. This passage should probably be removed from the docs or map should offer the advertised behavior.

vietj:
you are right, the doc should be updated

vietj:
fixed

PR #2103, opened 2017-08-24
Fixed docs

kerko:
this fixes the wrong docs mentioned in #2099 
Also fixes two instances of future where it should be async result


vietj:
thanks

PR #2107, opened 2017-08-31
Add Websocket Ping/Pong API methods and fix ping response (Issue #290)

dgoetsch:
In response to https://github.com/vert-x3/issues/issues/290

### DIFF
--- a/src/main/java/io/vertx/core/http/WebSocketBase.java
+++ b/src/main/java/io/vertx/core/http/WebSocketBase.java
@@ -21,6 +21,7 @@ import io.vertx.codegen.annotations.Fluent;
 import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.Nullable;
 import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.net.SocketAddress;
@@ -136,6 +137,39 @@ public interface WebSocketBase extends ReadStream<Buffer>, WriteStream<Buffer> {
   WebSocketBase writeTextMessage(String text);
 
   /**
+   * Writes a ping to the connection. This will be written in a single frame. Ping frames may be at most 125 bytes (octets).
+   * <p>
+   * This method should not be used to write application data and should only be used for implementing a keep alive or
+   * to ensure the client is still responsive, see RFC 6455 Section 5.5.2.
+   * <p>
+   * There is no pingHandler because RFC 6455 section 5.5.2 clearly states that the only response to a ping is a pong
+   * with identical contents.
+   *
+   * @param data the data to write, may be at most 125 bytes
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  WebSocketBase writePing(Buffer data);
+
+  /**
+   * Writes a pong to the connection. This will be written in a single frame. Pong frames may be at most 125 bytes (octets).
+   * <p>
+   * This method should not be used to write application data and should only be used for implementing a keep alive or
+   * to ensure the client is still responsive, see RFC 6455 Section 5.5.2.
+   * <p>
+   * There is no need to manually write a Pong, as the server and client both handle responding to a ping with a pong
+   * automatically and this is exposed to users.RFC 6455 Section 5.5.3 states that pongs may be sent unsolicited in order
+   * to implement a one way heartbeat.
+   *
+   * @param data the data to write, may be at most 125 bytes
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  WebSocketBase writePong(Buffer data);
+
+
+
+  /**
    * Set a close handler. This will be called when the WebSocket is closed.
    *
    * @param handler  the handler
@@ -173,6 +207,23 @@ public interface WebSocketBase extends ReadStream<Buffer>, WriteStream<Buffer> {
    */
   @Fluent
   WebSocketBase binaryMessageHandler(@Nullable Handler<Buffer> handler);
+
+  /**
+   * Set a pong message handler on the connection.  This handler will be invoked every time a pong message is received
+   * on the server, and can be used by both clients and servers since the RFC 6455 Sections 5.5.2 and 5.5.3 do not
+   * specify whether the client or server sends a ping.
+   * <p>
+   * Pong frames may be at most 125 bytes (octets).
+   * <p>
+   * There is no ping handler since pings should immediately be responded to with a pong with identical content
+   * <p>
+   * Pong frames may be received unsolicited.
+   *
+   * @param handler the handler
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  WebSocketBase pongHandler(@Nullable Handler<Buffer> handler);
 
   /**
    * Calls {@link #close()}
--- a/src/main/java/io/vertx/core/http/WebSocketFrame.java
+++ b/src/main/java/io/vertx/core/http/WebSocketFrame.java
@@ -63,6 +63,26 @@ public interface WebSocketFrame {
   }
 
   /**
+   * Create a ping WebSocket frame.  Will be a final frame. There is no option for non final ping frames.
+   *
+   * @param data the bytes for the frame, may be at most 125 bytes
+   * @return the frame
+   */
+  static WebSocketFrame pingFrame(Buffer data) {
+    return factory.pingFrame(data);
+  }
+
+  /**
+   * Create a pong WebSocket frame.  Will be a final frame. There is no option for non final pong frames.
+   *
+   * @param data the bytes for the frame, may be at most 125 bytes
+   * @return the frame
+   */
+  static WebSocketFrame pongFrame(Buffer data) {
+    return factory.pongFrame(data);
+  }
+
+  /**
    * Create a continuation frame
    *
    * @param data  the data for the frame
--- a/src/main/java/io/vertx/core/http/impl/ClientHandler.java
+++ b/src/main/java/io/vertx/core/http/impl/ClientHandler.java
@@ -23,6 +23,7 @@ import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpObject;
 import io.netty.handler.codec.http.HttpResponse;
 import io.netty.handler.codec.http.LastHttpContent;
+import io.netty.handler.codec.http.websocketx.PongWebSocketFrame;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.impl.ws.WebSocketFrameImpl;
 import io.vertx.core.http.impl.ws.WebSocketFrameInternal;
@@ -33,7 +34,6 @@ import io.vertx.core.spi.metrics.HttpClientMetrics;
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
 class ClientHandler extends VertxHttpHandler<ClientConnection> {
-
   private boolean closeFrameSent;
   private ContextImpl context;
   private ChannelHandlerContext chctx;
@@ -107,14 +107,12 @@ class ClientHandler extends VertxHttpHandler<ClientConnection> {
         case BINARY:
         case CONTINUATION:
         case TEXT:
+        case PONG:
           conn.handleWsFrame(frame);
           break;
         case PING:
           // Echo back the content of the PING frame as PONG frame as specified in RFC 6455 Section 5.5.2
-          chctx.writeAndFlush(new WebSocketFrameImpl(FrameType.PONG, frame.getBinaryData()));
-          break;
-        case PONG:
-          // Just ignore it
+          chctx.writeAndFlush(new PongWebSocketFrame(frame.getBinaryData().copy()));
           break;
         case CLOSE:
           if (!closeFrameSent) {
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -45,6 +45,7 @@ import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpRequestDecoder;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.LastHttpContent;
+import io.netty.handler.codec.http.websocketx.PongWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketHandshakeException;
 import io.netty.handler.codec.http.websocketx.WebSocketServerHandshaker;
 import io.netty.handler.codec.http.websocketx.WebSocketVersion;
@@ -638,14 +639,12 @@ public class HttpServerImpl implements HttpServer, Closeable, MetricsProvider {
           case BINARY:
           case CONTINUATION:
           case TEXT:
+          case PONG:
             conn.handleMessage(msg);
             break;
           case PING:
             // Echo back the content of the PING frame as PONG frame as specified in RFC 6455 Section 5.5.2
-            ch.writeAndFlush(new WebSocketFrameImpl(FrameType.PONG, wsFrame.getBinaryData()));
-            break;
-          case PONG:
-            // Just ignore it
+            chctx.writeAndFlush(new PongWebSocketFrame(wsFrame.getBinaryData().copy()));
             break;
           case CLOSE:
             if (!closeFrameSent) {
--- a/src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
+++ b/src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
@@ -17,6 +17,7 @@
 package io.vertx.core.http.impl;
 
 import io.netty.buffer.ByteBuf;
+import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.eventbus.Message;
@@ -32,7 +33,9 @@ import io.vertx.core.net.impl.ConnectionBase;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
 import javax.security.cert.X509Certificate;
+import java.util.ArrayDeque;
 import java.util.UUID;
+import java.util.concurrent.ArrayBlockingQueue;
 
 /**
  * This class is optimised for performance when used on the same event loop. However it can be used safely from other threads.
@@ -54,6 +57,7 @@ public abstract class WebSocketImplBase<S extends WebSocketBase> implements WebS
   private final MessageConsumer textHandlerRegistration;
   private Object metric;
   private Handler<WebSocketFrameInternal> frameHandler;
+  private Handler<Buffer> pongHandler;
   private Handler<Buffer> dataHandler;
   private Handler<Void> drainHandler;
   private Handler<Throwable> exceptionHandler;
@@ -61,6 +65,7 @@ public abstract class WebSocketImplBase<S extends WebSocketBase> implements WebS
   private Handler<Void> endHandler;
   protected final ConnectionBase conn;
   protected boolean closed;
+
 
   WebSocketImplBase(VertxInternal vertx, ConnectionBase conn, boolean supportsContinuation,
                               int maxWebSocketFrameSize, int maxWebSocketMessageSize) {
@@ -161,6 +166,18 @@ public abstract class WebSocketImplBase<S extends WebSocketBase> implements WebS
     }
   }
 
+  @Override
+  public S writePing(Buffer data) {
+    if(data.length() > maxWebSocketFrameSize || data.length() > 125) throw new IllegalStateException("Ping cannot exceed maxWebSocketFrameSize or 125 bytes");
+    return writeFrame(WebSocketFrame.pingFrame(data));
+  }
+
+  @Override
+  public S writePong(Buffer data) {
+    if(data.length() > maxWebSocketFrameSize || data.length() > 125) throw new IllegalStateException("Pong cannot exceed maxWebSocketFrameSize or 125 bytes");
+    return writeFrame(WebSocketFrame.pongFrame(data));
+  }
+
   private void writeMessageInternal(Buffer data) {
     checkClosed();
     writePartialMessage(FrameType.BINARY, data, 0);
@@ -233,17 +250,25 @@ public abstract class WebSocketImplBase<S extends WebSocketBase> implements WebS
         Buffer buff = Buffer.buffer(frame.getBinaryData());
         dataHandler.handle(buff);
       }
-
-      if (frameHandler != null) {
-        frameHandler.handle(frame);
+      switch(frame.type()) {
+        case PONG:
+          pongHandler.handle(frame.binaryData().copy());
+          break;
+        case TEXT:
+        case BINARY:
+        case CONTINUATION:
+          if (frameHandler != null) {
+            frameHandler.handle(frame);
+          }
+          break;
       }
     }
   }
 
   private class FrameAggregator implements Handler<WebSocketFrameInternal> {
-
     private Handler<String> textMessageHandler;
     private Handler<Buffer> binaryMessageHandler;
+
     private Buffer textMessageBuffer;
     private Buffer binaryMessageBuffer;
 
@@ -288,7 +313,6 @@ public abstract class WebSocketImplBase<S extends WebSocketBase> implements WebS
           textMessageHandler.handle(fullMessage);
         }
       }
-
     }
 
     private void handleBinaryFrame(WebSocketFrameInternal frame) {
@@ -345,6 +369,15 @@ public abstract class WebSocketImplBase<S extends WebSocketBase> implements WebS
         frameHandler = new FrameAggregator();
       }
       ((FrameAggregator) frameHandler).binaryMessageHandler = handler;
+      return (S) this;
+    }
+  }
+
+  @Override
+  public WebSocketBase pongHandler(Handler<Buffer> handler) {
+    synchronized (conn) {
+      checkClosed();
+      this.pongHandler = handler;
       return (S) this;
     }
   }
--- a/src/main/java/io/vertx/core/http/impl/WebSocketFrameFactoryImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/WebSocketFrameFactoryImpl.java
@@ -42,4 +42,13 @@ public class WebSocketFrameFactoryImpl implements WebSocketFrameFactory {
     return new WebSocketFrameImpl(FrameType.CONTINUATION, data.getByteBuf(), isFinal);
   }
 
+  @Override
+  public WebSocketFrame pingFrame(Buffer data) {
+    return new WebSocketFrameImpl(FrameType.PING, data.getByteBuf(), true);
+  }
+
+  @Override
+  public WebSocketFrame pongFrame(Buffer data) {
+    return new WebSocketFrameImpl(FrameType.PONG, data.getByteBuf(), true);
+  }
 }
--- a/src/main/java/io/vertx/core/spi/WebSocketFrameFactory.java
+++ b/src/main/java/io/vertx/core/spi/WebSocketFrameFactory.java
@@ -29,4 +29,8 @@ public interface WebSocketFrameFactory {
   WebSocketFrame textFrame(String str, boolean isFinal);
 
   WebSocketFrame continuationFrame(Buffer data, boolean isFinal);
+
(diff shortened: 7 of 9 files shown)


src/main/java/io/vertx/core/http/impl/ClientHandler.java
```diff
@@ -111,8 +110,11 @@ protected void handleMessage(ClientConnection conn, ContextImpl context, Channel
           break;
         case PING:
           // Echo back the content of the PING frame as PONG frame as specified in RFC 6455 Section 5.5.2
-          chctx.writeAndFlush(new WebSocketFrameImpl(FrameType.PONG, frame.getBinaryData()));
```
dgoetsch: I think this failed to work, because of encoding.

ChannelHandlerContext doesn't have encoding for the vertx Frame implementations.

The fix takes advantage of the encoder on the connection:
https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/Http1xConnectionBase.java#L52

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -638,14 +638,14 @@ protected void handleMessage(ServerConnection conn, ContextImpl context, Channel
...
           case CONTINUATION:
           case TEXT:
+          case PONG:
             conn.handleMessage(msg);
             break;
           case PING:
             // Echo back the content of the PING frame as PONG frame as specified in RFC 6455 Section 5.5.2
-            ch.writeAndFlush(new WebSocketFrameImpl(FrameType.PONG, wsFrame.getBinaryData()));
-            break;
-          case PONG:
```
dgoetsch: On the server, pong is moved so that the pongHandler is invoked.

Currently, the webSocketClient api exposes a pongHandler method, but the client does invoke it as the server does.  I would like to resolve the inconsistency, and request input on whether the client should expose that method.

src/main/java/io/vertx/core/http/impl/ClientHandler.java
dgoetsch: I think the same problem may exist a few lines down under `case CLOSE:`

dgoetsch:
This resolves the issue discussed here: https://github.com/vert-x3/issues/issues/290

I will also add an example once this passes code review that mirros this: https://github.com/dgoetsch/vertx-web-socket-test/blob/master/web-socket-server/src/main/kotlin/dev/yn/playground/web/socket/server/vertx/WebSocketEndpoint.kt

dgoetsch:
In case you are wondering.. I am working on an automated keep alive mechanism: https://github.com/dgoetsch/vert.x/tree/websocket-keep-alive

I will open a separate PR once this one is closed and I have done some testing.

src/main/java/io/vertx/core/http/impl/ClientHandler.java
```diff
@@ -107,14 +106,14 @@ protected void handleMessage(ClientConnection conn, ContextImpl context, Channel
...
           conn.handleWsFrame(frame);
           break;
         case PING:
           // Echo back the content of the PING frame as PONG frame as specified in RFC 6455 Section 5.5.2
-          chctx.writeAndFlush(new WebSocketFrameImpl(FrameType.PONG, frame.getBinaryData()));
-          break;
-        case PONG:
-          // Just ignore it
+          WebSocketFrameImpl pong = new WebSocketFrameImpl(FrameType.PONG, frame.getBinaryData());
+          conn.reportBytesWritten(pong.binaryData().length());
```
vietj: this should be tested

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -638,14 +638,14 @@ protected void handleMessage(ServerConnection conn, ContextImpl context, Channel
...
-            // Just ignore it
+            WebSocketFrameImpl pong = new WebSocketFrameImpl(FrameType.PONG, wsFrame.getBinaryData());
+            conn.reportBytesWritten(pong.binaryData().length());
```
vietj: this should be tested

dgoetsch:
@vietj I took a look at the HttpConnection, and attempted to mirror the api there.  I think the way it is right now has some problems that need to be worked through.  Currently, I've worked it out so that the API supports the same signature methods as other websocket methods for writeText/BinaryMessage and binary/textHandler as well as the style similar to HttpConnection.  With the current design, These two methods cannot be used simultaneously by the same instance, but that is not enforced by the API.  I don't think it is feasible to support using both styles of handling simultaneously, so I think the only solution to maintain a clean API is to keep only one of styles the API methods for writePing.  I'm biased towards maintaining consistency between the interface methods on WebSocket and keep pongHandler separate from writePing over attempting match a method in a different class.

Having the ArrayDeque of Pong handlers seems unecessary to me.  I can't think of an advantage of that over a single handler, as my strategy is just to restart a timer every time a pong is received.  I also don't think that having a queue of handlers improves the ability to detect absent pong responses.

Thoughts?

vietj:
@dgoetsch I think that the HttpConnection API should be simplified to just have a pong handler and send ping like you did, that would break the API though but it seems ok-ish to me as it's not a feature that is used a lot.

dgoetsch:
@vietj I will make it so.

src/main/java/io/vertx/core/http/impl/ClientHandler.java
```diff
@@ -107,14 +106,14 @@ protected void handleMessage(ClientConnection conn, ContextImpl context, Channel
         case BINARY:
         case CONTINUATION:
         case TEXT:
+        case PONG:
```
dgoetsch: Should we also handle ping here and expose a `pingHandler` method in `WebSocketBase`? We could default it to `buf -> { writePong(buf); }` to provide default behavior, but some users may want to log metrics (for example in a load test client).

src/main/java/io/vertx/core/http/impl/ClientHandler.java
dgoetsch: The unit tests I wrote in WebSocketTest do validate this functionality.  I reverted these lines and the tests failed.

I looked into writing a lower level test, but the API don't provide access to the underlying connection or clientHandler or `ChannelHandlerContext`.  Similarly, I cannot find a way to do this on the server API.  With enough time I could probably figure out to write an additional low level test. However, these tests would probably be very complex and the time investment/value proposition doesn't seem appropriate since the functionality is already validated by another UnitTest.

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
dgoetsch: Pease refer to above comment.

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -136,6 +137,31 @@
...
+   * Writes a pong to the connection. This will be written in a single frame, so the contents of data should be small.
+   * This method should not be used to write application data and should only be used for implementing a keep alive or
+   * to ensure the client is still responsive, see RFC 6455 Section 5.5.2. RFC 6455 Secion 5.5.3 states that pongs may
+   * be sent unsolicited in order to implement a one way heartbeat.
+   *
+   * @param data the data to write
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  WebSocketBase writePong(Buffer data);
```
vietj: I think we should be able to write a ping but not a pong.

Pong should be handled by the internal implementation when it receives a ping.

src/main/java/io/vertx/core/http/impl/ClientHandler.java
vietj: no, the case is different for websocket because there is no payload, so no need to receive specific data, in general the application is not interested to receive ping (as it would anyway response with a pong). However it is interested to send ping to check if the connection is still alive.

vietj:
ok, so forget what I said about changing HTTP/2 API, actually HTTP/2 and Websocket behavior are different : an HTTP/2 endpoint *must* respond to a PING frame with the same buffer, so the current API is fine and should not be modified.

vietj:
for Websockets:

1/ the websocket should be able to be sent ping frames
2/ the websocket should be able to have a pongHandler to receive pong frames that it sent and unsolicited pong frames
3/ the websocket should be able to send pong frames (unsolicited) - but I'm not sure this is really useful
4/ a websocket implementation should response to ping frame with a pong frame with the same content, so no need for pingHandler

what is your actual use case with this ?

vietj:
I should have look at the websocket ping/pong before and get myself a refresh on HTTP/2 as well.

src/main/java/io/vertx/core/http/WebSocketBase.java
dgoetsch: I exposed this method because this behavior is specifically allowed in RFC 6455 5.5.3 paragraph 4:

```
A Pong frame MAY be sent unsolicited.  This serves as a unidirectional heartbeat.  A response to an unsolicited Pong frame is not expected.
```

I can't imagine writing an application that only sends Pong so I'm in favor of removing it.  I'll remove it there is agreement that we should despite the specification.

dgoetsch:
@vietj When designing the API and making decisions I tried to correlate back to the specification.  I attempted to write clear javadoc where these correlations weremade, but it may not have translated as well as I had hoped.

1. This is exposed in the `writePing` method
2.  This is captured in the `pongHandler` method
3.  I personally do not understand why this is part of the specification.  One of the project goals is to remain unopinionated, and I don't know if its fair to put my opinion before the specification in this case. It will unquestionably be important to handle this behavior for the remote connection - its possible existing applications with which users will integrate use the pong in this way; this is covered by `pongHandler`.
4. agreed.

dgoetsch:
@vietj I believe I've responded to all requested changes and have reverted the Http2 changes.

src/main/java/io/vertx/core/http/WebSocketBase.java
vietj: Add a note saying that the websocket internally takes care of sending pong for ping and that the user should not use it for this purpose.

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -175,6 +201,19 @@
...
+   * on the server, and can be used by both clients and servers since the RFC 6455 Sections 5.5.2 and 5.5.3 do not
+   * specify whether the client or server sends a ping.
+   *
+   * There is no ping handler since pings should immediately be responded to with a pong with identical content
+   *
+   * @param handler the handler
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  WebSocketBase pongHandler(@Nullable Handler<Buffer> handler);
```
vietj: add a note saying that the pong handler will receive pong frames for sollicited ping but also unsolicited pong.

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
```diff
@@ -241,11 +254,13 @@ void handleFrame(WebSocketFrameInternal frame) {
...
 
   private class FrameAggregator implements Handler<WebSocketFrameInternal> {
-
     private Handler<String> textMessageHandler;
     private Handler<Buffer> binaryMessageHandler;
+    private Handler<Buffer> pongHandler;
+
     private Buffer textMessageBuffer;
     private Buffer binaryMessageBuffer;
+    private Buffer pongBuffer;
```
vietj: can pong frames be fragmented ?

vietj:
there should be a test for fragmented pong frames ? I can see code that handles this case.

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
dgoetsch: I don't think they actually can.  I'll be honest, I copy pasta'd this code to create the pongBuffer. I'll look and see if I can test anyway, or make changes if requested.

Per RFC 6455 5.5:

```
   Control frames are used to communicate state about the WebSocket.
   Control frames can be interjected in the middle of a fragmented
   message.

   All control frames MUST have a payload length of 125 bytes or less
   and MUST NOT be fragmented.
```

dgoetsch: I can remove the code, but I think I should throw an exception if the ping/pong frame being sent exceeds `WebSocketImplBase.maxWebSocketFrameSize`

dgoetsch: maybe it would be appropriate to also assert that the ping/pong frame size is less than 126 bytes.

vietj: I don't know what's best for now, but either way should have test :-)

dgoetsch: I've been testing this, It seems like Netty won't allow me to send a ping frame with more than 125 bytes of data.

Example test:
```  @Test
  public void testServerWebsocketReceivedPingExceeds125Bytes() {
    String pingBody = randomAlphaString(125);//fails if this is larger 
    Buffer pongMessage = Buffer.buffer(pingBody.length());
    Integer maxFrameSize = 256;
    server = vertx.createHttpServer(new HttpServerOptions().setIdleTimeout(1).setPort(HttpTestBase.DEFAULT_HTTP_PORT).setHost(HttpTestBase.DEFAULT_HTTP_HOST).setMaxWebsocketFrameSize(maxFrameSize));
    server.websocketHandler(ws -> {
      ws.pongHandler(pong -> {
        pongMessage.setBuffer(0, pong);
        ws.close();
      });
      ws.writeFrame(WebSocketFrame.pingFrame(Buffer.buffer(pingBody)));
    }).listen(ar -> {
      client.websocket(HttpTestBase.DEFAULT_HTTP_PORT, HttpTestBase.DEFAULT_HTTP_HOST, "/", ws -> {});
    });

    Long start = System.currentTimeMillis();
    while(System.currentTimeMillis() - start < 2000 && !pingBody.equals(pongMessage.toString())) { }
    assertEquals(pingBody, pongMessage.toString());
  }```

The "writeFrame" method appears to complete successfully, but the vertxHandler method on the client side is to  be invoked.  I haven't looked into the netty source yet, but my hunch is that the behavior is from netty and not from vertx

Comment (vietj) [reply src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java]: so we should not use a pongBuffer aggregation I think

Comment (dgoetsch) [reply src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java]: I'll have a slew of unit tests and error handling this evening.

File: src/main/java/io/vertx/core/http/WebSocketBase.java (hunk 6 above)
Comment (dgoetsch) [reply src/main/java/io/vertx/core/http/WebSocketBase.java]: I will also write a note about the 125Byte size limit :)

File: src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java [hunk 9]
```diff
@@ -288,7 +308,28 @@ private void handleTextFrame(WebSocketFrameInternal frame) {
           textMessageHandler.handle(fullMessage);
         }
       }
+    }
 
+    private void handlePongFrame(WebSocketFrameInternal frame) {
+      Buffer frameBuffer = Buffer.buffer(frame.getBinaryData());
```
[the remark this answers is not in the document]
dgoetsch: I can confirm that we will not be able to support this feature.  Netty throws an exception for fragmented control frames here:
https://github.com/netty/netty/blob/9ee4cc0ada3d8e46b139671dabfbfd35c8be3308/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocket08FrameDecoder.java#L207

I'm going to get rid of the frame handling for pong

dgoetsch:
@vietj I've added several unit tests to cover fragmentation and oversized ping/pong scenarios.

I got rid of the buffering for the pongFrames, since we can't send fragmented control frames through netty (see outdatted comment on `WebSocketImplBase`)

vietj:
@dgoetsch is it ready for final review ?

dgoetsch:
@vietj indeed, and thank you for all of the reviewing already done.

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -136,6 +137,41 @@
   WebSocketBase writeTextMessage(String text);
 
   /**
+   * Writes a ping to the connection. This will be written in a single frame, so the contents of data should be small.
+   * This method should not be used to write application data and should only be used for implementing a keep alive or
+   * to ensure the client is still responsive, see RFC 6455 Section 5.5.2.
+   *
```
vietj: add <p/>

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -136,6 +137,41 @@
...
+   * Ping frames must be less than 126 bytes (octets).
+   *
```
vietj: add <p/>

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -136,6 +137,41 @@
...
+   * There is no pingHandler because RFC 6455 section 5.5.2 clearly states that the only response to a ping is a pong
+   * with identical contents.
+   *
```
vietj: add line break tag (p)

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -136,6 +137,41 @@
...
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  WebSocketBase writePing(Buffer data);
+
+  /**
+   * Writes a pong to the connection. This will be written in a single frame, so the contents of data should be small.
+   * This method should not be used to write application data and should only be used for implementing a keep alive or
+   * to ensure the client is still responsive, see RFC 6455 Section 5.5.2.
+   *
```
vietj: add line break tag (p)

vietj: etc...

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -136,6 +137,41 @@
...
+   * with identical contents.
+   *
+   * @param data the data to write
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  WebSocketBase writePing(Buffer data);
+
+  /**
+   * Writes a pong to the connection. This will be written in a single frame, so the contents of data should be small.
```
vietj: contents -> size, please specify the max size if you can

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -136,6 +137,41 @@
...
+   * This method should not be used to write application data and should only be used for implementing a keep alive or
+   * to ensure the client is still responsive, see RFC 6455 Section 5.5.2.
+   *
+   * There is no need to manually write a Pong, as the server and client both handle responding to a ping with a pong
+   * automatically and this is exposed to users.RFC 6455 Secion 5.5.3 states that pongs may be sent unsolicited in order
```
vietj: Secion -> section

src/main/java/io/vertx/core/http/impl/ClientHandler.java
```diff
@@ -107,14 +107,12 @@ protected void handleMessage(ClientConnection conn, ContextImpl context, Channel
...
           conn.handleWsFrame(frame);
           break;
         case PING:
           // Echo back the content of the PING frame as PONG frame as specified in RFC 6455 Section 5.5.2
-          chctx.writeAndFlush(new WebSocketFrameImpl(FrameType.PONG, frame.getBinaryData()));
-          break;
-        case PONG:
-          // Just ignore it
+          chctx.channel().writeAndFlush(new PongWebSocketFrame(frame.getBinaryData().copy()));
```
vietj: use `chctx` for not `ctx.channe()` it is more efficient.

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -638,14 +639,12 @@ protected void handleMessage(ServerConnection conn, ContextImpl context, Channel
...
+          case PONG:
             conn.handleMessage(msg);
             break;
           case PING:
             // Echo back the content of the PING frame as PONG frame as specified in RFC 6455 Section 5.5.2
-            ch.writeAndFlush(new WebSocketFrameImpl(FrameType.PONG, wsFrame.getBinaryData()));
-            break;
-          case PONG:
-            // Just ignore it
+            chctx.channel().writeAndFlush(new PongWebSocketFrame(wsFrame.getBinaryData().copy()));
```
vietj: keep chctx and not `chctx.channel()`

src/main/java/io/vertx/core/http/WebSocketFrame.java
```diff
@@ -63,6 +63,26 @@ static WebSocketFrame textFrame(String str, boolean isFinal) {
   }
 
   /**
+   * Create a ping WebSocket frame.
```
vietj: mention it will be a finale frame

src/main/java/io/vertx/core/http/WebSocketFrame.java
```diff
@@ -63,6 +63,26 @@ static WebSocketFrame textFrame(String str, boolean isFinal) {
...
+    return factory.pingFrame(data);
+  }
+
+  /**
+   * Create a pong WebSocket frame.
+   *
+   * @param data the bytes for the frame
+   * @return the frame
+   */
+  static WebSocketFrame pongFrame(Buffer data) {
```
vietj: why is it useful to allow the user to create pong frames if we can't send them ? we should remove this

vietj: sorry, we allow user to write unsolicited pong frames

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
```diff
@@ -349,6 +371,18 @@ public S binaryMessageHandler(Handler<Buffer> handler) {
     }
   }
 
+  @Override
+  public WebSocketBase pongHandler(Handler<Buffer> handler) {
+    synchronized (conn) {
+      checkClosed();
+      if (frameHandler == null || frameHandler.getClass() != FrameAggregator.class) {
```
vietj: since pong frames are always final and not aggregated, I believe we don't need a frame aggregator

src/test/java/io/vertx/util/Container.java
```diff
@@ -0,0 +1,18 @@
+package io.vertx.util;
+
+public class Container<T> {
```
vietj: don't introduce this artificial class, instead use `AtomicReference` in your test which acts as a decent container, furthermore `AtomicReference` provides the same visibility semantics than volatile.

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
dgoetsch: If I don't call `channel()`, the PONG is never written and the unit tests fail.

src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
dgoetsch: doh.  I should have moved the variable to the class level.

src/test/java/io/vertx/util/Container.java
dgoetsch: Yes.

src/main/java/io/vertx/core/http/impl/ClientHandler.java
dgoetsch: See comment on HttpServerImpl

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
vietj: still we should use the chctx and not the channel for performance reason.

vietj: if that's never executed why do we have this code at all ?

src/main/java/io/vertx/core/http/impl/ClientHandler.java
vietj: please really use chctx : see http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#29.0

src/main/java/io/vertx/core/http/WebSocketFrame.java
vietj: so never mind on this one :-)

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
dgoetsch: I check this out. Not calling channel works. Using the netty class, `PongWebSocketFrame` works.

src/main/java/io/vertx/core/http/WebSocketBase.java
dgoetsch: line 164 reads `Pong frames must be less than 126 bytes (octets).`, i've moved this up to above the second paragraph.

src/main/java/io/vertx/core/http/WebSocketBase.java
dgoetsch: I don't see line breaks after the final paragraph in other docs (for example `binaryHandlerID`.  Should there be trailing line breaks?

src/main/java/io/vertx/core/http/impl/ClientHandler.java
dgoetsch: Yes, I changed this, per comment in ClientHandler; using `PongWebSocketFrame` is the real fix, not calling `channel()`

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
dgoetsch: Yes. I removed the call to `channel` and the tests still passed.  I think the change that fixed this was using the netty class `PongWebSocketFrame` rather than the vertx impl.

src/main/java/io/vertx/core/http/WebSocketBase.java
dgoetsch: I did have a note here... `Pong frames must be less than 126 bytes (octets).`  I moved it up to immediately after the first frame and changed the text to `Pong frames may be at most 125 bytes (octets).`

dgoetsch: Now I see what you mean.  Why did I just write small here?

vietj:
looks ok to me, before merging use the HTML line break tag `<b/>` in javadocs for line breaks so it displays fine.

dgoetsch:
@vietj
* I used <p> because that was what was used elsewhere in websocket base
* I don't have any multi-line comments in other files
* I did not include <p> tags after the last paragraph  because I did not see them in existing comments.
* isn't the line breaks tag <br />?

Please advise..

vietj:
ah sorry br :-) , confusion from me

dgoetsch:
Do you want me to edit existing comments in this file to also use br? What about other files? I chose p for consistency, but if we're going to depart from that I'd like to align on <br>.

vietj:
@dgoetsch I just wanted the initial separators to have `br` tag so they appear as line break in javadoc, nothing more

dgoetsch:
@vietj I really don't understand why you are asking for this or where precisely you want me to put the `<br>` tags.  I searched the project for other usages of the `<br>` tag, and I could only find them when there is an unordered list in the javadoc, so lines had to be broken up.  I don't think I have any use cases like this in the documentation I wrote.

Here is one of the comments that I wrote:

```
   * Writes a pong to the connection. This will be written in a single frame. Pong frames may be at most 125 bytes (octets).
   * <p>
   * This method should not be used to write application data and should only be used for implementing a keep alive or
   * to ensure the client is still responsive, see RFC 6455 Section 5.5.2.
   * <p>
   * There is no need to manually write a Pong, as the server and client both handle responding to a ping with a pong
   * automatically and this is exposed to users.RFC 6455 Section 5.5.3 states that pongs may be sent unsolicited in order
   * to implement a one way heartbeat.
```

Like other comments, it has a `<p>` tag between paragraphs.  I do have some extra line breaks, but ending those with `<br>` seems weird.  I don't actually care if those lines are collapsed in the javadoc - I only broke them up because I don't want the lines to be extremely long in my editor, but the browser will handle wrapping once the javadoc is rendered.

I see other comments following the exact same strategy I am, `writeBinaryMessage`, `writeTextMessage`, and `binaryHandlerID` to name a few.

What am I not understanding?  I could convert some of my comments to be more list-like, and that might make the line breaks appropriate, but at that point I should probably use a `<ul>` tag.

vietj:
looks ok

vietj:
I must admit my brain is a bit fried, it's been an intense week :-)

vietj:
thanks for the contribution @dgoetsch

dgoetsch:
I hope to have many more @vietj

PR #2112, opened 2017-09-04
Added run context close hooks when verticle deployment fails.

bhowell2:
Only needed to change two lines in source and added a test to ensure close hooks were run when verticle failed to deploy.  If anything else needs to be done for this let me know.


vietj:
can you sign the PR and the CLA ?

PR #2114, opened 2017-09-05
Fix issue 293 (https://github.com/vert-x3/issues/issues/293)

ylemoigne:
Let me know if any change is needed.


src/test/java/io/vertx/core/http/impl/Http2SettingInVertxConnectionHandlerBuilder.java
```diff
@@ -0,0 +1,38 @@
...
+    try {
+      new VertxHttp2ConnectionHandlerBuilder<Http2ClientConnection>(null) {
+        @Override
+        protected VertxHttp2ConnectionHandler<Http2ClientConnection> build(
+          Http2ConnectionDecoder decoder,
+          Http2ConnectionEncoder encoder,
+          io.netty.handler.codec.http2.Http2Settings nettySettings) throws Exception {
+
+          assertEquals(Long.valueOf(Integer.MAX_VALUE), nettySettings.get(Http2CodecUtil.SETTINGS_MAX_HEADER_LIST_SIZE));
+          throw new EndOfTest();
```
dgoetsch: why not just call `fail("message")`?

ylemoigne: If the assert is good, the test succeed. But the signature force me to return something.
I gave a `null` channel in constructor, and did not setup the builder completly, so calling `super` will throw an exception, and returning `null` will throw an exception.

dgoetsch: This seems really weird.  I would never expect a successful unit test to throw an exception, and down below I'm not sure how `EndOfTest` would be caught by the catch statement because it is not a subclass of `IllegalStateException`.

You could return a mock object, I've seen those used in a couple of places in this project. Alternatively you could also give it a channel to write to.

ylemoigne: Yes, it's weird... I have really no knowledge or vertx internal (nor netty). I've find the issue, investigate it and give my analysis when reporting.

It was politely asked if I can write a test case with the fix, and the place where the issue happens is very difficult to test. So I write it `as I can`.

The code calling the overloaded build method catch `Throwable` and wrap-it in `IllegalStateException`, that's why it work.

Now, maybe it's more clear/simple/acceptable if I just `return null` after the assert, and change the test annotation for `@Test(expected = NullPointerException.class)` ?

(I did not choose this way to express it, because throwing exception is not a part of the test contract, just a side effet of the construction used to test the value.....)

vietj:
looks good but you need to sign your commits and the Eclipse Contributor Agreement

src/main/java/io/vertx/core/http/Http2Settings.java
```diff
@@ -64,7 +64,7 @@
   /**
    * Default HTTP/2 spec value for {@link #getMaxHeaderListSize} : {@code Integer.MAX_VALUE}
```
vietj: please update the Javadoc which is now incorrect

vietj:
beside the javadoc, it looks good to me. However the PR cannot be processed because you did not sign commits and the Eclipse Contributor Agreement.

vietj:
please resubmit with IP validated

PR #2121, opened 2017-09-11
add idle timeout for sockets to leak

dinstone:
Channel missing link in the process of initialization, there may be links to the risk of leakage.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -270,6 +270,9 @@ public class HttpServerImpl implements HttpServer, Closeable, MetricsProvider {
                 if (DISABLE_HC2) {
                   configureHttp1(pipeline);
                 } else {
+                  if (options.getIdleTimeout() > 0) {
+                    pipeline.addLast("idle", new IdleStateHandler(0, 0, options.getIdleTimeout()));
+                  }
                   pipeline.addLast(new Http1xOrHttp2Handler());
                 }
               }
@@ -941,6 +944,9 @@ public class HttpServerImpl implements HttpServer, Closeable, MetricsProvider {
 
     @Override
     protected void configure(ChannelHandlerContext ctx, boolean h2c) {
+      if (options.getIdleTimeout() > 0) {
+        ctx.pipeline().remove("idle");
+      }
       if (h2c) {
         handleHttp2(ctx.channel());
       } else {


vietj:
@dinstone can you explain how you experience the issues ?

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -270,6 +270,9 @@ protected void initChannel(Channel ch) throws Exception {
                 if (DISABLE_HC2) {
                   configureHttp1(pipeline);
                 } else {
+                  if (options.getIdleTimeout() > 0) {
+                    pipeline.addLast("idle", new IdleStateHandler(0, 0, options.getIdleTimeout()));
```
dgoetsch: Please make "idle" a private static final String

dgoetsch:
Does the same thing need to be done for the client?

vietj:
can you come with a reproducer (test) that shows the issues ? we cannot patch this code without exhibiting the issue first.

dinstone:
We found this problem in a production environment, and frequent under the mobile client.In positioning, we can easily simulate and reconstruct the problem, steps are as follows:
1. The service side httpserver add idle timeout : new HttpServerOptions () setIdleTimeout (180);
2. Clients use Telnet connect http server : Telnet 10.25.76.170 8802
3. To disable the client network card, this time the telent service disconnected, and then enable the client network card;
4. After 180 seconds, on the server can also see a link from the client, for example
TCP 0 0 10.25.76.170:8802 10.25.76.15:61703 ESTABLISHED

vietj:
@dinstone ok thanks for reporting, we need an idle timeout test with just a client that uses a `NetClient` without sending data to reproduce this case to accept the patch.

dinstone:
I has an idle timeout test  with that code:
``
@RunWith(VertxUnitRunner.class)
public class HttpServerVerticleTest {

    private static Vertx vertx = Vertx.vertx();

    @Test
    public void testIdleTimeout(TestContext ctx) {
        final Async async = ctx.async();

        HttpServerOptions options = new HttpServerOptions().setIdleTimeout(5);
        vertx.createHttpServer(options).requestHandler(r -> {
            r.response().end("OK");
        }).listen(8081, server -> {
            NetClient netClient = vertx.createNetClient();
            netClient.connect(8081, "127.0.0.1", ar -> {
                if (ar.failed()) {
                    ctx.fail();
                    async.complete();
                    return;
                }

                NetSocket netSocket = ar.result();
                netSocket.handler(b -> {
                    String actual = b.toString();
                    System.out.println(actual);
                    String expected = "HTTP/1.1 200 OK\r\nContent-Length: 2\r\n\r\nOK";
                    ctx.assertEquals(expected, actual);

                    async.complete();
                });

                vertx.setTimer(10000, tid -> {
                    netSocket.write("GET / HTTP/1.1\r\n\r\n");
                });
            });
        });

        async.await();
    }

}
``

I set the idle timeout of the httpserver to 5 seconds, and then use the netclient connect to the server but not sending data, sending HTTP requests after 10 seconds, but the success have received a response.Obviously this is not to be expected.

dgoetsch:
@dinstone Reading your test, it looks like it completes successfully once it receives a message.  Am I reading correctly?  Wouldn't we expect it to complete by being closed (`closeHandler` invoked), and receiving a message should cause a failure?

Also, a test similar to this can probably go in Http1xTest, but all the unit tests in this project are in vanilla JUnit.  There are some nice methods in the tests to handle waiting and failing the test depending on conditions, but not all of the fancyness of

dinstone:
@dgoetsch Your reading of the test code is correct, but is not clear for the meaning of the test code.
I write this test shows that the idle timeout in the absence of disable h2c is a potential problem - don't release the connection.We expect after a TCP connection is established, if not send data, as long as idle timeout, we should close the connection, rather than not to close link now.

dinstone:
We should be in the TCP layer set idle timeout, rather than the HTTP this layer.In production, we found that many did in the mobile environment request only TCP connection is established, it somehow lost link, leading to the service side hold a lot of established TCP link cannot be released.

So I send this PR, modify the idle timeout setting in TCP, initialization of the Channel, to ensure that lost link correctly release.

vietj:
I think the IDLE timeout handler should be set at the initialization of the channel to be effective from the beginning. All cases should be tested, some are already.

vietj:
it has been fixed here https://github.com/eclipse/vert.x/issues/2138 can you check on your side it's ok ?

dinstone:
@vietj the fix is effective. thanks.

PR #2123, opened 2017-09-11
#2122 add WriteStream.end(Handler) 

### LINKED ISSUE #2122: WriteStream is missing callback for asynchronous end() completion
So far [WriteStream](/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/streams/WriteStream.java#L61) has only  synchronous 'end()` methods:
```
 void end();
 void end(T t) {
```
I.e. after I called `end()` it is safe to assume that stream is actually closed.

This means that [AsyncFile](/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java#L122) cannot properly implement this interface cause in order to end asyncfile's writestream you need to close file, which is asynchronous operation. This causes troubles when you download url to file and try to use file immediately after receiving download completion callback: [web-client #691](/vert-x3/vertx-web/issues/691) 
 
**Suggestion:**
1. Add method `void end(Handler<AsyncResult<Void>> handler)` to `WriteStream` interface (with default implementation to not break existing code).
2. Make AsyncFile use this method for asynchronously closing underlying file.

This will open a way to actually fix problem with web-client (by enabling `StreamingBodyCodec` to wait for actual `end()` completion event.

I made aforementioned changes locally. Can open a PR. Need a comment from maintainers whether it's a proper fix or I am missing something.

mkulak:
Closes #2122 

All work is done, **ready for review**


src/main/java/io/vertx/core/streams/WriteStream.java
```diff
@@ -68,6 +70,11 @@ default void end(T t) {
     end();
   }
 
+  default void end(Handler<AsyncResult<Void>> handler) {
+    end();
+    handler.handle(Future.succeededFuture());
+  }
+
   /**
```
juanavelez: Would you not need to handle the case where the end() call fails and therefore you need to pass a Future.failedFuture() to the handler?

mkulak: Since it's default implementation I tried to preserve original behaviour of `end()`. If it throws - caller get the exception. Same here in `end(Handler)`. Do you think I should do something like:
```java
  default void end(Handler<AsyncResult<Void>> handler) {
    Future<Void> fut = Future.<Void>future();
    try {
      end();
      fut.complete();
    } catch (Exception e) {
      fut.fail(e);  
    }
    handler.handle(fut);
  }

```

juanavelez: yes

vietj:
what's the motivation for this ?

vietj:
the title is wrong it should be `end` which is not the same thing at all

src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java
```diff
@@ -124,6 +124,11 @@ public void end() {
   }
 
   @Override
+  public void end(Handler<AsyncResult<Void>> handler) {
+    close(handler);
```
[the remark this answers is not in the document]
vietj: `end` and `close` are not the same thing.

- end means that no more item will be emitted from the perspective of the emitter
- close disposes the resource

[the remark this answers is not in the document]
vietj: handlers would rather used when something may end the stream on your behalf, usually the owner of the stream fully controls it and is ends it explicitly

src/main/java/io/vertx/core/streams/WriteStream.java
vietj: we should not provide a default implementation because `end()` is asynchronous, so the default implementation cannot assume that the callback can be called after the end() method call

src/main/java/io/vertx/core/streams/WriteStream.java
```diff
@@ -68,6 +70,11 @@ default void end(T t) {
     end();
   }
 
+  default void end(Handler<AsyncResult<Void>> handler) {
```
vietj: Missing javadoc.

src/main/java/io/vertx/core/streams/WriteStream.java
vietj: in addition, the write stream may be full and the actual end completion might be delayed until the stream is drained.  Check the `AsyncFile` implementation that delays the actual close until the pending buffers are written.

mkulak: Ok, I get: `end()` is asynchronous, it just didn't signal the actual completion. I think this should be reflected in the javadoc of this method (much like in one's for `write()`).

But original problem still stays: some classes (like `StreamingBodyCodec` in vertx-web-client) need to get notification when `WriteStream` end happens. And there is no way to achieve this with the current interface. That's why I'm proposing to change the interface: add `WriteStream.end(Handler)`. To me this looks like the most consistent way of solving the problem. But I admit that I'm not very familiar with Vert.x internals. Is there better way?
@vietj 
@dgoetsch

src/main/java/io/vertx/core/streams/WriteStream.java
mkulak: I'll address this one when we will reach consensus on whether this method should be added at all.

src/main/java/io/vertx/core/streams/WriteStream.java
vietj: can you provide a reproducer for the WebClient issue ?

dgoetsch: @mkulak that sounds like precisely the use case for an `endHandler`.

vietj: before speculating on a fix, it would be good to have a reproducer for the `WebClient` issue

mkulak: [Real world test](https://gist.github.com/mkulak/764536e26ade3f26e1daf38460b413d3) (in Kotlin)
Not sure if it will fail 100% of time on every machine. 

[Simulated test](https://gist.github.com/mkulak/764536e26ade3f26e1daf38460b413d3) (in Java)
It's synthetic but always fails. For this test I make my own implementation of `AsyncFile` with `end()` that calls `close()` and `close()` that calls `close((Handler) null)` (much like real `AsyncFileImpl`). But `close(Handler)` actually delays it's execution for 3 seconds.

src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java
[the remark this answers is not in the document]
mkulak: > handlers would rather used when something may end the stream on your behalf

Same argument is applicable for `AsyncFile` then. But for some reason `close(Handler)` is how `AsyncFileImpl` signals completion of it's close.

vietj:
it seems indeed useful to provide an overloaded completionHandler `WriteStream#end()` operation.

I believe we shall not provide a default implementation for it as `end()` is asynchronous, which means that we need to check the various project of the Vert.x stack implement it.

vietj:
So we should have tested implementations in Vert.x core for `HttpServerResponse`, `HttpClientRequest`, `NetSocket`, `WebSocket` and `MessageProducer`.

mkulak:
@vietj I will try to implement it in this PR. But I'm not sure that I will succeed.

mkulak:
All the implementations for `vertx-core` classes are here. 
Currently missing tests for:
`PacketWriteStreamImpl.java`
`MessageProducerImpl.java`
`Http2ServerResponseImpl.java`
`HttpClientRequestBase.java`
`HttpServerResponseImpl.java`
`VertxHttp2NetSocket.java`
`WebSocketImplBase.java`

vietj:
what's the currenet status ? I see progress, however I would like to have it ready for 3.5.0

mkulak:
@vietj 
Every implementation is in place, 3 tests left: 
Http2ServerResponseImpl.java
HttpServerResponseImpl.java
HttpClientRequestBase.java

What is the deadline for 3.5.0?

vietj:
I will review it

src/main/java/io/vertx/core/eventbus/impl/MessageProducerImpl.java
```diff
@@ -42,6 +43,7 @@
   private int maxSize = DEFAULT_WRITE_QUEUE_MAX_SIZE;
   private int credits = DEFAULT_WRITE_QUEUE_MAX_SIZE;
   private Handler<Void> drainHandler;
+  private Handler<Void> endHandler;
```
vietj: we should not have the `endHandler` to be a field of this class

src/main/java/io/vertx/core/eventbus/impl/MessageProducerImpl.java
```diff
@@ -127,10 +134,19 @@ public void end() {
...
 
   @Override
+  public synchronized void end(Handler<AsyncResult<Void>> handler) {
+    endHandler(v -> handler.handle(Future.succeededFuture()));
+    end();
+  }
+
+  @Override
   public void close() {
     if (creditConsumer != null) {
```
vietj: the code of close should be moved to `end(Handler<AsyncResult<Void>)` instead, so we should have something like

```
public synchronized void end(Handler<AsyncResult<Void>> completionHandler) {
   if (creditConsumer != null) {
      creditConsumer.unregister();
   }
   if (endHandler != null) {
      endHandler.handler(Future.succeededFuture());
   }
}
```

vietj: note: we should somehow take in account the pending queue before doing the effective end... (another issue)

vietj:
I haven't look at the entire PR, but I can see that calling end(handler) sets an endHandler field before calling end.

The application may have set endHandler and technically there is no need to set a field on the closed stream to achieve it.

I've shown an example for `MessageProducerImpl`, make sure the other don't follow this pattern

src/main/java/io/vertx/core/eventbus/impl/MessageProducerImpl.java
mkulak: In order to address issue with pending queue that you described below we need to keep this field.

src/main/java/io/vertx/core/eventbus/impl/MessageProducerImpl.java
mkulak: I made `MessageProducerImpl` wait till pending queue is empty before invoking endHandler(). Such implementation kinda invalidates your previous comment: logic is split between `close()` and `handleClosed()`

mkulak:
> calling end(handler) sets an endHandler field before calling end.
That's right. This is what `end(Handler)` is in a nutshell: `endHandler()` + `end()`.
 
> The application may have set endHandler
If you set `endHandler` and then call `end(Handler)` with another handler you will effectively overwrite previously set `endHandler`. Same as you manually set `endHandler` twice and then call `end`.

> there is no need to set a field on the closed stream to achieve it
In many cases we need to delay invocation of endHandler till some pending writes are completed. For this we need field.

> make sure the other don't follow this pattern
Other implementations follows exact same pattern. If you see this as bad design - please explain alternative solution for me. 
I would like to remind you why this issue exists: interface `WriteStream` doesn't  provide any callbacks for when `end` is actually completes. You can `end` it but you cannot receive notification when it's done.

src/main/java/io/vertx/core/eventbus/impl/MessageProducerImpl.java
vietj: drop it, we'll honor the pending queue later in another commit, it was not honoured before anyway

mkulak: I've already implemented it and made test for it. Should I rollback my last commits?
Regarding other points: I addressed it in [this comment](https://github.com/eclipse/vert.x/pull/2123#issuecomment-332031510). I would really like to know your position.

vietj:
@mkulak I think you are confusing completion handler (i.e Handler<AsyncResult> on a method) that signals the completion of this method and the general end/close handlers that are set by the application to signal when the resource is ended / closed

vietj:
for instance if I look at `WebSocketImplBase`, you are doing this:

```
public void end(Handler<AsyncResult<Void>> handler) {
   synchronized (conn) {
      endHandler(v -> handler.handle(Future.succeededFuture()));
      end();
   }
}
```

instead it should be:

in `WebSocketBase`:

```
  public void close(Handler<AsyncResult<Void>> completionHandler) {
    synchronized (conn) {
      checkClosed();
      conn.close(completionHandler);
      cleanupHandlers();
    }
  }

  public void end(Handler<AsyncResult<Void>> completionHandler) {
    close(completionHandler);
  }
```

and in `io.vertx.core.net.impl.ConnectionBase`

```
  public void close(Handler<AsyncResult<Void>> completionHandler) {
    // make sure everything is flushed out on close
    endReadAndFlush();
    ChannelFuture fut = chctx.channel().close();
    fut.addListener(v -> {
      context.executeFromIO(() -> {
        if (v.isSuccess()) {
          completionHandler.handle(Future.succeededFuture());
        } else {
          completionHandler.handle(Future.failedFuture(fut.cause()));
        }
      });
    });
  }
```

vietj:
ping @mkulak

mkulak:
@vietj Sorry for period of silence, didn't have time to OSS.

I kindly ask for additional explanation otherwise I'll not be able to proceed with this PR.
Let me write down my thought on this subject and you will tell me where my understanding is incorrect.

There are classes that model some resource: socket or file. Normally you can read from and write to this resource (with backpressure). Afterward you should close the resource (close usually happens asynchronously with respect to current write queue) . Examples are: `AsyncFileImpl`, `NetSocketImpl`, `WebSocketImplBase` (and more). 

There is `WriteStream` concept - something you can write into with backpressure. It serves to abstract out differences in previously mentioned classes and allows to create common read/write code (example: `BodyCodec.pipe`) that can work with any of those.
`WriteStream` also allows you to signal stream completion - you tell consumer that no more elements will be emitted (again, this happens asynchronously cause you need to wait till current write queue will be drained).

Observation: in most cases calling `end()` actually closes the resource (no need to leave resource hanging if you will not use it anymore).

When it comes to listening to end/close events things get complicated pretty quickly.
* `AsyncFileImpl` has `endHandler` (which is invoked only for reading from file but not for writing)  and `closedDeferred` (invoked when file is closed).
* `HttpResponseImpl` has `endHandler` (invoked both when ending and when closing response) and `closeHandler` (invoked on closing). 
* `WebSocketImplBase` has `endHandler` and `closeHandler`(both are invoked only on close and are in fact totally redundant) 
* `MessageProducerImpl` doesn't have any handlers (and no way to listen when it's actually closed). This is because MessageProducer doesn't really "own" underlying socket and if you close producer socket won't be closed.

It's kinda hard to find consistent rules how `end(Handler)` should be implemented for all cases. I tried to reuse existing handlers as much as possible to not overcomplicate things (simplest thing is to assume that `end(Handler)` equals to calling `endHandler(handler)` and then `end()`). 

How I understand your solution: you propose to add one additional "internal" close handler (without introducing additional fields where possible). In my opinion this will complicate implementation and I would like to know:
* Do I understand your proposal correctly?
* What benefit does it have over current implementation from this PR?

mkulak:
ping @vietj

vietj:
@mkulak will resume that soon, busy on 3.5.0 release atm

mkulak:
@vietj please give me some advice on this PR.
I'm trying to understand how to change this code in order to fix the original issue AND satisfy design requirements so that this PR can be merged.

vietj:
will have a look soon for 3.5.1

PR #2125, opened 2017-09-12
Issue 2096: queryString key should be case sensitive

jsight:
Relates to: https://github.com/eclipse/vert.x/issues/2096

### DIFF
--- a/src/main/java/io/vertx/core/CaseSensitiveMultiMap.java
+++ b/src/main/java/io/vertx/core/CaseSensitiveMultiMap.java
@@ -1,0 +1,418 @@
+/*
+ * Copyright (c) 2011-2013 The original author or authors
+ * ------------------------------------------------------
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Apache License v2.0 which accompanies this distribution.
+ *
+ *     The Eclipse Public License is available at
+ *     http://www.eclipse.org/legal/epl-v10.html
+ *
+ *     The Apache License v2.0 is available at
+ *     http://www.opensource.org/licenses/apache2.0.php
+ *
+ * You may elect to redistribute this code under either of these licenses.
+ */
+
+package io.vertx.core;
+
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Consumer;
+
+/**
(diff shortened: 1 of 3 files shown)


vietj:
this should have the same implementation than `CaseInsensitiveHeaders` and only change the equality check somehow.

jsight:
@vietj - That is basically what it does, just without a refactoring job to share the common bits of the code. If that approach is basically right, though, I will fix that. :)

jsight:
@vietj - I have refactored it a bit to completely share all of the code, other than the equality checks. This is really what I had intended to send the first time.

src/main/java/io/vertx/core/AbstractMultiMap.java
```diff
@@ -0,0 +1,416 @@
+/*
+ * Copyright 2017 jsight.
```
jponge: RedHat, Inc.

jsight: Fixed!

src/main/java/io/vertx/core/AbstractMultiMap.java
```diff
@@ -0,0 +1,416 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Consumer;
+
+/**
+ *
+ * @author <a href="mailto:jsightle@redhat.com">Jess Sightler</a>
```
vietj: you need rather credit the original author of the CaseInsensitiveHeaders class for this as most of this class is borrowed from it.

src/main/java/io/vertx/core/AbstractMultiMap.java
```diff
@@ -0,0 +1,416 @@
...
+ */
+public abstract class AbstractMultiMap implements MultiMap {
```
vietj: rather name it MultiMapBase rather and make it package-private so we don't expose it

vietj:
despite my comments, i believe we should not use inheritance for this and a simple boolean on the original class that say wether it's sensitive or not is enough.

jsight:
It will have to be renamed in that case, IMO. What name do you suggest?

vietj:
I think we cannot rename the existing class, so here is what we should do:

1/ make the base class MultiMapBase with all the logic overridable
2/ make CaseInsensitiveMultiMap extends it and override the hash/eq logic

Add a static create method in MultiMap#multimap() that returns the base implementation and is a general purpose case sensitive multimap.

PR #2129, opened 2017-09-15
Provide an exception handler on HttpServer and NetServer to process uncaught connection errors

### LINKED ISSUE #2128: Provide an exception handler on HttpServer and NetServer to process uncaught connection errors
motivation: currently the `HttpServer` forwards connection errors to the `HttpConnection` associated with the channel. It happens that the channel can fail before an `HttpConnection` is created, e.g SSL handshake failure. Application can have an interest to handle such failures for auditing purposes. In addition an existing internal handler already exists for testing purposes. The same can be done for `NetServer`.

changes: add an exception handler on `HttpServer` to process channel errors before an `HttpConnection` is created. Make sure also that `HttpConnection` is created after the SSL handshake for HTTP/1 as it was for HTTP/2 due to ALPN (the ALPN negotiation defers the creation of `HtppConnection` after the SSL handshake when the negotiated protocol is determined), which implements a consistent behavior. The same for `NetServer`.

vietj:
fixes #2128

motivation: currently the `HttpServer` forwards connection errors to the `HttpConnection` associated with the channel. It happens that the channel can fail before an `HttpConnection` is created, e.g SSL handshake failure. Application can have an interest to handle such failures for auditing purposes. In addition an existing internal handler already exists for testing purposes. The same can be done for `NetServer`.

changes: add an exception handler on `HttpServer` to process channel errors before an `HttpConnection` is created. Make sure also that `HttpConnection` is created after the SSL handshake for HTTP/1 as it was for HTTP/2 due to ALPN (the ALPN negotiation defers the creation of `HtppConnection` after the SSL handshake when the negotiated protocol is determined), which implements a consistent behavior. The same for `NetServer`.

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServer.java
+++ b/src/main/java/io/vertx/core/http/HttpServer.java
@@ -73,6 +73,16 @@ public interface HttpServer extends Measured {
   HttpServer connectionHandler(Handler<HttpConnection> handler);
 
   /**
+   * Set an exception handler called for socket errors happening before the HTTP connection
+   * is established, e.g during the TLS handshake.
+   *
+   * @param handler the handler to set
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  HttpServer exceptionHandler(Handler<Throwable> handler);
+
+  /**
    * Return the websocket stream for the server. If a websocket connect handshake is successful a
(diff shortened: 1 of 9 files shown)


tsegismont:
LGTM

Issue #2118, opened 2017-09-08
Feature request: Expose netty's SSL Handshake Timeout to the Vertx user


kpachar:
Hello,
This is a feature request; it will be nice to be able to set the ssl handshake timeout.

Thank you!

https://github.com/eclipse/vert.x/blob/38a19afe50917de3783bdd1512f2314b71cc1750/src/main/java/io/vertx/core/net/impl/NetClientImpl.java#L208

https://netty.io/4.0/api/io/netty/handler/ssl/SslHandler.html#setHandshakeTimeout-long-java.util.concurrent.TimeUnit-

tsegismont:
Isn't there a default timeout? Why isn't sufficient?

2017-09-08 17:17 GMT+02:00 Pradyumna Achar <notifications@github.com>:

kpachar:
@tsegismont Yes, there is a default, but it is insufficient for certain very low bandwidth networks.

tsegismont:
Would you like to contribute the change?

2017-09-10 17:29 GMT+02:00 Pradyumna Achar <notifications@github.com>:

vietj:
I'm wondering if that's not conflicting with the current default idle timeout, i.e the idle timeout set before the handshake could fill the same purpose

BillyYccc:
It looks like that the present idleTimeout implement only can be triggerd by the scenario neither read nor write was performed. https://github.com/eclipse/vert.x/blob/af1c846ea46b499b249c6829c4c32385b3022a23/src/main/java/io/vertx/core/net/impl/NetServerImpl.java#L146
I'm confused how to make idle timeout fill the same purpose as handshake timeout, could you give more specific details?

vietj:
I think actually it's better to differentiate both like you did, so you should not both of idle timeout

luengnat:
@vietj can you target this for 3.6.0?

BillyYccc:
Feel sorry for the delay... Please take a look at https://github.com/eclipse-vertx/vert.x/pull/2714.

PR #2131, opened 2017-09-18
add support for sslHandshakeTimeout in NetServer and NetClient Options

BillyYccc:
Fix Issue #2118 
add implement and some unit tests for sslHandshakeTimeout in NetClient Options
add an API to expose netty's SSL Handshake Timeout
Signed-off-by: BillyYccc <billy112487983@gmail.com>

### DIFF
--- a/src/test/java/io/vertx/test/core/NetTest.java
+++ b/src/test/java/io/vertx/test/core/NetTest.java
@@ -64,6 +64,7 @@ import io.vertx.core.net.ProxyType;
 import io.vertx.core.net.SSLEngineOptions;
 import io.vertx.core.net.SelfSignedCertificate;
 import io.vertx.core.net.SocketAddress;
+import io.vertx.core.net.TCPSSLOptions;
 import io.vertx.core.net.impl.SocketAddressImpl;
 import io.vertx.core.streams.ReadStream;
 import io.vertx.test.core.tls.Cert;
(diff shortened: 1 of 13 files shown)


vietj:
10 seconds is a big long for this test

vietj:
I'm wondering if that's not conflicting with the current default idle timeout, i.e the idle timeout set before the handshake could fill the same purpose

vietj:
also perhaps if we accept it, it should also apply not only to `NetClient`

src/test/java/io/vertx/test/core/NetTest.java
```diff
@@ -1261,6 +1288,20 @@ public void testSpecificTlsProtocolVersion() throws Exception {
   }
 
   @Test
+  // test client SSL handshakeTimeout option
+  public void testClientSslHandshakeTimeout() throws Exception {
+    testTLS(Cert.NONE, Trust.NONE, Cert.SERVER_JKS, Trust.NONE, false, true, true, false, new String[0],
```
vietj: I think SSL timeout should have specific test and not reuse the general one, for instance for server:

- setup an SSL server with a timeout with a small value for instance 400ms
- run a NetClient that connects to the server and does nothing
- wait for the SSL timeout to fail, i.e the client will get an error on connect

BillyYccc: Thanks for your helpful suggestions! I'm working on it

BillyYccc:
@vietj I think I'm ready now, but current version of netty does not support the setHandshakeTimeout in server side, see issue in **Netty** [issue #7230](https://github.com/netty/netty/issues/7230). We may need to wait for the fix of that issue in **Netty** and upgrade our dependency to let this make sense.

vietj:
@BillyYccc yes, I've followed this issue on the Netty side as well

BillyYccc:
Hi! After [issue #7230](https://github.com/netty/netty/issues/7230) of Netty has been fixed in Netty 4.1.17.Final. Recently Vert.x has upgraded netty dependency to 4.1.19. So this issue can be scheduled now. I have rewritten the commits and messages to make it cleaner.

BillyYccc:
Duplicate of https://github.com/eclipse-vertx/vert.x/pull/2714.

PR #2146, opened 2017-09-25
Serialize messages sent to the connection in the order of the synchronization monitor

vietj:
motivation: currently the `ConnectionBase` class does not  serialize the messages written with `writeToChannel` to the underlying channel when a message is written outside of the event loop followed by a message written from the event loop thread. This can cause out of order issues when ordering is assumed by the Vert.x code using `ConnectionBase`, for instance calling `HttpClientRequest#end()` outside of the event loop sends an last `HttpContent` message to the channel then release the connection to the pool (with pipelining), this connection might be provided then to a waiter that uses it and send pipelined messages, e.g an `HttpRequest` message.

changes: change the implementation of `writeToChannel` and write the message to the channel when the current thread is the event loop or there are no asynchronous message _write in progress_ otherwise a write task is scheduled causing further messages to be scheduled as well. Two sequential calls to `writeToChannel` (the method is synchronized) will have their messages in the same order the synchronization monitor (the class) allowed. In particular an asynchronous task calling `writeToChannel` started after a call to `writeToChannel` will have its message sent after.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
@@ -484,7 +484,7 @@ public class HttpServerResponseImpl implements HttpServerResponse {
       RandomAccessFile raf = null;
       try {
         raf = new RandomAccessFile(file, "r");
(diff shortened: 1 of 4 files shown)


tsegismont:
LGTM

PR #2147, opened 2017-09-26
SNI for TrustOptions

theoweiss:
This is the successor of #2040


vietj:
it looks good from my perspective, I need to make a more thorough review before merging.

vietj:
ip validation fails : please read the ECA

theoweiss:
Sorry for the delay, I was knocked out by a cold. I will try to fix the ip-validation now. As far as I understand I will have to squash the commits and add another Signed-off-by footer?

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
```diff
@@ -137,12 +135,20 @@ public static KeyStoreHelper create(VertxInternal vertx, TrustOptions options) t
...
     while (en.hasMoreElements()) {
       String alias = en.nextElement();
       Certificate cert = ks.getCertificate(alias);
+      if (ks.isCertificateEntry(alias) && ! alias.startsWith(DUMMY_CERT_ALIAS)){
+        KeyStore keyStore = KeyStore.getInstance("jks");
+        keyStore.load(null, null);
+        keyStore.setCertificateEntry("cert-1", cert);
+        trustMgrMap.put(alias, keyStore);
+      }
+      // TODO this should only be done if: ks.isKeyEntry(alias)
```
theoweiss: I've added this TODO to start the discussion with @vietj, it's off topic regarding the core feature of this PR, but I think the sni name gathering for server certificates shouldn't be done for CA certificate entries. What do you think.

theoweiss:
@vietj ip validation succeeded. I've added a further comment to the "files changed tab".

src/main/java/io/vertx/core/net/TrustOptions.java
```diff
@@ -44,4 +45,25 @@
...
   }
+
+  /**
+   * Returns a function that maps SNI server names to a {@link TrustManagerFactory} instance.
+   *
+   * The returned {@code TrustManagerFactory} must already be initialized and ready to use.
+   *
+   * The mapper is only used when the server has SNI enabled and the client indicated a server name.
+   * <p/>
+   * The returned function may return null in which case {@link #getTrustManagerFactory(Vertx)} is used as fallback.
```
vietj: {@code null}

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
vietj: can you elaborate ?

vietj:
I think it's ok, but we need documentation.

vietj:
ping @mkulak

vietj:
ping @theoweiss

theoweiss:
@vietj give me a hint were to add the docs.

vietj:
https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/net/package-info.java#L666

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
theoweiss: If the KeyStore is created from JKSOptions with a user provided TrustStore the store could contain CertificateEntries and KeyEntries at once. The loop currently also adds entries to the mgrMap for CNs from CertificateEntries.

theoweiss:
I've added some documentation.

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
vietj: I don't get it : the JKSOptions are different fields for keystore (`keyCertOptions`) and truststore (`trustOptions`) on `TCPSSLOptions`

theoweiss: A vert.x user creates a Java key store with a CertificateEntry and and a KeyEntry like this pseudo code and stores it in "myKeyStore" file:

```
 keyStore = KeyStore.getInstance("jks");
 keyStore.load(null, null);
 keyStore.setCertificateEntry("trustedCA", caCertBuf);
 PrivateKey key = privateKey;
 Certificate[] chain = certChain;
 keyStore.setEntry("keyCert", new KeyStore.PrivateKeyEntry(key, chain), new KeyStore.PasswordProtection("dummy".toCharArray()));
```

and uses this key store for TrustOptions and KeyCertOptions:

```
    NetServerOptions netServerOptions = new NetServerOptions().setTrustOptions(new JksOptions().setPath("myKeyStore"))
        .setKeyCertOptions(new JksOptions().setPath("myKeyStore"));
    vertx.createNetServer(netServerOptions);
```

vietj: ah ok but in reality those can be two different files, that was confusing

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
```diff
@@ -236,6 +242,20 @@ public X509KeyManager getKeyMgr(String serverName) {
     return getKeyMgrFactory().getKeyManagers();
   }
 
+  public TrustManagerFactory getTrustMgr(String serverName)  {
+    try {
+      KeyStore trustStore = trustMgrMap.get(serverName);
```
vietj: I would prefer this code to be done in the `KeyStoreHelper` constructor and keep a `Map<String, TrustManagerFactory >` so can be detected at this moment and this is also like how we do for the SNI key manager mapping.

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
vietj: you can implement the todo condition imho, it's better to do it.

theoweiss: done

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
theoweiss: thats good

theoweiss: is it ok now?

src/test/java/io/vertx/test/core/tls/Trust.java
```diff
@@ -29,5 +43,90 @@
...
+      index = end + 1;
+      pems.add(Base64.getDecoder().decode(content));
+    }
+    if (pems.isEmpty()) {
+      throw new RuntimeException("Missing " + beginDelimiter + " delimiter");
+    }
+    return pems;
+  }
+
+  static X509Certificate[] loadCerts(Buffer buffer) throws Exception {
```
vietj: isn't this code already in KeyStoreHelper ?

src/test/java/io/vertx/test/core/tls/Trust.java
```diff
@@ -29,5 +43,90 @@
...
+      }
+      ByteArrayOutputStream out = new ByteArrayOutputStream();
+      keyStore.store(out, password.toCharArray());
+      return Buffer.buffer(out.toByteArray());
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  static List<byte[]> loadPems(Buffer data, String delimiter) throws IOException {
```
vietj: isn't this code already in KeyStoreHelper ?

src/test/java/io/vertx/test/core/tls/Trust.java
```diff
@@ -29,5 +43,90 @@
...
+    certMap.put("xxx", "src/test/resources/tls/root-ca/ca-cert.pem");
+    return new JksOptions().setValue(loadCAs(certMap, DUMMY_PASSWORD)).setPassword(DUMMY_PASSWORD);
+  };
+  Trust<TrustOptions> SNI_SERVER_OTHER_CA_FALLBACK = () -> {
+    HashMap<String, String> certMap = new HashMap<>();
+    certMap.put("xxxx", "src/test/resources/tls/other-ca/ca-cert.pem");
+    return new JksOptions().setValue(loadCAs(certMap, DUMMY_PASSWORD)).setPassword(DUMMY_PASSWORD);
+  };
+
+  static Buffer loadCAs(Map<String, String> certMap, String password) {
```
vietj: isn't this code already in KeyStoreHelper ?

src/test/java/io/vertx/test/core/tls/Trust.java
theoweiss: Yes, these methods are more or less unchanged copies from __private__ methods in the KeyStoreHelper, which are not available for the tests. May be it would be nicer to share these methods with an Utility class. I just needed a way to create jks stores for the tests on the fly, also because it's self explaining how the key store is structured (the alias -> cert mapping).  Knowing the structure is important for understanding what the sni truststore test tests. Alternatively the keytool can be used to create the jks stores for the tests and add the files to the test resources. But I think then it's harder to get an idea what some of test do.

vietj: currently keystores/truststores are created manually (check ssl.txt) and it is very fine this way, is that something we can use instead ?

theoweiss: Yes sure. I will change it.

theoweiss: done

theoweiss:
@vietj when do you plan to proceed? After 3.5.0 release?

vietj:
yes after 3.5.0

theoweiss:
Any chance to get this into 3.5.1?

vietj:
yes it will

src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
```diff
@@ -71,6 +68,7 @@
 
   // Dummy password for encrypting pem based stores in memory
   private static final String DUMMY_PASSWORD = "dummy";
+  private static final String DUMMY_CERT_ALIAS = "cert-";
```
vietj: it's not really dummy is it ?

src/main/java/io/vertx/core/net/TrustOptions.java
```diff
@@ -44,4 +45,25 @@
...
+   * The returned {@code TrustManagerFactory} must already be initialized and ready to use.
+   *
+   * The mapper is only used when the server has SNI enabled and the client indicated a server name.
+   * <p/>
+   * The returned function may return {@code null} in which case {@link #getTrustManagerFactory(Vertx)} is used as fallback.
+   *
+   * @param vertx the vertx instance
+   * @return the trustManager
+   */
+  default Function<String, TrustManagerFactory> trustManagerMapper(Vertx vertx) throws Exception {
```
vietj: we should rather return null when not found and let the SSLHelper use the default value like it is for the `KeyCertOptions`

vietj: given that the SSLHelper actually seems to do that

PR #2149, opened 2017-09-26
Ability to reload CRL files/buffers at runtime

manish-panwar:
This PR contains following changes:

1 - Ability to reload CRL from file path or buffer at runtime in HttpClient
2 - Ability to reload CRL from file path or buffer at runtime in HttpServer

See - https://groups.google.com/forum/#!topic/vertx/fLQPPAxkftM

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/SSLHelper.java
+++ b/src/main/java/io/vertx/core/net/impl/SSLHelper.java
@@ -18,6 +18,7 @@ package io.vertx.core.net.impl;
 
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.handler.ssl.*;
+import io.vertx.core.Vertx;
 import io.vertx.core.VertxException;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.ClientAuth;
@@ -39,11 +40,7 @@ import io.vertx.core.net.TrustOptions;
 
 import javax.net.ssl.*;
 import java.io.ByteArrayInputStream;
-import java.security.cert.CRL;
-import java.security.cert.Certificate;
-import java.security.cert.CertificateException;
-import java.security.cert.CertificateFactory;
-import java.security.cert.X509Certificate;
+import java.security.cert.*;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
@@ -133,6 +130,7 @@ public class SSLHelper {
   private boolean trustAll;
   private ArrayList<String> crlPaths;
   private ArrayList<Buffer> crlValues;
+  private List<CRL> crls;
   private ClientAuth clientAuth = ClientAuth.NONE;
   private Set<String> enabledCipherSuites;
   private boolean openSsl;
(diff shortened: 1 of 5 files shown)


src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -133,6 +130,9 @@ public static SSLEngineOptions resolveEngineOptions(TCPSSLOptions options) {
   private boolean trustAll;
   private ArrayList<String> crlPaths;
   private ArrayList<Buffer> crlValues;
+  // TODO - HTTP Server has own instance of SSL Helper. How do we invalidate/reload CRLs for
+  // TODO - all instances of HTTP Server. For now making it static. Lets see what Julien think.
+  private static List<CRL> crls;
```
manish-panwar: Right now there is open question on how we invalidate in-memory CRL in HttpServer/HttpClient.

HttpServer instances (and SslHelper as well) are created per-thread, so user will have to ensure that they store HttpServer instance somewhere and call the reload method on each one of them in loop. I need recommendation from Vertx team on this. Right now I have kept it static for my testing purpose but it can't be kept static because SSL helper is shared among HttpClient & HttpServer.

manish-panwar:
I am also wondering if I should add unit test. I noticed that we aren't using Mockito etc which makes testing easy. I wonder if I should rather add a integration test.

vietj:
@manish-panwar yes you should

vietj:
there should be one test that do:

1. connects with a valid certificate
2. update the CRL list to add this certificate
3. try to reconnecti with the valid certificate and have an handshake failure

manish-panwar:
@vietj - Please bear with me on Integration tests. I got pulled into something at work, I will work on it ASAP.

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -321,27 +321,51 @@ private TrustManagerFactory getTrustMgrFactory(VertxInternal vertx) throws Excep
...
+  private List<CRL> generateCrl(final Stream<Buffer> stream) throws CertificateException, CRLException {
+    List<CRL> crls = new ArrayList<>();
+    CertificateFactory certificatefactory = CertificateFactory.getInstance("X.509");
+    for (Buffer crlValue : stream.collect(Collectors.toList())) {
+      crls.addAll(certificatefactory.generateCRLs(new ByteArrayInputStream(crlValue.getBytes())));
+    }
+    return crls;
+  }
+
+  private Stream<Buffer> readCrlFromFile(final VertxInternal vertx) {
```
vietj: this should be non blocking as the reload might be called from the event loop initially

vietj:
I would like to reconsider this feature by merging it with the `TrustOptions#getTrustManagerFactory` method, i.e include this rather in this interface and move this away from the `SSLHelper` implementation as it might have a better synergy with https://github.com/eclipse/vert.x/pull/2147

ndureiss:
Hi. Seems that CRLs are still not reload at runtime in 2k21. Do you planned to solve this fracture in next versions ? 
Thanks.

vietj:
@ndureiss I think we will implement it soon as this seems we need to provide a way to reload also certificates for some environments

vietj:

Issue #2141, opened 2017-09-21
ProxyErrorTest.testProxyHostUnknown() depends on normal dns, which may not work


alexlehm:
The test uses unknown.hostname to get a dns resolution error, this will not work when the providers dns returns an ad page for unknown hosts.
We will set up a fakedns instead

vietj:
yup

alexlehm:
issue belongs to this thread: https://groups.google.com/d/msg/vertx-dev/9RI99ZPM3Eo/MdeVLMpNAQAJ

I will implement a new test when i have time

alexlehm:
pr#2157

alexlehm:
closed with merge #2157

vietj:
thanks Alex!

PR #2157, opened 2017-10-01
issue #2141: use fakedns to avoid resolving unknown.hostname

alexlehm:
the test would fail if the provider's dns returns a adserver as unknown hostname

Signed-off-by: alexlehm <alexlehm@gmail.com>


alexlehm:
thank you

Issue #2158, opened 2017-10-04
AsyncResult improvements


ruslansennov:
I believe `AsyncResult.map()` must handle exceptions in mapper function. 
Also it seems that `AsyncResult.succeededAsyncResult()` and `AsyncResult.failedAsyncResult()` static factories will be useful.

If these proposals makes sense I can provide PR

vietj:
- what's your proposal for map ? I think this is debatable and it was discussed before (I can't find the issue). I think the main issue is that the result is lazy and it should be done with some care.

- there are method on `Future` for that which returns an optimized immutable Future that represent the async result so it would be more or less the same, see `Future#succeededFuture()`

ruslansennov:
Yes, I know about the immutable `Future` implementations, I use them and it is OK for me. But as I see, there is no other classes that implement `AsyncResult` interface. This means that` AsyncResult.map()` (which has a different behavior than `Future.map()` in part of checking exception) can't be used. We can make this method abstract or change it like overridden method in `Future` with additional static factories

vietj:
I think we should change the contract/implementation in `AsyncResult` to match the `Future#map` contract and satisfy `((AsyncResult)Future.succeededFuture().map(t -> throw new RuntimeException()))'

dgoetsch:
I see `AsyncResult` as being informally similar to a Try monad (see scala standard lib or kotlin funktionale library).
 
In scala Try, the `map` function handles functions as proposed here - it caches exceptions and wraps them in a `Failure`, or wraps the successful result in a `Success`;  Similarly, it may be useful to provide a `flatMap` method (or `compose` to align with `Future`): `flatMap(Function<T,AsyncResult<U>> mapper)`.

It also might be nice to have a couple of defualt implementations of AsyncResult (Success and Failure), and provide static methods on AsyncResult that return an instance of AsyncResult.Success or AsyncResult.Failure.  I've wanted those in a few projects and often end up implementing them myself in some gross Util object.

vietj:
@dgoetsch keep in mind that the goal of AsyncResult/Future is to not become a functionnal monads. Besides in the future we might want to replace them with more common types such as `CompletableFuture` instead, so I'm not keen on providing more than what it is necessary.

PR #2160, opened 2017-10-04
changes AsyncResult.map() and .otherwise() contracts 

ruslansennov:
towards #2158

### DIFF
--- a/src/main/java/io/vertx/core/AsyncResult.java
+++ b/src/main/java/io/vertx/core/AsyncResult.java
@@ -16,6 +16,8 @@
 
 package io.vertx.core;
 
+import io.vertx.core.impl.NoStackTraceThrowable;
+
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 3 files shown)


vietj:
please only keep the AsyncResult implementation and not add new factory method to not provide two ways of doing the same thing. People are used to the idiom Vertx.succeededFuture() to build an immutable AsyncResult and that's fine this way.

vietj:
however you can remove the map default method I think as the impl is lazy and reimplement them correctly in SucceededFuture/FailedFuture in a non lazy fashion

ruslansennov:
> you can remove the map default method

OK, it will be a new PR, but in this case all tests in `FutureTest` related to the `AsyncResult` do not make sense and can be deleted, right?

ruslansennov:
> you can remove the map default method

It is not works for uncompleted futures, because Future can has only one handler and this case will not work

```
    AsyncResult<Integer> map1 = f.map(String::length);
    AsyncResult<Integer> map2 = f.map(17);
```

So, if new static factories are undesirable, we must leave everything as is

vietj:
I mean that in `AsyncResult` you can remove the default implementation of the map method but keep the method.

vietj:
I mean this https://github.com/vietj/vert.x/commit/58238bc3ebb559cabf76c92ca6328bd49588f625

PR #2045, opened 2017-07-10
Websockets server compression

disparlure:
Hello,
I've added support for websockets compression for server and client. Server support is enabled by default (but the client must request it). Client side must be specifically enabled in options. Please let me know if there's anything that needs to be changed/added to make this complete. Thanks.

PR #2164, opened 2017-10-09
Enable websocket compression

disparlure:
Enable compression for websockets. Server and client side support included. Continuation of the work done in #2045 .


CodingFabian:
@disparlure the easiest way to get around the individual commits would be to just combine all your changes into a single commit like this:
https://stackoverflow.com/questions/17354353/git-squash-all-commits-in-branch-without-conflicting

disparlure:
Tried the suggested process, it does not seem to have worked as intended.

CodingFabian:
tried to help you out in #2165 - I have squashed all the changes for you into a single commit signed by us both.

disparlure:
Assuming we're closing this in favor of https://github.com/eclipse/vert.x/pull/2165.

PR #2165, opened 2017-10-10
Enable websocket compression

CodingFabian:
Server support is enabled by default (but the client must request it).
Client side must be specifically enabled in options.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -46,6 +46,11 @@ import io.netty.handler.codec.http.websocketx.PongWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketHandshakeException;
 import io.netty.handler.codec.http.websocketx.WebSocketServerHandshaker;
 import io.netty.handler.codec.http.websocketx.WebSocketVersion;
+import io.netty.handler.codec.http.websocketx.extensions.WebSocketServerExtensionHandshaker;
+import io.netty.handler.codec.http.websocketx.extensions.WebSocketServerExtensionHandler;
+import io.netty.handler.codec.http.websocketx.extensions.compression.DeflateFrameServerExtensionHandshaker;
+import io.netty.handler.codec.http.websocketx.extensions.compression.PerMessageDeflateServerExtensionHandshaker;
+import io.netty.handler.codec.compression.ZlibCodecFactory;
 import io.netty.handler.codec.http2.Http2CodecUtil;
 import io.netty.handler.codec.http2.Http2Settings;
 import io.netty.handler.logging.LoggingHandler;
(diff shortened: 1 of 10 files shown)


CodingFabian:
Trying to help out. I squashed #2164 for @disparlure - hope the double signing is correct ;)

CodingFabian:
@vietj let me know if i should amend the commit to remove my signature. think it gets overwritten anyway when you commit the change with yours.

bripkens:
It would be really great to land this. We are sending large messages to our users and leveraging widely supported per message deflate would be really helpful to improve the speed of our app.

bripkens:
Just compiled and tested it locally. The change is working fine (even in combination with vertx-web sockjs):

**Compression agreement**
headers

**Wireshark evidence that the content is truly compressed**
wireshark

vietj:
it shall be in 3.5.1

tsegismont:
@vietj what's left to do here? Only rebasing on master? I guess the conflict comes from the recent client pool changes.

CodingFabian:
I had a look if I can fix up the PR, but there is a problem with Http1xPool in ClientConnection that I lack the knowledge to fix. I just wanted to help squashing....

vietj:
I will merge it with the correct fixes

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -678,7 +711,7 @@ protected void handleMessage(ServerConnection conn, ContextImpl context, Channel
             if (!closeFrameSent) {
               // Echo back close frame and close the connection once it was written.
               // This is specified in the WebSockets RFC 6455 Section  5.4.1
-              ch.writeAndFlush(wsFrame).addListener(ChannelFutureListener.CLOSE);
+              ch.writeAndFlush(new CloseWebSocketFrame().replace(wsFrame.getBinaryData())).addListener(ChannelFutureListener.CLOSE);
```
vietj: can you explain this change ?

disparlure: It was assembling the PONG message empty. An empty PONG message doesn't get sent, the client doesn't see it and disconnects. This puts the content in the PONG.

vietj: ok, sure but then why not do instead:

`ch.writeAndFlush(new CloseWebSocketFrame(true, 0, wsFrame.getBinaryData())).addListener(ChannelFutureListener.CLOSE);`

vietj: this is a reply to a CLOSE msg not a PONG here

disparlure: Sorry, my mistake. I made that change at the same time as the PONG fix. I think it was a similar issue. There's some code further down the stack in netty (handleWebSocketFrame I think) that was silently dropping the echo of both the pong and close because they were not of one of the WebSocketFrame child classes.

kwangsu61:
@vietj when will you plan to release websocket compression feature? I miss it.

vietj:
it would be in 3.6.0, we might publish milestones of 3.6.0 you could use

elR1co:
@vietj Very interesting feature for me too. Would you need some help to integrate?

vietj:
rebasing on master would be great

CodingFabian:
merged via https://github.com/eclipse/vert.x/pull/2546

PR #2169, opened 2017-10-13
Fix multiple handler completion on deploy fail

sschmittBt:
When starting multiple verticles, if more than one fails it results in a duplicate call to the completionHandler. 
The boolean that should prevent duplicate calls is never set to true.

Also, in the catch block, the same check should be made as the same error could occur there.


vietj:
can you add a test and sign the commit ?

sschmittBt:
@vietj done

vietj:
does this test pass without the fix ?

sschmittBt:
no, it fails without the fix

vietj:
yes I checked, however it might pass in some cases, I updated a bit the test to complete it in a timer, so the timer should be executed on the same event loop after the completion handler callbacks

vietj:
thanks for the catch!

PR #2189, opened 2017-10-27
Add IPv4 addressType support to SocksProxy test class

philrykoff:
Would like to use this in vertx-eventbus-bridge-clients, but netty seems to convert domain name to IP automatically, so lets support those, too.

### DIFF
--- a/src/test/java/io/vertx/test/core/SocksProxy.java
+++ b/src/test/java/io/vertx/test/core/SocksProxy.java
@@ -30,7 +30,7 @@ public class SocksProxy extends TestProxyBase {
 
   private static final Buffer clientInit = Buffer.buffer(new byte[] { 5, 1, 0 });
   private static final Buffer serverReply = Buffer.buffer(new byte[] { 5, 0 });
-  private static final Buffer clientRequest = Buffer.buffer(new byte[] { 5, 1, 0, 3 });
+  private static final Buffer clientRequest = Buffer.buffer(new byte[] { 5, 1, 0 });
   private static final Buffer connectResponse = Buffer.buffer(new byte[] { 5, 0, 0, 1, 0x7f, 0, 0, 1, 0x27, 0x10 });
   private static final Buffer errorResponse = Buffer.buffer(new byte[] { 5, 4, 0, 1, 0, 0, 0, 0, 0, 0 });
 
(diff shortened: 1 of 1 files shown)


vietj:
can you give an explanation of the change ?

philrykoff:
Of course, this was accidentaly deleted by in my unsigned commit.

In vertx-eventbus-bridge-client, I'd like to add a test for SOCKS proxy connectivity. Previously, I had used a foreign dependancy, but then found `SocksProxy` class in vertx-core.

Somehow, netty always translates bridge hostnames to IPs before sending them to the SOCKS proxy, but `SocksProxy` did so far only work with hostnames, not IPs. This commit adds the capability to `SocksProxy` to connect to servers based on IPv4.

src/test/java/io/vertx/test/core/SocksProxy.java
```diff
@@ -30,7 +30,7 @@
 
   private static final Buffer clientInit = Buffer.buffer(new byte[] { 5, 1, 0 });
   private static final Buffer serverReply = Buffer.buffer(new byte[] { 5, 0 });
-  private static final Buffer clientRequest = Buffer.buffer(new byte[] { 5, 1, 0, 3 });
```
philrykoff: The removed fourth byte signals the address type. 1 = IPv4, 3 = hostname. Additionaly, there is IPv6, but I would not know how to test that and apparently, it's not needed anyhow.

vietj:
can we have this tested ?

vietj:
@alexlehm can you have a look at this ?

alexlehm:
when I try to use the unit for socks5 from NetTest to connect to 127.0.0.1, it still uses hostname parameter type (3), so netty doesn't map ip addresses to type 1 in our implementation.
@philrykoff Phil, do you know which client your test uses?

philrykoff:
In the vertx-eventbus-bridge-clients project, we use netty Bootstrap.

https://github.com/vert-x3/vertx-eventbus-bridge-clients/pull/18/files#diff-c0d551f2d294c5a6505e055ff16bf9c9R576

I have now added another fix in my local repository so that the Java vertx-eventbus-bridge-client does not resolve the hostname to an IP before sending the request to the proxy, as that is typically the proxy's job (in my coorporate environment, the client cannot even resolve DNS itself). But that is not yet included in my PR to vertx-eventbus-bridge-clients.

@vietj : how to proceed? Should I add that commit to the PR or would you like to accept this PR and I'll add the fix and several other things in a separate PR to vertx-eventbus-bridge-clients?

vietj:
@philrykoff it's not totally clear what you want to do, can you clarify ?

philrykoff:
Ideal for me would be this PR being accepted.

I will include my fix for vertx-eventbus-bridge-clients (hostnames not being resolved to IPs anymore when using Proxies) in another PR, so that I don't have to touch https://github.com/vert-x3/vertx-eventbus-bridge-clients/pull/18 again.

philrykoff:
Can this still be merged into 3.5.1 or are there any issues with it? Thanks!

vietj:
yes we can merge it

philrykoff:
Thanks!

PR #2198, opened 2017-11-02
Message forwarding in eventbus interceptors

jan-zajic:
This commit adds ability to forward message in eventbus interceptor to different address. 
You can route messages according to theirs headers or body content.

This is very useful feature, notably in multi tenant applications.

You can for example:
* have multiple implementations of same service, choose them by header
* deploy different number of target Verticle instances for diferrent tenants (based on their size), choose target address dynamicaly by tenant identification in header
* make asynchronous and clustered interceptors/proxies - against the code directly in interceptor you can forward message to diferrent address that somehow modify it, set flag in headers to not forward it again and forward it to original address
All behavior is transparent to original sender.

It's better than listen on target address in some other verticle and forward it manually because:
* can proxy existing library which have listen address hard coded
* don't break contract between sender and receiver, i.e:
  * don't break Backpressure handling, pause, resume etc. which can otherwise be very hard to implement
  * don't change request timeout and codec - this information can't be determined in message receiver
  * even if you choose dynamic forward address based on message headers (ie multiple forward addresses for same original address) producer communicate directly with consumer

Keep up this contract is very effective in clustered environment. If you use some proxy verticle instead, you must collocate this proxy instances with target verticles to minimize unnecessary inter node communication and forwarding, which can be very inefficient and difficult to implement.

We currently use all this use cases do this in interceptor via java reflection, which of course is very ugly.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -422,6 +422,7 @@ public class EventBusImpl implements EventBus, MetricsProvider {
       long timeout = options.getSendTimeout();
       String replyAddress = generateReplyAddress();
       message.setReplyAddress(replyAddress);
+      message.setReplyTimeout(timeout);
       Handler<Message<T>> simpleReplyHandler = convertHandler(replyHandler);
       HandlerRegistration<T> registration =
         new HandlerRegistration<>(vertx, metrics, this, replyAddress, message.address, true, replyHandler, timeout);
@@ -496,6 +497,14 @@ public class EventBusImpl implements EventBus, MetricsProvider {
     public Object body() {
       return message.sentBody;
     }
+    
+    @Override
+    public void forward(String address) {
+      message.address = address;
+      OutboundDeliveryContext<T> forwardedContext = new OutboundDeliveryContext<>(message, options, handlerRegistration, replierMessage);
+      forwardedContext.next();
+    }
+
   }
 
   private void unregisterAll() {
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -488,6 +488,15 @@ public boolean send() {
     public Object sentBody() {
       return message.sentBody;
     }
+
+    @Override
+    public void forward(String address) {
+      message.setAddress(address);
+      if(message.send)
+        handlerRegistration.setRepliedAddress(address);
```
vietj: this should not be changed, it is actually only used for providing more info for the metrics and by the time the value is changed, the registration will already have been done

vietj:
we had a kind of message forward feature at some plan and it seems related to it, but I can't find it anymore

vietj:
I'm wondering if instead the Message should not have a forward method on it instead of being on the SendContext

vietj:
and the forward would take care of the case the message is already delivered (and thus forward by the recipient) or if the message is in the interceptor and therefore act like your PR

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
jan-zajic: ok i removed registration modification

vietj:
@jan-zajic it looks great, I will integrate it for 3.6

jan-zajic:
@vietj it's fine, but current codebase is very different after you introduce delivery interception. So i remake it to preserve original simple implementation in OutboundDeliveryContext and implemented far more complex forwarding in InboundDeliveryContext. So please review modifed PR, thank you.

vietj:
@jan-zajic I tried to adapt your PR to latest change, and I think the main issue I had was the fact that I wanted to move the forward method from the `DeliveryContext` to the `Message` so it could be reused even outside of an interceptor, but I couldn't for some reason I don't remember. It would be much better if we could achieve this.

jan-zajic:
@vietj I try it but it couldn't be done in some elegant and clean way. And i personally prefer this variant with forward in DeliveryContext. I would even prefer to have it only in OutboundDeliveryContext (if it have different interface). So current state seems like a reasonable compromise for me, but it's just my opinion.

vietj:
@jan-zajic thanks for the feedback, I'll ahve a look over again

PR #2209, opened 2017-11-15
HttpClient improvements

vietj:
fixes https://github.com/eclipse/vert.x/issues/2207
fixes https://github.com/eclipse/vert.x/issues/2208


src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -0,0 +1,341 @@
...
+            long diff = concurrency - holder.concurrency;
+            capacity += diff;
+            if (holder.capacity == 0) {
+              available.add(holder);
+            }
+            holder.capacity += diff;
+            holder.concurrency = concurrency;
+            checkPending();
+          } else if (holder.concurrency > concurrency) {
+            throw new UnsupportedOperationException("Not yet implemented");
```
vietj: note, this is a case when an HTTP/2 connection send a max concurrent stream = 20 and then downgrade to a lower version, it should be supported at some point but it seems difficult and I don't believe it is an actual case

tsegismont: What would the Vert.x HTTP server do?

vietj: what do you mean ?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -0,0 +1,315 @@
...
+import io.vertx.core.logging.LoggerFactory;
+import io.vertx.core.spi.metrics.HttpClientMetrics;
+
+import java.util.*;
+import java.util.function.BiConsumer;
+
+/**
+ * The pool is a queue of waiters and a list of connections.
+ *
+ * The pools invariant:
```
tsegismont: minor: `pool's invariants`

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -0,0 +1,315 @@
...
+ * - the pool {@link #capacity} is the sum of the {@link Holder#capacity} of the {@link #available} list
+ * - a connection in the {@link #available} list has its {@code Holder#capacity > 0}
```
tsegismont: Why strictly bigger? Can't it be in the `available` dequeue without being borrowed at all?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -0,0 +1,315 @@
...
+ * - the {@link #weight} is the sum of all {@link Holder#weight}, i.e the connections in the {@link #all} set
+ *
+ * A connection is delivered to a {@link Waiter} on the connection's event loop thread, the waiter must take care of
+ * calling {@link io.vertx.core.impl.ContextInternal#executeFromIO} if necessary.
+ *
+ * Calls to the pool are synchronized on the pool to avoid race conditions and maintain its invariants. This pool can
+ * be called from different threads safely (although it is not encouraged for performance reasons), since it synchronizes
```
tsegismont: The second sentence is very long and it seems to conflate different ideas. Could you rephrase this paragraph?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -0,0 +1,315 @@
...
+ * To constrain the number of connections the pool maintains a {@link #weight} value that must remain below the the
+ * {@link #maxWeight} value to create a connection. Weight is used instead of counting connection because this pool
+ * can mix connections with different concurrency (HTTP/1 and HTTP/2) and this flexibility is necessary.
+ *
+ * When a connection is created an initial weight is returned by the {@link ConnectionProvider#connect} method and is
+ * added to the current weight. When the channel is connected the {@link ConnectionListener#onConnectSuccess} callback
+ * provides the initial weight returned by the connect method and the actual connection weight so it can be used to
+ * correct the current weight. When the channel fails to connect the {@link ConnectionListener#onConnectFailure} failure
+ * provides the initial weight so it can be used to correct the current weight.
+ *
```
tsegismont: Could you add a section about what happens to idle connections?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
tsegismont: You said 

> In practice the server will rather send a GO_AWAY frame.

 Is that what a Vert.x HTTP/2 server would do?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -0,0 +1,315 @@
...
+            checkPending();
+          }
+        } else {
+          synchronized (Pool.this) {
+            checkPending();
+          }
+        }
+      }
+      @Override
+      public void onConnectFailure(ContextImpl context, Throwable err, long weight) {
```
tsegismont: So if I understand correctly, when a `waiter` is picked up, and no connection is available, but `maxWeight` is not reached, we create a new connection. And if connection fails, the `waiter` handles the failure immediately. Correct?

Why not leaving the waiter in the list and pick it up only when a new connection is ready? Of course the waiter would timeout at some point. I'm not saying this strategy is better, just curious why you picked the other strategy.

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -0,0 +1,315 @@
...
+            boolean handled = deliverToWaiter(conn, waiter);
+            if (!handled) {
+              synchronized (Pool.this) {
+                recycleConnection(conn, 1,false);
+                checkPending();
+              }
+            }
+          } else {
+            synchronized (Pool.this) {
+              waiters.add(waiter);
```
tsegismont: Shouldn't we check the `waiters` queue constraint here (`queueMaxSize`)?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -0,0 +1,315 @@
...
+        waiter.initConnection(context, conn);
+        boolean usable;
+        synchronized (Pool.this) {
+          usable = initConnection(waiter, holder, context, concurrency, conn, channel, initialWeight, actualWeight);
+        }
+        if (usable) {
+          boolean consumed = deliverToWaiter(holder, waiter);
+          synchronized (Pool.this) {
+            if (!consumed) {
+              synchronized (this) {
```
tsegismont: Shouldn't the `recycleConnection` call be guarded by `Pool.this`?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
tsegismont: Could you also add a section about the pool getting closed whenever the last open connections closes? This behavior comes from the `ConnectionManager` implementation (with a map of endpoints) but if other Vert.x projects use this pool they must be aware of it.

tsegismont:
@vietj I just reviewed the Pool class. Is there any other class you want me to look at?

vietj:
@tsegismont you can have a global look around.

src/main/java/io/vertx/core/http/impl/pool/Pool.java
vietj: the nested second `synchronized (this)` shouldn't be here, thanks for spotting it

src/main/java/io/vertx/core/http/impl/pool/Pool.java
vietj: yes, I will rework the pool so it's better handled by removing the connection validity check that will be replaced by a simple pool eviction and avoid this check

src/main/java/io/vertx/core/http/impl/pool/Pool.java
vietj: a connection has `capacity == 0` means it cannot accept anymore connection

src/main/java/io/vertx/core/http/impl/pool/Pool.java
vietj: ok

src/main/java/io/vertx/core/http/impl/pool/Pool.java
vietj: because it is simpler this way

vietj:
@tsegismont I've improved the pool according to your review, can you make a new pass ?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -98,59 +97,73 @@ public Pool(ConnectionProvider<C> connector,
...
     this.available = new ArrayDeque<>();
     this.connectionAdded = connectionAdded;
     this.connectionRemoved = connectionRemoved;
   }
 
+  /**
+   * Get a connection for a waiter asynchronously.
+   *
+   * @param waiter the waiter
+   * @return wether the pool can satisfy the request
```
tsegismont: Typo: `whether`

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -30,16 +30,18 @@
  *
  * Pool invariants:
  * - the pool {@link #capacity} is the sum of the {@link Holder#capacity} of the {@link #available} list
- * - a connection in the {@link #available} list has its {@code Holder#capacity > 0}
+ * - a connection is in the {@link #available} list has its {@code Holder#capacity > 0}
```
tsegismont: Not sure why you added this

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -30,16 +30,18 @@
...
  *
  * Calls to the pool are synchronized on the pool to avoid race conditions and maintain its invariants. This pool can
- * be called from different threads safely (although it is not encouraged for performance reasons), since it synchronizes
- * on the pool and callbacks are asynchronous on the event loop thread of the connection without holding the pool
- * lock to avoid deadlocks. We benefit from biased locking which makes the overhead of synchronized near zero.
+ * be called from different threads safely (although it is not encouraged for performance reasons, we benefit from biased
+ * locking which makes the overhead of synchronized near zero), since it synchronizes on the pool.
+ *
+ * - acquisition success are on the event loop thread of the connection without holding the pool lock to avoid deadlocks.
+ * - acquisition failures are on an event loop too without holding the pool lock to avoid deadlocks.
```
tsegismont: Maybe rephrase both sentences to:

_In order to avoid deadlocks, acquisition events (success or failure) are handled on the event loop thread of the connection without holding the pool lock._

src/main/java/io/vertx/core/http/impl/pool/Pool.java
tsegismont: 👍

tsegismont:
@vietj just a few doc comments. Otherwise LGTM

PR #2221, opened 2017-11-27
Adding tracing of client HTTP calls

myok12:


vietj:
I think you can get tracing already with http://vertx.io/docs/vertx-core/java/#logging_network_activity

PR #2222, opened 2017-11-28
Adding tracing of client HTTP calls

myok12:
Following https://github.com/eclipse/vert.x/issues/2200.


myok12:
Still having IP-validation issues.

PR #2229, opened 2017-11-30
DecodeException

ukr15:
Throws a DecodeException for json strings containing trailing characters after an valid substring like "{\n    \"foo\": {\n    \"baz\" : \"bazValue\",\n    \"bar\" : \"barValue\"}\n    },\n     \"bla\" : [ \"a\" , \"b\", \"c\"],\n     \"xxx\" : 0.3\n}"

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -42,8 +42,8 @@ import static java.time.format.DateTimeFormatter.ISO_INSTANT;
  */
 public class Json {
 
-  public static ObjectMapper mapper = new ObjectMapper();
-  public static ObjectMapper prettyMapper = new ObjectMapper();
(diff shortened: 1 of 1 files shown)


vietj:
I'm not sure we want to change that, if you need it , you can modify the public static object mapper to achieve it.

ukr15:
As a consequence, Vertx will read partially correct config files and starts  a not properly configured verticle.

vietj:
@ukr15 why do you have invalid config files ?

ukr15:
Our customers have to change the configuration templates and make mistakes. Therefore, finding errors is sometimes a hell is they do not use José validators.

vietj:
why don't you pre-process this file before reading it ?

vietj:
is that for deploying verticles ?

that being said, you can change the ObjectMapper configuration as it is public and static easily.

PR #2235, opened 2017-12-12
DeploymentOptions#setMaxWorkerExecuteTime not honored if pool name is set

tsegismont:
Fixes vert-x3/issues#203

### DIFF
--- a/src/test/java/io/vertx/test/core/NamedWorkerPoolTest.java
+++ b/src/test/java/io/vertx/test/core/NamedWorkerPoolTest.java
@@ -19,25 +19,63 @@ package io.vertx.test.core;
 import io.vertx.core.AbstractVerticle;
 import io.vertx.core.Context;
 import io.vertx.core.DeploymentOptions;
+import io.vertx.core.Future;
 import io.vertx.core.Vertx;
 import io.vertx.core.WorkerExecutor;
-import io.vertx.core.impl.WorkerExecutorInternal;
 import org.junit.Test;
 
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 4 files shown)


src/test/java/io/vertx/test/core/NamedWorkerPoolTest.java
```diff
@@ -19,26 +19,64 @@
...
+        }, startFuture);
+      }
+    }, deploymentOptions, onSuccess(did -> {
+      testComplete();
+    }));
+    await();
+    String timeLimitMsg = "time limit is " + MILLISECONDS.convert(maxWorkerExecuteTime, NANOSECONDS);
+    assertTrue(Files.lines(Paths.get(System.getProperty("java.io.tmpdir"), "vertx.log"))
+      .anyMatch(line -> {
+        if (line.contains("has been blocked for")
```
vietj: how about return `expression` instead of `if (expression) return true else return false`

src/test/java/io/vertx/test/core/NamedWorkerPoolTest.java
```diff
@@ -74,7 +112,7 @@ public void testOrdered() {
     CountDownLatch submitted = new CountDownLatch(1);
     Context ctx = vertx.getOrCreateContext();
     ctx.runOnContext(v -> {
-      for (int i = 0;i < num;i++) {
+      for (int i = 0; i < num; i++) {
```
vietj: no need to reformat

src/test/java/io/vertx/test/core/NamedWorkerPoolTest.java
tsegismont: Done

src/test/java/io/vertx/test/core/NamedWorkerPoolTest.java
tsegismont: Done

tsegismont:
> can you encapsulate the log check and apply it to `BlockedThreadCheckerTest` ?

Done with a jUnit rule

PR #2241, opened 2017-12-15
Default configuration for vertx dns client

Sammers21:
Signed-off-by: Pavel Drankov <titantins@gmail.com>

A solution for https://github.com/eclipse/vert.x/issues/2239

### DIFF
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -221,7 +221,7 @@ public interface Vertx extends Measured {
   DnsClient createDnsClient(int port, String host);
 
   /**
-   * Create a DNS client to connect to automatically chosen DNS server
+   * Create a DNS client to connect to the DNS server configured by {@link VertxOptions#getAddressResolverOptions()}
    *
    * @return the DNS client
    */
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -221,6 +221,13 @@ public interface Vertx extends Measured {
   DnsClient createDnsClient(int port, String host);
 
   /**
+   * Create a DNS client to connect to automatically chosen DNS server
+   *
+   * @return the DNS client
+   */
+  DnsClient createDnsClient();
+
+  /**
    * Create a DNS client to connect to a DNS server
    *
    * @param options the client options
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -32,6 +32,7 @@ import io.vertx.core.http.HttpClient;
 import io.vertx.core.http.HttpClientOptions;
 import io.vertx.core.http.HttpServer;
 import io.vertx.core.http.HttpServerOptions;
+import io.vertx.core.impl.resolver.DnsResolverProvider;
 import io.vertx.core.metrics.Measured;
 import io.vertx.core.net.NetClient;
 import io.vertx.core.net.NetClientOptions;
@@ -222,6 +223,8 @@ public interface Vertx extends Measured {
 
   /**
    * Create a DNS client to connect to the DNS server configured by {@link VertxOptions#getAddressResolverOptions()}
+   * <p>
+   * DNS client takes the first configured resolver address provided by {@link DnsResolverProvider#nameServerAddresses()}}
    *
    * @return the DNS client
    */
--- a/src/main/java/io/vertx/core/impl/VertxImpl.java
+++ b/src/main/java/io/vertx/core/impl/VertxImpl.java
@@ -416,7 +416,7 @@ public class VertxImpl implements VertxInternal, MetricsProvider {
   @Override
   public DnsClient createDnsClient() {
     DnsResolverProvider provider = new DnsResolverProvider(this, addressResolverOptions);
-    InetSocketAddress address = provider.serverList.get(0);
+    InetSocketAddress address = provider.nameServerAddresses().get(0);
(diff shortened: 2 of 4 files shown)


Sammers21:
@vietj , do we actually should have a test for that?

vietj:
@Sammers21 of course! I recommend to configure the VertxOptions with the embedded test DNS server

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -406,6 +409,15 @@ public DnsClient createDnsClient(int port, String host) {
     return new DnsClientImpl(this, port, host, DnsClientOptions.DEFAULT_QUERY_TIMEOUT);
   }
 
+  @Override
+  public DnsClient createDnsClient() {
```
vietj: actually it should use `DnsResolverProvider` so it can be configured using `VertxOptions#addressResolverOptions` (that's how we can test it).

Sammers21: In `DnsResolverProvider`s constructor, we have a `serverList` so the list of DNS servers to use, but when creating DnsClient we are specifying only one sever. Should we create `DnsResolverProvider` with `addressResolverOptions` and then use the first element of serverList as a default DNS server with port=53?

Sammers21: I have just updated the PR. Is that what you mean?

src/main/java/io/vertx/core/impl/resolver/DnsResolverProvider.java
```diff
@@ -62,10 +62,10 @@
   private final Vertx vertx;
   private final List<ResolverRegistration> resolvers = Collections.synchronizedList(new ArrayList<>());
   private AddressResolverGroup<InetSocketAddress> resolverGroup;
+  public final List<InetSocketAddress> serverList = new ArrayList<>();
```
vietj: what's the purpose of this ?

src/main/java/io/vertx/core/impl/resolver/DnsResolverProvider.java
```diff
@@ -85,7 +85,7 @@ public DnsResolverProvider(VertxImpl vertx, AddressResolverOptions options) {
         }
       }
     } else {
-      DnsServerAddressStream stream = DnsServerAddresses.defaultAddresses().stream();
+      DnsServerAddressStream stream = DefaultDnsServerAddressStreamProvider.defaultAddresses().stream();
```
vietj: why this change ?

src/main/java/io/vertx/core/impl/VertxImpl.java
vietj: yes, we should add a method on DnsResolverProvider that provides the server address (the first one) and have only the stream nameServerAddresses to be a field of DnsResolverProvider with a method that returns it, so when building the client, it can use the first one rather than using a List

src/main/java/io/vertx/core/impl/resolver/DnsResolverProvider.java
Sammers21: since `DnsServerAddresses#defaultAddresses()` is deprecated

Sammers21: it is recommended to use  `DefaultDnsServerAddressStreamProvider#defaultAddresses()`

src/main/java/io/vertx/core/impl/resolver/DnsResolverProvider.java
Sammers21: that is where we are storing DNS server addresses

Sammers21: so the first one will be used as a default one

Sammers21:
@vietj , I have just added a test. So, what I m trying to achieve is forcing to use a fake DNS server as a default one and then check that the first element of the list returned by the _nameServerAddresses_ method is the fake server.

If something is wrong just let me know.

src/main/java/io/vertx/core/impl/resolver/DnsResolverProvider.java
vietj: ok

src/main/java/io/vertx/core/Vertx.java
```diff
@@ -220,6 +220,13 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
    */
   DnsClient createDnsClient(int port, String host);
 
+  /**
+   * Create a DNS client to connect to automatically chosen DNS server
```
vietj: Create a DNS client to connect to the DNS server configured by {@link VertxOptions#getAddressResolverOptions()}

Sammers21: changed

src/test/java/io/vertx/test/core/DNSTest.java
```diff
@@ -63,6 +66,27 @@ public void testIllegalArguments() throws Exception {
     dnsServer.stop();
   }
 
+  @Test
+  public void testDefaultDnsServer() throws Exception {
```
vietj: why isn't the test using createDnsClient() ?

Sammers21: In the test, I have tried to test that the first DNS server in a list is our fake DNS server. 

I m agree with you. We should test it with direct _createDnsClient()_ call.

I have just made some changes. Is that correct?

Sammers21:
@vietj , is it fine now?

vietj:
@Sammers21 one last thing : we need to say in the javadoc that the DNS client takes the first configured resolver address

Sammers21:
@vietj, done

PR #2244, opened 2017-12-18
Getting corresponding shared map

nolequen:
_in place of  outdated pull request #2094_

Sometimes it is very uncomfortable to choose an appropriate map explicitly, especialy if we don't know in advance whether the Vert.x instance clustered or not. E.g. when writing tests which may use cluster in some cases or not, choosing the "right" map leads to changes in production code.

The goal of the proposed changes is to make it easier to work with shared data without paying attention if the instance is clustered or not.

### DIFF
--- a/src/main/java/io/vertx/core/shareddata/SharedData.java
+++ b/src/main/java/io/vertx/core/shareddata/SharedData.java
@@ -80,4 +80,11 @@ public interface SharedData {
    */
   <K, V> LocalMap<K, V> getLocalMap(String name);
 
+  /**
+   * Get the cluster wide map if Vert.x instance is clustered or local otherwise.
+   *
+   * @param name  the name of the map
+   * @param resultHandler  the map will be returned asynchronously in this handler
+   */
+  <K, V> void getMap(String name, Handler<AsyncResult<AsyncMap<K, V>>> resultHandler);
 }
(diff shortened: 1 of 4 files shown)


vietj:
it would be simpler if LocalMapImpl implements directly AsyncMap

nolequen:
but the number of LocalMapImpl methods will double and additional dependency of instance will appear (to support methods with TTL)

tsegismont:
@nolequen I had started a branch last week to have AsyncMap in local/cluster mode. I just finished impl and tests (and doc updates)

See https://github.com/tsegismont/vert.x/commits/local-async-map

I am going to write an email about in Vert.x dev. It slightly differs from your solution:

- no new method in SharedData: it deprecates `getClusterWideMap` and replaces it with `getAsyncMap`: this what users have been asking in the past
- local async map have their own namespace, so no possible conflict with a local map having the same name

Please follow-up on vert-dev after I sent my email if you have any comment

nolequen:
@tsegismont this is the second time you're overtaking me at the finish line :)

Hope, your changes will be released in the near releases.

tsegismont:
Yes this is unfortunate. :-(

I had started last week on Friday and found your PR this morning.

2017-12-18 18:50 GMT+01:00 Anton Potsyus <notifications@github.com>:

nolequen:
This PR is the copy of outdated one: #2094 (opened on 8 Aug).

tsegismont:
I missed it, sorry

2017-12-19 8:28 GMT+01:00 Anton Potsyus <notifications@github.com>:

PR #2249, opened 2017-12-21
Local async map

### LINKED ISSUE #2137: SharedData#getMap() feature that provides LocalMap wrapped as AsyncMap when not clustered
Currently SharedData provides methods for dealing with locks, counters and maps. Locks and Counters can be easily aquired with getLock() and getCounter(). Vertx decides if you get the local or clustered version. For Maps we have no such feature. We have to know if our Vertx is clustered and use the appropriate method getLocalMap() or getClusterWideMap(). I think it will be nice to have also a method called getMap() which is similar to getLock() and getCounter().

tsegismont:
Fixes #2137 

Users often develop and test verticles in local mode and later deploy them in clustered mode. With shared data APIs this can lead to strange behavior:

- counters and locks can be used in local/clustered
- async map can be used in clustered mode only

We got questions/issues about this quite a few times so here's what I did:

- Added #getAsyncMap; the new one works regardless of the cluster status;
- reworked the Javadoc and the doc
- made sure AsyncMap is tested in both local/clustered mode
- implemented a local AsyncMap wich supports TTL correctly (ie an entry stays in the map even if previous binding had a TTL); the FakeClusterManager reuses this implemation

### DIFF
--- a/src/main/java/io/vertx/core/shareddata/SharedData.java
+++ b/src/main/java/io/vertx/core/shareddata/SharedData.java
@@ -26,6 +26,10 @@ import io.vertx.core.Handler;
  *   <li>asynchronous counters (local or cluster-wide)</li>
  * </ul>
  * <p>
+ * <p>
+ *   <strong>WARNING</strong>: In clustered mode, asynchronous maps/locks/counters rely on distributed data structures provided by the cluster manager.
+ *   Beware that the latency relative to asynchronous maps/locks/counters operations can be much higher in clustered than in local mode.
+ * </p>
  * Please see the documentation for more information.
  *
  * @author <a href="http://tfox.org">Tim Fox</a>
@@ -46,6 +50,10 @@ public interface SharedData {
   /**
    * Get the {@link AsyncMap} with the specified name. When clustered, the map is accessible to all nodes in the cluster
    * and data put into the map from any node is visible to to any other node.
+   * <p>
+   *   <strong>WARNING</strong>: In clustered mode, asynchronous shared maps rely on distributed data structures provided by the cluster manager.
+   *   Beware that the latency relative to asynchronous shared maps operations can be much higher in clustered than in local mode.
+   * </p>
    *
    * @param name the name of the map
    * @param resultHandler the map will be returned asynchronously in this handler
--- a/src/main/java/io/vertx/core/shareddata/SharedData.java
+++ b/src/main/java/io/vertx/core/shareddata/SharedData.java
@@ -20,10 +20,10 @@ import io.vertx.core.Handler;
  * <p>
  * Shared data provides:
  * <ul>
- *   <li>Cluster wide maps which can be accessed from any node of the cluster</li>
- *   <li>Cluster wide locks which can be used to give exclusive access to resources across the cluster</li>
- *   <li>Cluster wide counters used to maintain counts consistently across the cluster</li>
- *   <li>Local maps for sharing data safely in the same Vert.x instance</li>
+ *   <li>synchronous shared maps (local)</li>
+ *   <li>asynchronous maps (local or cluster-wide)</li>
+ *   <li>asynchronous locks (local or cluster-wide)</li>
+ *   <li>asynchronous counters (local or cluster-wide)</li>
  * </ul>
  * <p>
  * Please see the documentation for more information.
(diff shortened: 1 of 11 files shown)


src/main/java/io/vertx/core/shareddata/impl/LocalAsyncMapImpl.java
```diff
@@ -0,0 +1,233 @@
+/*
```
vietj: Use the right new license header

src/main/java/io/vertx/core/shareddata/impl/SharedDataImpl.java
```diff
@@ -67,6 +68,24 @@ public SharedDataImpl(VertxInternal vertx, ClusterManager clusterManager) {
...
   }
 
+  @Override
+  public <K, V> void getAsyncMap(String name, Handler<AsyncResult<AsyncMap<K, V>>> resultHandler) {
+    Objects.requireNonNull(name, "name");
+    Objects.requireNonNull(resultHandler, "resultHandler");
+    if (clusterManager == null) {
+      getLocalAsyncMap(name, resultHandler);
+    } else {
+      clusterManager.<K, V>getAsyncMap(name, ar -> {
```
vietj: delegate to the existing getClusterWideMap instead

src/main/java/io/vertx/core/shareddata/package-info.java
```diff
@@ -41,16 +45,16 @@
  * {@link examples.SharedDataExamples#example1}
  * ----
  *
- * === Cluster-wide asynchronous maps
+ * === Asynchronous maps
```
vietj: rather `Asynchronous shared maps`

shared being either between event loops or cluster wide, it's important to keep in mind that this is not local

vietj:
I believe that the renames are not appropriate we go from cluster-wide to asynchronous which is a real semantic change. The renaming should at least convey the idea that data is shared and that it in case of clustered vertx, using these methods has an impact on the latency of the operation (i.e it's not a local operation that operates only on memory and it involves distributed coordination).

tsegismont:
Would "Async shared map" as you suggested be fine.?

2017-12-22 10:45 GMT+01:00 Julien Viet <notifications@github.com>:

vietj:
I think we should get wider consensus on this on vertx-dev as I'm not the actual expert :-)

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncMapImpl.java
tsegismont: Done

vietj:
Async map is still not a good name : it's just say it is async and does not really carry the latency / failures expectations from a remote call

tsegismont:
`AsyncMap` has always been named `AsyncMap`. The change is only on the method name: `getClusterWideMap` to `getAsyncMap`. It's consistent with the other methods (`getCounter`, `getLock`) from both a naming and behavior point of views.

I don't get why `AsyncMap` retrieval method naming should be handled differently (I prefer consistency and the _shared_ aspect is already present in the interface name) but:

1. I was/am fine with `getAsyncSharedMap` as you suggested
2. I asked on vertx-dev but there is either no interest or no concern

In any case, let's not prevent the feature from getting merged because of a method name.

src/main/java/io/vertx/core/shareddata/SharedData.java
```diff
@@ -39,11 +39,21 @@
    *
    * @param name  the name of the map
    * @param resultHandler  the map will be returned asynchronously in this handler
+   * @throws IllegalStateException if the parent {@link io.vertx.core.Vertx} instance is not clustered
    */
   <K, V> void getClusterWideMap(String name, Handler<AsyncResult<AsyncMap<K, V>>> resultHandler);
 
   /**
-   * Get a cluster wide lock with the specified name. The lock will be passed to the handler when it is available.
+   * Get the {@link AsyncMap} with the specified name. When clustered, the map is accessible to all nodes in the cluster
```
vietj: Add a note about cluster wide latency like it is in doc

src/main/java/io/vertx/core/shareddata/SharedData.java
```diff
@@ -20,10 +20,10 @@
...
  * Shared data provides:
  * <ul>
- *   <li>Cluster wide maps which can be accessed from any node of the cluster</li>
- *   <li>Cluster wide locks which can be used to give exclusive access to resources across the cluster</li>
- *   <li>Cluster wide counters used to maintain counts consistently across the cluster</li>
- *   <li>Local maps for sharing data safely in the same Vert.x instance</li>
+ *   <li>synchronous shared maps (local)</li>
+ *   <li>asynchronous maps (local or cluster-wide)</li>
+ *   <li>asynchronous locks (local or cluster-wide)</li>
+ *   <li>asynchronous counters (local or cluster-wide)</li>
```
vietj: Add an overall note about latencies impact when clustered.

vietj:
Ok for the naming proposal, I've added a couple of todo to have more warnings in the doc (i.e one should not have to read the full doc when browsing the javadoc to be aware of latency differences).

src/main/java/io/vertx/core/shareddata/SharedData.java
tsegismont: Done

src/main/java/io/vertx/core/shareddata/SharedData.java
tsegismont: Done

tsegismont:
@vietj Just pushed a commit with more Javadoc

PR #2259, opened 2018-01-04
Provide user-friendly way to test an HTTP header value

### LINKED ISSUE #2258: Provide user-friendly way to test an HTTP header value
There is no easy way to test that an HTTP header is present and has a given value:

```java
    String value = HttpHeaders.WEBSOCKET.toString();
    boolean upgrade = headers.getAll(HttpHeaders.UPGRADE).stream()
      .anyMatch(val -> val.equalsIgnoreCase(value));
```

Drawbacks:

1. Testing a header value is a common task, especially if you write edge services or API gateways
2. The header value in this example should be tested in a case-insensitive way; so the `HttpHeaders` constant must be first converted to `String` (no `equalsIgnoreCase` on `CharSequence`)
3. A list of all headers needs to be created before testing

tsegismont:
Fixes #2258

There is no easy way to test that an HTTP header is present and has a given value:

```java
    String value = HttpHeaders.WEBSOCKET.toString();
    boolean upgrade = headers.getAll(HttpHeaders.UPGRADE).stream()
      .anyMatch(val -> val.equalsIgnoreCase(value));
```

Drawbacks:

1. Testing a header value is a common task, especially if you write edge services or API gateways
2. The header value in this example should be tested in a case-insensitive way; so the `HttpHeaders` constant must be first converted to `String` (no `equalsIgnoreCase` on `CharSequence`)
3. A list of all headers needs to be created before testing

### DIFF
--- a/src/main/java/io/vertx/core/MultiMap.java
+++ b/src/main/java/io/vertx/core/MultiMap.java
@@ -20,6 +20,7 @@ import io.vertx.core.http.CaseInsensitiveHeaders;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
 
 /**
  * This class represents a MultiMap of String keys to a List of String values.
@@ -91,6 +92,34 @@ public interface MultiMap extends Iterable<Map.Entry<String, String>> {
    */
   @GenIgnore
   boolean contains(CharSequence name);
+
+  /**
+   * Check if there is a header with the specified {@code name} and {@code value}.
+   *
+   * If {@code caseInsensitive} is {@code true}, value is compared in a case-insensitive way.
+   *
+   * @param name The name to search for
+   * @return true if at least one entry is found
+   */
+  default boolean contains(String name, String value, boolean caseInsensitive) {
+    return getAll(name).stream()
+      .anyMatch(val -> caseInsensitive ? val.equalsIgnoreCase(value) : val.equals(value));
+  }
+
+  /**
+   * Like {@link #contains(String, String, boolean)} but accepting {@code CharSequence} parameters.
+   */
+  @GenIgnore
+  default boolean contains(CharSequence name, CharSequence value, boolean caseInsensitive) {
+    Predicate<String> predicate;
+    if (caseInsensitive) {
+      String valueAsString = value.toString();
+      predicate = val -> val.equalsIgnoreCase(valueAsString);
+    } else {
+      predicate = val -> val.contentEquals(value);
+    }
+    return getAll(name).stream().anyMatch(predicate);
+  }
 
   /**
    * Return true if empty
--- a/src/main/java/io/vertx/core/MultiMap.java
+++ b/src/main/java/io/vertx/core/MultiMap.java
@@ -96,10 +96,11 @@ public interface MultiMap extends Iterable<Map.Entry<String, String>> {
   /**
    * Check if there is a header with the specified {@code name} and {@code value}.
    *
-   * If {@code caseInsensitive} is {@code true}, value is compared in a case-insensitive way.
+   * If {@code caseInsensitive} is {@code true}, {@code value} is compared in a case-insensitive way.
    *
-   * @param name The name to search for
-   * @return true if at least one entry is found
+   * @param name the name to search for
+   * @param value the value to search for
+   * @return {@code true} if at least one entry is found
    */
   default boolean contains(String name, String value, boolean caseInsensitive) {
     return getAll(name).stream()
--- a/src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
@@ -11,9 +11,7 @@
 package io.vertx.core.http.impl;
 
 import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http2.DefaultHttp2Headers;
 import io.netty.handler.codec.http2.Http2Headers;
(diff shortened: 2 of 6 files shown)


tsegismont:
@vietj can you please take a look? I split the change so that each commit message explains the motivation.

src/main/java/io/vertx/core/MultiMap.java
```diff
@@ -92,6 +93,34 @@ static MultiMap caseInsensitiveMultiMap() {
   @GenIgnore
   boolean contains(CharSequence name);
 
+  /**
+   * Check if there is a header with the specified {@code name} and {@code value}.
+   *
+   * If {@code caseInsensitive} is {@code true}, value is compared in a case-insensitive way.
```
vietj: _{@code value}_

src/main/java/io/vertx/core/MultiMap.java
```diff
@@ -92,6 +93,34 @@ static MultiMap caseInsensitiveMultiMap() {
...
+   *
+   * @param name The name to search for
+   * @return true if at least one entry is found
```
vietj: _{@code true}_

src/main/java/io/vertx/core/MultiMap.java
```diff
@@ -92,6 +93,34 @@ static MultiMap caseInsensitiveMultiMap() {
   @GenIgnore
   boolean contains(CharSequence name);
 
+  /**
+   * Check if there is a header with the specified {@code name} and {@code value}.
+   *
+   * If {@code caseInsensitive} is {@code true}, value is compared in a case-insensitive way.
+   *
+   * @param name The name to search for
```
vietj: no _@param_ for value ?

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -23,12 +25,18 @@
...
 
+import static io.netty.util.AsciiString.*;
+
 /**
  * @author <a href="mailto:nmaurer@redhat.com">Norman Maurer</a>
  */
 public class Http2HeadersAdaptor implements MultiMap {
 
   static CharSequence toLowerCase(CharSequence s) {
+    if (s instanceof AsciiString) {
```
vietj: how is this related to this PR ?

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
```diff
@@ -228,6 +236,24 @@ public boolean contains(CharSequence name) {
     return headers.contains(toLowerCase(name));
   }
 
+  @Override
+  public boolean contains(String name, String value, boolean caseInsensitive) {
+    if (headers instanceof DefaultHttp2Headers) {
+      DefaultHttp2Headers h = (DefaultHttp2Headers) headers;
```
vietj: perhaps we should submit a patch for Netty `Http2Headers` and add this method ? this look a bit awkward - wdyt @normanmaurer

src/main/java/io/vertx/core/MultiMap.java
tsegismont: @vietj Just fixed the Multimap Javdoc

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
tsegismont: It's not strictly related to this PR. But while working on this (and thinking about unnecessary conversions) I stumbled upon this method.

By the way, this property of AsciiString is important to note, if we want to resurrect #1878 : having the HttpHeaders constant all lower case avoids repeated conversion for HTTP2

src/main/java/io/vertx/core/http/impl/Http2HeadersAdaptor.java
normanmaurer: What exactly you want to submit ?

vietj: to have `Http2Headers` declare the method `contains(String name, String value, boolean caseInsensitive)` or with CharSequence

tsegismont: @normanmaurer I just pushed a commit in my fork to give you an idea: https://github.com/tsegismont/netty/commit/2fb2d359586bfe54f6f03445a76a1b1527e924fc It would be great if you can take a look.

tsegismont: @vietj since the `Http2HeadersAdaptor` changes are the only thing left discussed, how about reverting them and merge this PR? We can revisit if the commit above lands in Netty and, in the meantime, it will work in HTTP2 with the default impl from `Multimap`

tsegismont: @normanmaurer how can we move the Netty contrib forward? Should I open an issue on your side?

tsegismont: > since the Http2HeadersAdaptor changes are the only thing left discussed, how about reverting them and merge this PR?

@vietj ^^

normanmaurer: @tsegismont just reviewed your patch... please open a PR and sorry for the delay.

tsegismont:
@vietj can you take another look before we close the 3.5.1 round? Thanks

tsegismont:
@vietj done

PR #2261, opened 2018-01-04
small fix to exception logging

kurtostfeld:
Signed-off-by: Kurt Ostfeld <kurt@samba.tv>

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
+++ b/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
@@ -199,7 +199,7 @@ public abstract class ConnectionBase {
     if (exceptionHandler != null) {
       exceptionHandler.handle(t);
     } else {
-      log.error(t);
+      log.error("exception", t);
     }
   }
 


vietj:
can you describe _what_ it fixes ?

kurtostfeld:
The existing log line will do an slf4j `logger.error(exception.toString())`. The new log line should do an slf4j `logger.error("exception", exception)` and if logging is configured accordingly, it will provide full stack trace output, which is what I wanted to see.

vietj:
I believe that you need to configure your logger to see the full stack trace and there is no need to add an artificial "exception" message for this, what is your logger ?

kurtostfeld:
Line 116 here: https://github.com/eclipse/vert.x/blob/master/src/test/java/io/vertx/test/core/SocksProxy.java

is already doing the same thing. Sure, the "exception" string is not useful. The io.vertx.core.logging.Logger class has the `error` method is `error(final Object message, final Throwable t)`; you want to pass the exception as the `Throwable t`, not as the `Object message`. My logging is configured correctly.

vietj:
same as https://github.com/eclipse/vert.x/pull/2277

Issue #2226, opened 2017-11-30
Use new SniCompletionEvent from Netty


vietj:
and try to remove `VertxSniHandler` - see https://github.com/netty/netty/pull/7444

Sammers21:
Hi, I m interested in working on the issue. Can you point me out?

vietj:
look at `VertxSniHandler#replaceHandler` method, the idea is to use instead the event fired to obtain the hostname rather than the current override of `replaceHandler` to set the `VertxSniHandler.SERVER_NAME_ATTR` on the channel.

vietj:
keep in mind this requires a later version than the actual netty version

PR #2262, opened 2018-01-06
Use new SniCompletionEvent from Netty

Sammers21:
A solution for #2226

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/VertxSniHandler.java
+++ b/src/main/java/io/vertx/core/net/impl/VertxSniHandler.java
@@ -20,18 +20,13 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.handler.ssl.SniCompletionEvent;
 import io.netty.handler.ssl.SniHandler;
-import io.netty.handler.ssl.SslContext;
-import io.netty.handler.ssl.SslHandler;
 import io.netty.util.Attribute;
 import io.netty.util.AttributeKey;
-import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.DefaultPromise;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.Promise;
 import io.vertx.core.impl.VertxInternal;
-
-import javax.net.ssl.SSLEngine;
 
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/net/impl/VertxSniHandler.java
```diff
@@ -54,6 +51,23 @@ protected EventExecutor executor() {
         if (ctx == null) {
           throw new IllegalStateException();
         }
+        ctx.pipeline().addFirst(new ChannelInboundHandlerAdapter() {
```
Sammers21: Not sure is that correct place on the handler in the pipeline...

Sammers21:
@vietj , should I actually submit a PR to https://github.com/vert-x3/vertx-dependencies with Netty version upgrade to _4.1.18.Final_

vietj:
@Sammers21 no, I will take care of merging this one as I'll do a dependency upgrade early next week. Besides we should drop using vertx-dependencies in theory because otherwise we would need a CQ for it.

vietj:
I'm planning a full dependency update early next week (cf vertx-dev list)

vietj:
@Sammers21 can you try to drop the vertx sni handler class and do that using an event when the HTTP pipeline is assembled ?

vietj:
@Sammers21 any news about this ? Netty 4.1.19.Final support has been merged in master

Sammers21:
@vietj, great. But VertxSniHandler is also used in _NetServerImpl_, _NetSocketImpl_ .... if we would drop VertxSniHandler does it may cause code duplication. Is not it worth it to keep VertxSniHandler?

vietj:
good point, let's keep it this way

Sammers21:
@vietj, is something wrong?

vietj:
yes it fails in CI, I had to revert.

Sammers21:
where can I watch build results?

Sammers21:
I would like to fix it

vietj:
the build actually froze , I could reproduce on my machine, so just grab the master after I merged and test it

Sammers21:
@vietj, I have set up Travis CI on my vertx fork. And when building the latest version of master... it fails... here is details: https://api.travis-ci.org/v3/job/326756484/log.txt . Is it expected?

Sammers21:
and here is a link to the job: https://travis-ci.org/Sammers21/vert.x/jobs/326756484

Sammers21:
while the PR job without your revert failed timeout: https://travis-ci.org/Sammers21/vert.x/jobs/326740521

vietj:
wdym ?

Sammers21:
I m trying to figure out where is the problem with my code. To begin with i m trying to ensure that all test passed without my changes made in this PR, but when i m running `mvn clean install` on latest version of master branch it [fails](https://api.travis-ci.org/v3/job/326756484/log.txt). I don't understand is it expected or not? Is my approach is right?

Sammers21:
is it clear? :-)

Sammers21:
But it works fine on my laptop.

vietj:
I think your code is adding an handler but is not removing it, so I've mixed feelings about this approach after all because it's more complex than what I thought.

Sammers21:
So now you think that it is better to leave it as is? And it is not worth it to make it works?

Sammers21:
Or I should try one more time? :))

vietj:
I think it's better to leave it as is, I would have tried and failed as well if that makes you more comfortable :-)

PR #2270, opened 2018-01-10
Add a close method to AsyncMultiMap

tsegismont:
This allows cluster managers to free resources they use to maintain the state.

Needed for vert-x3/issues#316

### DIFF
--- a/src/main/java/io/vertx/core/spi/cluster/AsyncMultiMap.java
+++ b/src/main/java/io/vertx/core/spi/cluster/AsyncMultiMap.java
@@ -26,6 +26,11 @@ import java.util.function.Predicate;
  * The cluster implementation should ensure that any entries placed in the map from any node are available on any
  * node of the cluster.
  *
+ * <p>
+ *   <strong>Warning</strong>: always call {@link #close()} when you are done with an {@link AsyncMultiMap} instance.
+ *   Cluster managers may hold resources to maintain its state.
+ * </p>
+ *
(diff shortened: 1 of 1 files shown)


tsegismont:
@vietj PTAL

vietj:
how did it work before without a close method ?

tsegismont:
> how did it work before without a close method ?

`ClusteredEventBus` just call it once. The problem would happen only if a user creates many multimaps for his own purposes.

tsegismont:
> can we try instead to use java refs ?

Sorry I can't see what you mean... :(

vietj:
weak references, look at NetClientImpl

tsegismont:
Not sure how we could. We have to unregister listeners on the HZ map and
the ISPN cache. See the related cluster manager PRs

2018-01-10 16:40 GMT+01:00 Julien Viet <notifications@github.com>:

vietj:
I think you can on the map proxy itself as only the client has a ref to it, can't you ?

tsegismont:
Sorry I don't understand. In any case, HZ or ISPN may still hold resources
related to the listener. So unregistering it seems mandatory

2018-01-10 16:52 GMT+01:00 Julien Viet <notifications@github.com>:

tsegismont:
@vietj can you take another look before we close the 3.5.1 round? Thanks

tsegismont:
vert-x3/issues#316 closed (won't fix)

PR #2274, opened 2018-01-12
correct claim io.vertx.core.http.HttpClient#websocketStream to be fluent

wu-lee:
See https://github.com/vert-x3/issues/issues/319

Signed-off-by: wu-lee <eclipse.wu-lee@noodlefactory.co.uk>


tsegismont:
@wu-lee thanks!

Issue #2230, opened 2017-11-30
HTTP/1.x server connection back-pressure management might reorder request body chunks


st-h:
As discussed [here](https://groups.google.com/forum/#!topic/vertx/7EPAUv73CtY) previously, starting with vert.x 3.5.0 I was noticing failing integration tests when uploading large (at least a few MB) files. The issue so far only appeared when running against mongo in a docker container or with travis ci. Downgrading to vert.x 3.4.0 seems to fix the issue. This issue does not appear on every build.
I once examined the differences between the uploaded and the persisted file and it showed that about the first 1054400 bytes were correctly persisted, but subsequent bytes were corrupt. I could not find any issues with file download. Just the upload seems to encouter this issue.

I have put a branch together which explicitly compares the md5 checksums of the uploaded file in addition to the actual bytes with vertx 3.5.1-SNAPSHOT [travis available here](https://travis-ci.org/st-h/vertx-mongo-streams/jobs/309572086) - [source](https://github.com/st-h/vertx-mongo-streams/tree/vertx-3.5.x)

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.github.sth.vertx.mongo.streams.GridFSInOutStreamIT
Cluster created with settings {hosts=[localhost:27017], mode=SINGLE, requiredClusterType=UNKNOWN, serverSelectionTimeout='30000 ms', maxWaitQueueSize=500} 
Opened connection [connectionId{localValue:1, serverValue:1}] to localhost:27017 
Monitor thread successfully connected to server with description ServerDescription{address=localhost:27017, type=STANDALONE, state=CONNECTED, ok=true, version=ServerVersion{versionList=[3, 4, 10]}, minWireVersion=0, maxWireVersion=5, maxDocumentSize=16777216, roundTripTimeNanos=4467832} 
Opened connection [connectionId{localValue:2, serverValue:2}] to localhost:27017 
uploaded file md5: e875acd2c89e7db444f04c54c5b01460
java.lang.AssertionError: uploaded file md5 does not match md5 calculated on server. Not equals : 31C48A5A246C9352E0EB846FEFD0A498 != E875ACD2C89E7DB444F04C54C5B01460 
Cluster created with settings {hosts=[localhost:27017], mode=SINGLE, requiredClusterType=UNKNOWN, serverSelectionTimeout='30000 ms', maxWaitQueueSize=500} 
Opened connection [connectionId{localValue:3, serverValue:3}] to localhost:27017 
[...]

PR #2275, opened 2018-01-12
wip #2230 refactor buffer Http1xServerConnection 

st-h:
as discussed in #2230 this PR splits the single queue from Http1xServerConnection into two separate
queues. I tried to move stuff that belongs to the Request or the Response into the appropriate places, but it turned out to be difficult within the current structure to separate concerns clearly. 
There are still some things which could probably be cleaned up, but I wanted to submit this by now as I am unsure if splitting the queues really reduces complexity. And I am not 100% sure if the vertx context related stuff makes sense.


vietj:
Actually  I do have code for that and I would like rather for 3.6 (I can push the code I have).

vietj:
please see this commit https://github.com/vietj/vert.x/commit/19f60b8b1fe06c450c4e03fd9ab0df43244637a0

vietj:
the main motivation for this refactor (look at the previous commit) is to have an improvements of `ReadStream` so when it is _paused_, a new `read` method can read a requested number of items from the stream.

The implementation of this is the `EventQueue` that allows to _push_ items in the queue and have the `ReadStream` delegate the item management to the queue. Without this `HttpServerRequestImpl ` using it is very difficult.

The main overall purpose of this change are:

- simplify the implementation of Reactive-Streams from `ReadStream` (so basically a `ReadStream` in paused mode behave like a unicast publisher) : the code in vertx-reactive-streams, vertx-rx-java and vertx-rx-java2 would be simplified and possibly simpler (merging the implicit `ReadStream` queue + the adapter queue into a single queue)
- provide a base class for implementing `ReadStream` (`EventQueueBase`) that basically is a queue with an `handle` method

vietj:
and that would be for 3.6 I think - for now I'm just trying to implement in vertx-core for a few important cases such as http server, async file, then try to see how it can be used in vertx-rxjava2 and reactive-streams.

st-h:
> Actually I do have code for that and I would like rather for 3.6

Well, that must have been quite a misunderstanding. Feel free to close this then.

vietj:
n/p :-) , I t hin that before 3.6 we will ahve to check it still works with vertx-mongo-streams

vietj:
thanks for the time spent

PR #2277, opened 2018-01-16
Properly log the exception, showing the trace

ctron:
This change adds a log message to the log output. Which also results in logging the stack trace of the exception. Otherwise only the exception name will be logged which isn't really helpful as the `ConnectionBase` is already missing some handling of the exception in the first place. Finding the root cause of this exception can be rather hard.

Signed-off-by: Jens Reimann <jreimann@redhat.com>

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
+++ b/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2018 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -199,7 +199,7 @@ public abstract class ConnectionBase {
     if (exceptionHandler != null) {
       exceptionHandler.handle(t);
     } else {
-      log.error(t);
+      log.error("Unhandled exception", t);
     }
   }
 


src/main/java/io/vertx/core/net/impl/ConnectionBase.java
```diff
@@ -199,7 +199,7 @@ protected synchronized void handleException(Throwable t) {
     if (exceptionHandler != null) {
       exceptionHandler.handle(t);
     } else {
-      log.error(t);
+      log.error("Unhandled exception", t);
```
vietj: shall we rather have the exception message to be used ?

vietj:
isn't it a logging issue btw ?

vietj:
I mean shouldn't we have isntead:

```
public void error(final Throwable error) { ... }
```

in the loggger API ?

ctron:
This could he hard to translate to other logging frameworks. Aside from that I think having the additional information that this is an unhandled exception is actually useful.

tsegismont:
@ctron is right, logging backends don't have such a method. Which makes sense if you think about it: it's better to provide some context instead of simply dumping the stacktrace

vietj:
@tsegismont, the point is that calling `Logger#log(Object)` with `logger.log(throwable)` is simply wrong, and saying "Unhandled exception" as context isn't a lot of context

vietj:
I can count none the less than 10 instances of `Logger#log(Object)` called with a throwable argument in Vert.x Core

tsegismont:
> the point is that calling Logger#log(Object) with logger.log(throwable) is simply wrong

Agreed

> and saying "Unhandled exception" as context isn't a lot of context

Agreed. I was talking about the reason why logging backend don't have a log(throwable) method. Not saying that "Unhandled exception" is enough context

ctron:
> I can count none the less than 10 instances of Logger#log(Object) called with a throwable argument in Vert.x Core

So I think fixing the root cause would make more sense to me, than to legitimize the wrong behavior.

> and saying "Unhandled exception" as context isn't a lot of context

What more context would you like me to provide at this point?

vietj:
@ctron the subclass of connection base that uses it for instance

vietj:
well for now let's be pragmatic and merge this one and we need to consider adding a log(Throwable) error in the logger as well

PR #2279, opened 2018-01-17
Harmonizing SSL/TLS protocol options

philrykoff:
Implementation for https://github.com/eclipse/vert.x/issues/2179:

> Right now, `SSLHelper` is 
> 
> 1. taking the `DEFAULT_ENABLED_PROTOCOLS` as a basis protocol list
> 2. removing those protocols from that protocol list which are not already enabled in the `SSLEngine` by default (effectively leaving only those protocols in place which were enabled in the `SSLEngine` already)
> 3. if the developer added protocols to the `options`, the current protocol list is filtered to only leave those
> 
> This means that there is currently no way to add protocols not already enabled in both `SSLHelper` and `SSLEngine` by default - for my Java 8 setup, that is SSLv3 (disabled by both) and SSLv2Hello (enabled by `SSLHelper`, but disabled by `SSLEngine`).
> 
> As I need SSLv3 and SSLv2Hello support for a customer's legacy devices, I'd like to bring this forward. Let's not discuss the security implications here ☹️ 
> 
> I propose to move the `DEFAULT_ENABLED_PROTOCOLS` field from `SSLHelper` to `TCPSSLOptions`, where all other `DEFAULT_*` are. There, this field will be documented, so that developers know what the defaults are. A `addEnabledSecureTransportProtocol(String protocol)` method is already in place, a `removeEnabledSecureTransportProtocol(String protocol)` should be added. While the names are quite long, they are quite descriptive and the add-method is backwards compatible.
> 
> `SSLHelper` can then simply get that list, remove those protocols not supported by `SSLEngine` and tell `SSLEngine` to use the remainder of them.
> 
> `TCPSSLOptions.addEnabledSecureTransportProtocol` and `getEnabledSecureTransportProtocols` are being invoked/overwritten in several places:
> 
> - `ClientOptionsBase`, `HttpClientOptions`, `HttpServerOptions`, `NetClientOptions`, `NetServerOptions`: `removeEnabledSecureTransportProtocol` needs to be added
> - `EventBusOptionsConverter` and `TCPSSLOptionsConverter`: nothing to do
> - `NetExamples`: make it clearer by comments and code what is enabled and what not
> - `TCPSSLOptions` constructor: take over `other` values
> - `SSLHelper`: as described above
> - `SSLHelperTest`, `HttpTLSTest`, `NetTest`: analyze how they worked until now; add tests for additional protocols if possible

Signed-off-by: Philipp Lehmann <github@phil.to>

### DIFF
--- a/src/main/java/io/vertx/core/net/TCPSSLOptions.java
+++ b/src/main/java/io/vertx/core/net/TCPSSLOptions.java
@@ -71,7 +71,7 @@ public abstract class TCPSSLOptions extends NetworkOptions {
    *
    * SSLv3 is NOT enabled due to POODLE vulnerability http://en.wikipedia.org/wiki/POODLE
    */
-  public static final String[] DEFAULT_ENABLED_SECURE_TRANSPORT_PROTOCOLS = {"SSLv2Hello", "TLSv1", "TLSv1.1", "TLSv1.2"};
+  public static final List<String> DEFAULT_ENABLED_SECURE_TRANSPORT_PROTOCOLS = Collections.unmodifiableList(Arrays.asList("SSLv2Hello", "TLSv1", "TLSv1.1", "TLSv1.2"));
 
   /**
    * The default TCP_FASTOPEN value = false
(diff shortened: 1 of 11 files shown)


src/main/java/io/vertx/core/net/TCPSSLOptions.java
```diff
@@ -66,6 +66,13 @@
    */
   public static final SSLEngineOptions DEFAULT_SSL_ENGINE = null;
 
+  /**
+   * The default ENABLED_SECURE_TRANSPORT_PROTOCOLS value = { "SSLv2Hello", "TLSv1", "TLSv1.1", "TLSv1.2" }
+   *
+   * SSLv3 is NOT enabled due to POODLE vulnerability http://en.wikipedia.org/wiki/POODLE
+   */
+  public static final String[] DEFAULT_ENABLED_SECURE_TRANSPORT_PROTOCOLS = {"SSLv2Hello", "TLSv1", "TLSv1.1", "TLSv1.2"};
```
vietj: Rather use an unmodifiable List<String> that is copied in the constructor

philrykoff:
Thanks for your review! I added a commit hopefully fixing it.

philrykoff:
I just now tested my implementation with an old client that can only speak SSLv3. Using this patch, it works.

vietj:
@johnoliver can you have a look too ?

vietj:
@johnoliver can you confirm it is ok for you ?

johnoliver:
Certainly looks good to me

philrykoff:
Thanks! Just wondering: are there no automated tests for PRs against eclipse vert.x?

vietj:
cloudbees does not support it

Issue #2276, opened 2018-01-16
WebSocket API close frame payload


vietj:
The WebSocket API should allow to set and receive the close frame payload.

```
ws.close(1000); // Close with a status code
ws.close(1000, "some reason");
```

Close frame should be propagated on the client, i.e before calling the close handler the websocket should forward a close frame with the status and reason.

PR #2280, opened 2018-01-17
Websocket close payload

slinkydeveloper:
Refeers to #2276


src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -239,6 +239,15 @@
...
   void close();
 
+  /*
+   * Close sending a close frame with specified status code and reason. You can give a look at various close payloads
+   * here: <a href="https://tools.ietf.org/html/rfc6455#section-7.4.1">RFC6455 Section 7.4.1</a>
+   *
+   * @param statusCode Status code
+   * @param reason reason of closure
+   */
+  void closeWithReason(short statusCode, @Nullable String reason);
```
vietj: no need to call `closeWithReason`, just `close` is enough

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -676,6 +676,7 @@ protected void handleMessage(Http1xServerConnection conn, ContextImpl context, C
             chctx.writeAndFlush(new PongWebSocketFrame(wsFrame.getBinaryData().copy()));
             break;
           case CLOSE:
+            conn.handleMessage(msg);
```
vietj: rather do that after following statement, so a failure in the conn.handleMessage does not prevent the frame echo to be sent.

vietj:
how can the client be aware of the status code and reason ? I think we should have them on `WebSocketFrame` so the user can access directly.

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -239,6 +239,15 @@
...
   void close();
 
+  /*
+   * Close sending a close frame with specified status code and reason. You can give a look at various close payloads
+   * here: <a href="https://tools.ietf.org/html/rfc6455#section-7.4.1">RFC6455 Section 7.4.1</a>
+   *
+   * @param statusCode Status code
+   * @param reason reason of closure
+   */
+  void close(short statusCode, @Nullable String reason);
```
vietj: can we have an overload with only `close(short)` reason ?

src/main/java/io/vertx/core/http/impl/ws/WebSocketCloseFrameCode.java
```diff
@@ -0,0 +1,37 @@
...
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+
+import java.nio.charset.Charset;
+
+/**
+ * @author Francesco Guardiani @slinkydeveloper
+ */
+public enum WebSocketCloseFrameCode {
```
vietj: I don't think we need a special enum for that, can you make a single static method on HttpUtils instead ?

src/main/java/io/vertx/core/http/impl/WebSocketFrameFactoryImpl.java
```diff
@@ -11,11 +11,16 @@
...
 
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.buffer.UnpooledDirectByteBuf;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.WebSocketFrame;
 import io.vertx.core.http.impl.ws.WebSocketFrameImpl;
 import io.vertx.core.spi.WebSocketFrameFactory;
 
+import java.nio.charset.Charset;
```
vietj: this file does not seem to be modified, we need to avoid modifications in it (imports and  blank line)

src/main/java/io/vertx/core/http/impl/ws/WebSocketCloseFrameCode.java
slinkydeveloper: The initial idea was to provide a enum with status code. I'm removing it

Issue #2298, opened 2018-01-28
Make LogDelegateFactory configurable via VertxOptions to avoid classloading issues in ContextClassLoader sensitive environments such as OSGi


vorburger:
The use of:

```java
String className = System.getProperty("vertx.logger-delegate-factory-class-name");
Thread.currentThread().getContextClassLoader().loadClass(className);
```

in `io.vertx.core.logging.LoggerFactory` can be problematic in "context ClassLoader sensitive environments", such as OSGi (but this could also affect use of Vert.x within other environments that do funky class loading); here are some links with background about why that is so:

* http://njbartlett.name/2010/08/30/osgi-readiness-loading-classes.html
* http://wiki.eclipse.org/Context_Class_Loader_Enhancements
* https://helpx.adobe.com/experience-manager/kb/OsgiClassLoading3Party.html
* https://issues.apache.org/jira/browse/FELIX-2951

I've actually run into this problem myself, see the detailed error below (which probably won't really make all that sense, without more context ; it's from https://github.com/vorburger/minecraft-storeys-maker/ running under https://github.com/vorburger/ch.vorburger.minecraft.osgi ... but ignore that).

I was able to work-around this in my code by instead of doing this:

```java
System.setProperty("vertx.logger-delegate-factory-class-name", SLF4JLogDelegateFactory.class.getName());
vertx = Vertx.vertx(new VertxOptions().setWorkerPoolSize(23));
```

using this beauty:

```java
ClassLoader tccl = Thread.currentThread().getContextClassLoader();
Thread.currentThread().setContextClassLoader(this.getClass().getClassLoader());
try {
    System.setProperty("vertx.logger-delegate-factory-class-name", SLF4JLogDelegateFactory.class.getName());
    vertx = Vertx.vertx(new VertxOptions().setWorkerPoolSize(23));
} finally {
    Thread.currentThread().setContextClassLoader(tccl);
}
```

but in an ideal world... this kind of hand-stand should not be required?

One possible solution to this could be to offer an alternative to the system property and ClassLoader mess by simply letting users of Vert.x specify an instance of their desired `io.vertx.core.spi.logging.LogDelegateFactory` programmatically - I guess this would naturally fit into the `io.vertx.core.VertxOptions.VertxOptions` ?

[...]

PR #2299, opened 2018-01-29
make LoggerFactory try its own ClassLoader if TCCCL fails, see #2298

vorburger:
attempts to use LoggerFactory.class.getClassLoader(); as a fallback if
Thread.currentThread().getContextClassLoader() fails to be able to load
the LogDelegateFactory.

This is because the use of getContextClassLoader() can be problematic in
"context ClassLoader sensitive environments", such as OSGi; see e.g.
http://njbartlett.name/2010/08/30/osgi-readiness-loading-classes.html or
http://wiki.eclipse.org/Context_Class_Loader_Enhancements

PS, minor: The renaming of delegateFactory variable to
newDelegateFactory just for clarity and to avoid IDE/javac warning re.
possible confusion with static field of same name.

Signed-off-by: Michael Vorburger <vorburger@redhat.com>

### DIFF
--- a/src/main/java/io/vertx/core/logging/LoggerFactory.java
+++ b/src/main/java/io/vertx/core/logging/LoggerFactory.java
@@ -33,7 +33,7 @@ public class LoggerFactory {
   }
 
   public static synchronized void initialise() {
-    LogDelegateFactory delegateFactory;
+    LogDelegateFactory newDelegateFactory;
 
     // If a system property is specified then this overrides any delegate factory which is set
     // programmatically - this is primarily of use so we can configure the logger delegate on the client side.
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/logging/LoggerFactory.java
```diff
@@ -46,18 +46,27 @@ public static synchronized void initialise() {
...
     if (className != null) {
-      ClassLoader loader = Thread.currentThread().getContextClassLoader();
+      ClassLoader tcccl = Thread.currentThread().getContextClassLoader();
       try {
-        Class<?> clz = loader.loadClass(className);
-        delegateFactory = (LogDelegateFactory) clz.newInstance();
-      } catch (Exception e) {
-        throw new IllegalArgumentException("Error instantiating transformer class \"" + className + "\"", e);
+        newDelegateFactory = loadClass(className, tcccl);
+      } catch (IllegalArgumentException e) {
```
cescoffier: Why catching an IllegalArgumentException here? ClassNotFoundException should be fine, no?

cescoffier: Sorry, got it (didn't read the `loadClass` method.

src/main/java/io/vertx/core/logging/LoggerFactory.java
```diff
@@ -46,18 +46,27 @@ public static synchronized void initialise() {
...
-    LoggerFactory.delegateFactory = delegateFactory;
+    LoggerFactory.delegateFactory = newDelegateFactory;
+  }
+
+  private static LogDelegateFactory loadClass(String className, ClassLoader classLoader) {
+    try {
+      Class<?> clz = classLoader.loadClass(className);
+      return (LogDelegateFactory) clz.newInstance();
+    } catch (Exception e) {
+      throw new IllegalArgumentException("Error instantiating transformer class \"" + className + "\"", e);
```
cescoffier: why _transformer_?

cescoffier:
I believe this PR is superseded by #2300

vorburger:
> I believe this PR is superseded by #2300 

if you guys in Vert.x don't squash upon merge then #2300 with 2 commits including this instead of this is fine; but I wanted to keep the two things separate, and filed them separately because I'm too using to a Gerrit instead of a GitHub Pull Request workflow... I'll amend #2300 ASAP!

PR #2305, opened 2018-01-29
Limit the size of the cached request body for following redirections …

### LINKED ISSUE #2304: HttpClient should not cache request body for redirection purpose
motivation: currently the http client redirection following feature will cache the sent body along with the request, this may is not desirable

change: remove the request body cache

vietj:
…- fixes #2304

### DIFF
--- a/src/main/java/examples/HTTPExamples.java
+++ b/src/main/java/examples/HTTPExamples.java
@@ -17,7 +17,6 @@ import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.file.AsyncFile;
 import io.vertx.core.http.*;
-import io.vertx.core.net.JksOptions;
 import io.vertx.core.net.ProxyOptions;
 import io.vertx.core.net.ProxyType;
 import io.vertx.core.streams.Pump;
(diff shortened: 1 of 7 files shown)


tsegismont:
@vietj I'm not comfortable with this. As a default, we should follow the W3C recommendations:

>If the 301 status code is received in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might change the conditions under which the request was issued. 

>If the 302 status code is received in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might change the conditions under which the request was issued.

>If the 307 status code is received in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might change the conditions under which the request was issued. 

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

This is what other HTTP clients do (I checked Apache and OkHttp)

vietj:
so we should change the implementation of the default handler and never cache body ?

tsegismont:
I believe so.

2018-01-30 10:57 GMT+01:00 Julien Viet <notifications@github.com>:

vietj:
ok better fix the cause than the symptoms, I'll work on a PR for that today

vietj:
how about 303 code ?

tsegismont:
>The response to the request can be found under a different URI and SHOULD be retrieved using a GET method on that resource. This method exists primarily to allow the output of a POST-activated script to redirect the user agent to a selected resource. The new URI is not a substitute reference for the originally requested resource.

vietj:
@tsegismont please see https://github.com/eclipse/vert.x/pull/2308

PR #2308, opened 2018-01-30
HttpClient redirection fixes - fixes #2307 - fixes #2304

### LINKED ISSUE #2304: HttpClient should not cache request body for redirection purpose

### LINKED ISSUE #2307: Fix HttpClient default redirection policy
motivation: the current redirection policy does not follow strictly https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html. It performs a redirection on any HTTP method for the 301, 302 and 307 status instead of only GET and HEAD. In addition for 301 and 302 status it changes the method to GET
change: perform redirection for 301, 302 and 307 status code only when the method is GET or HEAD and do not change the method

vietj:


vietj:
yes I planned to do so

PR #2309, opened 2018-01-30
Redirect handler should not decode query params

### LINKED ISSUE #2263: HttpClientImple DEFAULT_HANDLER uses uri.getQuery instead of getRawQuery when extract redirection uri from response location header
Hi,

I found that Vertx HttpClient DEFAULT_HANDLER uses java.net.URI.getQuery instead of getRawQuery when extracts a redirection uri from response location header.

Even Vertx HttpUitls.resolveURIReference uses java.net.URI.getQuery and then redirects it with decoded query.

java.net.URI.getQuery returns decoded query according to documents.
```
String java.net.URI.getQuery()

Returns the decoded query component of this URI.

The string returned by this method is equal to that returned by the getRawQuery method except that all sequences of escaped octets are decoded.
```

In my case, there are many cases that location header value includes URIEncoded query because my service covers not only english alphabet but also ascii unsafe language. 

Redirection target server still needs URIEncoded query for ascii unsafe language.

I've made a patch for that referring to Apache httpClient where uses getRawQuery and then applied it with HttpClient.setRedirectinHandler.

I guess this issue not only for mine.
I think java.net.URI.getRawQuery is proper rather than getQuery.

Do you have any reason for using getQuery I'm not aware of?

tsegismont:
Fixes #2263

### DIFF
--- a/src/test/java/io/vertx/test/core/HttpTest.java
+++ b/src/test/java/io/vertx/test/core/HttpTest.java
@@ -644,7 +644,7 @@ public abstract class HttpTest extends HttpTestBase {
 
   @Test
   public void testParamUmlauteDecoding() throws UnsupportedEncodingException {
-    testParamDecoding("äüö");
+    testParamDecoding("\u00e4\u00fc\u00f6");
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


tsegismont:
@vietj please review

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -3813,6 +3813,44 @@ private void testFoo(String location, String expected) throws Exception {
     }
   }
 
+  @Test
+  public void testFollowRedirectEncodedUrl() throws Exception {
+    String value1 = "한글", value2 = "A B+C";
```
vietj: `value1` should be escaped, i.e with \u00 , we should also do one value with the Euro sign as it's a 3 digits UTF8 encoded value

tsegismont: Done

tsegismont:
@vietj PTAL

PR #2319, opened 2018-02-13
Do not invoke exception handler with wrong exception for every response

gmokki:
For every HTTP/2 connection response has been succesfully written then `handleEnded(false)` method is invoked.

Commit https://github.com/eclipse/vert.x/commit/fdd4c7bdd4cd34ad3e9005e9b94579dfb5035cdd changed added code to invoke the `exceptionHandler` unconditionally with message `"Connection was closed"`. This happens even if the response has been succesfully sent an the connection has not been closed.

The exception should be only generated and reported if `handleEnded` is invoked with `failed`=`true` argument.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
@@ -456,7 +456,7 @@ public class Http2ServerResponseImpl implements HttpServerResponse {
           conn.metrics().responseEnd(metric, this);
         }
       }
-      if (exceptionHandler != null) {
+      if (failed && exceptionHandler != null) {
         conn.getContext().runOnContext(v -> exceptionHandler.handle(new VertxException("Connection was closed")));
       }
       if (endHandler != null) {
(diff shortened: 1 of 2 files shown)


vietj:
can you sign the Eclipse CA and we absolutely need a test for this case ?

gmokki:
Sorry for the long delay.
I modified the test case that was added when the wrong functionality was added to detect this new case.
The test now sends 2 requests, but only the latter one is abruptly closed by the client. The exception handler must be only called once. Reverting the fix the assert fails because the exception handler is called twice.

gmokki:
PS. while trying making the tests work with both HTTP1 and HTTP2 I noticed that the HTTP2 seems to call closehandler for each server response, and HTTP1 only when the actual tcp socket is closed.

vietj:
@gmokki that's what the docs says:

- For HTTP/1.x it is called when the connection is closed before {@code end()} is called, therefore it -    - For HTTP/2 it is called when the related stream is closed, and therefore it will be always be called.

vietj:
perhaps we should revisit to have the same behavior for HTTP/2 than HTTP/1, i.e it should only be closed when the actual response is not ended and the underlying connection is closed.

PR #2330, opened 2018-03-04
Implemented RFC6901 - Json Pointer including tests and docs

slinkydeveloper:


jponge:
@slinkydeveloper don't forget file headers (license, etc)

jponge:
(see https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md#copyright-notice-headers)

pmlopes:
Should this be in core?

slinkydeveloper:
@pmlopes I think this should be in core because is specific of Json support and creating a new package for one/two classes doesn't make sense.

vietj:
at some point we discussed of an "utils" package that would contain things not directly related to core

slinkydeveloper:
A utils package could be interesting for this feature and not only

vietj:
yes, things like ReadStream / WriteStream implementations for testing, etc...

pmlopes:
@slinkydeveloper does this allow searching JSON? @Stwissel has implemented a helper that does searches: https://gist.github.com/Stwissel/58638b219c1ee91ff957995117bee194 are these 2 things related?

Perhaps (if they are) we could align the 2 and it seems that this is a common task users need and keep re-inventing all the time as there is no such thing from core?

Stwissel:
The sticky point in processing JSON is the traversal of Arrays. Especially arrays of objects when you are interested in (returning an array of) values inside each object. How did you model array handling ?

slinkydeveloper:
@Stwissel what do you mean with search?
I've implemented traversal of JsonObject and JsonArray. Give a look at tests:

- https://github.com/eclipse/vert.x/pull/2330/files#diff-9d2ecaf9e18cc243531968a7537144e1R169
- https://github.com/eclipse/vert.x/pull/2330/files#diff-9d2ecaf9e18cc243531968a7537144e1R91

The implementation of query takes an Object as input https://github.com/eclipse/vert.x/pull/2330/files#diff-5173c1f2cd3aee210e63cfd9c0bc2f00R134 .

Note that this is an implementation of RFC6901, allowing traversal based on object/array indexes. I've not implemented nothing similar to "JsonPath", like index regexes, search in array, etc... Because there is no JsonPath spec actually (all implementations refeers to XMLPath spec).

jponge:
On a related note, did you see https://github.com/schibsted/jslt ? It only depends on Jackson, so it may be a good alternative.

slinkydeveloper:
@jponge there are also https://github.com/json-path/JsonPath and https://github.com/jsurfer/JsonSurfer that implements a JsonPath language. I think that, if we want to support one of these alternatives, we should put it out in a separate package, because they bring new deps and because they are not implementing any standard (maybe in reactiverse?)

jponge:
reactiveverse would be a good fit indeed.

PR #2333, opened 2018-03-05
Provide a bit more context to the exception

ctron:
Following up on #2277 this PR adds a bit more context.

Signed-off-by: Jens Reimann <jreimann@redhat.com>


ctron:
Ok, I think there is a bug in the logger implementation of vertx. Or I didn't understand it correctly. The change results in the following exception output:

~~~
11:57:39.488 [vert.x-eventloop-thread-3] ERROR i.vertx.core.net.impl.ConnectionBase - Unhandled exception for connection: {}
java.io.IOException: Connection reset by peer
	at sun.nio.ch.FileDispatcherImpl.read0(Native Method)
	at sun.nio.ch.SocketDispatcher.read(SocketDispatcher.java:39)
	at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223)
	at sun.nio.ch.IOUtil.read(IOUtil.java:192)
	at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:380)
	at io.netty.buffer.PooledUnsafeDirectByteBuf.setBytes(PooledUnsafeDirectByteBuf.java:288)
	at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1106)
	at io.netty.channel.socket.nio.NioSocketChannel.doReadBytes(NioSocketChannel.java:373)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:123)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:645)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:580)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:497)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:459)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
	at java.lang.Thread.run(Thread.java:748)
~~~

I would have expected `{}` to be replaced with the first entry of the parameters. But it looks like it is empty.

vietj:
I'm not sure that {} works for all loggers, @cescoffier ?

ctron:
So what would be the correct what to use parameters?

vietj:
I think you need to append strings

cescoffier:
@vietj @ctron, unfortunately, the `{}` syntax does not work for some logging framework. Vert.x just delegates the message to the underlying logger. So, depending the logging framework you use the syntax may be different:

* Log4J 2 supports `{}`
* SLF4J supports `{}`
* JUL uses `{0}`, `{1}` ...
* Log4J 1 does not support parameterization (so you need to concatenate or use String.format)

ctron:
Seriously? Why is that feature there in the first place?

Ok, I will just re-write this.

ctron:
As there isn't even `isErrorEnabled()` available to check if concatenating can calling `toString()` makes sense I split this up into two branches now. If info is enabled you get the full info. If info is not active you get an error message telling you how to get more information.

vietj:
@ctron we don't try to abstract the underlying logging framework specificities

vietj:
that's also the sorry state of logging in the JVM

PR #2336, opened 2018-03-11
Support setting local port for http and tcp clients

wkgcass:
In current version vert, the local port is randomly picked by the OS, (in `io.vertx.core.net.impl.transport.Transport`, the port is set to `0`), however some company strictly restricts the use of ports, which requires the developer to specify local port even in clients.
This pr adds a `localPort` field in `XXOptions` classes, and sets the port into netty bootstrap instance. Test cases are added as well.

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/transport/Transport.java
+++ b/src/main/java/io/vertx/core/net/impl/transport/Transport.java
@@ -214,7 +214,9 @@ public class Transport {
     setOption("TCK_FASTOPEN", options.isTcpFastOpen(), setter);
     setOption("SO_REUSEPORT", options.isReusePort(), setter);
     if (options.getLocalAddress() != null) {
-      bootstrap.localAddress(options.getLocalAddress(), 0);
+      bootstrap.localAddress(options.getLocalAddress(), options.getLocalPort());
+    } else if (options.getLocalPort() != 0) {
+      bootstrap.localAddress(options.getLocalPort());
     }
     bootstrap.option(ChannelOption.TCP_NODELAY, options.isTcpNoDelay());
     if (options.getSendBufferSize() != -1) {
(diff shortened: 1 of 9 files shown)


vietj:
any news about this on going PR ?

src/main/java/io/vertx/core/net/impl/transport/Transport.java
```diff
@@ -214,7 +214,9 @@ public void configure(ClientOptionsBase options, Bootstrap bootstrap) {
     setOption("TCK_FASTOPEN", options.isTcpFastOpen(), setter);
     setOption("SO_REUSEPORT", options.isReusePort(), setter);
     if (options.getLocalAddress() != null) {
-      bootstrap.localAddress(options.getLocalAddress(), 0);
+      bootstrap.localAddress(options.getLocalAddress(), options.getLocalPort());
+    } else if (options.getLocalPort() != 0) {
```
julianladisch: To be consistent with the previous line you should replace `else if (options.getLocalPort() != 0) {` by `else {`

leotu:
Currently, I face the problem without local port method (net/http client), I try to write  same
example NAT traversal:
https://en.wikipedia.org/wiki/NAT_traversal
https://www2.cs.arizona.edu/~bzhang/nat/

vietj:
we can do something for `NetClient` but we cannot really for `HttpClient` at the moment because the client uses pooling. We could support it for `HttpClient` if we have the ability to recycle a `NetSocket` into an `HttpClient` (which is something I would like to have in Vert.x 4 at some point).

vietj:
I think this PR is valid but was never happy with the fact it uses ClientOptions, now we do have ConnectOptions in vertx 5 that make sense to apply at, please reopen a PR with that feature

PR #2337, opened 2018-03-12
Http client request timeout with hashedWheelTimer

pifuant:


vietj:
can you explain what is the benefit to use this instead of vert.x timer ?

vietj:
what's your motiviation ? are you experiencing issues ?

pifuant:
@vietj 

I think hashedWheelTimer is better for io timeout, so I make a PR.

vietj:
the actual timers in Vert.x are done on the event loop thread and your current patch uses a timer that has its own thread, so will break the actual concurrent model that expects all tasks to be fired on the same thread

pifuant:
We can modify it  on the event loop thread, what do you think?

pifuant:
https://github.com/netty/netty/blob/48df2f66b858f3ce2a0a069dafb1332da55c47f9/common/src/main/java/io/netty/util/HashedWheelTimer.java#L565

Let the event loop thread timer trigger next tick, not thread sleep.

vietj:
this thread sleep in inherent to the HashedWheelTimer

pifuant:
implement new hashedWheelTimer with  the event loop thread timer driven.

low level ----------------------------------------------> high level
the event loop thread timer  ->  HashedWheelTimer ->   io timeout

vietj:
that's not really possible @pifuant since the event loop thread must not block

PR #2343, opened 2018-03-19
Created andThen methods for Future

slinkydeveloper:
I've created two new methods for Future, inspired to Scala one. This is useful when you need to handle both completed and failure state of Future at same time (for example inverting a future state from completed to failed)

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -468,6 +468,80 @@ public interface Future<T> extends AsyncResult<T>, Handler<AsyncResult<T>> {
     return (Future<T>) AsyncResult.super.otherwiseEmpty();
   }
 
+  /**
+   * Apply a mapper to this future with a {@code mapper} function.<p>
+   *
+   * Both when this future (the one on which {@code andThen} is called) succeeds or fail, the {@code mapper} will be called with
+   * the actual async result and this mapper returns another future object. This returned future completion will complete
+   * the future returned by this method call.<p>
+   *
+   * If the {@code mapper} throws an exception, the returned future will be failed with this exception.<p>
+   *
+   * @param mapper the mapper function
+   * @return the composed future
+   */
+  default <U> Future<U> andThen(Function<AsyncResult<T>, Future<U>> mapper) {
+    if (mapper == null) {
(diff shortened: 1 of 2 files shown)


vietj:
I don't think this code compiles with maven

slinkydeveloper:
uops sorry, error with code generation, now fixed

vietj:
isn't the later similar to the former ?

vietj:
can you point similar method in other libraries like Scala, Vavr, CompletableFuture, etc... ?

slinkydeveloper:
https://www.scala-lang.org/api/2.12.3/scala/concurrent/Future.html#andThen[U](pf:PartialFunction[scala.util.Try[T],U])(implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[T]

vietj:
how about other libraries ?

as side note : I'm a bit reluctant to have this because then when we add new such methods, adding new methods never ends :-) (there will be always a next one) and we are wanting to switch over `CompletionStage`

slinkydeveloper:
Also akka has this method: https://doc.akka.io/api/akka/2.0/#akka.dispatch.Future

I understand your thoughts, but I think this is an important method, because sometimes is useful to both manage the successful case and recover at same time without chaining like:

```
future.recover().map()
```

Just think when you want to switch the result of a future, like I've done in tests https://github.com/eclipse/vert.x/pull/2343/files#diff-8eb0378a65d21c09378e95ea4996eb15R1236 . This method is the only way to do that

jponge:
I think we should have `andThen` that delegates to `compose`, because `andThen` reads better than `compose` and clearly marks sequentiality.

Ideally we'd mimic `then` / `catch` / `finally` like JavaScript. This is simple, clear and elegant, while the more elaborated asynchronous composition cases work great with RxJava.

pmlopes:
I've solve this on JavaScript by having my own future implementation registered with the future factory. I extend the original method with a `then` method so it can be used on async await tasks.

I think specific languages should add the extra methods on their side as the factory allows this.

vietj:
done as https://github.com/eclipse-vertx/vert.x/commit/95f82b52bd68c28fe05c4cf0d2c9aaa88f358ced

PR #2347, opened 2018-03-20
HttpServerRequest provides a bytesRead() method

fzakaria:
This change is to expose `bytesRead` from `HttpServerRequest` similar to `bytesWritten` for the response.

The use case is the following:
 
Suppose I have a `loggerHandler` (middleware) where I would like to record the bytes read to a log file.
ex. `router.route().handler(loggerHandler);`

The logger handle wants to record bytesRead (not content length). If it isn't a middleware solution I _could_ wrap my request handler ReadStream with a `CountingReadStream` -- however for a more general middleware solution (see below) when `context.next()` is called, downstream handlers won't be using the wrapped CountingReadStream.

```java
@Override
public void handle(RoutingContext context) {
    CountingReadStream<Buffer> rs = new CountingReadStream<>(context.request());
    context.addBodyEndHandler(event -> {
         //the value from CountingReadStream is not correct here !
    });
    context.next();
}
```

With the exposed `bytesRead` I can log that information in the `addBodyEndHandler` in my middleware.

```java
@Override
public void handle(RoutingContext context) {
    context.addBodyEndHandler(event -> {
         //do whatever you want with context.request().bytesRead()
    });
    context.next();
}
```

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerRequestImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerRequestImpl.java
@@ -84,6 +84,7 @@ public class HttpServerRequestImpl implements HttpServerRequest {
   private MultiMap attributes;
   private HttpPostRequestDecoder decoder;
   private boolean ended;
+  private long bytesRead;
 
 
   HttpServerRequestImpl(Http1xServerConnection conn,
@@ -158,6 +159,13 @@ public class HttpServerRequestImpl implements HttpServerRequest {
   @Override
   public @Nullable String host() {
     return getHeader(HttpHeaderNames.HOST);
+  }
+
+  @Override
+  public long bytesRead() {
+    synchronized (conn) {
+      return bytesRead;
+    }
   }
 
   @Override
@@ -373,6 +381,7 @@ public class HttpServerRequestImpl implements HttpServerRequest {
 
   void handleData(Buffer data) {
     synchronized (conn) {
+      bytesRead += data.getByteBuf().readableBytes();
       if (decoder != null) {
         try {
           decoder.offer(new DefaultHttpContent(data.getByteBuf()));
--- a/src/main/java/io/vertx/core/http/impl/HttpServerRequestImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerRequestImpl.java
@@ -381,7 +381,7 @@ public class HttpServerRequestImpl implements HttpServerRequest {
 
   void handleData(Buffer data) {
     synchronized (conn) {
-      bytesRead += data.getByteBuf().readableBytes();
+      bytesRead += data.length();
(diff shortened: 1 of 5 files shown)


vietj:
it owuld be better to keep it as you are doing and make it an utility ReadStream, the bytesRead is a detail implementation and it does not even exists in HTTP/2

fzakaria:
@vietj I'm not sure what you mean to keep doing it as I'm doing. 
There is no solution available at the moment to solve it in a middleware because downstream handlers won't be using the wrapped ReadStream.

I'm happy to just even add the method to HTTP 1x implenrtation and just do a cast but I'm not sure why it's not not accurate for http2. 

I just want to expose a way to get the bytes consumed as part of the body.

vietj:
Can't this done in vertx web rather ?

fzakaria:
@vietj no. 

Our use case specifically is "billing". 
We want to charge customers for all data sent. 
We can't use content length because we want to charge for what is sent (imagine they canceled the request). 

Rather than changing all handlers I would like to solve it with a simple middleware at the root.
That's the crux of the problem as there is no way to get that. 

HttpResponse exposes `bytesWritten` btw

fzakaria:
This change is an attempt in parity with `bytesWritten` which was helpful for our billing needs.

vietj:
alright, this makes sense.

fzakaria:
@vietj great!
Please let me know anything else needed to help get this merged --

vietj:
the test should be in `HttpTest` so it's ran for HTTP1 and H2

fzakaria:
@vietj done -- I've moved the test.

src/main/java/io/vertx/core/http/impl/HttpServerRequestImpl.java
```diff
@@ -373,6 +381,7 @@ public HttpConnection connection() {
 
   void handleData(Buffer data) {
     synchronized (conn) {
+      bytesRead += data.getByteBuf().readableBytes();
```
vietj: rather use `data.length()`

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -4084,6 +4084,29 @@ public void testContainsValueCharSequenceIgnoreCase() {
...
+    server.requestHandler(req -> {
+      req.bodyHandler(buffer -> {
+        assertEquals(req.bytesRead(), length);
+        req.response().end();
+      });
+    });
+    startServer();
+    client.post(DEFAULT_HTTP_PORT, DEFAULT_HTTP_HOST, DEFAULT_TEST_URI, resp -> {
+      resp.bodyHandler(buff -> {
+        complete();
```
vietj: it should be `testComplete()`

fzakaria:
@vietj addressed feedback.

PR #2348, opened 2018-03-21
Allow access to host() field

### LINKED ISSUE #2346: Make HttpClientRequest#host() returns the actual server origin of the request instead of the host header
I was looking into metrics and wanted to group HTTP client request metrics by host. Unfortunately the original host is not easily accessible and before I PR something I wanted to point out that the whole host handling is somewhat confusing right now and start a discussion about naming. 

So `HttpClientRequestBase` has a `host` field, that will be populated in the constructor by the default host or the host from the actual request depending on the use case. Albeit it does not have a direct access method for it.

There is a protected `hostHeader()` method, that does depending on the ssl type of the request return the host or the host plus port. Otherwise the field is completely encapsulated and only used for the `absoluteURI` method. 

To make things a little confusing the `HttpClientRequestImpl` has a `getHost()` method that returns the content of the `hostHeader` field in that class. Furthermore it overrides the `hostHeader()` method to give precedence to this field, but that is still protected. 

My naive approach would have been to add a `getRawHost()` or `getBaseHost()` to `HttpClientRequestBase` but that feels fishy at least. Any suggestions?

marcust:
This closes #2346

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientRequestBase.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientRequestBase.java
@@ -89,6 +89,10 @@ abstract class HttpClientRequestBase implements HttpClientRequest {
     return uri;
   }
 
+  public String host() {
+    return host;
+  }
+
   @Override
   public io.vertx.core.http.HttpMethod method() {
     return method;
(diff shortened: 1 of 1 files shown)


vietj:
can you sign the Eclipse Contributor Agreement?

marcust:
Done

PR #2354, opened 2018-03-22
Allow to store mutable objects in local map

### LINKED ISSUE #2353: There is no proper way to store mutable objects in a local map
It is possible only for JsonObject, byte array and some other copyable objects known to vert.x.
It would be usefull to have mechanism to store mutable user objects in a local map if they can be copied.

nolequen:
fixes #2353

### DIFF
--- a/src/main/java/io/vertx/core/buffer/Buffer.java
+++ b/src/main/java/io/vertx/core/buffer/Buffer.java
@@ -19,6 +19,7 @@ import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.ServiceHelper;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
+import io.vertx.core.shareddata.Shareable;
 import io.vertx.core.shareddata.impl.ClusterSerializable;
 import io.vertx.core.spi.BufferFactory;
 
@@ -36,7 +37,7 @@ import java.nio.charset.Charset;
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
 @VertxGen
-public interface Buffer extends ClusterSerializable {
+public interface Buffer extends ClusterSerializable, Shareable {
 
   /**
    * Create a new, empty buffer.
--- a/src/main/java/io/vertx/core/shareddata/LocalMap.java
+++ b/src/main/java/io/vertx/core/shareddata/LocalMap.java
@@ -24,13 +24,12 @@ import java.util.function.Function;
 /**
  * Local maps can be used to share data safely in a single Vert.x instance.
  * <p>
- * The map only allows immutable keys and values in the map, OR certain mutable objects such as {@link io.vertx.core.buffer.Buffer}
- * instances which will be copied when they are added to the map.
+ * By default the map allows immutable keys and values.
+ * Custom keys and values should implement {@link Shareable} interface. The map returns their copies.
  * <p>
  * This ensures there is no shared access to mutable state from different threads (e.g. different event loops) in the
  * Vert.x instance, and means you don't have to protect access to that state using synchronization or locks.
  * <p>
- * Your own objects can be marked as immutable by implementing the {@link Shareable} interface.
  *
  * Since the version 3.4, this class extends the {@link Map} interface. However some methods are only accessible in Java.
  *
--- a/src/main/java/io/vertx/core/shareddata/Shareable.java
+++ b/src/main/java/io/vertx/core/shareddata/Shareable.java
@@ -12,13 +12,15 @@
 package io.vertx.core.shareddata;
 
 /**
- * A marker interface which allows you to put arbitrary objects into a {@link io.vertx.core.shareddata.LocalMap}.
+ * An interface which allows you to put arbitrary objects into a {@link io.vertx.core.shareddata.LocalMap}.
  * <p>
- * Normally local maps only allow immutable objects or other copiable objects such as {@link io.vertx.core.buffer.Buffer}
- * instances in order to avoid shared access to mutable state.
+ * Normally local maps only allow immutable or copiable objects in order to avoid shared access to mutable state.
  * <p>
  * However if you have an object that you know is thread-safe you can mark it with this interface and then you
  * will be able to add it to {@link io.vertx.core.shareddata.LocalMap} instances.
+ * <p>
+ * Mutable object that you want to store in a {@link io.vertx.core.shareddata.LocalMap}
+ * should override {@link Shareable#copy()} method.
  * <p>
  * Use this interface with caution.
  * <p>
@@ -26,4 +28,12 @@ package io.vertx.core.shareddata;
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
 public interface Shareable {
+
+  /**
+   * Returns a copy of the object.
+   * Only mutable objects should provide a custom implementation of the method.
+   */
+  default Shareable copy() {
+    return this;
+  }
 }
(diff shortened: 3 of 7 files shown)


nolequen:
Hello @vietj,
could you please review the PR and merge it if possible?

vietj:
I would rather avoid as it seems to not be a general use case enough and I think you should not store mutable state in such map

nolequen:
But if I'm getting the map using method `SharedData#getAsyncMap` I can't determine is it `local` or not (I consider `instance of` as a bad idea) and there is no simple way to avoid runtime bugs especially if the codebase is big.
The PR helps to use `local` maps in the same way as `clustered` with expectation of the same behaviour.

vietj:
if you need a local map then you need to use `getLocalMap()`

nolequen:
As you know, there might be some cases, when it is very uncomfortable to choose an appropriate map explicitly, especialy if we don't know in advance whether the Vert.x instance clustered or not. E.g. when writing tests which may use cluster in some cases or not, choosing the "right" map leads to changes in production code. This was the main reason to introduce `getAsyncMap` method.
The goal of the PR is to make the behaviour of local maps more predictable and similar to clustered ones.

vietj:
I think we could reasonably support it by adding the `copy()` method in `Shareable` and providing a default implementation that returns `this`.

nolequen:
@vietj done.

vietj:
hi, this is not yet a proper contribution : it needs to be properly tested and the BufferImpl could be retrofited to implements Shareable and save the instanceof check

nolequen:
Ok. But how about `JsonObject` and `JsonArray`? It seems they could implement `Shareable` too.

src/main/java/io/vertx/core/shareddata/Shareable.java
```diff
@@ -26,4 +26,8 @@
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
 public interface Shareable {
+
+  default Shareable copy() {
```
vietj: missing javadoc

vietj:
@tsegismont what's your take on this ?

nolequen:
Hi @vietj, @tsegismont,
Is there any progress with PR? Should I do something else?

vietj:
I would like to have the review and thoughts of @tsegismont , he's busy, the work review is just queued in his todo list I think

tsegismont:
Sorry, I just missed the notif from last week... Will review today

nolequen:
> how is this related to #2353 ? Shared#getAsyncMap returns an AsyncMap, not a LocalMap.
> So why not just getAsyncMap?
Could you please consider both cases (if the Vert.x is clustered and not) for the pattern:
`getAsyncMap() -> get(key) -> value.setSomeField()`.
Value will not be changed in the map if the Vert.x is **clustered**.
On the other hand updated value will be remain in the map if Vert.x is **not clustered**.

nolequen:
Anyway, is there anything else I could do for the PR to be merged?

nolequen:
> Can you explain how these changes will help with the problem discussed above?
As is mentioned before the PR fixes #2353:

> It is possible only for JsonObject, byte array and some other copyable objects known to vert.x.
> It would be usefull to have mechanism to store mutable user objects in a local map if they can be copied.

nolequen:
> Yes, this happens with any caching library when going from local to clustered mode. In clustered mode, if you want changes to be propagated, you need to invoke put again.

The problem is that there are some cases where it is impossible to determine in proper way is it clustered mode or not. The problem was described in #2244.

src/main/java/io/vertx/core/buffer/Buffer.java
```diff
@@ -36,7 +37,7 @@
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
 @VertxGen
-public interface Buffer extends ClusterSerializable {
+public interface Buffer extends ClusterSerializable, Shareable {
```
tsegismont: `Shareable` could also be applied to `JsonObject` and `JsonArray`. Then the code in `Checker` could be further simplified.

src/main/java/io/vertx/core/shareddata/Shareable.java
```diff
@@ -26,4 +26,11 @@
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
 public interface Shareable {
+
+  /**
+   * Returns a copy of the object.
```
tsegismont: Can you update the comment to explain that only mutable objects need to provide an implementation of this method?

src/test/java/io/vertx/test/core/ShareableTest.java
```diff
@@ -0,0 +1,66 @@
...
+import io.vertx.core.shareddata.LocalMap;
+import io.vertx.core.shareddata.Shareable;
+import org.junit.Test;
+
+import java.util.function.BiFunction;
+
+/**
+ * @author <a href="http://nlq.su">Anton Potsyus</a>
+ */
+public class ShareableTest extends VertxTestBase {
```
tsegismont: It would be better to add new tests in the existing `LocalSharedDataTest`.

src/main/java/io/vertx/core/buffer/Buffer.java
tsegismont: I had not noticed that comment. But I see no reason why we'd do it for `Buffer` but not `JsonObject` and `JsonArray`.

src/main/java/io/vertx/core/shareddata/LocalMap.java
```diff
@@ -30,7 +30,7 @@
  * This ensures there is no shared access to mutable state from different threads (e.g. different event loops) in the
  * Vert.x instance, and means you don't have to protect access to that state using synchronization or locks.
  * <p>
- * Your own objects can be marked as immutable by implementing the {@link Shareable} interface.
+ * Custom key and value objects should implement {@link Shareable} interface.
```
tsegismont: Could you please rather rephrase all the 3 paragraphs? The 2 previous ones will be confusing after this change.

src/main/java/io/vertx/core/shareddata/Shareable.java
```diff
@@ -12,18 +12,26 @@
...
 /**
- * A marker interface which allows you to put arbitrary objects into a {@link io.vertx.core.shareddata.LocalMap}.
+ * An interface which allows you to put arbitrary objects into a {@link io.vertx.core.shareddata.LocalMap}.
  * <p>
  * Normally local maps only allow immutable objects or other copiable objects such as {@link io.vertx.core.buffer.Buffer}
  * instances in order to avoid shared access to mutable state.
  * <p>
- * However if you have an object that you know is thread-safe you can mark it with this interface and then you
- * will be able to add it to {@link io.vertx.core.shareddata.LocalMap} instances.
+ * However if you have a mutable object that you want to store in a {@link io.vertx.core.shareddata.LocalMap}
```
tsegismont: Same here, the previous paragraph will be confusing after this change, since it's now also a `Shareable` object.

src/main/java/io/vertx/core/shareddata/impl/Checker.java
```diff
@@ -32,7 +32,6 @@ static void checkType(Object obj) {
         obj instanceof Byte ||
         obj instanceof Character ||
         obj instanceof byte[] ||
-        obj instanceof Buffer ||
         obj instanceof JsonObject ||
```
vietj: I think that if JsonObject and JsonArray implement it, we need to remove them from this list as well

src/main/java/io/vertx/core/buffer/Buffer.java
vietj: actually I did not see that copyIfRequired was copying them, I thought only buffer was copied

nolequen:
@tsegismont I have done all the requested changes. Could you check them please?
Is there something else I can do for the PR to be merged?

tsegismont:
@nolequen I was off for a week. Just checked the latest changes. Looks good to me. Thank you!

PR #2356, opened 2018-03-27
HTTP client connection keep alive timeout support

vietj:
Provides configurable keep alive timeout for the HTTP client pool, either provided by the `Keep-Alive: timeout=XYZ` header (for HTTP/1.x) or defaulted to a configurable value (for HTTP/1.x and HTTP/2).


vietj:
rebased on master

src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
```diff
@@ -95,6 +103,9 @@ public void createStream(HttpClientRequestImpl req, Handler<AsyncResult<HttpClie
     synchronized (this) {
       try {
         Http2Connection conn = handler.connection();
+
```
tsegismont: 3 blank lines

src/main/java/io/vertx/core/http/impl/HttpUtils.java
```diff
@@ -478,4 +479,47 @@ static HttpVersion toNettyHttpVersion(io.vertx.core.http.HttpVersion version) {
       return io.vertx.core.http.HttpMethod.OTHER;
     }
   }
+
+  private static final AsciiString TIMEOUT_EQ = AsciiString.of("timeout=");
+
+  public static int parseKeepAliveHeaderTimeout(CharSequence value) {
```
tsegismont: Meh, I trust you. There wasn't any code in Netty which we could have reused?

src/main/java/io/vertx/core/http/impl/pool/ConnectionListener.java
```diff
@@ -21,9 +21,8 @@
 public interface ConnectionListener<C> {
 
   /**
-   * Callback to signal the connection succeeded and provide all the info requires to manage the connection
-   *
-   * @param conn the connection
+   * Signal the connection succeeded, provide all the info requires to manage the connection
```
tsegismont: _requires_ to _required_

src/main/java/io/vertx/core/http/impl/pool/ConnectionListener.java
```diff
@@ -36,30 +35,29 @@ void onConnectSuccess(C conn,
...
+   * Signal the connection failed.
+   *
    * @param context the context
    * @param err the error
    */
   void onConnectFailure(ContextInternal context, Throwable err);
 
   /**
-   * Signals the connrection changed to the {@code concurrency} value.
+   * Signals the connection concurrent changed to the {@code concurrency} value.
```
tsegismont: _connection concurrent_?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -195,6 +200,27 @@ private boolean acquireConnection(Waiter<C> waiter) {
...
+  /**
+   * Close all unused connections with a {@code timestamp} greater than expiration timestamp.
+   *
+   * @param timestamp the timestamp value
+   * @return the number of closed connections when calling this method
+   */
+  public synchronized int closeIdle(long timestamp) {
+    int removed = 0;
+    if (waitersQueue.isEmpty() && available.size() > 0) {
+      List<Holder<C>> copy = new ArrayList<>(available);
```
tsegismont: Why do you need to work on a copy here?

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
```diff
@@ -53,6 +51,18 @@
     this.version = version;
   }
 
+  synchronized void start(boolean checkExpired) {
+    this.timerID = checkExpired ? client.getVertx().setTimer(1, id -> checkExpired()) : -1;
+  }
+
+  private synchronized void checkExpired() {
+    if (timerID >= 0) {
```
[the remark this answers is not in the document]
vietj: when `close()` is called it sets the value to `-1`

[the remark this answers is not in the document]
vietj: normally this should not be necessary though as the method are exclusive via the lock

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -61,6 +61,7 @@
 
   private final ConnectionListener<HttpClientConnection> listener;
   private final HttpClientImpl client;
+  private final HttpClientOptions options;
```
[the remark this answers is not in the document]
vietj: it is just a reference to the options, we'll remove it if somebody complains :-)

src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
vietj: I'll address that after the other review because it will create unnecessary conflicts when rebasing http_improvements pr

src/main/java/io/vertx/core/http/impl/HttpUtils.java
vietj: no

src/main/java/io/vertx/core/http/impl/pool/Pool.java
vietj: because `closeConnection(conn.connection)` will modify the `available` , I could have used an `Iterator` to remove the elements but it means to update other state (like the weight) and therefore I prefer to copy this code (that is clearly not an hotspot) and avoid duplicate code that modifies the pool state.

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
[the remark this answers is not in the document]
vietj: it should not be necessary

vietj:
feedback addressed

PR #2358, opened 2018-03-28
Set SO_REUSEPORT to ServerBootstrap instead of child options

guoyu511:
Signed-off-by: guoyu511 <guoyu.511@gmail.com>


vietj:
good catch, is there any pragmatic way to test this ?

guoyu511:
I have made a project to reproduce this issue:
[vertx-native-transport-test](https://github.com/guoyu511/vertx-native-transport-test)
and it can be fixed by upgrade vertx-core with this commit.

We needs at least two processes to check whether we can reuse the port across processes.
So I think it is hardly to make it testable in junit way.

vietj:
yes, I was actually looking for a way to do it. your project could be reused for integration tests

PR #2360, opened 2018-03-29
Vertx context improvements

### LINKED ISSUE #2359: Context improvements
A few improvements and simplifications can be done in Vert.x `Context`

- remove the internal `ContextTask` functional interface (which is a `Runnable` that can throw) as we can use `Handler<Void>` in practice which unified the internal `Context#wrapTask` method
- remove the internal `Action` functional interface (which is similar to a `Callable`) as we can use the `Handler<Future<T>>` idiom instead 
- add a `ContextInternal#executeFromIO(T,Handler<T>)` so an handler task can be passed an argument
- decompose the implementation of `ContextImpl#wrapTask` so the logic can be reused by `executeFromIO` without creating a unnecessary lambda
- handle the worker task metric in `WorkerContext` instead of `ContextImpl#wrapTask`
- avoid unnecessary lambda captures in `VertxHandler` when handling a message from Netty

vietj:
Fixes #2359


tsegismont:
@vietj looks good to me. Have you prepared PRs for cluster managers?

vietj:
no, I'll do it soon

vietj:
I also need to check the effect on benchmarks using JMH before merging btw

vietj:
I believe it will improve performance because it creates less lambdas, but a check is important to do

vietj:
will push the cluster manager changes as well soon

Issue #2365, opened 2018-04-04
HTTP server connection keeps processing pending pipelined requests when connection closed


vietj:
Currently the HTTP server keeps processing pipelined HTTP messages when the connection is closed during the processing of a request and there are pending HTTP request in the server pipeline (i.e the response has not been sent). This is due to a bug that when a message has not been processed during a check, the connection redo the check and it must not do it.

Changes: we only trigger an message processing check from the message check itself when a message has been effectively processed.

PR #2366, opened 2018-04-04
HTTP server connection keeps processing pending pipelined requests

vietj:
Fix for #2365

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
@@ -81,7 +81,6 @@ public class Http1xServerConnection extends Http1xConnectionBase implements Http
 
   private static final int CHANNEL_PAUSE_QUEUE_SIZE = 5;
 
-  private final Deque<Object> pending = new ArrayDeque<>(8);
   private final String serverOrigin;
   private final SSLHelper sslHelper;
   final HttpServerOptions options;
@@ -94,12 +93,15 @@ public class Http1xServerConnection extends Http1xConnectionBase implements Http
   private HttpServerRequestImpl currentRequest;
   private HttpServerResponseImpl pendingResponse;
   private ServerWebSocketImpl ws;
-  private boolean channelPaused;
-  private boolean paused;
-  private boolean sentCheck;
   private long bytesRead;
   private long bytesWritten;
+
+  // (pending == true) <=> (paused && pending.size() > 0)
+  private final Deque<Object> pending = new ArrayDeque<>(8);
+  private boolean paused;
+  private boolean sentCheck;
   private boolean queueing;
+  private boolean channelPaused;
 
   public Http1xServerConnection(VertxInternal vertx,
                                 SSLHelper sslHelper,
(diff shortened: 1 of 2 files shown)


vietj:
@tsegismont the server connections receives pipelined requests messages (HttpRequest, HttpLastContent, HttpRequest, etc...)

The first requests does not end the first response the field `pendingResponse` remains assigned. and it closes the connection.

When the second requests arrives (it arrives in the same channel read operation since it's pipelined and aggregated at the TCP level) then `processMessage(Object)` returns false and the request object is enqueued. Asynchronous the first processing  message triggered a call to `checkNextTick()`, when the asynchronous `checkNextTick()` runs it tries to process the message and fails since `pendingResponse ` is still assigned, the message is put back on the queue and since the queue is not empty `checkNextTick()` is called again and so on.

vietj:
now I'm wondering about the current code:

```
  synchronized void handleMessage(Object msg) {
    if (queueing) {
      enqueue(msg);
    } else {
      if (processMessage(msg)) {
        checkNextTick();
      } else {
        enqueue(msg);
      }
    }
  }
```

that seems correct, but I think that it's not necessary to call `checkNextTick()` if the request process the message, i.e:

```
  synchronized void handleMessage(Object msg) {
    if (queueing) {
      enqueue(msg);
    } else {
      if (processMessage(msg)) {
        // Since we processed the message it means that the queue is empty so we don't need to check the next tick
        checkNextTick();
      } else {
        enqueue(msg);
      }
    }
  }
```

vietj:
so we could have instead

```
  synchronized void handleMessage(Object msg) {
    if (queueing || !processMessage(msg)) {
      enqueue(msg);
    }
  }
```

vietj:
@tsegismont I've updated the PR to not call `checkNextTick()` when a message is processed in `handleMessage(Object)` and improved the related tests

vietj:
the microbenchmark for HTTP server shows significant improvements likely related to the removal of the unecessary `checkNextTick` invocation in `handleMessage`:

### this branch

```
HttpServerHandlerBenchmark.netty     thrpt   10  840.095 ± 132.653  ops/ms
HttpServerHandlerBenchmark.vertx     thrpt   10  649.268 ±  10.494  ops/ms
HttpServerHandlerBenchmark.vertxOpt  thrpt   10  711.100 ±   9.458  ops/ms
```

### master branch

```
HttpServerHandlerBenchmark.netty     thrpt   10  845.225 ± 85.887  ops/ms
HttpServerHandlerBenchmark.vertx     thrpt   10  620.237 ±  9.065  ops/ms
HttpServerHandlerBenchmark.vertxOpt  thrpt   10  690.147 ± 11.093  ops/ms
```

tsegismont:
>@tsegismont the server connections receives pipelined requests messages (HttpRequest, HttpLastContent, HttpRequest, etc...)
>
>The first requests does not end the first response so the field pendingResponse remains assigned and it closes the connection.
>
>When the second requests arrives (it arrives in the same channel read operation since it's pipelined and aggregated at the TCP level) then processMessage(Object) returns false and the request object is enqueued. The first processing message triggered a call to checkNextTick(), when the asynchronous checkNextTick() runs it tries to process the message and fails since pendingResponse is still assigned, the message is put back on the queue and since the queue is not empty checkNextTick() is called again and so on.

I understand that but I can't see why the new test would fail without the fix.

vietj:
ok, so what will fail is not the test itself but the shutdown of the corresponding Vert.x instance because the event loop will continue to process tasks and the vertx.close(Handler<AsyncResult>> will not complete in time.

I can add a note in the test about it.

tsegismont:
>I can add a note in the test about it.

:+1:

PR #2367, opened 2018-04-04
Http improvements

vietj:
A set of HTTP improvements that fix bugs on the client and server and simplifies / improves the concurrency main on the client.

This PR is rebased already on several other PRs:
- https://github.com/eclipse/vert.x/pull/2366
- https://github.com/eclipse/vert.x/pull/2356

Noticeable refactoring have been necessary to
- being conformant for the h2c upgrade protocol
- improve and simplify (make easier to reason about) the client concurrency

Each commit can be examined separately to ease reasoning about the changes and likely examining the PR as a whole can be more challenging.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
+++ b/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
@@ -12,6 +12,9 @@
 package io.vertx.core.http.impl;
 
 import io.netty.channel.Channel;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
 import io.vertx.core.http.HttpVersion;
 import io.vertx.core.http.impl.pool.Pool;
 import io.vertx.core.spi.metrics.HttpClientMetrics;
@@ -110,9 +113,7 @@ class ConnectionManager {
     }
   }
 
-  void getConnection(String peerHost, boolean ssl, int port, String host,
-                     Function<HttpClientConnection, Boolean> onSuccess,
-                     Consumer<Throwable> onFailure) {
+  void getConnection(String peerHost, boolean ssl, int port, String host, Handler<AsyncResult<HttpClientConnection>> handler) {
     EndpointKey key = new EndpointKey(ssl, port, peerHost, host);
     while (true) {
       Endpoint endpoint = endpointMap.computeIfAbsent(key, targetAddress -> {
@@ -155,15 +156,12 @@ class ConnectionManager {
             metrics.dequeueRequest(endpoint.metric, metric);
           }
 
-          boolean claimed = onSuccess.apply(conn);
-          if (!claimed) {
-            conn.recycle();
-          }
+          handler.handle(Future.succeededFuture(conn));
         } else {
           if (metrics != null) {
             metrics.dequeueRequest(endpoint.metric, metric);
           }
-          onFailure.accept(ar.cause());
+          handler.handle(Future.failedFuture(ar.cause()));
         }
       })) {
         break;
--- a/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
+++ b/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
@@ -21,9 +21,6 @@ import io.vertx.core.spi.metrics.HttpClientMetrics;
 
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.Consumer;
-import java.util.function.Function;
 
 /**
  * The connection manager associates remote hosts with pools, it also tracks all connections so they can be closed
@@ -137,18 +134,9 @@ class ConnectionManager {
       } else {
         metric = null;
       }
+
       if (endpoint.pool.getConnection(client.getVertx().getOrCreateContext(), ar -> {
         if (ar.succeeded()) {
-          /*
-        @Override
-        public void initConnection(ContextInternal ctx, HttpClientConnection conn) {
-          if (connectionHandler != null) {
-            ctx.executeFromIO(v -> {
-              connectionHandler.handle(conn);
-            });
-          }
-        }
-           */
 
           HttpClientConnection conn = ar.result();
 
--- a/src/main/java/io/vertx/core/http/impl/Http2UpgradedClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2UpgradedClientConnection.java
@@ -1,0 +1,389 @@
+/*
+ * Copyright (c) 2011-2018 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.http.impl;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.channel.ChannelPipeline;
+import io.netty.handler.codec.http.*;
+import io.vertx.codegen.annotations.Nullable;
+import io.vertx.core.*;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.*;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.http.HttpVersion;
+import io.vertx.core.http.impl.pool.ConnectionListener;
+import io.vertx.core.impl.ContextInternal;
+import io.vertx.core.net.NetSocket;
+import io.vertx.core.net.SocketAddress;
+
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLSession;
+import javax.security.cert.X509Certificate;
+
+/**
+ * An HTTP/2 connection in clear text that upgraded from an HTTP/1 upgrade.
+ */
+public class Http2UpgradedClientConnection implements HttpClientConnection {
+
+  private HttpClientImpl client;
+  private HttpClientConnection current;
+
+  private Handler<Void> closeHandler;
+  private Handler<Void> shutdownHandler;
+  private Handler<GoAway> goAwayHandler;
+  private Handler<Throwable> exceptionHandler;
+  private Handler<Buffer> pingHandler;
+  private Handler<Http2Settings> remoteSettingsHandler;
+
+  Http2UpgradedClientConnection(HttpClientImpl client, Http1xClientConnection connection) {
+    this.client = client;
+    this.current = connection;
+  }
+
+  @Override
+  public Channel channel() {
+    return current.channel();
+  }
+
+  @Override
+  public void close() {
+    current.close();
+  }
+
+  /**
+   * The first stream that will send the request using HTTP/1, upgrades the connection when the protocol
+   * switches and receives the response with HTTP/2 frames.
+   */
+  private class UpgradingStream implements HttpClientStream {
+
+    private HttpClientRequestImpl request;
+    private Http1xClientConnection conn;
+    private HttpClientStream stream;
+
+    UpgradingStream(HttpClientStream stream, Http1xClientConnection conn) {
+      this.conn = conn;
+      this.stream = stream;
+    }
+
+    @Override
+    public HttpClientConnection connection() {
+      return current == null ? conn : current;
+    }
+
+    /**
+     * HTTP/2 clear text upgrade here.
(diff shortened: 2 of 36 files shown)


vietj:
rebased on master

vietj:
checked that h2c upgrade works correctly:

```
OKMBP-de-julien:vert.x julien$ nghttp -u -v http://localhost:8000
[  0.002] Connected
[  0.002] HTTP Upgrade request
GET / HTTP/1.1
host: localhost:8000
connection: Upgrade, HTTP2-Settings
upgrade: h2c
http2-settings: AAMAAABkAAQAAP__
accept: */*
user-agent: nghttp2/1.31.0


[  0.003] HTTP Upgrade response
HTTP/1.1 101 Switching Protocols
connection: upgrade
upgrade: h2c
content-length: 0


[  0.003] HTTP Upgrade success
[  0.003] send SETTINGS frame <length=12, flags=0x00, stream_id=0>
          (niv=2)
          [SETTINGS_MAX_CONCURRENT_STREAMS(0x03):100]
          [SETTINGS_INITIAL_WINDOW_SIZE(0x04):65535]
[  0.003] send PRIORITY frame <length=5, flags=0x00, stream_id=3>
          (dep_stream_id=0, weight=201, exclusive=0)
[  0.003] send PRIORITY frame <length=5, flags=0x00, stream_id=5>
          (dep_stream_id=0, weight=101, exclusive=0)
[  0.003] send PRIORITY frame <length=5, flags=0x00, stream_id=7>
          (dep_stream_id=0, weight=1, exclusive=0)
[  0.003] send PRIORITY frame <length=5, flags=0x00, stream_id=9>
          (dep_stream_id=7, weight=1, exclusive=0)
[  0.003] send PRIORITY frame <length=5, flags=0x00, stream_id=11>
          (dep_stream_id=3, weight=1, exclusive=0)
[  0.003] send PRIORITY frame <length=5, flags=0x00, stream_id=1>
          (dep_stream_id=11, weight=16, exclusive=0)
[  0.004] recv SETTINGS frame <length=12, flags=0x00, stream_id=0>
          (niv=2)
          [SETTINGS_MAX_CONCURRENT_STREAMS(0x03):100]
          [SETTINGS_MAX_HEADER_LIST_SIZE(0x06):8192]
[  0.004] recv (stream_id=1) :status: 200
[  0.004] recv (stream_id=1) content-length: 2
[  0.004] recv HEADERS frame <length=9, flags=0x24, stream_id=1>
          ; END_HEADERS | PRIORITY
          (padlen=0, dep_stream_id=0, weight=16, exclusive=0)
          ; First response header
OK[  0.004] recv DATA frame <length=2, flags=0x01, stream_id=1>
          ; END_STREAM
[  0.004] send GOAWAY frame <length=8, flags=0x00, stream_id=0>
          (last_stream_id=0, error_code=NO_ERROR(0x00), opaque_data(0)=[])
```

and

```
MBP-de-julien:vert.x julien$ curl --http2 http://localhost:8000
OK
```

with 

```java
public class Main {
    Vertx vertx = Vertx.vertx();
    vertx.createHttpServer()
      .requestHandler(request -> request.response().end("OK"))
      .listen(8000);
  }
}
```

vietj:
rebased on master

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -1036,6 +1098,8 @@ public int hashCode() {
     result = 31 * result + (sendUnmaskedFrames ? 1 : 0);
```
earezki: Objects.hashCode(...) ?

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
```diff
@@ -53,6 +51,18 @@
     this.version = version;
   }
 
+  synchronized void start(boolean checkExpired) {
+    this.timerID = checkExpired ? client.getVertx().setTimer(1, id -> checkExpired()) : -1;
+  }
+
+  private synchronized void checkExpired() {
+    if (timerID >= 0) {
```
earezki: Would you please elaborate on the use of this checkExpired()? what exactly is expired, and why Is it synchronized, is this a stop whole app synchronized. (If It's the timeID, the you may consider AtomicInteger ?). What do you think?

vietj: it checks the `HttpClientConnection` which have been in the pools for too long so these connections are closed

synchronized is actually better because we are used biased locking and always the same thread, it's fast because not contended

src/main/java/io/vertx/core/http/HttpClientOptions.java
vietj: actually I think this code should be generated like we do generate code for toJson/fromJson which would alleviate us from writing hashcode/equals

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
vietj: it is to evict unused keep-alive connections that have been idle for a long time

so actually, it may be called from any thread, but we expect it will be always called from the same thread when it is used inside a verticle

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
```diff
@@ -112,14 +106,12 @@
   private static final boolean DISABLE_WEBSOCKETS = Boolean.getBoolean(DISABLE_WEBSOCKETS_PROP_NAME);
   private static final String DISABLE_H2C_PROP_NAME = "vertx.disableH2c";
   private final boolean DISABLE_H2C = Boolean.getBoolean(DISABLE_H2C_PROP_NAME);
-  private static final String[] H2C_HANDLERS_TO_REMOVE = { "idle", "flashpolicy", "deflater", "chunkwriter" };
```
tsegismont: We will not remove those anymore now?

src/main/java/io/vertx/core/http/impl/HttpUtils.java
```diff
@@ -380,6 +380,15 @@ static Http2Settings decodeSettings(String base64Settings) {
     return null;
   }
 
+  public static String encodeSettings(io.vertx.core.http.Http2Settings settings) {
```
tsegismont: This is for test only right? Shouldn't it move to test dir then?

src/main/java/io/vertx/core/http/impl/VertxHttp2ConnectionHandler.java
```diff
@@ -397,6 +396,22 @@ public void onSettingsRead(ChannelHandlerContext ctx, Http2Settings settings) th
     }
   }
 
+  @Override
+  public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+    if (msg instanceof Http2StreamFrame) {
+      // Handle HTTP/2 clear text upgrade request
```
tsegismont: I don't get this comment. If we have an `Http2StreamFrame`, isn't the connection already upgraded?

src/test/java/io/vertx/test/core/Http1xTest.java
```diff
@@ -41,6 +41,7 @@
 import io.vertx.core.net.TrustOptions;
 import io.vertx.core.parsetools.RecordParser;
 import io.vertx.core.streams.Pump;
+import org.junit.Ignore;
```
tsegismont: unused

tsegismont:
The other commits look good to me.

src/main/java/io/vertx/core/http/impl/VertxHttp2ConnectionHandler.java
vietj: that's the frame actually sent by `Http2UpgradeHandler` which are translated from HTTP/2 `HttpObject`, e.g

```
            HttpContent content = (HttpContent) msg;
            ByteBuf buf = VertxHandler.safeBuffer(content.content(), ctx.alloc());
            boolean end = msg instanceof LastHttpContent;
            ctx.fireChannelRead(new DefaultHttp2DataFrame(buf, end, 0));
```

that is in `Http2UpgradeHandler`

vietj: normally the `VertxHttp2ConnectionHandler` expects `ByteBuf` that it will transform into such frames, during the upgrade we directly receive these frames from the upgrade handler

src/main/java/io/vertx/core/http/impl/HttpUtils.java
vietj: I think it's fine because it's in impl

src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
vietj: they are actually all removed during the upgrade now in `Http2UpgradeHandler`

```
              ChannelPipeline pipeline = ctx.pipeline();
              Iterator<Map.Entry<String, ChannelHandler>> iterator = pipeline.iterator();
              while (iterator.hasNext()) {
                Map.Entry<String, ChannelHandler> handler = iterator.next();
                if (handler.getValue() instanceof Http2ConnectionHandler) {
                  // Continue
                } else {
                  iterator.remove();
                }
              }
```

src/main/java/io/vertx/core/http/impl/Http2UpgradedClientConnection.java
```diff
@@ -0,0 +1,389 @@
...
+import io.vertx.core.net.SocketAddress;
+
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLSession;
+import javax.security.cert.X509Certificate;
+
+/**
+ * An HTTP/2 connection in clear text that upgraded from an HTTP/1 upgrade.
+ */
+public class Http2UpgradedClientConnection implements HttpClientConnection {
```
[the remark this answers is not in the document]
vietj: we need to wrap because the `HttpConnection` returned by `connection()` needs to be `this` since the real connection is the `current` field that is initially an `Http1xClientConnection` and then will be an `Http2ClientConnection`.

when the user has set a `HttpClientRequest#connectionHandler` he will receive this instance and keep a reference to it, if we pass directly the HTTP/1 connection then the reference will be invalid after upgrade.

src/main/java/io/vertx/core/http/impl/Http2UpgradedClientConnection.java
```diff
@@ -0,0 +1,389 @@
...
+          super.channelRead(ctx, msg);
+        }
+      }
+      VertxHttp2ClientUpgradeCodec upgradeCodec = new VertxHttp2ClientUpgradeCodec(client.getOptions().getInitialSettings()) {
+        @Override
+        public void upgradeTo(ChannelHandlerContext ctx, FullHttpResponse upgradeResponse) throws Exception {
+
+          // Now we need to upgrade this to an HTTP2
+          ConnectionListener<HttpClientConnection> listener = conn.listener();
+          VertxHttp2ConnectionHandler<Http2ClientConnection> handler = Http2ClientConnection.createHttp2ConnectionHandler(client, conn.endpointMetric(), listener, conn.getContext(), (conn, concurrency) -> {
```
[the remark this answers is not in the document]
vietj: yes

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -3300,6 +3298,7 @@ public void testClientLogging() throws Exception {
     }
   }
 
+  @Ignore
```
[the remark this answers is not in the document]
vietj: just lost in the commits :-) I will renable it again

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -406,23 +407,20 @@ private void testSimpleRequest(String uri, HttpMethod method, Handler<HttpClient
...
+      req = client.requestAbs(method, (ssl ? "https://" : "http://") + DEFAULT_HTTP_HOST + ":" + DEFAULT_HTTP_PORT + uri, handler);
     } else {
       req = client.request(method, DEFAULT_HTTP_PORT, DEFAULT_HTTP_HOST, uri, handler);
     }
-    testSimpleRequest(uri, method, req, absolute);
+    testSimpleRequest(uri, method, req, absolute, ssl);
   }
 
-  private void testSimpleRequest(String uri, HttpMethod method, HttpClientRequest request, boolean absolute) {
+  private void testSimpleRequest(String uri, HttpMethod method, HttpClientRequest request, boolean absolute, boolean ssl) {
```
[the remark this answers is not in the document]
vietj: right, it uses to be used, I will remove it.

absolute is not used either, so I will remove it

src/test/java/io/vertx/test/core/Http2MetricsTest.java
```diff
@@ -0,0 +1,86 @@
...
+public class Http2MetricsTest extends HttpMetricsTestBase {
+
+  public Http2MetricsTest() {
+    super(HttpVersion.HTTP_2);
+  }
+
+  @Override
+  public void setUp() throws Exception {
+    super.setUp();
+    client = vertx.createHttpClient(createBaseClientOptions());
```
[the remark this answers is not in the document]
vietj: I think the tests could be improved, I'll do it separately

PR #2383, opened 2018-04-10
The HTTP client uses a LIFO strategy for connection pooling instead of a FIFO strategy

michaelpog:
Signed-off-by: Michael Pogrebinsky <smichpog2@gmail.com>

https://github.com/vert-x3/issues/issues/350

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpClientOptions.java
+++ b/src/main/java/io/vertx/core/http/HttpClientOptions.java
@@ -13,7 +13,6 @@ package io.vertx.core.http;
 
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.buffer.Buffer;
-import io.vertx.core.http.impl.pool.Pool;
 import io.vertx.core.impl.Arguments;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.net.*;
@@ -164,7 +163,7 @@ public class HttpClientOptions extends ClientOptionsBase {
   /**
    * Default policy on how to recycle HTTP connections in the connection POOL
    */
-  public static final Pool.ConnectionRecyclePolicy DEFAULT_CONNECTION_RECYCLE_POLICY = Pool.ConnectionRecyclePolicy.FIFO;
+  public static final RecyclePolicy DEFAULT_POOL_RECYCLE_POLICY = RecyclePolicy.FIFO;
 
   private boolean verifyHost = true;
   private int maxPoolSize;
@@ -194,7 +193,7 @@ public class HttpClientOptions extends ClientOptionsBase {
   private int maxRedirects;
   private boolean forceSni;
   private int decoderInitialBufferSize;
-  private Pool.ConnectionRecyclePolicy connectionRecyclePolicy;
+  private RecyclePolicy poolRecyclePolicy;
 
   /**
    * Default constructor
@@ -238,7 +237,7 @@ public class HttpClientOptions extends ClientOptionsBase {
     this.maxRedirects = other.maxRedirects;
     this.forceSni = other.forceSni;
     this.decoderInitialBufferSize = other.getDecoderInitialBufferSize();
-    this.connectionRecyclePolicy = other.getConnectionRecyclePolicy();
+    this.poolRecyclePolicy = other.getPoolRecyclePolicy();
   }
 
   /**
@@ -291,7 +290,7 @@ public class HttpClientOptions extends ClientOptionsBase {
     maxRedirects = DEFAULT_MAX_REDIRECTS;
     forceSni = DEFAULT_FORCE_SNI;
     decoderInitialBufferSize = DEFAULT_DECODER_INITIAL_BUFFER_SIZE;
-    connectionRecyclePolicy = DEFAULT_CONNECTION_RECYCLE_POLICY;
+    poolRecyclePolicy = DEFAULT_POOL_RECYCLE_POLICY;
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -56,7 +56,7 @@
  * When a waiter asks for a connection, it is either added to the queue (when it's not empty) or attempted to be
  * served (from the pool or by creating a new connection) or failed. The {@link #waitersCount} is the number
  * of total waiters (the waiters in {@link #waitersQueue} but also the inflight) so we know if we can close the pool
- * or not. The {@link #waitersCount} is incremented when a waiter wants to acquire a connection succesfully (i.e
+ * or not. The {@link #waitersCount} is incremented when a waiter wants to acquire a connection successfully (i.e
```
vietj: make another issue for this

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -176,6 +182,7 @@
   private int maxRedirects;
   private boolean forceSni;
   private int decoderInitialBufferSize;
+  private Pool.ConnectionRecyclePolicy connectionRecyclePolicy;
```
vietj: named it instead `poolRecyclePolicy`, a pool recycles connections

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -81,6 +81,16 @@
     long weight;         // The weight that participates in the pool weight
 
   }
+
+  /**
+   * Defines the policy in which connections will be reused in the pool
+   *
+   */
+  public enum ConnectionRecyclePolicy {
```
vietj: move it to the `io.vertx.core.http` package instead and rename it just `RecyclePolicy`

vietj:
we need to have tests for this.

the test could be :

1. the client performs two simultaneous requests and the server completes one after the other
2. the client perform another request
3. in the third request the server checks wether it the first or the second connections is used for this request

michaelpog:
@vietj  I think I addressed all your comments and added unit tests

vietj:
I'm wondering if we should not avoid to use the RecyclePolicy enum since there are two values, a boolean might suffice

vietj:
have you by chance been able to test it with your app ?

michaelpog:
else you'd like me to address?

michaelpog:
Also @vietj when do you think this will be released? 
We really need this.
In fact I am working on forking Vert.x right now for our internal uses just for this purpose.
We had to roll back production code multiple times this week because of the way connections are managed.
If the size of the pool is too high, then some of our partners end up with too many open connections and they just cannot handle that ( and they do not need those connections, it's just the way the are recycled which is the problem). My recent solution which limited this pool size, caused problems with other partners that had a lot more traffic, we ended up having not enough connections. Waiter objects queued up in in the ArrayDeque. We started getting a lot of timeouts because of that as well as those HttpClientRequestImpl objects started getting promoted by the GC to older generation and created high pause time in the GC

vietj:
@michaelpog as said on vertx-dev list there will be a patch release with this patch within a few weeks - this will be back-ported to 3.5.2 created from the 3.5.1 tag and should contain this patch along with other HTTP patches for bugs that have been found in 3.5.1 HTTP server

vietj:
in the meantime if you can get this patch tested in your env, it would be great so we can measure it.

have you tested the recent keep alive timeout value that is now in master branch ?

michaelpog:
@vietj thank you. I will try to test this. I cannot reproduce the same environment in production but I will create a small reproducer to make sure it does what is expected.

michaelpog:
@vietj  I was able to validate the pull request
https://github.com/michaelpog/vertx-recycle-policy-test

I could not figure out how to build Vert.x core with all its dependencies so I did a bit of a hack, you can see my description in the README

The results 
[results.zip](https://github.com/eclipse/vert.x/files/1908810/results.zip)

You can see that with LIFO:
If I follow the 0 stream, we can se that the connection is created with the first request at time 0.
And then stops being used after 860 ms
And because it is idle it is closed after 60 seconds by the HTTP client as configured
screen shot 2018-04-13 at 1 22 07 pm
Stream 1 stays open and is used for all the connections
screen shot 2018-04-13 at 1 25 02 pm

So because I send a request every second, one connection is enough and the two connections that I forced to be opened become one


With FIFO:
Stream 0: is used all the time every other request ( every 2 seconds)
screen shot 2018-04-13 at 1 26 34 pm

And Stream 1 is used all the time every other 2 seconds:
screen shot 2018-04-13 at 1 27 07 pm

I think the results are pretty conclusive

vietj:
since LIFO seems to behave better than FIFO, I'm wondering if that should not be the actual behaviour without having this configurable.

vietj:
similar concern was raised here https://issues.apache.org/jira/browse/HTTPCLIENT-1108

michaelpog:
We can definitely change the default.
I wanted to keep the default as it is, so that if somebody upgrades to a new version, they do not suddenly get a different behavior from what they had before.
If you're one of those who benefit more from FIFO, and you upgrade your Vert.x version, and suddenly things look worst for you, it would be painful to dig through all the patches and find out why that happened.

I guess every application/env has their constrains, in our case we have constraints on the number of connections, someone else can be happy to create as many as possible to make sure they don't get a timeout because of those bursts that were described in AsyncHttpClient/async-http-client#1184 .

So
1. I think it's good to add both options, that's for sure.
2. Maybe first release we keep the default as FIFO, and after a few releases we can change it? That will give users some time to upgrade, and notice the new option. This may minimize surprises. 
That's my suggestion.

vietj:
@manish-panwar no there is always a default, I don't think most users won't notice if we change to LIFO

vietj:
interesting comment found here:

https://github.com/linkedin/rest.li/blob/master/r2-core/src/main/java/com/linkedin/r2/transport/http/client/AsyncPoolImpl.java#L146

```
MRU is sensible for communicating with a single host in order
   * to minimize the number of idle pool objects.
   *
   * LRU, in combination with a minimum pool size, is sensible for
   * communicating with a hardware load balancer that directly maps
   * persistent connections to hosts. In this case, the AsyncPoolImpl
   * balances requests evenly across the pool.
```

michaelpog:
@vietj so you're going to create a pull request to make it un-configurable?
If you still want it configurable let me know as for that I think we'd need a pull request for Vertx. Web to add it to the WebClientOptions.

PR #2387, opened 2018-04-13
HTTP client requests should always prefix request URI with / for absolute queries

hradecek:
Related to https://github.com/vert-x3/vertx-web/issues/908

Note, that `http://127.0.0.1:8080?test=something` is valid URL. Provided PR prefixes URL `path` with '/' in case of empty path, because based on HTTP1 spec https://tools.ietf.org/html/rfc2616#section-5.1.2 it should be so.

Netty implements this behaviour as well: https://github.com/netty/netty/issues/719

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
@@ -495,6 +495,7 @@ public class HttpClientImpl implements HttpClient, MetricsProvider {
     URL url = parseUrl(absoluteURI);
     Boolean ssl = false;
     int port = url.getPort();
+    String relativeUri = url.getPath().isEmpty() ? "/" + url.getFile() : url.getFile();
     String protocol = url.getProtocol();
     if ("ftp".equals(protocol)) {
       if (port == -1) {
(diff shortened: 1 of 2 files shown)


pmlopes:
From the RFC:

```
Request-URI    = "*" | absoluteURI | abs_path | authority
```

`*` means does not apply to a resource so it won't be handled by vertx-web
`absoluteURI` | `abs_path` these **must** start with `/`
`authority` this is authN feature that we don't support on web either 

So when you do:

```
http://127.0.0.1:8080?test=something
```
The `path` is not valid as the part: `?test=something` would probably be handled as an `authority`

vietj:
@hradecek looks like a good fix for http client - can you sign the Eclipse Contributor Agreement ?

hradecek:
@pmlopes - You might be right! I see there's much more confusion about URL, that I thought so e.g.
 [stackoverflow.com/ok-to-skip-slash-before-query-string](https://stackoverflow.com/questions/1617058/ok-to-skip-slash-before-query-string)

Most browsers even adds `/` to URL when you make a request to `http://127.0.0.1:8080?test=something`.

@vietj - Done.

vietj:
curl does it as well

vietj:
note that we don't intent to address the other cases than `abs_path ` with the absolute form

vietj:
question to @hradecek how is it related to vert-x3/vertx-web#908 ?

vietj:
ah ok this issue was discovered in this vertx-web issue.

PR #2392, opened 2018-04-16
Fix race condition in handler registration update

### LINKED ISSUE #1531: Possible concurrency issue in Event Bus
It looks like, there is possible concurrency issue in Event Bus, when we're adding and removing local registrations.

**addLocalRegistration**: https://github.com/eclipse/vert.x/blob/5338f6353a49a4e7e9a6350e4878bcc5f11028a0/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java#L232
**removeLocalRegistration**: https://github.com/eclipse/vert.x/blob/5338f6353a49a4e7e9a6350e4878bcc5f11028a0/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java#L277

There is very very little possibility (but non 0), that when we're removing last handler in one thread and adding the new one for the same address in another, we can lost handlers list for that address.
Is this real issue or not, because of it's very little possibility, or I'm missing some additional synchronization?

Topic in user group: https://groups.google.com/forum/#!topic/vertx/T3F3CuG1n3w

vietj:
Fixes #1531 

The ConcurrentMap updates has been modified to use a compute operation in which the Handlers list is copied and updated. The compute
operation ensures that there is no modification race, the Handlers list modification is side effect free and the copied list is safely published
using final HandlerHolder array in the constructor.

Since the HandlerHolder is not anymore accessible, the HandlerRegistration has been refactored so the HandlerHolder reference is kept
instead on the HandlerRegistration where it replaces the registered boolean field instead. So now unregistration is done
by providing the HandlerHolder instead of the HandlerRegistration reference.

A test reproducing the issue has been added, the test performs intensive concurrency handler registration/unregistration of the same
address by two threads


src/main/java/io/vertx/core/eventbus/impl/Handlers.java
```diff
@@ -11,21 +11,57 @@
...
 /**
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
-public class Handlers {
+public class Handlers implements Iterable<HandlerHolder> {
 
-  private final AtomicInteger pos = new AtomicInteger(0);
-  public final List<HandlerHolder> list = new CopyOnWriteArrayList<>();
+  private final AtomicInteger pos;
+  private final HandlerHolder[] elements;
```
tsegismont: Why not simply an ArrayList? It seems to me code below would be simplified without performance or memory penalty. Also you get the iterator for free without wrapping the array in a list any time you publish a message.

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -227,11 +227,12 @@ protected MessageImpl createMessage(boolean send, String address, MultiMap heade
     return msg;
   }
 
-  protected <T> void addRegistration(String address, HandlerRegistration<T> registration,
+  protected <T> HandlerHolder<T> addRegistration(String address, HandlerRegistration<T> registration,
                                      boolean replyHandler, boolean localOnly) {
     Objects.requireNonNull(registration.getHandler(), "handler");
-    boolean newAddress = addLocalRegistration(address, registration, replyHandler, localOnly);
-    addRegistration(newAddress, address, replyHandler, localOnly, registration::setResult);
+    LocalRegistrationResult result = addLocalRegistration(address, registration, replyHandler, localOnly);
```
tsegismont: Should be `LocalRegistrationResult<T> result = `

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -240,8 +241,17 @@ protected MessageImpl createMessage(boolean send, String address, MultiMap heade
...
+  private static class LocalRegistrationResult<T> {
+    final HandlerHolder<T> holder;
+    final boolean newAddress;
+    public LocalRegistrationResult(HandlerHolder<T> holder, boolean newAddress) {
+      this.holder = holder;
+      this.newAddress = newAddress;
+    }
+  }
+
+  private <T> LocalRegistrationResult addLocalRegistration(String address, HandlerRegistration<T> registration,
```
tsegismont: Should be `private <T> LocalRegistrationResult<T>`

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -227,11 +227,12 @@ protected MessageImpl createMessage(boolean send, String address, MultiMap heade
...
   }
 
-  protected <T> void addRegistration(String address, HandlerRegistration<T> registration,
+  protected <T> HandlerHolder<T> addRegistration(String address, HandlerRegistration<T> registration,
                                      boolean replyHandler, boolean localOnly) {
     Objects.requireNonNull(registration.getHandler(), "handler");
-    boolean newAddress = addLocalRegistration(address, registration, replyHandler, localOnly);
-    addRegistration(newAddress, address, replyHandler, localOnly, registration::setResult);
+    LocalRegistrationResult result = addLocalRegistration(address, registration, replyHandler, localOnly);
+    addRegistration(result.newAddress, address, replyHandler, localOnly, registration::setResult);
```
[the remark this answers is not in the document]
vietj: so `onRegister` is fine ?

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
```diff
@@ -34,12 +70,21 @@ public HandlerHolder choose() {
         pos.set(0);
       }
       try {
-        return list.get(p);
-      } catch (IndexOutOfBoundsException e) {
+        return elements[p];
+      } catch (ArrayIndexOutOfBoundsException e) {
```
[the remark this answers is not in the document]
vietj: we need to reason about it, I would say it might because of a concurrent update with the Atomic position but I haven't checked

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
vietj: ArrayList has mutable array field, using final we are sure the object array is safely published

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: yes

vietj: but I think the main issue is that Handlers does not have a generic type for holders

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -252,63 +262,42 @@ protected MessageImpl createMessage(boolean send, String address, MultiMap heade
...
-      handlers = new Handlers();
-      Handlers prevHandlers = handlerMap.putIfAbsent(address, handlers);
-      if (prevHandlers != null) {
-        handlers = prevHandlers;
-      }
-      newAddress = true;
-    }
-    handlers.list.add(holder);
+    Handlers handlers = new Handlers(holder);
+    Handlers actual = handlerMap.merge(address, handlers, (prev, current) -> prev.add(current.first()));
```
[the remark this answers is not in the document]
vietj: yes

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: I will try to use generic if possible

vietj: we might lose <T> because of Map<String, Handlers<?>>

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
vietj: will try

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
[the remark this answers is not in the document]
vietj: I think now it's still needed because two threads might race on 

```
      int p = pos.getAndIncrement();
      if (p >= size - 1) {
        pos.set(0);
      }
```

and have `pos >= size`

to avoid it we would need to do a CAS certainly

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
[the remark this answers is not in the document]
tsegismont: :+1:

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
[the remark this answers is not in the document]
tsegismont: :+1: for CAS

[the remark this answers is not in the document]
vietj: CAS is not free :-)

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
tsegismont: But it is no longer published. That was the purpose of making `Handlers` implement `Iterable`, wasn't it?

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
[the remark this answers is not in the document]
tsegismont: Neither is catching a checked exception. Anyway, `getAndIncrement` is already using CAS, isn't?

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
vietj: that's for the stop sequence when the event bus is closed

tsegismont: No, it's also used if you _publish_ (see `deliverMessageLocally`)

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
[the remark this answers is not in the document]
tsegismont: OK, but isn't `getAndIncrement` already using CAS ?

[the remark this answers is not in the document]
vietj:
```
        int p = pos.getAndIncrement();
        if (p >= len) {
          p = p % len;
          if (p == 0) {
            pos.addAndGet(-len);
          }
        }
        return (T) elements[p];
```
is more accurate (correct distribution) and scale better concurrently across different sizes of holders (tested from 2 to 10)

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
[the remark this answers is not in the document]
vietj: actually I will not rename them because it will break cluster managers and I want to backport this to 3.5.2 so it will be easier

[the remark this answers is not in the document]
tsegismont: It won't break anything in cluster managers. This is only overridden in ClusteredEventbus.

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
tsegismont: I got you. Read the comments stream again 😀

src/main/java/io/vertx/core/eventbus/impl/Handlers.java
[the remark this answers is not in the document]
tsegismont: ok

vietj:
here is the PR update

vietj:
@tsegismont can you have a final look at the issue ? I'd like to merge the PR and backport to 3.5 branch

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -264,15 +265,18 @@ public LocalRegistrationResult(HandlerHolder<T> holder, boolean newAddress) {
 
     HandlerHolder<T> holder = new HandlerHolder<>(metrics, registration, replyHandler, localOnly, context);
 
-    Handlers handlers = new Handlers(holder);
-    Handlers actual = handlerMap.merge(address, handlers, (prev, current) -> prev.add(current.first()));
+    CyclicSequence<HandlerHolder> handlers = new CyclicSequence<HandlerHolder>().add(holder);
+    CyclicSequence<HandlerHolder> newHandlers = handlerMap.merge(
+      address,
+      handlers,
+      (oldHandlers, current) -> oldHandlers.add(current.first()));
```
tsegismont: `current` is still a wrong name fir this param

src/main/java/io/vertx/core/impl/utils/CyclicSequence.java
```diff
@@ -0,0 +1,137 @@
...
+ * A sequence is immutable and mutations uses {@link #add(Object)} and {@link #remove(Object)}
+ * to return a modified copy of the current instance.
+ * <p/>
+ * The iterator uses a volatile index, so it can be incremented concurrently by several
+ * threads with locking.
+ *
+ * @author <a href="http://tfox.org">Tim Fox</a>
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+public class CyclicSequence<T> implements Iterable<T>, Iterator<T> {
```
tsegismont: This is a critical piece in EventBus implementation. Why changing completely, especially if you want to backport it in 3.5.2 ?

src/main/java/io/vertx/core/impl/utils/CyclicSequence.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private final AtomicInteger pos;
+  private final Object[] elements;
```
tsegismont: I am still convinced that using an array instead of list makes the code more complex without significant gain in return.

vietj: we use arrays because we can make them final and be sure the array is published correctly with respect to the java memory model. An ArrayList does not have its internal array final (also it removes one extra indirection). That's what CopyOnWriteArrayList does as well internally.

vietj: so the gain is to have safe publication

src/main/java/io/vertx/core/impl/utils/CyclicSequence.java
vietj: I can make it internal again but then I have to remove the corresponding CyclicSequence test because it requires to use HandlerHolder to test it.

tsegismont: It is changed: before it only implemented `Iterable`, not it also implements `Iterator`.

src/main/java/io/vertx/core/impl/utils/CyclicSequence.java
tsegismont: Where is this field published? In your previous implementation, `Handlers` only provided an Iterator for the content. That is not publishing the array itself.

vietj: it's published in the JMM meaning (https://shipilev.net/blog/2014/safe-public-construction/). Previously Handlers was using a CopyOnWriteArrayList that has it's internal array to be volatile. In this case we are using a final value.

vietj: the CyclicList is copied and then put in the ConcurrentHashMap, we use final to ensure that the copied array is published in CyclicList that is then made available to the concurrent hashmap via merge/compute operations

Issue #2391, opened 2018-04-16
error time unit


okou19900722:
https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/Vertx.java#L559
https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/impl/BlockedThreadChecker.java#L46

it's must be ms of maxExecuteTime  unit

vietj:
more generally this code should use TimeUnit to convert the value and not make a division - can you provide a signed PR ?

okou19900722:
yes, i will do it

PR #2394, opened 2018-04-16
fix typo of comment and support custom time unit of max execute time

okou19900722:
#2391
now ,you can invoke method createSharedWorkerExecutor to create a named worker executor with custom time unit


src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -1053,6 +1053,11 @@ public synchronized WorkerExecutorImpl createSharedWorkerExecutor(String name, i
     return namedExec;
   }
 
+  @Override
+  public synchronized WorkerExecutorImpl createSharedWorkerExecutor(String name, int poolSize, TimeUnit maxExecuteTimeUnit, long maxExecuteTime) {
```
vietj: the pattern is rather (long, TimeUnit) than (TimeUnit, long) .

vietj:
actually I'm not sure this PR is correct.

the default unit is really ns (nano seconds).

when the warning is displayed, the value is divided by 1000000 which converts nano seconds to microseconds.

so overall we can't change this (as it would be breaking change).

okou19900722:
oh sorry, i made a mistake, I change the comment of createSharedWorkerExecutor. But the wrong was in
the comment of Examples:[CoreExamples](https://github.com/eclipse/vert.x/blob/master/src/main/java/examples/CoreExamples.java#L101)

```java
public void workerExecutor3(Vertx vertx) {
    //
    // 10 threads max
    int poolSize = 10;

    // 2 minutes
    long maxExecuteTime = 120000;

    WorkerExecutor executor = vertx.createSharedWorkerExecutor("my-worker-pool", poolSize, maxExecuteTime);
  }
```
Converting nanoseconds into 2 minutes should be 120,000,000,000 instead of 120000

Maybe it's the old version，But nanoseconds are really hard to use. so in the Examples, it can replace with
```java
WorkerExecutor executor = vertx.createSharedWorkerExecutor("my-worker-pool", poolSize, 2, TimeUnit.MINUTES);
```

okou19900722:
ok, it's done. It is correct in the class `VertxImpl` and unit tests `NamedWorkerPoolTest`

vietj:
yes they are, we should try to change the default format in Vert.x 4 I think

vietj:
thanks for the contribution, this ns is indeed misleading.

I think we should also then modify the same configuration in VertxOptions and have a `maxWorkerExecuteTimeUnit`, `maxEventLoopExecuteTimeUnit` and `warningExceptionTimeUnit` settings

okou19900722:
I encountered a problem。
If I add a time unit field for maxEventLoopExecuteTime, then the getMaxEventLoopExecuteTime method will be like this
```java
public long getMaxEventLoopExecuteTime() {
    return maxEventLoopExecuteTimeUnit.toNanos(maxEventLoopExecuteTime);
}
```
But the toJson method in the VertxOptionsConverter class calls the getMaxEventLoopExecuteTime method
https://github.com/eclipse/vert.x/blob/ed75286d4e6a206244829fdbd01775d9516f6a70/src/main/generated/io/vertx/core/VertxOptionsConverter.java#L124
So if I set the unit of maxEventLoopExecuteTime to milliseconds and set maxEventLoopExecuteTime to 1,The following unit test illustrates the problem 
```java
@Test
public void test(){
    VertxOptions options = new VertxOptions();
    options.setMaxEventLoopExecuteTimeUnit(TimeUnit.MILLISECONDS);
    options.setMaxEventLoopExecuteTime(1);

    JsonObject json = new JsonObject();
    json.put("maxEventLoopExecuteTimeUnit", options.getMaxEventLoopExecuteTimeUnit());
    json.put("maxEventLoopExecuteTime", options.getMaxEventLoopExecuteTime());
    options = new VertxOptions(json);
    assertNotEquals(1, options.getMaxEventLoopExecuteTime());
    assertNotEquals(TimeUnit.MILLISECONDS.toNanos(1), options.getMaxEventLoopExecuteTime());
    assertEquals(TimeUnit.MILLISECONDS.toNanos(TimeUnit.MILLISECONDS.toNanos(1)), options.getMaxEventLoopExecuteTime());
}
```
Although VertxOptions does not have a toJson method, and the toJson method of VertxOptionsConverter is not accessible, there is a hidden danger here.

Can you give some advice?Thank you!

okou19900722:
How compatible If other people call getMaxEventLoopExecuteTime or other methods

okou19900722:
I made another mistake. If he didn't set a time unit, the default time unit would be nanoseconds.

okou19900722:
Yes, I understand, but if the third-party library uses this method, and the user uses the new api, then this will be a bug

vietj:
why ? changing the checker period is not somebody does often

okou19900722:
ok, i'll commit later

okou19900722:
it's done

src/main/java/examples/CoreExamples.java
```diff
@@ -98,7 +98,7 @@ public void workerExecutor3(Vertx vertx) {
     int poolSize = 10;
 
     // 2 minutes
-    long maxExecuteTime = 120000;
+    long maxExecuteTime = 120000000000L;
```
vietj: I think we should change this example to call 2 minutes with `TimeUnit.SECONDS` instead

src/test/java/io/vertx/test/core/BlockedThreadCheckerTest.java
```diff
@@ -110,4 +112,82 @@ public void start() throws InterruptedException {
     await();
     blockedThreadWarning.expectMessage("vert.x-worker-thread", maxWorkerExecuteTime);
   }
+
+
+  @Test
+  public void testBlockCheckExceptionTimeLimitWithTimeUnit() throws Exception {
```
vietj: can you refactor the tests to avoid code duplication with testBlockCheckExceptionTimeLimit, same for other methods

vietj:
we also need to address the `DeploymentOptions` that have a maxWorkerExecuteTime field

okou19900722:
it's ok

vietj:
can you add in the comment "Like {@link #createSharedWorkerExecutor(String, int, long, TimeUnit)} but with the {@link VertxOptions#setMaxWorkerExecuteTimeUnit} {@code maxExecuteTimeUnit} (default is ns)." so user know it is nano seconds by default

src/main/java/io/vertx/core/DeploymentOptions.java
```diff
@@ -329,23 +329,25 @@ public DeploymentOptions setWorkerPoolSize(int workerPoolSize) {
   }
 
   /**
-   * Get the value of max worker execute time, in {@link DeploymentOptions#getMaxWorkerExecuteTimeUnit} {@code maxWorkerExecuteTimeUnit}.
+   * Get the value of max worker execute time, in {@link DeploymentOptions#setMaxWorkerExecuteTimeUnit} {@code maxWorkerExecuteTimeUnit}.
+   * The value of {@link DeploymentOptions#setMaxWorkerExecuteTimeUnit} {@code maxWorkerExecuteTimeUnit} is {@link TimeUnit#NANOSECONDS}
```
vietj: missing _default_

vietj:
I think we need the same correction on VertxOptions as well

okou19900722:
I am sorry for my sloppy

PR #2395, opened 2018-04-17
Hydrate vert.x cache for directory on first access

### LINKED ISSUE #2126: FileSystem readDir does not list all entries when resolving from Classpath
```
final FileSystem fs = vertx.fileSystem();
fs.propsBlocking("test/one,txt");
fs.readDirBlocking("test/");    // Bump: Only lists "one.txt" - even if there are more files on test/
```
Problem is: FileResolver copies-out resources from Classpath to a temporary directory in filesystem (i.e. .vertx/file-cache-[uuid]/... seen as a _cache_).
If you **first access a single file**, then only this one file is copied-out. A dir-listing hereafter then only lists this single file. Vertx seems to rely on the cache content here.

If you do list the directory contents _before_ accessing a single file, i.e:
```
final FileSystem fs = vertx.fileSystem();
fs.readDirBlocking("test/");    // Yeah - this now lists all files in test/
```
then you see the the whole subtree copied-out of the classpath.

fzakaria:
When a file is first access -- when caching is enabled, vert.x will copy
the file to a cache directory.

Subsequent listing of the directory will match the cache however the
contents of the directory will be only the recently accessed file.

To circumvent this bug -- when a file is first accessed, the
FileResolver will cache all other files in the same directory.

Fixes #2126

### DIFF
--- a/src/main/java/io/vertx/core/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/impl/FileResolver.java
@@ -22,13 +22,11 @@ import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.UncheckedIOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.net.URLDecoder;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.Enumeration;
 import java.util.UUID;
@@ -129,24 +127,24 @@ public class FileResolver {
         //Cache all elements in the parent directory if it exists
         //this is so that listing the directory after an individual file has
         //been read works.
-        if (file.getParent() != null) {
-          URL directoryContents = cl.getResource(file.getParent());
+        String parentFileName = file.getParent();
+        if (parentFileName != null) {
+          URL directoryContents = cl.getResource(parentFileName);
           if (directoryContents != null) {
-            unpackUrlResource(directoryContents, file.getParent());
+            unpackUrlResource(directoryContents, parentFileName, cl);
           }
         }
 
         URL url = cl.getResource(fileName);
         if (url != null) {
-          return unpackUrlResource(url, fileName);
+          return unpackUrlResource(url, fileName, cl);
         }
       }
     }
     return file;
   }
 
-  private File unpackUrlResource(URL url, String fileName) {
-    ClassLoader cl = getClassLoader();
+  private File unpackUrlResource(URL url, String fileName, ClassLoader cl) {
     String prot = url.getProtocol();
     switch (prot) {
       case "file":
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/impl/FileResolver.java
```diff
@@ -122,27 +124,45 @@ public File resolveFile(String fileName) {
         if (NON_UNIX_FILE_SEP) {
           fileName = fileName.replace(FILE_SEP, "/");
         }
+
+        //https://github.com/eclipse/vert.x/issues/2126
+        //Cache all elements in the parent directory if it exists
+        //this is so that listing the directory after an individual file has
+        //been read works.
```
vietj: cache `file.getParent()` in a variable

src/main/java/io/vertx/core/impl/FileResolver.java
```diff
@@ -122,27 +124,45 @@ public File resolveFile(String fileName) {
...
+        if (file.getParent() != null) {
+          URL directoryContents = cl.getResource(file.getParent());
+          if (directoryContents != null) {
+            unpackUrlResource(directoryContents, file.getParent());
```
cescoffier: pass the classloader as param

vietj:
@fzakaria @cescoffier will review your PR but it would be next week as we are busy this week with a conference

fzakaria:
Thanks @vietj  & @cescoffier 
I've addressed current feedback -- 

Feel free to review at your leisure following the conference (enjoy it!)
I've just wanted to contribute more (I've been seeing enough vert.x internals that I'm feeling comfortable) and saw your email on the dev mailing list so I took a crack at it.

vietj:
@cescoffier I would like to have this validated for backporting to 3.5.2

fzakaria:
Let me know if you want any more work on it.

cescoffier:
Merged in #2471

vietj:
thanks @cescoffier

vietj:
@MichaelMih thanks for reporting

MichaelMih:
Yes.
The main problem is that any antivirus program is going crazy when it sees
so many files created, especially with extension of .js .
It scans all of them, causing very slow times.
I my case, there are 4,800 files, only 70MB in total. When they are created
in .vertx by FileResolver it takes 15 seconds without antivirus and 90
seconds with antivirus.

vietj:
@MichaelMih I've looked at the code and I don't see any other way of doing this (otherwise we'll get the bug that this issue fixes). Perhaps there is a better way to handle this ?

One way you could solve this issue is by accessing one of the resource before your application starts (it delays the start but will behave fine at runtime), and disable the anti virus scan for the directory youur process run in.

MichaelMih:
Maybe you can add a system property, say vertx.enable_directory_listing
with default of true. In my case I will set it to false, disabling dir
listing and also disabling parent folder copy.

MichaelMih:
Another approach may be using enableCaching property. If it is false then do not run this fix, something like:
```
if (enableCaching ) { <<<<<this one is the proposed one
        String parentFileName = file.getParent();
        if (parentFileName != null) {
          URL directoryContents = cl.getResource(parentFileName);
          if (directoryContents != null) {
            unpackUrlResource(directoryContents, parentFileName, cl, true);
          }
        }
}
```

PR #2399, opened 2018-04-19
Fix bytesWritten when sending files with range

zsiegel:
This fix sets the proper contentLength internally so that `bytesWritten` which is used by other classes is accurate.

I came across this issue while inspecting output from the `io.vertx.ext.web.handler.impl.LoggerHandlerImpl` which uses `bytesWritten` to log the response size in bytes.

When sending requests that use the HTTP range header one can request chunked portions of a response. On the first request asking for `bytes=0-512` the request log correctly shows the following.

```
INFO: 127.0.0.1 - - [Thu, 19 Apr 2018 02:21:27 GMT] "GET /static/small HTTP/1.1" 206 513 "-" "okhttp/3.8.0"
```

However on requests that do not start at byte 0 the `contentLength` on the `HttpServerResponseImpl` does not properly take into account the offset which means  `bytesWritten` is not correctly set and next the logger logs the following as one requests chunks further along in the file.

```
INFO: 127.0.0.1 - - [Thu, 19 Apr 2018 02:21:29 GMT] "GET /static/small HTTP/1.1" 206 1026 "-" "okhttp/3.8.0"
INFO: 127.0.0.1 - - [Thu, 19 Apr 2018 02:21:31 GMT] "GET /static/small HTTP/1.1" 206 1539 "-" "okhttp/3.8.0"
INFO: 127.0.0.1 - - [Thu, 19 Apr 2018 02:21:32 GMT] "GET /static/small HTTP/1.1" 206 2052 "-" "okhttp/3.8.0"
```

The content length logged for all these requests should match the `Content-Length: 513` header being returned but the offset was not being properly accounted for.

Signed-off-by: zsiegel <zsiegel87@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
@@ -440,7 +440,7 @@ public class HttpServerResponseImpl implements HttpServerResponse {
         return;
       }
 
-      long contentLength = Math.min(length, file.length() - offset);
+      long contentLength = Math.min(length - offset, file.length() - offset);
       bytesWritten = contentLength;
       if (!headers.contentTypeSet()) {
         String contentType = MimeMapping.getMimeTypeForFilename(filename);
--- a/src/test/java/io/vertx/test/core/HttpTest.java
+++ b/src/test/java/io/vertx/test/core/HttpTest.java
@@ -1864,6 +1864,7 @@ public abstract class HttpTest extends HttpTestBase {
       res.response().sendFile("webroot/somefile.html", 6);
     }).listen(onSuccess(res -> {
       vertx.createHttpClient(new HttpClientOptions()).request(HttpMethod.GET, 8080, "localhost", "/", resp -> {
+        assertEquals(resp.headers().get("Content-Length"), String.valueOf(24));
         resp.bodyHandler(buff -> {
           assertTrue(buff.toString().startsWith("<body>blah</body></html>"));
           testComplete();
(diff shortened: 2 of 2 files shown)


vietj:
I'm wondering how this test can pass since it was changed, the HTTP/1 server has been fixed but not the HTTP/2 server.

can you explain this ?

vietj:
actually the test is incorrect for HTTP/2 which explains this, I will fix it myself

zsiegel:
Sorry about that! Was having trouble late last night figuring out which tests it might touch because some passed and some didn't.

tsegismont:
It did not fail on my box...

2018-04-19 14:59 GMT+02:00 Zac Siegel <notifications@github.com>:

tsegismont:
I will take care of it, I have the fix ready. The test was not started for
Http2 because instead of using the pre-configured HTTP/2 server and client,
it creates new ones... for HTTP/1.1

vietj:
actually the test is incorrect, it should be fixed first and I think the patch is incorrect too (need to check)

zsiegel:
In the default cause that is what you want though right? 

For example in the default case its asking to send the entire file starting from the `offset` going to `Integer.MAX_VALUE` - and in that case you want it to keep the second value I think.

The first case is only for when you are asking for a small chunk of the file.

I think this handling could be re-written to be easier to understand as it is but I think thats the proper behavior.

zsiegel:
Basically the idea is that the `Math.min` will always choose the `file.length()` or the `length` property when the `length` property is user defined in the `Range` header. Whenever the `Interger.MAX_VALUE` is passed in it will always choose the smaller value. Which I think is correct.

If you encounter a file with a size larger than `Integer.MAX_VALUE` then there would be a problem I think.

I can open up another PR and do some further cleanup to maybe make this simpler to read?

vietj:
please make a new PR that updates the test to run with HTTP/2 as well, i.e remove

`new HttpServerOptions().setPort(8080)` and replace `vertx.createHttpClient(new HttpClientOptions())` by just `client` in the tests (so that it runs with HTTP/2).

finally in your patch, use `Math.min(length, file.length()) - offset` instead

vietj:
also the `Http2ServerResponse` will need to be patched as well

zsiegel:
Will do. Ill send a PR this evening.

Thanks

tsegismont:
>keeps the second value file.length() - offset because length = Integer.MAX_VALUE

there are two tests, `testSendOpenRangeFileFromClasspath` and `testSendRangeFileFromClasspath`. The latter makes sure this part of the code is invoked with `length!=length = Integer.MAX_VALUE`

>so for now I'll rollback this and I would like to have a clean PR that fixes everything

There was a fix for HTTP/2 in #2401

But if you think it's cleaner to revert this one and create a new one, fine...

PR #2401, opened 2018-04-19
Fix bytesWritten when sending files with range HTTP/2

tsegismont:
Follows-up on #2399

Applied same fix as for HTTP/1.1
Test did not run for HTTP/2 because it creates new HTTP/1.1 client and server instead of reusing the preconfigured ones.


tsegismont:
Should be replaced by another PR. See comments in #2399

PR #2404, opened 2018-04-19
Fix bytesWritten when using Range headers

zsiegel:
This is an updated PR to address issues found with #2399

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServerResponse.java
+++ b/src/main/java/io/vertx/core/http/HttpServerResponse.java
@@ -366,7 +366,7 @@ public interface HttpServerResponse extends WriteStream<Buffer> {
    *
    * @param filename path to the file to serve
    * @param offset the offset to serve from
-   * @param length the length to serve to
+   * @param length the number of bytes to send
    * @param resultHandler  handler that will be called on completion
    * @return a reference to this, so the API can be used fluently
    */
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
@@ -572,7 +572,7 @@ public class Http2ServerResponseImpl implements HttpServerResponse {
         if (resultHandler != null) {
           resultCtx.runOnContext((v) -> resultHandler.handle(Future.failedFuture(new FileNotFoundException())));
         } else {
-           log.error("File not found: " + filename);
+          log.error("File not found: " + filename);
         }
         return this;
       }
--- a/src/test/java/io/vertx/core/http/HttpTest.java
+++ b/src/test/java/io/vertx/core/http/HttpTest.java
@@ -1943,8 +1943,10 @@ public abstract class HttpTest extends HttpTestBase {
       res.response().sendFile("webroot/somefile.html", 6);
     }).listen(testAddress, onSuccess(res -> {
       client.request(HttpMethod.GET, testAddress, DEFAULT_HTTP_PORT, DEFAULT_HTTP_HOST, DEFAULT_TEST_URI, onSuccess(resp -> {
+        assertEquals(String.valueOf(24), resp.headers().get("Content-Length"));
         resp.bodyHandler(buff -> {
           assertTrue(buff.toString().startsWith("<body>blah</body></html>"));
+          assertEquals(24, buff.toString().length());
           testComplete();
         });
       })).end();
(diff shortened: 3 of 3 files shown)


src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -1875,10 +1876,11 @@ public void testSendOpenRangeFileFromClasspath() {
 
   @Test
   public void testSendRangeFileFromClasspath() {
-    vertx.createHttpServer(new HttpServerOptions().setPort(8080)).requestHandler(res -> {
-      res.response().sendFile("webroot/somefile.html", 6, 6);
+    server.requestHandler(res -> {
+      res.response().sendFile("webroot/somefile.html", 6, 12);
```
vietj: why do you need to specify 12 here, the second value is a length to and end offset

zsiegel: The documentation for that method says it's the "length to serve to" which I took to mean then "end" not "length to send". 

This would also line up correctly with how Range headers work so I think this is correct. If you wanted to send the entire from that's why the shorter method uses Int.MAX right?

Am I reading the the documentation right?

zsiegel: Sorry it uses Long.Max not Int.Max but the intent is the same. It should be the end byte to send inclusively

vietj: the test does `sendFile(path, 6, 6)` which means to send 6 bytes at offset 6.

zsiegel: I understand what the prior test was trying to do but I think there is a bit of miscommunication here so let me try and work through what I am seeing.

Currently the documentation for sendFile reads as follows...
```
  /**
   * Ask the OS to stream a file as specified by {@code filename} directly
   * from disk to the outgoing connection, bypassing userspace altogether
   * (where supported by the underlying operating system.
   * This is a very efficient way to serve files.<p>
   * The actual serve is asynchronous and may not complete until some time after this method has returned.
   *
   * @param filename  path to the file to serve
   * @param offset offset to start serving from
   * @param length length to serve to
   */
  @Fluent
  HttpServerResponse sendFile(String filename, long offset, long length);
```

the `@param length` says the `length to serve to`. In my head that does not mean `how many bytes do you want to send` it means `at what byte should i stop sending bytes`. This lines up with how the http Range header works where you are asking to send in `Range: $start-$end`. 

I may be reading the documentation incorrectly here but thats just how I am interpreting it. Based on your feedback and how the old test was written it seems you are indicating to me that the `length` is the `amount of bytes to send` which is fair (and might be the intention of the API). 

However before I ended up here I was looking at the `StaticFileHandler` in vert.x-web to see how it handles the `Range` http header. [StaticHandler](https://github.com/vert-x3/vertx-web/blob/1f597d407971a18d39bc493bfc18f9436f83d2e3/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/StaticHandlerImpl.java#L391)

In this case the header is parsed as `Range: $start-$end` and passed into sendFile with `start` as `offset` and `end` as `length`. This is incompatible and does not match the way the test calls the API hence the confusion here.

The changes I have made below will be incorrect for the usage `sendFile(6,6)` since it will return 0 for the contentLength. 

```
long contentLength = Math.min(length, file.length()) - offset; // 6 - 6 = 0
```

So what I am trying to understand is what does the `length` mean in the `sendFile()` API? 

If it means `the number of bytes to send` then the `StaticFileHandler` in the other project may break because it is sending the `end byte` of the `Range` header as the length.

Thanks for working through this. However the `sendFile` method is determined to be I will make the necessary changes.

Thanks!

vietj: usually in java when you give the _length_, it means how many you want, otherwise _end_ is used, for instance in `java.lang.String`:

```
/*
 * @param  length The number of bytes to decode
 */
public String(byte bytes[], int offset, int length, String charsetName);
```

we could modify the javadoc so it's clearer and use instead a similar sentence like _the number of bytes to send_

zsiegel: ok great just wanted that clarification. i am going to update the javadoc on this PR and fix the offset calculation to reflect that behavior.

I will also do a PR for the vertx-web repo and link this over so its clear what change is being made.

zsiegel:
Ok this is all ready to go. I have updated the documentation and augmented the tests slightly to check the content-length headers are correct

vietj:
can you resubmit a PR and sign the ECA ? currently the ECA tools does not validates your email address

zsiegel:
Yes I can do that. When I re-submit a new PR will a bot or something prompt me? Would you like me to reference this PR? Are you going to close this one?

zsiegel:
I was already signed into the ECA just needed to re-validate. Should be all set!

vietj:
cherry picked in master

PR #2408, opened 2018-04-20
Fixes #2406

### LINKED ISSUE #2406: Cleanup expired HTTP client connection every second instead of milliseconds
https://github.com/eclipse/vert.x/blob/5945cd3aabbfad8ef8b79fab3fd0feca98f5ee35/src/main/java/io/vertx/core/http/impl/ConnectionManager.java#L62
starts a Timer with 1 _milli_-second delay to search for closeable/evictable connections.

Is it really required to do this so often? - Especially as we can define the keep-alive time only as whole seconds.

oli-h:
Don't search for timed out keep-alive Connections every Millisecond. This burns CPU.
Just search every 1000ms
Fixes #2406

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
+++ b/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
@@ -53,13 +53,13 @@ class ConnectionManager {
   }
 
   synchronized void start(boolean checkExpired) {
-    this.timerID = checkExpired ? client.getVertx().setTimer(1, id -> checkExpired()) : -1;
+    this.timerID = checkExpired ? client.getVertx().setTimer(1000, id -> checkExpired()) : -1;
   }
 
   private synchronized void checkExpired() {
     long timestamp = System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


vietj:
thanks, that's actually a good idea

have you checked all tests are passing ?

oli-h:
Unfortunately the tests don't work here in our company network.
A few tests fail - but it looks like they fail for other reason and not due to this PR. E.g.
- Http1xTest#requestAbsNoPort fails becaus it taks to www.google.com - we are firewalled
- FileSystemTest#testCopyNoFollowLinks fails because of insufficient access rights to a directory
- RedeployTest#testStartingApplicationInRedeployModeWithInlineConf2 fails because of an illegal character in some JSON
- HostnameResolutionTest#testNetSearchDomain fails because it can't resolve "host1"
- etc...

But of course: I'm not 100% sure. Sorry

vietj:
yes some of these tests might not work in these conditions.

I was implying for the related tests in Http1xTest and not the one that needs to resolve google.com <http://google.com/> :-)

PR #2420, opened 2018-04-25
Add a listen variant with a SocketAddress on HttpServer

tcataldo:
With this I can create http servers over a unix socket.

### DIFF
--- a/src/test/java/io/vertx/test/core/HttpTest.java
+++ b/src/test/java/io/vertx/test/core/HttpTest.java
@@ -33,7 +33,9 @@ import io.vertx.core.http.HttpServerOptions;
 import io.vertx.core.http.HttpServerResponse;
 import io.vertx.core.http.HttpVersion;
 import io.vertx.core.http.impl.HeadersAdaptor;
+import io.vertx.core.net.NetClient;
 import io.vertx.core.net.NetSocket;
+import io.vertx.core.net.SocketAddress;
 import io.vertx.test.netty.TestLoggerFactory;
 import org.junit.Assume;
 import org.junit.Ignore;
@@ -86,7 +88,7 @@ public abstract class HttpTest extends HttpTestBase {
     super.setUp();
     testDir = testFolder.newFolder();
   }
-
+    
   protected HttpServerOptions createBaseServerOptions() {
     return new HttpServerOptions().setPort(DEFAULT_HTTP_PORT).setHost(DEFAULT_HTTP_HOST);
   }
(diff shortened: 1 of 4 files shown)


tcataldo:
just signed the ECA with this email address.

vietj:
we need at least one test with an inet socket address in HttpTest for this.

tcataldo:
Test added as requested.

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -80,13 +81,22 @@
   public TemporaryFolder testFolder = new TemporaryFolder();
 
   protected File testDir;
+  protected SocketAddress testAddress;
+  private File tmp;
 
   @Override
   public void setUp() throws Exception {
     super.setUp();
+    if (USE_DOMAIN_SOCKETS) {
```
vietj: actually we don't need this because it cannot work since we only have HTTP server with domain sockets and not HTTP client, that's why only a simple test is necessary

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -129,6 +139,17 @@ public void testClientChaining() {
     await();
   }
 
+  @Test
+  public void testListenSocketAddress() {
+    server = vertx.createHttpServer().requestHandler(noOpHandler());
+    server.listen(testAddress, onSuccess(server -> {
+      testComplete();
+    }));
+
```
vietj: we should also have a client and complete the test when the server gets the request

vietj:
I think we need also an integration test that uses NetClient with domain sockets in `TransportTest`, there are already `testDomainSocketServer` and `testDomainSocketClient`.

Basically we start an Http server with a domain socket address and we connect a NetClient on this address and it performs a simple HTTP GET and expects a 200 OK response.

tcataldo:
I just looked at TransportTest to add a test there et just use SocketAddress.inetAddress() in the HttpTest but transportTest seems to only check that domain socket do not work (they seem to expect a failure)

vietj:
@tcataldo ok, let's see how we can create such test then, I confused myself

vietj:
so I think for now we should have in HttpTest a test that uses a domain socket address with the NetClient wihtout using the mechanism like in NetTest, i.e it create a domain socket address.

at the beginning of the test we use a https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html so the test only run with socket domain enabled. later we will remove this test when the HttpClient is able to use domain sockets

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -129,6 +131,40 @@ public void testClientChaining() {
...
+        sock.write("GET / HTTP/1.1\r\n");
+      }));
+    }));
+
+    await();
+  }
+
+  @Test
+  public void testListenDomainSocketAddress() throws Exception {
+    Vertx vx = Vertx.vertx(new VertxOptions().setPreferNativeTransport(true));
```
vietj: you need to close the created vertx in a finally block otherwise we leak resources in tests

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -129,6 +131,40 @@ public void testClientChaining() {
...
+    Assume.assumeTrue("Native transport must be enabled", vx.isNativeTransportEnabled());
+    NetClient netClient = vx.createNetClient();
+    HttpServer httpserver = vx.createHttpServer().requestHandler(req -> req.response().end());
+    File sockFile = TestUtils.tmpFile("vertx", ".sock");
+    SocketAddress sockAddress = SocketAddress.domainSocketAddress(sockFile.getAbsolutePath());
+    httpserver.listen(sockAddress, onSuccess(server -> {
+      netClient.connect(sockAddress, onSuccess(sock -> {
+        sock.handler(buf -> testComplete());
```
vietj: check the response starts with the usual HTTP 200 OK \r\n

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -129,6 +131,40 @@ public void testClientChaining() {
     await();
   }
 
+  @Test
+  public void testListenSocketAddress() {
+    NetClient netClient = vertx.createNetClient();
```
vietj: close NetClient in a finally block

vietj:
looks good, we need at the documentation to be updated to mention it with an example and that will be perfect.

tcataldo:
Is the doc somewhere in this repo so I can propose something ? may I squash/reword the commits ?

tcataldo:
Doc is updated & commits are squashed.

vietj:
can you rebase this on the master branch instead ? that it where it will be merged. Beware the documentation is handled slightly differently in master

tcataldo:
I close this PR as the version for master is here https://github.com/eclipse/vert.x/pull/2426

PR #2421, opened 2018-04-28
[modify]get the first char from a String using charAt() directly, rather than transfer a String into char array.

glomie:
[modify]get the first char from a String using charAt() directly, rather than transfer a String into char array.


vietj:
can you sign the ECA ?

glomie:
so.. Is it easy to pass the sign？

PR #2422, opened 2018-04-29
[modify repull ]get the first char from a String using charAt() directly, rather than transfer a String into char array.

glomie:
try again


glomie:
I have signed the ECA, but I still can't pass the eclipse valiadation.

vietj:
just redo a new PR and close this one

PR #2440, opened 2018-05-11
ClusteredEventBus racy initialization - fixes #2438 - fixes #2439

### LINKED ISSUE #2438: ClusteredEventBus racy initialization
Hi,

It seems as if there is a case in the ClusteredEventBus (3.5.1) where a NPE can appear. After trying out some different cluster manager implementations the following NPE appear from time to time on a new instance when it is being launched. This is the NPE that appears:
```
2018-05-11 16:14:32:531 +0200 [vert.x-eventloop-thread-1] ERROR ContextImpl - Unhandled exception
java.lang.NullPointerException
	at io.vertx.core.eventbus.impl.clustered.ClusteredEventBus.lambda$setClusterViewChangedHandler$9(ClusteredEventBus.java:283)
	at io.vertx.core.impl.HAManager.lambda$checkSubs$10(HAManager.java:516)
	at io.vertx.core.impl.HAManager.lambda$runOnContextAndWait$11(HAManager.java:525)
	at io.vertx.core.impl.ContextImpl.lambda$wrapTask$2(ContextImpl.java:339)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:404)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:463)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:886)
	at java.lang.Thread.run(Thread.java:748)
```

Looking at the code for the ClusteredEventBus it seems as if the member "subs" can be null when the view changed handler is called before the ClusteredEventBus has started completely. The view changed handler is set in the constructor but the "subs" member is only given a value in the method "start(...)". I briefly looked at previous versions of the ClusteredEventBus and there has previously been a null-check in the view changed handler.

Regards

### LINKED ISSUE #2439: Vertx racy initialization
Motivation: Vertx instances are not properly initialized in local or cluster mode (simple `this` passed in the constructor, race clustered event bus initialization). We need to have a proper initialization of Vertx that will not suffer racy construction and initialization.
Changes: rework the creation of Vertx that leads to a non racy initialisation. Avoid to pass `this` in constructor, break the startup sequence into the constructor and an init method that allow to properly orchestrate the bootstrap of the vertx instance.

vietj:
Assigned to you @tsegismont , I believe it will be fine for cluster manager implementations as this does not touch anything specific to them. In particular pay attention to the init sequence:

1. init vertx
2. join cluster manager
3. instantiate the ha manager that needs the cluster manager to be joined
4. start the event bus
5. init the ha manager on event bus start callback so that the cluster manager listener are registered after the event bus is initialized

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
@@ -72,7 +72,6 @@ public class ClusteredEventBus extends EventBusImpl {
   private static final String SUBS_MAP_NAME = "__vertx.subs";
 
   private final ClusterManager clusterManager;
-  private final HAManager haManager;
   private final ConcurrentMap<ServerID, ConnectionHolder> connections = new ConcurrentHashMap<>();
   private final Context sendNoContext;
 
@@ -85,14 +84,11 @@ public class ClusteredEventBus extends EventBusImpl {
 
   public ClusteredEventBus(VertxInternal vertx,
                            VertxOptions options,
-                           ClusterManager clusterManager,
-                           HAManager haManager) {
+                           ClusterManager clusterManager) {
     super(vertx);
     this.options = options.getEventBusOptions();
     this.clusterManager = clusterManager;
-    this.haManager = haManager;
     this.sendNoContext = vertx.getOrCreateContext();
-    setClusterViewChangedHandler(haManager);
   }
 
   private NetServerOptions getServerOptions() {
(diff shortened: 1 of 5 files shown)


vietj:
can you merge and check it works with cluster manager ?

tsegismont:
> can you merge and check it works with cluster manager ?

:+1:

tsegismont:
Merging, tested with HZ and ISPN cluster managers successfully.

PR #2448, opened 2018-05-14
Vertx http headers refactor

vietj:
Refactor to remove duplicate methods and use the optimal implementation that can possibly uses `AsciiString` optimised headers.


vietj:
@tsegismont no implementation changed, only shuffled the methods and rearranged code, but it's better to have another pair of eyes validating this.

vietj:
it's probably better to review only the last commits as the first one reshuffle the method order.

vietj:
why not , there are no reasons people would extend this class

PR #2453, opened 2018-05-15
Missing @Nullable in AsyncMap, Context, Vertx and WorkerExecutor

### LINKED ISSUE #2434: Missing @Nullable annotations
Building this list so that we can track any missing `@Nullable` annotation until 3.6 is released

[ ] `Vertx.executeBlocking`

[ ] `Context.executeBlocking`

[ ] `WorkerExecutor.executeBlocking`

[ ] `AsyncMap.get`
[ ] `AsyncMap.putIfAbsent`
[ ] `AsyncMap.remove`
[ ] `AsyncMap.replace`

------ _Original comment_

Currently, the `executeBlocking` result handler type is not annotated with `@Nullable`.

As a result, the generated Rxified API returns a `Single<T>`. But it's quite common to use `Future<Void>` for a blocking task which does not need to return any result.
Then if the user simply invoke `future.complete()`, the program fails at runtime.

As a workaround, the user can use a `Future<Object>` and invoke `complete()` with any non-null value.

tsegismont:
Fixes #2434

Improves Vert.x RxJava2 API (returns Maybe instead of Single)

With executeBlocking, the future provided to the blocking handler could be completed with null.

In AsyncMap, there are a few cases (get, remove, replace, putIfAbsent) when the result can be null as well.

### DIFF
--- a/src/main/java/io/vertx/core/Context.java
+++ b/src/main/java/io/vertx/core/Context.java
@@ -116,7 +116,7 @@ public interface Context {
    *                 guarantees
    * @param <T> the type of the result
    */
-  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<T>> resultHandler);
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<@Nullable T>> resultHandler);
 
   /**
    * Invoke {@link #executeBlocking(Handler, boolean, Handler)} with order = true.
@@ -124,7 +124,7 @@ public interface Context {
    * @param resultHandler  handler that will be called when the blocking code is complete
    * @param <T> the type of the result
    */
-  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<T>> resultHandler);
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<@Nullable T>> resultHandler);
 
   /**
    * If the context is associated with a Verticle deployment, this returns the deployment ID of that deployment.
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -12,11 +12,7 @@
 package io.vertx.core;
 
 import io.netty.channel.EventLoopGroup;
-import io.vertx.codegen.annotations.CacheReturn;
-import io.vertx.codegen.annotations.Fluent;
-import io.vertx.codegen.annotations.GenIgnore;
-import io.vertx.codegen.annotations.Nullable;
-import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.codegen.annotations.*;
 import io.vertx.core.datagram.DatagramSocket;
 import io.vertx.core.datagram.DatagramSocketOptions;
 import io.vertx.core.dns.DnsClient;
@@ -520,12 +516,12 @@ public interface Vertx extends Measured {
    *                 guarantees
    * @param <T> the type of the result
    */
-  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<T>> resultHandler);
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<@Nullable T>> resultHandler);
 
   /**
    * Like {@link #executeBlocking(Handler, boolean, Handler)} called with ordered = true.
    */
-  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<T>> resultHandler);
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<@Nullable T>> resultHandler);
 
   /**
    * Return the Netty EventLoopGroup used by Vert.x
--- a/src/main/java/io/vertx/core/WorkerExecutor.java
+++ b/src/main/java/io/vertx/core/WorkerExecutor.java
@@ -11,6 +11,7 @@
 
 package io.vertx.core;
 
+import io.vertx.codegen.annotations.Nullable;
 import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.metrics.Measured;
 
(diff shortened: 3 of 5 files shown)


src/main/java/io/vertx/core/Context.java
```diff
@@ -116,15 +116,15 @@ static boolean isOnVertxThread() {
...
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<@Nullable T>> resultHandler);
 
   /**
    * Invoke {@link #executeBlocking(Handler, boolean, Handler)} with order = true.
    * @param blockingCodeHandler  handler representing the blocking code to run
    * @param resultHandler  handler that will be called when the blocking code is complete
    * @param <T> the type of the result
    */
-  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<T>> resultHandler);
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<@Nullable T>> resultHandler);
```
vietj: it needs also to be `Handler<Future<@Nullable T>>`

src/main/java/io/vertx/core/Vertx.java
```diff
@@ -519,12 +519,12 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
    *                 guarantees
    * @param <T> the type of the result
    */
-  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<T>> resultHandler);
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<@Nullable T>> resultHandler);
```
vietj: it needs also to be `Handler<Future<@Nullable T>>`

src/main/java/io/vertx/core/Vertx.java
```diff
@@ -519,12 +519,12 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
...
 
   /**
    * Like {@link #executeBlocking(Handler, boolean, Handler)} called with ordered = true.
    */
-  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<T>> resultHandler);
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<@Nullable T>> resultHandler);
```
vietj: it needs also to be `Handler<Future<@Nullable T>>`

src/main/java/io/vertx/core/WorkerExecutor.java
```diff
@@ -47,12 +48,12 @@
    *                 guarantees
    * @param <T> the type of the result
    */
-  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<T>> resultHandler);
+  <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, boolean ordered, Handler<AsyncResult<@Nullable T>> resultHandler);
```
vietj: it needs also to be `Handler<Future<@Nullable T>>`

src/main/java/io/vertx/core/WorkerExecutor.java
```diff
@@ -47,12 +48,12 @@
...
 
   /**
    * Like {@link #executeBlocking(Handler, boolean, Handler)} called with ordered = true.
    */
-  default <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<T>> resultHandler) {
+  default <T> void executeBlocking(Handler<Future<T>> blockingCodeHandler, Handler<AsyncResult<@Nullable T>> resultHandler) {
```
vietj: it needs also to be `Handler<Future<@Nullable T>>`

src/main/java/io/vertx/core/shareddata/AsyncMap.java
```diff
@@ -81,15 +82,15 @@
    * @param ttl  The time to live (in ms) for the entry
    * @param completionHandler  the handler
    */
-  void putIfAbsent(K k, V v, long ttl, Handler<AsyncResult<V>> completionHandler);
+  void putIfAbsent(K k, V v, long ttl, Handler<AsyncResult<@Nullable V>> completionHandler);
```
vietj: the second param `V` should also be `@Nullable`

src/main/java/io/vertx/core/shareddata/AsyncMap.java
```diff
@@ -108,7 +109,7 @@
    * @param v  the new value
    * @param resultHandler  the result handler will be passed the previous value
    */
-  void replace(K k, V v, Handler<AsyncResult<V>> resultHandler);
+  void replace(K k, V v, Handler<AsyncResult<@Nullable V>> resultHandler);
```
vietj: the second param `V` should also be `@Nullable`

src/main/java/io/vertx/core/shareddata/AsyncMap.java
tsegismont: No, null values are forbidden in `AsyncMap`.

src/main/java/io/vertx/core/shareddata/AsyncMap.java
tsegismont: No, null values are forbidden in AsyncMap.

tsegismont:
@vietj good point, I only thought about RxJava2, but alt langs implementation like Scala could also benefit.

Nevertheless, the build fails with  a `Handler<Future<@Nullable T>>` parameter. I will create an issue in codegen. Can I go ahead and merge this one in the meantime?

src/main/java/io/vertx/core/shareddata/AsyncMap.java
vietj: ah ok makese sense then, perhaps we should say it in the javadoc as well then

vietj:
+1 for issue in codegen, however I think we shall keep this issue open until we fix it in codegen

src/main/java/io/vertx/core/shareddata/AsyncMap.java
vietj: ah ok makese sense then, perhaps we should say it in the javadoc as well then

tsegismont:
@vietj all the changes are in

tsegismont:
@vietj I just rebased and fixed this one after closing vert-x3/vertx-codegen#180

Can you please take another look (just a couple of classes and a few lines changes)? Thanks

vietj:
ideally we should declare

```
<@Nullable T> void executeBlocking(Future<T> fut, Handler<AsyncResult<T>> handler)
```

but codegen does not support this declaration (yet)

so for now it's fine the accept the PR as is.

Issue #2441, opened 2018-05-13
All DNS requests are always sent with the flag 'RecursionDesired' set to true


stephanebastian:
We should probably add a property to the class DnsClientOptions to support sending non recursive DNS requests. something like DnsClientOptions.isRecursionDesired() and DnsClientOptions.setRecursionDesired(boolean newValue)

PR #2455, opened 2018-05-16
implements #2441 - Dns client recursion desired

stephanebastian:
@vietj  Julien, could you please review? Thanks


stephanebastian:
forgot to mention in the comment that it fixes #2441 
(there is no PR showing up in the issue, just the commit I made)

Issue #2461, opened 2018-05-18
IllegalStateException(WebSocket is closed) may occur after stop writing in endHandler


jg513:
In `NetSocketImpl`: 
```
  @Override
  protected synchronized void handleClosed() {
    checkContext();
    if (endHandler != null) {
      endHandler.handle(null);
    }
    super.handleClosed();
    if (vertx.eventBus() != null) {
      registration.unregister();
    }
  }
```
In `WebSocketImplBase`:
```
  void handleClosed() {
    synchronized (conn) {
      cleanupHandlers();
      if (endHandler != null) {
        conn.getContext().runOnContext(endHandler);
      }
      if (closeHandler != null) {
        conn.getContext().runOnContext(closeHandler);
      }
    }
  }
```
I set a closed flag in the `endHandler`. I will stop write to the socket, if the socket has the closed flag. However `IllegalStateException: WebSocket is closed` will be thrown by WebSocket sometimes. I modify the code snippet in `WebSocketImplBase` according to `NetSocketImpl`, Then the `IllegalStateException` never occurs again :
```
  void handleClosed() {
    synchronized (conn) {
      cleanupHandlers();
      if (endHandler != null) {
        endHandler.handle(null);
      }
      if (closeHandler != null) {
        conn.getContext().runOnContext(closeHandler);
      }
    }
  }
```
Is it a proper fix?

vietj:
actually I think the callbacks should be done outside of the synchronized block (same for NetSocket but that's another issue).

```
void handleClosed() {
    Handler<Void> endHandler = this.endHandler;
    synchronized (conn) {
      cleanupHandlers();
      Handler<Void> closeHandler = this.closeHandler;
      if (closeHandler != null) {
        conn.getContext().runOnContext(closeHandler);
      }
    }
    if (endHandler != null) {
       endHandler.handle(null);
    }
  }
```

1. handlers are not called in the the synchronized block
2. we read the closeHandler ref and cache it because it might have changed concurrently when executed on context
3. if the end handler fails it will not impact the close handler that will still be called

jg513:
Great! I have tested it. It solves my problem, and is more robust. Thanks!

vietj:
do you mind filing a PR with a test ?

jg513:
#2462 Invoke endHandler immediately when a webSocket connection is closed.

PR #2462, opened 2018-05-18
Invoke endHandler immediately when a webSocket connection is closed.

jg513:
#2461


src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
```diff
@@ -424,15 +424,17 @@ void handleException(Throwable t) {
   }
 
   void handleClosed() {
+    Handler<Void> endHandler = this.endHandler;
```
vietj: endHandler must be read in the synchronized block

jg513:
Check failed: ip-validation — the pull request did not pass eclipse validation
So I close this pull request.

PR #2469, opened 2018-05-21
Add VertxOption to specify worker pool queue capacity

### LINKED ISSUE #2465: Add ability to set queueing strategy for WorkerPool
At the moment the _workerPool_ in _VertxImpl_ is initialized as follows:

```java
ExecutorService workerExec = Executors.newFixedThreadPool(options.getWorkerPoolSize(),
        new VertxThreadFactory("vert.x-worker-thread-", checker, true, options.getMaxWorkerExecuteTime()));
```

This creates an underlying _ThreadPoolExecutor_ with an unbounded _LinkedBlockingQueue_
When our HTTP server is under immense load the queue grows which causes our latencies to grow unbounded.
We would prefer to fail fast or at least set a fixed size for the worker queue.

How do you feel about a new _VertxOptions#workerPoolQueueCapacity_ ?
I could provide the accompanying PR as well.

fzakaria:
Add the ability to set the maximum capacity for the worker pool queue at
which point trying to enqueue additional tasks when its full will result
in a RejectedExecutionException

Previously the capacity of the queue was Integer.MAX_VALUE (unbounded)
which can result in a death-spiral for systems (i.e. http services) that
are processing more requests than they can handle.

For instance, as work tasks begin to queue for the workerPool, clients
that have yet to be serviced start retrying their operations. The queue
then continues to grow with clients continuing to retry their operation.
Death spiral.

Fixes #2465 

Signed-off-by: Farid Zakaria <farid.m.zakaria@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/VertxOptions.java
+++ b/src/main/java/io/vertx/core/VertxOptions.java
@@ -38,6 +38,11 @@ public class VertxOptions {
   public static final int DEFAULT_EVENT_LOOP_POOL_SIZE = 2 * CpuCoreSensor.availableProcessors();
 
   /**
+   * The default capacity for the queue into the worker pool executor service
+   */
+  public static final int DEFAULT_WORKER_POOL_QUEUE_CAPACITY = Integer.MAX_VALUE;
+
+  /**
    * The default number of threads in the worker pool = 20
    */
   public static final int DEFAULT_WORKER_POOL_SIZE = 20;
@@ -151,6 +156,7 @@ public class VertxOptions {
 
   private int eventLoopPoolSize = DEFAULT_EVENT_LOOP_POOL_SIZE;
   private int workerPoolSize = DEFAULT_WORKER_POOL_SIZE;
+  private int workerPoolQueueCapacity = DEFAULT_WORKER_POOL_QUEUE_CAPACITY;
   private int internalBlockingPoolSize = DEFAULT_INTERNAL_BLOCKING_POOL_SIZE;
   private long blockedThreadCheckInterval = DEFAULT_BLOCKED_THREAD_CHECK_INTERVAL;
   private long maxEventLoopExecuteTime = DEFAULT_MAX_EVENT_LOOP_EXECUTE_TIME;
@@ -184,6 +190,7 @@ public class VertxOptions {
   public VertxOptions(VertxOptions other) {
     this.eventLoopPoolSize = other.getEventLoopPoolSize();
     this.workerPoolSize = other.getWorkerPoolSize();
+    this.workerPoolQueueCapacity = other.getWorkerPoolQueueCapacity();
     this.blockedThreadCheckInterval = other.getBlockedThreadCheckInterval();
     this.maxEventLoopExecuteTime = other.getMaxEventLoopExecuteTime();
     this.maxWorkerExecuteTime = other.getMaxWorkerExecuteTime();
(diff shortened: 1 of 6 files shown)


vietj:
@jponge yes it would be better as the developer has to care about this result more than an runtime exception in the execute blocking

src/main/java/io/vertx/core/VertxOptions.java
```diff
@@ -37,6 +37,11 @@
    */
   public static final int DEFAULT_EVENT_LOOP_POOL_SIZE = 2 * CpuCoreSensor.availableProcessors();
 
+  /**
+   * The default capacity for the queue into the worker pool executor service
+   */
+  public static final int DEFAULT_WORKER_POOL_QUEUE_CAPACITY = Integer.MAX_VALUE;
```
vietj: it should be unbounded by default as this is the actual behavior.

vietj:
queues should be unbounded by default  (the actual case) and when it's unbounded we should not create a LinkedBlockingQueue because it's not necessary.

vietj:
also LinkedBlockingQueue should not be used at all since it can block the event loop

fzakaria:
@vietj @jponge Thank you for the review -- here are are responses:

In regards to LinkedBlockingQueue, the original implementation called `Executors.newFixedThreadPool` which is implemented as:
```java
public static ExecutorService newFixedThreadPool(int nThreads, ThreadFactory threadFactory) {
        return new ThreadPoolExecutor(nThreads, nThreads,
                                      0L, TimeUnit.MILLISECONDS,
                                      new LinkedBlockingQueue<Runnable>(),
                                      threadFactory);
    }
```
I kept the implementation the same as it just uses the same implementation but now specifies the max capacity of the queue (when no value is specified it defaults to _Integer.MAX_VALUE_)

In regards to the exception:
The class _ContextImpl.java_ already throws **RejectedExecutionException**

```java
catch (RejectedExecutionException e) {
      // Pool is already shut down
      if (metrics != null) {
        metrics.rejected(queueMetric);
      }
      throw e;
    }
```java

I did not want to change the callers expectation as at the moment they are expected to handle that exception when calling _executeBlocking_

jponge:
I don't think it's right for `executeBlocking` to throw an exception, as it is better for the handler of an asynchronous result to deal with it.

Also, consider the case of generated API like https://vertx.io/docs/apidocs/io/vertx/reactivex/core/Vertx.html#rxExecuteBlocking-io.vertx.core.Handler- where it makes more sense to have the error handling on the returned `Single` than on the method call site.

So perhaps we should modify the behaviour in `vert-core`, what do you think @vietj?

jponge:
On the `LinkedBlockingQueue` comment I think @fzakaria is right. It's the queue of jobs, and posting to it will either enqueue or raise an overfull exception, so it shouldn't block the event loop.

vietj:
@fzakaria we can change the behavior, the only way before to get rejection was during vertx shutdown (executors are closed) and it was not during application lifetime

src/main/java/io/vertx/core/VertxOptions.java
fzakaria: Responded in the comments section -- this was the previous implementation default.

fzakaria:
made the change.
@vietj to minimize any breaking changes to current consumers -- I handle the RejectedExecutionException only on the Handler if the executorService is not shutdown.

vietj:
I'm not sure that actually this PR is useful as it is because there is a special task queue in front of the executor for ordered execution, so in all I think we need to put more thoughts in this feature than just trivially bounding queues in executor.

the test is actually working because it specifies 10 max value but if you set an higher value (higher than 2 x your number of cores), it shall fail.

fzakaria:
I think I see what you mean -- I'd have to bound the TaskQueue as well ?
(for ordered execution)

Looks like the original proposition then to just have the _RejectedExecutionException_ throw like normally from the _executeBlocking_ call wouldn't also work because of that too.

What about when the TaskQueue is initialized it takes the same capacity as the workerPool?

fzakaria:
I think I see what you mean -- I'd have to bound the TaskQueue as well ?
(for ordered execution)

Looks like the original proposition then to just have the _RejectedExecutionException_ throw like normally from the _executeBlocking_ call wouldn't also work because of that too.

What about when the TaskQueue is initialized it takes the same capacity as the workerPool?

vietj:
I'm saying that for now we should sit down and think about what we really want to do and the problem we want to solve.

vietj:
have you experienced a problem in production system where you need this to solve your problem ?

vietj:
the thing is that the task queues are per context to keep ordering, so having a global configuration like in the proposal will not work at all.

fzakaria:
This issue has come from a production problem. 
The goal is try to and reduce or bound all queues in the system to achieve a consistent avg. response time.

vietj:
@fzakaria I'm not convinced that this would solve your issues at the moment. I think the best option is to throttle instead your incoming requests (for instance using a counter) when you receive them instead of going to the execute blocking and fail some. In addition if you use several execute blocking in the same workflow, some might succeed and some might be rejected which would be a very inefficient usage of the system. I will add this as a request feature so we can properly implement a solution for this.

vietj:
declining this PR with wontfix for now

vietj:
thanks for the contribution

PR #2472, opened 2018-05-23
Issue 368 - Allow to set connection exception handler for all request from HttpClientImpl

michaelpog:
Signed-off-by: Michael Pogrebinsky <smichpog2@gmail.com>

https://github.com/vert-x3/issues/issues/368

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpClient.java
+++ b/src/main/java/io/vertx/core/http/HttpClient.java
@@ -1422,6 +1422,13 @@ public interface HttpClient extends Measured {
   HttpClient redirectHandler(Function<HttpClientResponse, Future<HttpClientRequest>> handler);
 
   /**
+   * Set connection exception handler
+   * @param handler - handler to be called if a connection exception is raised
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  HttpClient connectionExceptionHandler(Handler<Throwable> handler);
+  /**
    * @return the current redirect handler.
    */
   @GenIgnore
--- a/src/main/java/io/vertx/core/http/HttpClient.java
+++ b/src/main/java/io/vertx/core/http/HttpClient.java
@@ -1422,12 +1422,12 @@ public interface HttpClient extends Measured {
   HttpClient redirectHandler(Function<HttpClientResponse, Future<HttpClientRequest>> handler);
 
   /**
-   * Set connection exception handler for all HTTP requests
+   * Set uncaught connection exception handler for all HTTP requests
    * @param handler - handler to be called if a connection exception is raised
    * @return a reference to this, so the API can be used fluently
    */
   @Fluent
-  HttpClient exceptionHandler(Handler<Throwable> handler);
+  HttpClient uncaughtExceptionHandler(Handler<Throwable> handler);
   /**
    * @return the current redirect handler.
    */
--- a/src/main/java/io/vertx/core/http/HttpClient.java
+++ b/src/main/java/io/vertx/core/http/HttpClient.java
@@ -1422,12 +1422,12 @@ public interface HttpClient extends Measured {
   HttpClient redirectHandler(Function<HttpClientResponse, Future<HttpClientRequest>> handler);
 
   /**
-   * Set connection exception handler
+   * Set connection exception handler for all HTTP requests
    * @param handler - handler to be called if a connection exception is raised
    * @return a reference to this, so the API can be used fluently
    */
   @Fluent
-  HttpClient connectionExceptionHandler(Handler<Throwable> handler);
+  HttpClient exceptionHandler(Handler<Throwable> handler);
   /**
    * @return the current redirect handler.
    */
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -106,6 +106,7 @@ public class Http1xClientConnection extends Http1xConnectionBase implements Http
     this.version = version;
     this.endpointMetric = endpointMetric;
     this.keepAliveTimeout = options.getKeepAliveTimeout();
+    this.exceptionHandler(client.exceptionHandler());
   }
 
   Object endpointMetric() {
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -865,7 +865,7 @@ public class Http1xClientConnection extends Http1xConnectionBase implements Http
     }
 
     if(exceptionHandler() == null) {
-      client.exceptionHandler().handle(e);
+      client.uncaughtExceptionHandler().handle(e);
     }
   }
 
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -106,7 +106,6 @@ public class Http1xClientConnection extends Http1xConnectionBase implements Http
     this.version = version;
     this.endpointMetric = endpointMetric;
     this.keepAliveTimeout = options.getKeepAliveTimeout();
-    this.exceptionHandler(client.exceptionHandler());
   }
 
   Object endpointMetric() {
@@ -863,6 +862,10 @@ public class Http1xClientConnection extends Http1xConnectionBase implements Http
       ws.handleException(e);
     } else {
       failStreams(e);
+    }
+
+    if(exceptionHandler() == null) {
+      client.exceptionHandler().handle(e);
     }
   }
 
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -59,7 +59,7 @@ import static io.vertx.core.http.HttpHeaders.*;
  *
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
-public class Http1xClientConnection extends Http1xConnectionBase implements HttpClientConnection {
+class Http1xClientConnection extends Http1xConnectionBase implements HttpClientConnection {
 
   private static final Logger log = LoggerFactory.getLogger(Http1xClientConnection.class);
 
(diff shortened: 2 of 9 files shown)


src/main/java/io/vertx/core/http/HttpClient.java
```diff
@@ -1413,6 +1413,13 @@ HttpClient websocket(String requestURI, MultiMap headers, WebsocketVersion versi
   @Fluent
   HttpClient redirectHandler(Function<HttpClientResponse, Future<HttpClientRequest>> handler);
 
+  /**
+   * Set connection exception handler
+   * @param handler - handler to be called if a connection exception is raised
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  HttpClient connectionExceptionHandler(Handler<Throwable> handler);
```
vietj: just call it exceptionHandler

vietj:
we need a few tests with it that shows it is used.

michaelpog:
Ok I think actually put it in the wrong place earlier since the connection actually gets created after the end() or write() is called. @vietj I also added some unit tests, let me know if this is right

michaelpog:
@vietj any additional work I need to do in this pr?

vietj:
no time yet to address it

src/test/java/io/vertx/test/core/HttpTest.java
```diff
@@ -3767,6 +3769,63 @@ public void testFollowRedirectHost() throws Exception {
     await();
   }
 
+  @Test
+  public void testSetExceptionHandler() throws Exception {
```
vietj: it would be better to test this without casting and instead trigger an exception on the stream, like the server closes the stream during a response transfer

michaelpog: Trying to find an example on how to do it...

michaelpog: @vietj sorry, I cannot figure out how to get raise that exception on the stream. Would you mind pointing me to what you are talking about?

michaelpog:
@vietj I've spent a few days trying to figure out what exactly you mean, and how to test it the way you expect it but I could not figure out what you had in mind.
Would you mind elaborating on what you see as a good test coverage for this feature and help me add those tests. We have a lot of those log lines in production, and that extra writing to disk may impact our performance. I am reluctant to completely disable logging from Vert.x, as some exceptions are important to catch.
Your help on this would be highly appreciated

vietj:
I was I think meaning that you set an exception handler on the HttpClient handler and simply trigger an error in the client by having the server sending a garbage HTTP response.

I think we should also check that connection close does not raise an exception in this particular handler.

michaelpog:
I see, yeah what you're proposing makes sense, I just don't understand what you mean by garbage HTTP response from the server.
I've tried setting all kinds of invalid combinations and values of HTTP headers, but I don't see any exception. Would you mind giving me an example please?

vietj:
@michaelpog just create a `NetServer` that sends an incorrect HTTP response, here are a few examples

_invalid http version_

```
HTTP/4.1 200 OK\r\n
\r\n
```

_invalid chunk_

```
HTTP/1.1 200 OK\r\n
transfer-encoding: chunked\r\n
\r\n
4\r\n
012\r\n
```

etc...

michaelpog:
@vietj I was able to add the test for the connection exception handler. Thanks for your help.
However I also wrote a test (not in this pr yet) where I close the connection from the netServer, and it does raise that connection exception handler. I'm wondering why you don't think it's an expected behavior?
I think this is exactly the behavior we are looking for as most of my logs are flooded with "connection closed by peer", which is the exception I want to handle and suppress

michaelpog:
Ok, I will just add a test that shows that a connection close from the server does raise that exception and the pr should be complete. Thanks @vietj

michaelpog:
@vietj please review my pr when you have a chance, I had a test to check that a connection close from the net server triggers the connection exception handler. For some reason that test works by itself, but fails whenever I run it with other tests. I'm totally clueless to why, so it's not included in this pr.

src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -440,59 +440,64 @@ protected String hostHeader() {
   }
 
   private synchronized void connect(Handler<HttpVersion> headersHandler) {
-    if (!connecting) {
+    if (connecting) {
```
vietj: can you avoid changing the control flow in the PR, so the differences with the actual code are easy to see, otherwise it really adds lot of noise to the actual review work

michaelpog: @vietj done. Sorry about that. I just thought that would make the code more readable. Fixed in the last commit

vietj: these should be done in separate PR's

vietj:
I think the actual callback should be instead in `Http1xClientConnection#handleException(Throwable)` method so any exception is actually reported:

```
  protected synchronized void handleException(Throwable e) {
    super.handleException(e);
    for (StreamImpl r = responseInProgress;r != null;r = r.next) {
      r.handleException(e);
    }
  }
```

michaelpog:
@vietj I am not sure I understand the last comment. The exception handler is set in the HTTP client and the exception is raised in the connection object. So I'm not really following, sorry

vietj:
the HttpClient#handler should catch any exception happening in the client.

eventually it should catch any unprocessed exception, that's somethign we need to decide.

michaelpog:
@vietj I think I'm starting to understand what you're saying.
So following the code, when a connection exception happens this method : 
```
protected synchronized void handleException(Throwable e) {
    super.handleException(e);
    for (StreamImpl r = responseInProgress;r != null;r = r.next) {
      r.handleException(e);
    }
  }
```
is called.
And the handler that was passed from the HttpClient is called in the super class (ConnectionBase).
So please let me know if that's what you mean:
You want that HttpClient handler to be called directly from the derived class inside that handleException method that I mentioned above, instead of in  the base class?

vietj:
yes, the entry point for exception is the ConnectionBase and the goal of the HttpClient handler is to catch them all.

that being said we could also think that the HttpClient handler catches only the uncaught exceptions from HttpConnection, i.e if there is an HttpConnection exception handler then it handles the throwable, however if there is no such handler then the HttpClient handler handles it.

michaelpog:
@vietj but this is exactly the implementation that I am proposing here, in this pr.
(Unless I misunderstood) you, the HttpClient exception handler overrides the ConnectionBase exception handler. If it does not that the HttpConnection exception handler takes care of the exception. 
The only thing is if the HttpClient Exception Handler is not set at all. Then we will log an error instead.

vietj:
@michaelpog I see however I think it's still the wrong place, it should instead happen in

- `Http1xClientHandler#handlerAdded(ChannelHandlerContext)`
- `Http2ClientConnection#createHttp2ConnectionHandler(...)`

so the handler is set once just after the connection begins it's life cycle.

vietj:
@michaelpog what is the use case for this feature ?

vietj:
can you rebase on latest master ? unit tests have been refactored (just package changes), doing simple

```
git pull -r vertx master
```

should do the job

michaelpog:
@vietj rebased.
The main use case is that currently Vertx Web Client doesn't have any way to pass a connection exception handler at all. So once this is merged, I can add an exception handler to WebClient to be passed to this HttpClient exception handler.
In addition, it gives a way to set an connection exception once instead of per request.
We are using Vert.x Web Client and we have our logs flooded with "connection reset by peer" exceptions. So this will allow us to potentially suppress that particular exception and log others.

michaelpog:
@vietj I'm not sure why CI is failing. I ran the tests locally and they all pass

michaelpog:
Ok, looks like it passed now.

michaelpog:
@vietj looks like it's good to go?

vietj:
@michaelpog I need to think about the name of the handler, it's called `exceptionHandler` and perhaps it should be `uncaughtExceptionHandler` ?

vietj:
also I think that the connection handler shouldn't be initialized with it, instead if the connection handler is null then it should escalate to the http client handler, the case is the following:

1/ open a new connection
2/ set an exception handler after a while : it overrides the connection handler
3/ set this exception handler to null

in this case we want the client handler to handle the exception.

perhaps that the handler should catch any exception caught and uncaught.

that's the concerns I have with the PR :-)

michaelpog:
Super valid concerns! I addressed all of them.

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -59,7 +59,7 @@
  *
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
-class Http1xClientConnection extends Http1xConnectionBase implements HttpClientConnection {
+public class Http1xClientConnection extends Http1xConnectionBase implements HttpClientConnection {
```
vietj: does still this needs to be changed to public visibility ?

vietj:
@michaelpog considering your use case, I think it should be better if we add on `HttpClient` a global connection handler that is called for every new connection, when the `HttpClientRequest` has not set one. So this way you can reliably handle new connection in WebClient and create a WebClient by wrapping an HttpClient. I will contribute this.

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
michaelpog: Nope, forgot to remove that

michaelpog:
@vietj yep. I rebase onto master after your changes.

vietj:
for now I want to leave this aside until the scope of the handler is figured out

michaelpog:
ok:(

vietj:
@michaelpog the HttpClient#connectionHandler allows you to build a WebClient by passing an HttpClient on which you can set the connection handler, i.e:

```
WebClient client = WebClient.wrap(vertx.createHttpClient().connectionHandler(conn -> {
  conn.exceptionHandler(err -> ...);
});
```

michaelpog:
Oh... so do we want to keep this pr? The same functionality can be achieved without it now with the new addition you made. But I guess it's a bit more explicit. What do you think?

vietj:
yes keep it, that feature might make it at some point in 4.0

PR #2478, opened 2018-05-26
Ssl cert refactoring

dachaac:
Updates instructions on how to create certificates and keys for testing.

Changes expiration date for certificates to be 10 years as existing certificates were already expired.

Changes CA names to be more descriptive.

Changes client certificate to be named as "client" to differentiate it from DNS based certificate names.

Removed the need to use external tools (apart from keytool and openssl) when recreating certs.


dachaac:
Used following helper script to auto-create certs:
```
echo "set -xe" > recreate-ssl.sh
echo "rm -rf root-ca int-ca other-ca | true" >> recreate-ssl.sh
echo "rm *.p12 *.jks *.pem *.crt *.key | true" >> recreate-ssl.sh
cat ssl.txt | grep -v -e ^# | grep -v -e ^[1-5T\(] | grep -v -e ^\- >> recreate-ssl.sh
```

dachaac:
Tried to verify that all tests pass like before.

There are changes to output but seems to be because certs were previously expired.

vietj:
so keyeplorer is not needed anymore ?

vietj:
thanks @dachaac !

PR #2482, opened 2018-05-30
Test acquiring different locks on same event loop

### LINKED ISSUE #2418: Possible deadlock when trying to get multiple locks
Here is the code, which illustrates the problem.
The `Consumer` verticle acuires `lock1` and gets stuck.
I have reproduced the same problem using `HazelcastClusterManager` and `IgniteClusterManager`.

```java
    Vertx.clusteredVertx(
        new VertxOptions().setClusterManager(new HazelcastClusterManager()),
        result -> {
          final Vertx vertx = result.result();

          final int messagesToSend = 10;

          final Consumer consumer = new Consumer();
          vertx.deployVerticle(
              consumer,
              new DeploymentOptions(),
              event -> {
                if (event.succeeded()) {
                  vertx.deployVerticle(
                      new Supplier(messagesToSend),
                      new DeploymentOptions()
                  );
                }
              }
          );

          vertx.setTimer(
              messagesToSend * DEFAULT_LOCK_TIMEOUT,
              timer -> {
                final long failsNumber = messagesToSend - consumer.counter.sum();
                if (failsNumber != 0) {
                  System.err.println("Failed to get " + failsNumber + " locks");
                }
                vertx.close();
              }
          );
        }
    );
```

```java
  class Supplier extends AbstractVerticle {
    private final long messagesToSend;

    public Supplier(long messagesToSend) {
      this.messagesToSend = messagesToSend;
    }

    @Override
    public void start() {
      final LongAdder sent = new LongAdder();
      vertx.setPeriodic(1, timer -> {
        vertx.eventBus().send(TEST_ADDRESS, "message");
        sent.increment();
        System.out.println(sent.sum() + " messages sent");
        if (sent.sum() >= messagesToSend) {
          vertx.cancelTimer(timer);
        }
      });
    }
  }
```

```java
class Consumer extends AbstractVerticle {
    private final LongAdder counter = new LongAdder();

    @Override
      vertx.eventBus().consumer(
          TEST_ADDRESS,
          event -> {
            vertx.sharedData().getLock("lock1", lock1 -> {
              if (lock1.failed()) {
                System.out.println("Failed to get lock1: " + lock1.cause());
              } else {
                System.out.println("Succeeded to get lock1");
                vertx.sharedData().getLock("lock2", lock2 -> {
                  if (lock2.failed()) {
                    System.out.println("Failed to get lock2: " + lock2.cause());
                  } else {
                    System.out.println("Succeeded to get lock2");
                    counter.increment();
                    System.out.println(counter.sum() + " messages consumed");
                    lock2.result().release();
                  }
                  lock1.result().release();
                });
              }
            });
          }
      ).completionHandler(startFuture);
    }
  }
```

tsegismont:
Closes #2418

### DIFF
--- a/src/test/java/io/vertx/test/core/AsynchronousLockTest.java
+++ b/src/test/java/io/vertx/test/core/AsynchronousLockTest.java
@@ -19,6 +19,7 @@ import io.vertx.core.shareddata.SharedData;
 import org.junit.Test;
 
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static io.vertx.test.core.TestUtils.*;
@@ -83,6 +84,32 @@ public class AsynchronousLockTest extends VertxTestBase {
           });
         });
       });
+    });
+    await();
+  }
+
+  @Test
+  public void testAcquireDifferentLocksOnSameEventLoop() {
+    Vertx vertx = getVertx();
+    Context context = vertx.getOrCreateContext();
+    SharedData sharedData = vertx.sharedData();
+    AtomicInteger stage = new AtomicInteger();
+    context.runOnContext(v -> {
+      sharedData.getLock("foo", onSuccess(foo -> {
+        assertTrue(stage.compareAndSet(0, 1));
+        // Create another lock request
+        sharedData.getLock("foo", onSuccess(foo1 -> {
+          assertEquals(2, stage.get());
+          foo1.release();
+          testComplete();
+        }));
+        // Should not be blocked by second request for lock "foo"
+        sharedData.getLock("bar", onSuccess(bar -> {
+          assertTrue(stage.compareAndSet(1, 2));
+          foo.release();
+          bar.release();
+        }));
+      }));
     });
     await();
   }
@@ -179,5 +206,4 @@ public class AsynchronousLockTest extends VertxTestBase {
     });
     await();
   }
-
 }
--- a/src/test/java/io/vertx/test/core/ClusteredAsynchronousLockTest.java
+++ b/src/test/java/io/vertx/test/core/ClusteredAsynchronousLockTest.java
@@ -32,7 +32,7 @@ public class ClusteredAsynchronousLockTest extends AsynchronousLockTest {
     return new FakeClusterManager();
   }
 
-  protected final int numNodes = 2;
+  protected final int numNodes = 3;
 
   public void setUp() throws Exception {
     super.setUp();


tsegismont:
@vietj here is the test in Vert.x core. There are PRs for all cluster managers.

vietj:
I am actually not in favour of this change.

In case case of N concurrent acquisitions of the same lock inside the same VM, all contenders will wait and block on a worker thread, so no deadlock but it consumes N worker threads.

My proposal is to have one ephemeral TaskQueue per lock which sounds to me like a good trade-off  : we consume at most one worker thread as we order on the lock and we don't avoid deadlocks.

vietj:
actually this is already the case I think currently, so I'm proposing to merge this and open an issue to serialise the lock acquisition per named lock.

that being said, I would like to see extra TCK test(s), in particular one that acquires several times the same lock concurrently in the same VM and make sure that all ownerships are satisfied and serials.

tsegismont:
>My proposal is to have one ephemeral TaskQueue per lock which sounds to me like a good trade-off : we consume at most one worker thread as we order on the lock and we don't avoid deadlocks.

@vietj this will introduce a lot of complexity for something which is really just an edge/fabricated case. 

For example, you could use a Vert.x lock to make sure just one verticle instance in the cluster will update your relational database with the new schema.

If acquiring different locks in order in a multi-threaded process is pretty common, I'm not sure what's the use case for distributed locks in Vert.x.

> actually this is already the case I think currently

Currently, the internal ordered pool is used. So at most one thread is used. OTOH, waiting for a lock might block internal pool tasks (like Vert.x `FileSytemp` ops). Not great either.


Consequently, I'm in favor of keeping the changes in the different cluster managers.

Note that Infinispan offers a `CompletableFuture` API for locks, so the implementation could be improved to simply avoid `executeBlocking` in this cluster manager.

vietj:
@tsegismont yes fine, I think nobody complained so far that this can drain worker pools, so I guess it does not happen in practice

see my later comment

tsegismont:
> see my later comment

which one?

vietj:
the one that contains _I would like to see extra TCK test(s), in particular one that acquires several times the same lock concurrently in the same VM and make sure that all ownerships are satisfied and serials._

tsegismont:
What do you want to check exactly? I may miss the point but I believe we'd only add a test for something which is probably already in the Hazelcast/Infinispan/... etc test suite

vietj:
can you point it out ?

tsegismont:
I haven't checked their test suites and I'm not really excited about doing so :)

My point is that we have a test for acquisition of the same lock on the same event loop already: `AsynchronousLockTest#testAcquireOnSameEventLoop`

We can do it for several requests of course. But:

- how is it related to the problem described in this issue?
- we could do this for locks, maps as well

vietj:
does it mean that in `AsynchronousLockTest#testAcquireOnSameEventLoop` the cluster manager lock acquisition will be concurrently executed by two threads ?

vietj:
So I've looked at your new test and it does not test what I meant, here is the test I wrote:

```
  @Test
  public void testConcurrentAcquisitionOfSameLock() {
    int n = 3;
    waitFor(n);
    Vertx vertx = getVertx();
    SharedData sharedData = vertx.sharedData();
    AtomicInteger current = new AtomicInteger();
    for (int i = 0;i < n;i++) {
      sharedData.getLock("foo", onSuccess(lock -> {
        assertEquals(0, current.getAndIncrement());
        vertx.setTimer(500, id -> {
          assertEquals(0, current.decrementAndGet());
          lock.release();
          complete();
        });
      }));
    }
    await();
  }
```

tsegismont:
@vietj the new test in this PR (`testAcquireDifferentLocksOnSameEventLoop`) makes sure the problem described in the issue (#2418) is fixed.

It is indeed not related to what you're talking about. I would prefer not to block this PR (and PRs in cluster managers) with something which is unrelated.

If you have some new testing code ready perhaps you could simply push it to master? I had a look at your test code snippet by the way. It does not test what you say it tests: the same thread (`main`) tries to acquire a single lock. It is not different from what's in `testAcquireOnSameEventLoop` (set aside that one test tries to acquire the lock twice and the other three times).

vietj:
actually it does because, because each cluster manager will have executeBlocking creating an implicit Vertx `Context`.

tsegismont:
> actually it does because, because each cluster manager will have executeBlocking creating an implicit Vertx Context.

Right, but then it's really white box testing. And we wouldn't be testing our code, but just the underlying tool.

PR #2501, opened 2018-06-14
Internal references to local Locks taken using SharedData.getLock() are not cleaned up

### LINKED ISSUE #2484: Internal references to local Locks taken using SharedData.getLock() are not cleaned up
Every time a unique lock name is taken via `SharedData.getLock(...)`, it gets added to a Map inside of `SharedDataImpl`. Currently, entries added to this Map are never removed, even if all references to the lock have been released, and there are no pending waiters. This means the Map can grow infinitely if an application takes locks using generated names, rather than hard-coded ones. To avoid this, entries could be removed from the internal map at the point the lock is released, as long as there are no pending waiters.

tsegismont:
Fixes #2484 

Refactored local locks implementation to avoid keeping references to lock with no waiters.


vietj:
here is the initial feedback after a first pass:

1/ there is a race condition with timedOut / acquired in timeout() / acquire(AsyncLock lock) methods so that the handler can be called twice with a lock granted and an acquisition.
It should be exclusive and if you are using a volatile synchronization, the best would be a single status (waiting, timedOut, acquired) atomic field with a CAS operation.

2/ AsyncLock should not implement Lock directly and instead should provide a delegating Lock implementation that has a field preventing misuse of the Lock,
i.e release() can be called only once otherwise throw an IllegalStateException

3/ in forNextWaiter() the list filtering with timed out waiter can lead to a race condition (filter(LockWaiter::notTimedOut)).

- waiter 1 acquires the lock from a non event loop thread with a timeout, is the first in the AsyncLock list and gets its AsyncLock view
- thread waiter 1 is unscheduled
- waiter 1 timer fires on the random event loop thread (obtained from setTimer() from non vertx thread) and update the waiter 1 state to be timedOut
- waiter 2 acquires the lock and filters the view, it gets an AsyncLock view that is itself (because of filtering) and will call AsyncLock#acquire() because firstWaiter() returns true
- waiter 1 thread is now rescheduled it will call also acquire(), as it is timed out, it will call release() and call AsyncLock#acquire() a second timed

tsegismont:
@vietj fixed 1/ and 2/

Looking in to 3/

tsegismont:
Actually, I think 3/ should not happen in the new impl. Or am I missing something?

vietj:
what is the new impl ?

vietj:
something you committed after my initial review ?

tsegismont:
Yes

vietj:
as you changed the implementation I need to make a 2nd review, I'll do that later

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
```diff
@@ -0,0 +1,142 @@
...
+    int size() {
+      return waiters.size();
+    }
+
+    LockWaiter first() {
+      return waiters.get(0);
+    }
+
+    WaitersList add(LockWaiter waiter) {
+      return new WaitersList(Stream.concat(waiters.stream(), Stream.of(waiter)).collect(toList()));
```
vietj: simply create an ArrayList with the exact size as it is determine (Collectors#toList() also creates an ArrayList)

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    WaitersList removeStale() {
```
vietj: remove this `removeStale`

it adds a cleanup mechanism that is already taken care of by acquireLock() already

it makes the algorithm simpler and smaller

tsegismont: I don't understand. `acquireLock()` invokes this method to remove the first waiter and the following ones if they timed out.

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
tsegismont: Done

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
vietj: I mean it could be simpler

- since `acquireLock ` already takes care of skipping the lock if that's timed out, each lock will be either executed or skipped, so there is no need to cleanup the stale elements
- when `compute()` is called on the map, it returns the rest of the list or `null` if the list was a single element

I don't think it buys much to filter stale locks

tsegismont: Ha. I see your point now. Filtering out eagerly the waiters removes the need to recompute the waiters list N times if there are N timed-out waiters.

tsegismont: I just pushed a commit:

* removed WaitersList abstraction
* do not skip timed out waiters when recomputing the list

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
```diff
@@ -0,0 +1,118 @@
...
+      timerId = timeout != Long.MAX_VALUE ? context.owner().setTimer(timeout, tid -> timeout()) : null;
+    }
+
+    boolean isWaiting() {
+      return status.get() == Status.WAITING;
+    }
+
+    void timeout() {
+      if (status.compareAndSet(Status.WAITING, Status.TIMED_OUT)) {
+        handler.handle(Future.failedFuture(new VertxException("Timed out waiting to get lock")));
```
vietj: use instead the string version in order to avoid creating a costly stack trace creation with the exception.

tsegismont: Done

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
```diff
@@ -0,0 +1,118 @@
...
+    final String lockName;
+    final AtomicBoolean invoked = new AtomicBoolean();
+
+    AsyncLock(String lockName) {
+      this.lockName = lockName;
+    }
+
+    @Override
+    public void release() {
+      if (invoked.compareAndSet(false, true)) {
```
vietj: `java.util.concurrent.locks.Lock` throws `java.lang.IllegalMonitorStateException` when you release an not owned lock, perhaps we should do the same here ?

tsegismont: I don't believe we should. But this PR is supposed to fix #2484 , so I'd prefer we stick to that.

vietj: fine

PR #2503, opened 2018-06-19
Add a IterableMessageCodec for eventbus

jo5ef:
This is a message codec for iterables. When sending iterables (e.g. `List<String>`, `Set<Integer>`, etc.) over the eventbus, common practice is to first convert them to a `JsonArray`. This `IterableMessageCodec` allows to send iterables directly, provided an appropriate message codec for the item type exists.

### Example

```
vertx.eventBus().<Iterable<String>>consumer("topicname", msg -> {
  Iterable<String> items = msg.body();
});
vertx.eventBus().send("topicname", List.of("foo", "bar", "eek"));
```

### Why?

In our vertx applications, we use mostly primitive types and pojos (for ease of use, type safety, speed, etc.). For us, converting to json (and back) just for the eventbus is inefficient. I've added a benchmark to the PR that shows a ~30% performance gain when using the `IterableMessageCodec` instead.

```
Benchmark                                                    Mode  Cnt    Score    Error   Units
IterableMessageCodecBenchmark.iterableCodec                 thrpt   10  478.171 ± 18.934  ops/ms
IterableMessageCodecBenchmark.jsonArrayCodec                thrpt   10  394.110 ±  6.332  ops/ms
IterableMessageCodecBenchmark.jsonArrayCodecWithConversion  thrpt   10  374.585 ± 14.168  ops/ms
```

### Remarks
* iterable items must be of equal type
* custom message codecs for items not (yet) supported

Please let me know what you think.


eutkin:
@vietj Hello. Will this feature be implemented? It is very necessary to optimize the operation of Vertx Service Proxy

vietj:
we will think of a solution, we are currently working on improving this

eutkin:
@vietj see https://github.com/eclipse-vertx/vert.x/pull/2966 please. This feature is needed to implement support for message codecs in the Vert.x Service Proxy. My version is slightly different from this one, but for the Service Proxy project this is a really important feature.

vietj:
@eutkin I don't understand how this is important for service proxies because this PR cares about Iterable and the service proxy encoding uses JSON before sending the message over the bus, can you elaborate ?

eutkin:
with this feature, we can serialize the method arguments not in json format (java object -> json -> binary), but directly into binary (java object -> binary) in the way that the user chooses

```java
public class SomeDatabaseServiceVertxEBProxy implements SomeDatabaseService {

  public SomeDatabaseServiceVertxEBProxy (Vertx vertx) {
       vertx.eventBus()
              .registerDefaultCodec(SampleDataObject .class, new SampleDataObjectMessageCodec());
  }

  @Override
  public  void save(String collection, SampleDO dataObject, Handler<AsyncResult<Void>> result) {
    if (closed) {
      result.handle(Future.failedFuture(new IllegalStateException("Proxy is closed")));
      return;
    }
   /* Serialize data in text format
    JsonObject args = new JsonObject();
    args.put("collection", collection);
    args.put("dataObject", dataObject.toJson());
   */
  // now, serialize data in binary format, supporting user's codecs (default codec).
  // and DataObject may not have Constructor(JsonObject json) and toJson method.
   Iterable<Object> args = asList(collection, document);

    DeliveryOptions _deliveryOptions = (_options != null) ? new DeliveryOptions(_options) : new DeliveryOptions();
    _deliveryOptions.addHeader("action", "save");
    _vertx.eventBus().<Void>send(_address, args, _deliveryOptions, res -> {
      if (res.failed()) {
        result.handle(Future.failedFuture(res.cause()));
      } else {
        result.handle(Future.succeededFuture(res.result().body()));
      }
    });
  }
}
```

vietj:
so you are using a custom service proxy project ?

eutkin:
@vietj Not. I want to modify the existing one. But in order to do this, I need to serialize the method arguments without using text formats (json).

1. This feature merges
2. I do a pull request in Vert.x Service Proxy to change proxy generators
3. ???
4. Profit

eutkin:
@vietj This feature will eliminate the need to have a constructor with JsonObject and toJson () method in DTO. And this is important if we want to use models from third-party dependencies, which can be serialized for transmission to eventBus via message codecs.

eutkin:
@vietj Why you do not want to merge this feature?

vietj:
we do have a work in progress about json serialization currently and I would like to address this rather in this effort.

eutkin:
Why do you think that serialization in json will be better? At the moment, it creates more inconvenience than advantages in comparing message codec. Or something I do not understand?

vietj:
@eutkin what about serialising to BSON ?

eutkin:
@vietj The problem is not the binary data, but the limitations of the json format. Some data can be serialized to a binary format, but cannot be serialized to json. Moreover, external json message codecs are not currently supported, for example, if we have a third-party model (Message.class from telegram client), we cannot transmit it through a service proxy without creating a duplicate model with the @DataObject annotation and methods for de / serialization in json. The use of MessageCodec will provide much more opportunities than encoding information in json (bjson). While I do not see the advantages of the json format over the message codec (from event bus), the more globally the changes will suffer only proxy in the service proxy project. This is not a global change. I do not propose to abandon the json format, I propose to give the user (developer) a choice of what to use.

jo5ef:
I don't mean to interrupt your discussion but I want to point out that this PR is not about whether JSON is good or bad on the event bus. It is about adding a built-in codec for iterables *of any type*.

In my opinion sending multiple items (of any type) in one message is a pretty common use case and vertx users shouldn't have to implement their own iterable codec every time or - worse - encode and decode their items in `JsonArray`s just for the trip through the event bus.

vietj:
yes so it's more an issue of service-proxy than vertx itself right ?

vietj:
i.e service-proxy should provide a way to do that at some point

vietj:
@jo5ef I perfectly hear you :-) hence my last comment to @eutkin about the $topic, I'm currently dedicating some time to encoding in general (json, etc...) and trying to gather all the issues related to it to move forward with something coherent, this issue is one of them

jo5ef:
@vietj roger : )

eutkin:
@jo5ef Your pull request (or [mine](https://github.com/eclipse-vertx/vert.x/pull/2966), because I didn’t notice yours) will allow replacing json serialization with message codec in the vertx service proxy project, so it has much more meaning than just “regular message codec for collections”

vietj:
we don't aim replacing json serialization in vertx proxy, vertx proxy use direct serialization of known types and no lookups

eutkin:
@vietj I know and therefore service proxy is so uncomfortable

eutkin:
any progress?

eutkin:
up

DarwinsBuddy:
is this still in review? would be a nice feature :)

PR #2507, opened 2018-06-20
Add instant deserializer

EmadAlblueshi:

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -14,12 +14,7 @@ package io.vertx.core.json;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.DeserializationContext;
-import com.fasterxml.jackson.databind.JsonDeserializer;
-import com.fasterxml.jackson.databind.JsonSerializer;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
-import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.*;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import io.netty.buffer.ByteBufInputStream;
 import io.vertx.core.buffer.Buffer;
@@ -35,7 +30,7 @@ import java.util.Map;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
-import static java.time.format.DateTimeFormatter.*;
+import static java.time.format.DateTimeFormatter.ISO_INSTANT;
 
 /**
  * @author <a href="http://tfox.org">Tim Fox</a>
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -14,12 +14,16 @@ package io.vertx.core.json;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.module.SimpleModule;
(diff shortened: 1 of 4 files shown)


pmlopes:
I understand the PR as we support instant at the JsonObject/JsonArray level, but there it is only a read operation that converts from text to Instant to be used in code. If we allow this (and then probably the same with the bas64 strings) we could avoid many parsing/encoding cycles as we would keep types in the right state in memory (not the text representations).

I think in order to make sure this works, we need to see how would this work with eventbus/clustering serialization. I know for example on vertx-web sessions this would fail as it is quite strict with the json types.

Question for @vietj should we go this way and avoid encode/decode and probably have to adapt web session data to be a pure JSON object? or do you have a good reason not to go?

vietj:
I'm not sure to understand what you mean @pmlopes I thought this only pojo mapping by default and not related to json object

pmlopes:
JsonObject/Array implementation follows the json spec + 1 rfc that specifies decoding only from string to instant or base64. So in memory, there's never an object of type Buffer or Instant. 

When we do getInstant() a parse from string is performed, a setInstant() encodes to string. So you can imagine if these types are used often lots of gc is generated. 

If we add this pr, then we would just keep instant objects in memory, only encode to string when crossing boundaries. 

Now my concern is the borders as I known that it will fail on web session but I've no idea about the behavior on the eventbus.

Personally I think it will make things simple but it requires a deep look before merge

vietj:
what could happen wrong on web sessions ? how are json object / array serialized ?

EmadAlblueshi:
@vietj @pmlopes we have the ability to decode from JSON string to Pojo this PR includes Instant like other JSON data types. I think it completes what is missing instead of adding extra Jackson packages for temporal java 8 types.

pmlopes:
@vietj ah! :) the whole issue with sessions is that for some reason the object is not a json object but a Java map so it uses custom serialization. 

But I agree with @EmadAlblueshi it simplifies things but needs checks across the stack, I'm fine in proposing a change from map to JsonObject in web sessions but technically it's a breaking one...

vietj:
but I still don't see the relationship between session and json object since session uses map, I guess that json object are serialized when put in web session by encoding them to a string

pmlopes:
@vietj the relation is the following, currently the map serializer allows the exact same types as JSON + ClusterSerializable, so imagine that you have a Json like this:

```json
{ "d": "2011-12-19T15:28:46.493Z" }
```

you can do something like:

```java
// store the date into the session by value
session.put("date", json.getValue("d"))
// store it in the storage
storage.put("sessionId", session);
```

This works as currently we don't have Instant types in JSON types (internally they're always Strings), now assume we allow Instants, at the store time you'd get an `IllegalParameterException` or something like that (I'm typing without reading the code) because the serializer assumes no Instant type is allowed.

So, my point here is that I'm OK with allowing this **but** we must check over the stack for issues like this and probably, replace the custom serialization by Jackson so we're always consistent all over the stack.

pmlopes:
Also note that `JsonObject` uses jackson underneath for parsing (e.g.: `new JsonObject("{ \"d\": \"2011-12-19T15:28:46.493Z\" }")'`, so it could assume wrongly that a instant is stored as a string internally when doing `put` and `get` operations` perhaps this PR should that that too.

vietj:
so we should check that this PR only affects POJOs and not `JsonObject` / `JsonArray` and that is fine

vietj:
@EmadAlblueshi can you check if there is a test that checks that the content (via getValue()) of a JsonObject / JsonArray instant remain `String` and are not `Instant` object ?

if there is not, then we just need an extra test with that.

EmadAlblueshi:
@vietj sure 👍

EmadAlblueshi:
@pmlopes @vietj Tests for `JsonObject` / `JsonArray` have been added.

vietj:
@pmlopes is that ok for you ?

EmadAlblueshi:
@vietj I think it's ok shall we proceed :) ?

vietj:
yes, I didn't notice paulo's ok because there is no notification for thumbs up :-)

vietj:
that being said, I'm seying we are supporting serializer for : Instant, JsonArray, JsonObject, and byte[]

shouldn't we also have deserializer for these classes as well to have somehting symetric ?

EmadAlblueshi:
@vietj well I think we don’t need for Json[Object/Array] but the `byet[]` we need know reasonable use cases. because we need to ask are we going to deserialize from `BASE64` only  ? 

I think it could be from different type :)

`Instant` is not supported by default in jackson deseralizer and heavily being used in vertx `Json[Object/Array]` as the only temporal data type in the core.

pmlopes:
@EmadAlblueshi @vietj for byte[] to make things consistent it should be to and from `Buffer` this way it's consistent with the rest of the API

vietj:
there is already a byte[] serializer

tsegismont:
@EmadAlblueshi thanks for the patch. I have incorporated your work in https://github.com/eclipse-vertx/vert.x/pull/2754. This other PR adds the byte array handling. I'm sorry I couldn't keep your author email in the commit (due to Eclipse IP check).

PR #2516, opened 2018-06-26
Add discriminator to internal eventbus addresses

### LINKED ISSUE #1240: Prefix internal event bus handlers with "__vertx."
For instance the NetSocket handlers, that would allow the Metrics SPI to distinguish them easily.

jotak:
Addresses are now like: "__vertx.[net|ws|reply].randomUUID"

Fixes #1240

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -407,7 +407,7 @@ public class EventBusImpl implements EventBus, MetricsProvider {
   }
 
   protected String generateReplyAddress() {
-    return Long.toString(replySequence.incrementAndGet());
+    return "__vertx.reply." + Long.toString(replySequence.incrementAndGet());
   }
 
   private <T> HandlerRegistration<T> createReplyHandlerRegistration(MessageImpl message,
--- a/src/main/java/io/vertx/core/net/impl/NetSocketImpl.java
+++ b/src/main/java/io/vertx/core/net/impl/NetSocketImpl.java
@@ -84,7 +84,7 @@ public class NetSocketImpl extends ConnectionBase implements NetSocketInternal {
                        SSLHelper helper, TCPMetrics metrics) {
     super(vertx, channel, context);
     this.helper = helper;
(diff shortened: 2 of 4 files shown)


src/main/java/io/vertx/core/impl/EventbusAddresses.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.UUID;
+
+/**
+ * @author Joel Takvorian
+ */
+public final class EventbusAddresses {
+  private EventbusAddresses() {
+  }
+
+  public static String generateInternal(String module) {
```
vietj: move it to another class (like MessageImpl), no need to create an extra class just for this purpose

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -407,7 +406,7 @@ protected void checkStarted() {
   }
 
   protected String generateReplyAddress() {
-    return Long.toString(replySequence.incrementAndGet());
+    return EventbusAddresses.generateInternal("reply");
```
vietj: keep the sequence, but just prefix with "__vertx.reply."

src/main/java/io/vertx/core/net/impl/NetSocketImpl.java
```diff
@@ -84,7 +84,7 @@ public NetSocketImpl(VertxInternal vertx, ChannelHandlerContext channel, SocketA
                        SSLHelper helper, TCPMetrics metrics) {
     super(vertx, channel, context);
     this.helper = helper;
-    this.writeHandlerID = UUID.randomUUID().toString();
+    this.writeHandlerID = EventbusAddresses.generateInternal("net");
```
vietj: just do "__vertx.reply" + UUID.randomUUID().toString()

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
jotak: Ah, is the sequence used for something?

vietj: I don't know, but I think it's less expensive than creating an UUID :-)

PR #2518, opened 2018-06-27
Disable HttpClient TCP idle timeout when connection is in the pool - …

### LINKED ISSUE #2517: Disable HttpClient TCP idle timeout when connection is in the pool
I noticed that in a corner case, we have a lot of test failure where the connection is not used very often. For example, I make a timeout to be of 10 seconds and I make a call every 10 seconds exactly. Every once in while, my request fails with VertxException (Connection is closed) since the checkout connection is closed right after the connection is polled from the connection pool. I only noticed this in 3.5.2 but not in 3.4.2

vietj:
…fixes #2517


vietj:
will be back-ported to 3.5.x

vietj:
another one for you @tsegismont

PR #2523, opened 2018-07-03
HttpClient response chunk handling rework

### LINKED ISSUE #2522: HttpClient response chunk handling rework
Rework the `HttpClient` request chunk handling that is split between `HttpClientResponse` and `HttpClientStream` so it is fully managed by the stream. The logic is now encapsulated in the stream and it simplifies the logic of the client response that does not care anymore of paused trailers and paused end chunk (that was only used by HTTP/1).

vietj:
Fixes #2522

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -38,7 +38,6 @@ import io.vertx.core.spi.metrics.HttpClientMetrics;
 
 import java.net.URI;
 import java.util.ArrayDeque;
-import java.util.Deque;
 import java.util.Map;
 import java.util.Queue;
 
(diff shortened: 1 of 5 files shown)


src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -212,6 +270,29 @@ private void sendRequest(
...
+      synchronized (conn) {
+        if (paused) {
+          if (pausedChunk == null) {
+            pausedChunk = buff.copy();
+          } else {
+            pausedChunk.appendBuffer(buff);
+          }
+          return;
+        } else {
+          if (pausedChunk != null) {
```
codepitbull: Just for my understanding:
This part is only here to support the multithreaded case.
In the case of regular EventLoop usage the doResume should have taken care of everything before handleChunk could be hit again.

vietj: actually it is necessary because there is a possible race condition since the actual resuming uses `runOnContext` to deliver the chunk asynchronously (`doResume` might be called from non vertx thread): so it's possible to have `paused == false` and `pausedChunk != null`.

This part is actually going away with a PR I'm currently working on (https://github.com/vert-x3/wiki/wiki/Vert.x-Roadmap#stream-improvements). The idea of this PR is to prepare the http client connection so the stream behaviour is replaced by the next PR (and also in all other places of vertx).

vietj: In my working branch the new implementation is simply:

```
    private boolean handleChunk(Buffer buff) {
      return queue.push(buff);
    }
```

codepitbull: Ah crap, yes, overlooked the runOnContext.

codepitbull:
LGTM

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
vietj: that's why we want to encapsulate such racy prone code in a single class :-)

PR #2528, opened 2018-07-04
JsonObject.mergeIn helping method

mystdeim:
Method JsonObject.mergeIn(T object) can be helpfull when we pass json in the object constructor, for example:

```java
@DataObject
public class Account {

    UUID id;
    String person;

    public Account(JsonObject obj) {
        obj.mergeIn(this);
    }

    public JsonObject toJson() {
        return JsonObject.mapFrom(this);
    }

    //...
}
```


mystdeim:
Why I couldn't pass ip-validation?

PR #2549, opened 2018-07-24
Fix FileResolver for OSGi bundle resources

### LINKED ISSUE #2548: `FileResolver.unpackFromBundleURL()` does not detect a directory unless the requested path has a trailing '/'
As can be seen from [this line](https://github.com/eclipse/vert.x/blob/407001830066c8f819ce8d8f5a10188e21c0c3c7/src/main/java/io/vertx/core/file/impl/FileResolver.java#L279) of `FileResolver.java`, `unpackFromBundleURL()` does not account for the possibility of a resource path to be a directory - even if it doesn't end with a '`/`'.

For example, consider the following sample OSGi bundle structure (excluding manifests etc.):
```
META-INF/
└── resources
    └── webjars
        └── swagger-ui
            └── 3.17.4
                ├── favicon-16x16.png
                ├── favicon-32x32.png
                ├── index.html
                ├── index.html.gz
                ├── oauth2-redirect.html
                ├── oauth2-redirect.html.gz
                ├── swagger-ui-bundle.js
                ├── swagger-ui-bundle.js.gz
                ├── swagger-ui-bundle.js.map
                ├── swagger-ui.css
                ├── swagger-ui.css.gz
                ├── swagger-ui.css.map
                ├── swagger-ui.js
                ├── swagger-ui.js.gz
                ├── swagger-ui.js.map
                ├── swagger-ui-standalone-preset.js
                ├── swagger-ui-standalone-preset.js.gz
                └── swagger-ui-standalone-preset.js.map

4 directories, 18 files
```
If `unpackFromBundleURL()` is called for `META-INF/resources/webjars/swagger-ui/3.17.4/`, it will correctly reach the directory code [here](https://github.com/eclipse/vert.x/blob/407001830066c8f819ce8d8f5a10188e21c0c3c7/src/main/java/io/vertx/core/file/impl/FileResolver.java#L280). However, if this method is called for `META-INF/resources/webjars/swagger-ui/3.17.4`, the `else` branch (for file resources) will be taken and the cache will be corrupted.

See #2549 for a fix.

devplayer0:
The use of `FileResolver` for a directory without a trailing '`/`' will fail since `unpackFromBundleURL()` does not handle this case correctly.

By trying `ClassLoader.getResource()` on such a resource path with an appended '`/`', it can be determined whether or not the resource is really a directory since a URL will only be returned if the resource exists and is a directory.

Fixes #2548.

### DIFF
--- a/src/main/java/io/vertx/core/file/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/file/impl/FileResolver.java
@@ -272,6 +272,18 @@ public class FileResolver {
   }
 
   /**
+   * It is possible to determine if a resource from a bundle is a directory based on whether or not the ClassLoader
+   * returns null for a path (which does not already contain a trailing '/') *and* that path with an added trailing '/'
+   *
+   * @param url      the url
+   * @return if the bundle resource represented by the bundle URL is a directory
+   */
+  private boolean isBundleUrlDirectory(URL url) {
+    return url.toExternalForm().endsWith("/") ||
+      getClassLoader().getResource(url.getPath().substring(1) + "/") != null;
+  }
+
+  /**
    * bundle:// urls are used by OSGi implementations to refer to a file contained in a bundle, or in a fragment. There
    * is not much we can do to get the file from it, except reading it from the url. This method copies the files by
    * reading it from the url.
@@ -283,7 +295,7 @@ public class FileResolver {
     try {
       File file = new File(cacheDir, url.getHost() + File.separator + url.getFile());
       file.getParentFile().mkdirs();
-      if (url.toExternalForm().endsWith("/")  || isDir) {
+      if (isBundleUrlDirectory(url)  || isDir) {
         // Directory
         file.mkdirs();
       } else {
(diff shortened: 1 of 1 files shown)


devplayer0:
I have just signed the required ECA.

pmlopes:
sorry for the inconvenience, @devplayer0 can you amend your commit so the eclipse validation will rerun?

devplayer0:
@pmlopes Done.

vietj:
@cescoffier can you review this please ?

devplayer0:
@vietj @cescoffier Any possibility of getting this merged?

vietj:
I think we should get it reviewed and merged

devplayer0:
@cescoffier If `getClassLoader()` could be `null` in that environment I can add a check. As of now I have only tested this under Felix, I'll try and get to testing it on others.

devplayer0:
@cescoffier I have confirmed this fix works on Equinox as well as Felix, and I have added a `null` check on `getClassLoader()`.

cescoffier:
Thanks @devplayer0 !

I agree that the `null` check is weird, but well... it's the only way I've found to detect this case.

Issue #2550, opened 2018-07-24
Use VertxOptions to configure class path resolving


MarkArbogast:
Class path resolving is currently enabled or disabled via [System property](https://github.com/eclipse/vert.x/blob/407001830066c8f819ce8d8f5a10188e21c0c3c7/src/main/java/io/vertx/core/file/impl/FileResolver.java#L59).

In environments with multiple Vert.x-based applications the property will be set to the same value for all the applications running in the same JVM.

Adding a VertxOption field to override the System property would allow each Vert.x application in a JVM to decide individually whether to allow file class path resolving.

The use-case driving this issue is that we would like to use Vert.x as a client without any file system side effects (such as creating a `.vertx` directory), but we don't want to interfere with other Vert.x-based applications that may be running in the same JVM.

MarkArbogast:
Addressed in #2561

PR #2554, opened 2018-07-24
Add file class path resolving option to VertxOptions

MarkArbogast:
Addresses: #2550 

Signed-off-by: Mark Arbogast <mark.t.arbogast@gmail.com>


vietj:
I think we should use a `FileSystemOptions` as we may have other related options in the future.

MarkArbogast:
@vietj I added `FileSystemOptions` in #2561. Does that look closer to what we want? If so I will close this PR.

vietj:
yes that sounds better in the #2561 PR.

PR #2568, opened 2018-07-31
EventBus local only delivery

### LINKED ISSUE #1764: Performance penalty of localConsumer() when in -cluster mode
Using vertx 3.3.3

There seems to be a performance penalty when using localConsumer() in clustered mode.
There is cases where you want to mix and match how various verticles are consumed. Some you may only want them as local and others clustered. So a localConsumer() should not incur any performance loss. 

See conversation with Julien and Thomas here: https://groups.google.com/forum/#!topic/vertx/HlWmc4ZuLtA

The reproducer code...
http server: http://pastebin.com/hA4T8Nyy
verticle: http://pastebin.com/wwzRiAut

Basically either run the code as:
1- Non clustered: java -jar ....
2- Clsutered: java  -Dvertx.hazelcast.config=cluster.xml -jar ... -cluster -cluster-host xxx.xxx.xxx

Check the performance between the two commands.

tsegismont:
When deploying verticles across the cluster, it can be useful to force delivery to a local consumer.

For example, you might deploy audit verticles on all nodes, with local consumers only.
Forcing local delivery avoids paying the price of a cluster manager lookup.

This fixes #1764

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/DeliveryOptions.java
+++ b/src/main/java/io/vertx/core/eventbus/DeliveryOptions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2018 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/eventbus/DeliveryOptions.java
```diff
@@ -188,4 +197,25 @@ private void checkHeaders() {
...
+   */
+  public boolean isLocalOnly() {
+    return localOnly;
+  }
+
+  /**
+   * Whether a message should be delivered to local consumers only. Defaults to {@code false}.
+   *
+   * <p>
+   * <strong>This option is effective in clustered mode only and does not apply to reply messages</strong>.
```
vietj: I think you should have use <blink> instead

vietj:
do we need to update the section in the eventbus doc ?

src/main/java/io/vertx/core/eventbus/DeliveryOptions.java
tsegismont: lol

tsegismont:
>do we need to update the section in the eventbus doc ?

I don't believe so: it's an advanced topic and local consumers are not covered either.

Issue #2493, opened 2018-06-10
DnsException improvements


vietj:
- shall not capture stack trace
- provide a meaningful to string

PR #2570, opened 2018-08-02
DnsException improvements #2493

KowalczykBartek:
With this PR I wanted to provide fix for https://github.com/eclipse/vert.x/issues/2493

### DIFF
--- a/src/main/java/io/vertx/core/dns/DnsException.java
+++ b/src/main/java/io/vertx/core/dns/DnsException.java
@@ -11,6 +11,7 @@
 
 package io.vertx.core.dns;
 
+import io.vertx.core.impl.NoStackTraceThrowable;
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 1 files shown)


vietj:
what's the motivation ?

KowalczykBartek:
Oops, sorry :D updated.

vietj:
right, I forgot it :-), thanks for the quickwin

vietj:
for the stacktrace I think we want to modify the exception such as it does not actually capture a stack trace, the easy way to achieve is to extend rather `NoStackTraceThrowable` class

KowalczykBartek:
ok, true, setting empty as stacktrace was a little bit nasty :D I updated code according to your suggestion. One thing, it is now not instance of Exception, is it ok ?

src/main/java/io/vertx/core/dns/DnsException.java
```diff
@@ -19,11 +20,13 @@
...
  * @author <a href="mailto:nmaurer@redhat.com">Norman Maurer</a>
  */
-public final class DnsException extends Exception {
+public final class DnsException extends NoStackTraceThrowable {
 
+  private static final String ERROR_MESSAGE_TEMPLATE = "DNS query error occurred: %s";
   private DnsResponseCode code;
 
   public DnsException(DnsResponseCode code) {
+    super(String.format(ERROR_MESSAGE_TEMPLATE, code));
```
vietj: just use string concat instead of format

PR #2573, opened 2018-08-06
Add missing @Fluent annotations to HttpServerResponse methods

DavidGregory084:
Adds missing fluent annotations to some methods in HttpServerResponse. It seems like without these the methods have `MethodKind.OTHER` when they are processed by vertx-codegen due to the condition [here](https://github.com/vert-x3/vertx-codegen/blob/master/src/main/java/io/vertx/codegen/ClassModel.java#L868).


vietj:
thanks!

PR #2589, opened 2018-08-14
Use simple tracer in continuation POC

pavolloffay:
Use simple tracer implementation in instrumentation tests.

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

### DIFF
--- a/src/test/java/io/vertx/test/core/instrumentation/MetricsTest.java
+++ b/src/test/java/io/vertx/test/core/instrumentation/MetricsTest.java
@@ -10,7 +10,6 @@
  */package io.vertx.test.core.instrumentation;
 
 import io.vertx.core.Handler;
-import io.vertx.core.MultiMap;
 import io.vertx.core.VertxOptions;
 import io.vertx.core.http.*;
 import io.vertx.core.metrics.MetricsOptions;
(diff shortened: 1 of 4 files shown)


src/test/java/io/vertx/test/core/instrumentation/MetricsTest.java
```diff
@@ -205,16 +167,25 @@ public void testHttpServerRequest() throws Exception {
     awaitLatch(latch);
     HttpClientRequest req = client.get(8080, "localhost", "/1", resp -> {
       assertEquals(200, resp.statusCode());
+      List<Span> finishedSpans = tracer.getFinishedSpans();
+      // client request to /1, server request /1, client request /2, server request /2
+      assertEquals(4, finishedSpans.size());
+      assertOneTrace(finishedSpans);
```
pavolloffay: @vietj now it breaks here.

The metrics impl looks suspicious, when something is stored in thread local (scope created) it should be removed once processing is done - e.g. request is finished.

vietj: agreed, that's part of the thing we should try to improve

Issue #2560, opened 2018-07-26
RecordParser should have an option to limit the record size


richardf:
When using RecordParser's delimited mode, it would be nice to somehow limit the maximum buffer size, to prevent misbehaving/malicious clients from exhausting server resources.

This could be achieved by exposing the buffer length, like so:
```java
RecordParser rp = RecordParser.newDelimited("\n", System.out::println);
socket.handler(buffer -> {
   rp.handle(buffer);
   if (rp.getBufferSize() > config().getInteger("maxRecordSize")) {
      System.out.println("big record, eh?");
      socket.close();
   }
});
```

Another option would be to check the record size in RecordParserImpl.parseDelimited() and throw an exception, that could be handled inside the exceptionHandler:
```java
RecordParser rp = RecordParser.newDelimited("\n", System.out::println);
rp.setMaxRecordSize(10);
rp.exceptionHandler(throwable -> {
   System.out.println("big record, eh?");
   socket.close();
});
```

Thoughts?

tsegismont:
Option /2 sounds good to me. Netty's line based frame decoder does not even let you create an instance without providing a max length for frames.

richardf:
Thanks for the feedback. Going to work on a PR

PR #2611, opened 2018-09-03
RecordParser: Limit the record size

richardf:
Implements #2560.
If the record is too long, the RecordParser's exception handler will be called. After detecting a record whose size exceeds the maxRecordSize, the implementation will discard any data until a delimiter is found.


src/main/java/io/vertx/core/parsetools/impl/RecordParserImpl.java
```diff
@@ -152,12 +155,33 @@ public void fixedSizeMode(int size) {
...
 
+  /**
+   * Set the maximum allowed size for a record when using the delimited mode.
+   * The delimiter itself does not count for the record size.
+   * <p>
+   * If a record is longer than specified, a RecordTooLongException will be thrown.
+   *
+   * @param size the maximum record size
+   */
+  public void maxRecordSize(int size) {
```
vietj: make it fluent

vietj:
I think we should do a simpler implementation, i.e instead of checking the boundary in `handleParsing ` and `parseDelimited` method, we should simply check that the remaining buffer size is lesser than the max size after `handleParsing` is called in the `handle` method.

richardf:
@vietj The downside of this approach is that it will be difficult to resume the data ingestion after detecting a record that exceeds maxRecordSize. For my use case it is ok since I'll just close the socket and let the client reconnect. Wanna go that route?

richardf:
@vietj updated as discussed.
The ip-validation failed because I committed from a different machine with other git configs. How do I fix that?

vietj:
@richardf make a new clean PR from a clean commit

richardf:
done. created #2635

vietj:
thanks

PR #2613, opened 2018-09-04
Adding metrics SPI methods for the event loop

marcust:
Continuation of https://github.com/eclipse/vert.x/pull/2355 because of employer change

Signed-off-by: Marcus Thiesen <mt@hubrick.com>


vietj:
instead of this, we will provide a new callback to the VertxMetrics SPI called when the vertx instance has been initialised. This allows to get the `EventLoopGroup` from the instance and obtain the event loops in the same way. See https://github.com/eclipse-vertx/vert.x/issues/2695

Issue #2600, opened 2018-08-23
Support for proxy hosts black listing in ProxyOptions


haithkris:
Hi, we have this issue when we want to ignore the proxy and when it comes to call certain components of our internal network. Since vertx ignores JVM proxy arguments like `http.nonProxyHosts` there is no real workaround.

The proxy configuration is set at the `WebClient` level. ProxyOptions object should have an equivalent to `http.nonProxyHosts`.

Vertx web version: 3.5.1
Vertx core version: 3.5.1

vietj:
is that something you can contribute ?

haithkris:
I would if had proper time and proper knowledge on how vertx works in the internal. But I was hoping you guys can pull it through.

vietj:
ok, i'm keeping this as feature request

marcottedan:
Actually I believe it's more "whitelisting" than "blacklisting" since the noProxy feature of the JVM (and docker network has the same feature) is that everything goes to the proxy except the `noProxy` hosts.

KowalczykBartek:
Hi, I crafted something today, 
of course there is a lot of to do in this PR still (only HttpClientImpl covered), but I wanted to ask what do you think about API, 
cc  @haithkris
```java
new ProxyOptions()
  .setType(ProxyType.HTTP)
  .exludeHost("localhost") //
  .exludeHost("localhost2") //
  .setHost("google.com")
  .setPort(proxy.getPort());
```

haithkris:
looks good for me!

vietj:
the setter should be `setExcludedHosts(List<String>)` and there should be a method `addExcludedHost(String)` that adds element to the list.

Look at how other options handle this.

KowalczykBartek:
What with NetClient ? should ProxyOptions include option to exclude SocketAddress ?
```java
NetClient connect(SocketAddress remoteAddress, Handler<AsyncResult<NetSocket>> connectHandler);
```

vietj:
that's a concern I had indeed, but ProxyOptions don't support `@VertxGen` so I think the best would be to have the excluded host to begin with `/` instead and we should parse them accordingly

KowalczykBartek:
> so I think the best would be to have the excluded host to begin with / instead and we should parse them accordingly when we build the proxy

can you explain it a little bit ?

vietj:
well actually I don't think we need to do that because one can still create 2 net clients if he wants to use domain sockets.

[...]

PR #2614, opened 2018-09-04
Allow to exclude host from being proxying #2600

KowalczykBartek:
Initial PR for #2600

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpChannelConnector.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpChannelConnector.java
@@ -27,6 +27,7 @@ import io.vertx.core.http.impl.pool.ConnectResult;
 import io.vertx.core.http.impl.pool.ConnectionListener;
 import io.vertx.core.http.impl.pool.ConnectionProvider;
 import io.vertx.core.impl.ContextImpl;
+import io.vertx.core.net.ProxyOptions;
 import io.vertx.core.net.ProxyType;
 import io.vertx.core.net.SocketAddress;
 import io.vertx.core.net.impl.ChannelProvider;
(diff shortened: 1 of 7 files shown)


src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -1030,6 +1030,17 @@ private HttpClientRequest createRequest(HttpMethod method, String protocol, Stri
     return req;
   }
 
+  private boolean useProxy(boolean isSsl, ProxyOptions proxyOptions, String targetHost) {
```
xkr47: This functionality is duplicated in `HttpChannelConnector.doConnect()` that you also updated. Please move this helper method to e.g. a new class `src/main/java/io/vertx/core/http/impl/ProxyHelper.java`. Then reuse that helper in `doConnect()` as well. (You'll have to extract the proxy type frokm the proxyOptions instead, but that's fine.)

src/main/java/io/vertx/core/http/impl/HttpChannelConnector.java
```diff
@@ -129,7 +137,7 @@ private void doConnect(
 
     ChannelProvider channelProvider;
     // http proxy requests are handled in HttpClientImpl, everything else can use netty proxy handler
-    if (options.getProxyOptions() == null || !ssl && options.getProxyOptions().getType()== ProxyType.HTTP ) {
+    if (options.getProxyOptions() == null || !ssl && options.getProxyOptions().getType() == ProxyType.HTTP || isExcluded(options.getProxyOptions(), host)) {
```
xkr47: See comment for `HttpClientImpl.useProxy()`.

src/main/java/io/vertx/core/net/ProxyOptions.java
```diff
@@ -46,6 +46,7 @@
   private String username;
   private String password;
   private ProxyType type;
+  private Set<String> exludedHosts;
```
xkr47: Typo - please rename to `excludedHosts`.

src/main/java/io/vertx/core/http/impl/HttpChannelConnector.java
```diff
@@ -116,6 +117,13 @@ public void deactivate(HttpClientConnection conn) {
     }
   }
 
+  private boolean isExcluded(ProxyOptions options, String targetProxy) {
```
xkr47: Remove this as it's no longer needed; see comment for `HttpClientImpl.useProxy()`.

KowalczykBartek:
Thanks @xkr47 for the review - unfortunately, this PR will never be merged - I will close it to note confuse anyone .

PR #2623, opened 2018-09-13
Fixes #2619: properly handle plus signs when resolving paths

### LINKED ISSUE #2619: FileResolver fails when path contains `+` character
### Version

* vert.x core: 3.5.3

### Context

` io.vertx.core.impl.FileResolver` fails to extract content from JARs containing `+` in a file name (e.g. `vertx-static-content-0.0.1+test-fat.jar`).
Most likely `java.net.URLDecoder.decode(String, String)` replaces `+` with ` `  (space)

```
java -jar vertx-static-content-0.0.1+test-fat.jar
io.vertx.core.Starter
INFO: Succeeded in deploying verticle
io.vertx.ext.web.impl.RoutingContextImplBase
SEVERE: Unexpected exception in route
io.vertx.core.VertxException: java.io.FileNotFoundException: /home/user/workspace/vertx-static-content/target/vertx-static-content-0.0.1 test-fat.jar
        at io.vertx.core.impl.FileResolver.unpackFromJarURL(FileResolver.java:249)
        at io.vertx.core.impl.FileResolver.unpackUrlResource(FileResolver.java:153)
        at io.vertx.core.impl.FileResolver.resolveFile(FileResolver.java:134)
        at io.vertx.core.impl.VertxImpl.resolveFile(VertxImpl.java:776)
        at io.vertx.core.http.impl.HttpServerResponseImpl.doSendFile(HttpServerResponseImpl.java:432)
        at io.vertx.core.http.impl.HttpServerResponseImpl.sendFile(HttpServerResponseImpl.java:352)
        at io.vertx.core.http.HttpServerResponse.sendFile(HttpServerResponse.java:327)
        at io.vertx.core.http.HttpServerResponse.sendFile(HttpServerResponse.java:313)
        at test.Web.lambda$start$1(Web.java:15)
```

### Do you have a reproducer?

https://github.com/zaytsev/vertx-static-content

pmlopes:
Fixes #2619 

And the new utility is exposed on the public package `io.vertx.net` so it can be reused from `vertx-web` where it originally was implemented.


vietj:
- I'm wondering why there are so few tests for code that does not seem trivial ?
- can you move this in `net.impl` package I would like avoiding to supporting it in API as it does not seem required

pmlopes:
The reason it's outside implementation is because I'd like to use it on Web where it originally came from. Avoiding 2 identical implementations.

 I can port the tests from Web too.

vietj:
I think vertx web can use it from an impl package without a problem.

please port all tests to avoid breakage.

src/main/java/io/vertx/core/net/impl/URIDecoder.java
```diff
@@ -0,0 +1,126 @@
...
+              + " of: " + s);
+        }
+        c = (char) (c * 16 + c2);
+        // shouldn't check for plus since it would be a double decoding
+        buf[pos++] = (byte) c;
+      } else {
+        buf[pos++] = (byte) (plus && c == '+' ? ' ' : c);
+      }
+    }
+    return new String(buf, 0, pos, CharsetUtil.UTF_8);
```
vietj: Use `java.nio.charset.StandardCharsets` instead of `CharsetUtil`

src/main/java/io/vertx/core/net/impl/URIDecoder.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    final int size = s.length();
+    boolean modified = false;
+    int i;
+    for (i = 0; i < size; i++) {
+      final char c = s.charAt(i);
+      if (c == '%' || (plus && c == '+')) {
+        modified = true;
+        break;
```
vietj: just `return s`, the modified flag is unnecessary

src/main/java/io/vertx/core/net/impl/URIDecoder.java
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+    if (!modified) {
+      return s;
+    }
+    final byte[] buf = s.getBytes();
```
vietj: Use an encoding when getting the bytes otherwise it will use the default platform charset

src/main/java/io/vertx/core/net/impl/URIDecoder.java
```diff
@@ -0,0 +1,126 @@
...
+    int pos = i;  // position in `buf'.
+    for (; i < size; i++) {
+      char c = s.charAt(i);
+      if (c == '%') {
+        if (i == size - 1) {
```
vietj: use `>=` as safety guard instead of `==`

src/main/java/io/vertx/core/net/impl/URIDecoder.java
vietj: Actually since we only write in buff, we should create a `byte[]`, using `getBytes()` will result in unnecessary CPU usage.

src/main/java/io/vertx/core/net/impl/URIDecoder.java
pmlopes: the loop is inverted, if the flag is `true` means we're not on the happy path and we need to start deciding from the current `i`

src/main/java/io/vertx/core/net/impl/URIDecoder.java
pmlopes: I reuse the generated `byte[]` from the `i` where we detected the first `%` char. So I could either create a new `byte[]` and loop over the string to copy bytes (and i need bytes not chars) or let this do some extra work by copying the bytes and waste the decoded bytes from `i` to the end...

src/main/java/io/vertx/core/net/impl/URIDecoder.java
vietj: oh right :-)

PR #2626, opened 2018-09-14
Add PKCS#11 like Java Keystore

aicfr:


vietj:
I think we are missing tests for this, please look at the `ssl.txt` file and how to create pkcs11 version of a store

vietj:
we also need corresponding tests with this.

vietj:
can we have a test that does not use a native library ?

aicfr:
> can we have a test that does not use a native library ?

I think is not posiible with PKCS11

pmlopes:
Just a question? Why pcks11? Since java9 if I'm not mistaken, the jdk standard is pcks12 which requires no sun classes or native libs. Why not just tell users how to migrate to pcks12? As that already works out of the box?

vietj:
missing test is real issue and as @pmlopes is there a real benefit ? this could be added externally since we do support pluggable TrustOption / KeyCertOption

pmlopes:
Pcks11 are hardware stores, sun / oracle jdk supports them but if I'm not mistaken, most of the work is configuration, and a keystore only requires a password to unlock it. So why not just ensure that the jks options have a way to specify the keystore format? Wouldn't that suffice instead of a specific method and locking to sun / oracle jdk?

PR #2635, opened 2018-09-23
RecordParser: Limit the record size

richardf:
A simple implementation of  #2560.


src/main/java/io/vertx/core/parsetools/RecordTooLongException.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.parsetools;
+public class RecordTooLongException extends RuntimeException {
```
vietj: can you just use a `VertxException` instead ?

vietj: or better an `IllegalStateException`

richardf: Sure

vietj:
back-port to 3.5.4

PR #2639, opened 2018-09-24
Test lock release twice

meggarr:
When an async lock is acquired, and release multiple
    times, there should only be one pending lock request
    be acquired. In some implementation, the multiple release
    will make multiple pending lock requests be all acquired.

    Please check detail in https://github.com/vert-x3/vertx-hazelcast/issues/94

    Signed-off-by: Richard MENG <mengangr@foxmail.com>


meggarr:
@tsegismont

PR #2652, opened 2018-10-04
For internal keystores prefer PKCS12 (industry standard) instead of p…

pmlopes:
…roprietary JKS, this is compatible with GraalVM


pmlopes:
`PCKS12` also has the benefit of being more secure, for more info: http://openjdk.java.net/jeps/229

PR #2656, opened 2018-10-08
Support retrieving X509 certificates CN since javax.naming package is unavailable in Android

hjames9:
…Support retrieving X509 certificates CN if javax.naming package is unavailable.  Android does not contain this package.

Signed-off-by: Hayden James <hayden.james@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
+++ b/src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
@@ -11,6 +11,7 @@
 
 package io.vertx.core.net.impl;
 
+import io.netty.util.internal.PlatformDependent;
 import io.vertx.core.VertxException;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.impl.VertxInternal;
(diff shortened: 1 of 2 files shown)


vietj:
have you tried running Vert.x on Android ? is there a way we could test this ?

hjames9:
Yes, I've run this against Android with TLS connections and X509 certificates.  That's how I initially ran into the problem and subsequently worked around the issue with this change.

vietj:
which android os did you run with ? I'm thinking we should abstract these in a platform dependent class similar to what netty has

hjames9:
I'm using Android 8.1 and also Android Things 1.0.  API 27 on both systems.  However the change should work on every Android version as the javax.naming.* package never existed in any version.  Let me know if you need application logs or any other data.

hjames9:
Also, I can do the change how you'd like, but it's written in a way not to be Android specific.  However I can understand the urge to do it that way.

vietj:
can we avoid this code by determining we are on android with loading a class ? this will not work in native vertx

vietj:
try to use Netty's `PlatformDependant#isAndroid()` static method

hjames9:
@vietj Updated and tested the code using PlatformDependant.isAndroid() instead of checking for class existence.  Code change is now smaller and works just as well.

vietj:
that was me speaking with @vertx-ci

hjames9:
Yes, will do.  Change coming shortly

hjames9:
@vietj Refactored into a separate function.  Should be a lot cleaner and concise now.  Thanks for the feedback!

PR #2661, opened 2018-10-11
Create additional write interface to NetSocket for notifications after socket writes are executed

hjames9:
Add writeHandler interface to NetSocket for notifications after socket writes are executed.  Since socket writes are asynchronous, this allows notifications for when they actually occur.

Signed-off-by: Hayden James <hayden.james@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/VertxHttp2ConnectionHandler.java
+++ b/src/main/java/io/vertx/core/http/impl/VertxHttp2ConnectionHandler.java
@@ -203,19 +203,29 @@ class VertxHttp2ConnectionHandler<C extends Http2ConnectionBase> extends Http2Co
     encoder().writeHeaders(chctx, stream.id(), headers, 0, end, chctx.newPromise());
   }
 
-  void writeData(Http2Stream stream, ByteBuf chunk, boolean end) {
+  void writeData(Http2Stream stream, ByteBuf chunk, boolean end, Handler<AsyncResult<Void>> handler) {
     EventExecutor executor = chctx.executor();
+    ChannelPromise promise = chctx.newPromise();
+    if (handler != null) {
+      promise.addListener((future) -> {
+        if(future.isSuccess()) {
+          handler.handle(Future.succeededFuture());
+        } else {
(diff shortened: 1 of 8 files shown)


vietj:
actually what we want is different, we want to overload the write(Buffer) method with a write(Buffer, Handler<AsyncResult<Void>>) method and the handler is called when the buffer has been sent. This callback might be delayed when the buffer is full. We also need to test that when the buffer is not delivered and the socket is closed then the callback happens with a failure.

vietj:
Basically as it is, it is not useful because the application that is writing to the NetSocket is itself aware that it wrote something, so the user can eventually make a NetSocket wrapper that will result in the same effect.

src/main/java/io/vertx/core/http/impl/VertxHttp2ConnectionHandler.java
```diff
@@ -205,17 +205,37 @@ private void _writeHeaders(Http2Stream stream, Http2Headers headers, boolean end
...
+    if (executor.inEventLoop()) {
+      _writeData(stream, chunk, end, promise);
+    } else {
+      executor.execute(() -> {
+        _writeData(stream, chunk, end, promise);
+      });
+    }
+  }
+
+  void writeData(Http2Stream stream, ByteBuf chunk, boolean end, Handler<AsyncResult<Void>> handler) {
```
vietj: you can keep a single `writeData` method and add the handler as listener of the promise when it's not null and remove `void writeData(Http2Stream stream, ByteBuf chunk, boolean end)`

vietj:
we need more testing:

- test a write failure
   - the client connects and pause the socket
   - the server writes until writeQueueFull is true
   - the server writes one extra buffer (we are sure it's queued)
   - the client closes the socket abruptly
   - the server gets a callback that the write failed
- test also for HTTP/2 that has its implementation using a multiplexed HTTP/2 connection

hjames9:
@vietj I've covered the test cases you've suggested.  Please take a look.  Thanks!

hjames9:
@vietj In case you missed this due to the Github outage

Issue #2662, opened 2018-10-11
Websocket client HTTP header response callback


swinterberger:
Would the vert.x team be open to a PR which exposes yet another callback to be passed into httpClient.websocket that allows a user to receive the HTTP headers returned by the sever after the websocket handshake succeeds?

The issue is that we load balance on the server end based on a cookie, the cookie is auto generated by HAProxy when not present in a request and it includes the name of the backend node the request should route to.  We need this cookie for session identity because we occasionally open multiple websocket channels from a single client.  It makes our lives easier if secondary websocket channels can be forwarded to the same backend node as the original, or if we have to reconnect the initial channel.

I was unable to get this information back even via reflection because it is quickly discarded so I had to fork and modify the code to add a handler.  Since I see no other issues related to this, maybe we are an oddity therefore I am curious if the PR would be welcome.  Feel free to close if not and we will maintain our fork.

swinterberger:
@vietj that was actually the first way I had it but wasnt sure if keeping that data attached to the WebSocketBase indefinitely was ideal.  I will switch it back to a getter and I have to add some unit tests.  

I will try to find time to wrap that up today or over the weekend.

vietj:
@swinterberger just make it present during the Handler<AsyncResult<WebSocket>> callback and drop the data after, say it in the javadoc. It's fine.

vietj:
response headers are now provided on `WebSocketBase`

PR #2667, opened 2018-10-13
Websocket Client HTTP response headers

swinterberger:
expose http response headers to websocketBase during the connection handler.  Clear headers after handler to prevent resource useage when the user does not care about the headers

#2662

### DIFF
--- a/src/main/java/io/vertx/core/http/WebSocketBase.java
+++ b/src/main/java/io/vertx/core/http/WebSocketBase.java
@@ -13,6 +13,7 @@ package io.vertx.core.http;
 
 import io.vertx.codegen.annotations.*;
 import io.vertx.core.Handler;
+import io.vertx.core.MultiMap;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.net.SocketAddress;
 import io.vertx.core.streams.ReadStream;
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -90,6 +91,15 @@
    */
   String subProtocol();
 
+  /**
+   *  Returns the HTTP response headers during the websocket connection handler.
+   *  <p/>
+   *  After the connection handler has completed the response headers will be null.
```
vietj: _After the completion handler callback has completed the response headers will be {@code null}_

src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -90,6 +91,15 @@
...
+   *
+   * @return the response headers
+   */
+  MultiMap responseHeaders();
```
vietj: call it `headers()` instead and provide also the request headers on the server, that is surely useful

swinterberger:
Regarding the null header after callback test, that is the point of the assertNull after the await.  Not sure how else to better verify it is nullified after handler.

`    assertNull(websocketRef.get().responseHeaders());
`

vietj:
you can do something like in the handler:

vertx.runOnContext(v -> {
   assertNull(websocket.responseHeaders());
   testComplete();
});

PR #2668, opened 2018-10-14
Websocket client response headers

swinterberger:
expose http response headers to websocketBase during the connection handler. Clear headers after handler to prevent resource useage when the user does not care about the headers


vietj:
@swinterberger can you submit a new PR that can pass IP validation (requires you to sign off commit and accept CLA)

PR #2694, opened 2018-11-02
Port the utils removeDots and normalizePath from web to core, as they…

pmlopes:
… are usefull not just to web routing but HTTP in general

Follow up from: https://github.com/vert-x3/vertx-web/pull/1040

Once this is merged, the Utils on web can be clean as most of it's code is now on core.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpUtils.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpUtils.java
@@ -82,6 +82,85 @@ public final class HttpUtils {
   }
 
   /**
+   * Normalizes a path as per <a href="http://tools.ietf.org/html/rfc3986#section-5.2.4>rfc3986</a>.
+   *
+   * There are 2 extra transformations that are not part of the spec but kept for backwards compatibility:
+   *
+   * double slash // will be converted to single slash and the path will always start with slash.
+   *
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/http/impl/HttpUtils.java
```diff
@@ -81,6 +81,85 @@ private static boolean matches(CharSequence path, int start, String what, boolea
...
+   * Normalizes a path as per <a href="http://tools.ietf.org/html/rfc3986#section-5.2.4>rfc3986</a>.
+   *
+   * There are 2 extra transformations that are not part of the spec but kept for backwards compatibility:
+   *
+   * double slash // will be converted to single slash and the path will always start with slash.
+   *
+   * @param pathname raw path
+   * @return normalized path
+   */
+  public static String normalizePath(String pathname) {
```
vietj: I think we should actually disallow `null` here

pmlopes: `null` paths will return `null` and docs are updated stating that nothing can be assumed about null so they stay null.

vietj: ok, anyway this is internal

Issue #2749, opened 2018-12-04
empty vertx server metrics with http2


jo5ef:
When using dropwizard metrics and vertx 3.6.0, the vertx server metrics are empty (values are 0). The following test *passes* using vertx 3.5.4 and *fails* with vertx 3.6.0. It also passes with vertx 3.6.0 when switching the protocol version to `HTTP_1_1`.

```
@RunWith(VertxUnitRunner.class)
public class TestMetrics {

  private static final String SERVER_REQUEST_METRIC = "vertx.http.servers.0.0.0.0:0.requests";
  private Vertx vertx;

  @Before
  public void setUp() {
    vertx = Vertx.vertx(new VertxOptions()
      .setMetricsOptions(new DropwizardMetricsOptions()
        .setEnabled(true)));
  }

  @Test
  public void testMetric(TestContext ctx) {
    startServer()
      .compose(this::makeRequest)
      .map(r -> {

        var m = MetricsService.create(vertx)
          .getMetricsSnapshot(SERVER_REQUEST_METRIC)
          .getJsonObject(SERVER_REQUEST_METRIC);

        ctx.assertEquals(1, m.getInteger("count"));  // check request was tracked
        ctx.assertTrue(m.getDouble("max") > 0);

        return r;
      })
      .setHandler(ctx.asyncAssertSuccess());
  }

  private Future<Integer> startServer() {
    var f = Future.<HttpServer>future();
    vertx.createHttpServer(new HttpServerOptions())
      .requestHandler(r -> r.response().end())
      .listen(0, f);
    return f.map(HttpServer::actualPort);
  }

  private Future<Void> makeRequest(int port) {
    var f = Future.<Void>future();
    vertx.createHttpClient(new HttpClientOptions()
      .setDefaultPort(port)
      .setProtocolVersion(HttpVersion.HTTP_2))
      .getNow("/", r -> f.complete());
    return f;
  }
}
```

It seems that this is due to the `failed` flag passed to `Http2ServerResponseImpl.handleEnded`.

vietj:
can you update the test to be correct ?

jo5ef:
How do you mean?
I wrote this test to demonstrate that metrics no longer work with HTTP2 & vertx 3.6.0. Changing this test won't bring back our metrics : )

vietj:
ah sorry it's a new test, I will use it for a fix then :-)

vietj:
so this uses dropwizard metrics right ?

[...]

PR #2751, opened 2018-12-04
fix empty vertx server metrics with http2

jo5ef:
I've described the problem in issue #2749. This is my temporary quick fix that works for us for now, but I haven't changed/added tests and don't have a deep understanding of the new vertx http code in general.

### DIFF
--- a/src/test/java/io/vertx/core/http/HttpMetricsTestBase.java
+++ b/src/test/java/io/vertx/core/http/HttpMetricsTestBase.java
@@ -237,7 +237,7 @@ public abstract class HttpMetricsTestBase extends HttpTestBase {
       assertFalse(metric.failed.get());
       req.response().closeHandler(v -> {
         assertNull(metrics.getMetric(req));
-        assertFalse(metric.failed.get());
+        assertEquals(protocol == HttpVersion.HTTP_1_1, metric.failed.get());
         testComplete();
       });
     });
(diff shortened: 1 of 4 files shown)


src/test/java/io/vertx/core/http/HttpMetricsTestBase.java
```diff
@@ -227,6 +228,7 @@ public void testClientConnectionClosed() throws Exception {
   }
 
   @Test
+  @Ignore
```
vietj: why is that necessary ?

vietj:
can you add a metrics test too ?

jo5ef:
I think the existing tests are good, I re-enabled & fixed the one that was failing due to my change.

src/test/java/io/vertx/core/http/HttpMetricsTestBase.java
jo5ef: re-enabled & fixed now, see latest commit

vietj:
I don't see what should have failed in the tests that your patch fixes, can you elaborate ?

jo5ef:
The `HttpServerResponse.closeHandler` behaves differently for HTTP/1.x and HTTP/2. For HTTP/1.x it is called when the connection is closed, for HTTP/2 it is called when the stream is closed, i.e. it is called for *every* request.

The test uses a `closeHandler` to assert hat the metric has `failed`, which is correct for HTTP/1.x. For HTTP/2 however this is not correct, otherwise there would never be any metrics (reason for my issue & PR). Since this test is in a base class for both `Http2MetricsTest` and `Http1xMetricsTest`, it failed for HTTP/2.

I made another change to improve the test to account for both cases.

jo5ef:
Works again after the changes in the http code. Tested with 3.6.3. Thanks!

PR #2754, opened 2018-12-05
JsonObject cannot map to class with Instant or byte array field

### LINKED ISSUE #2753: JsonObject cannot map to class with Instant or byte array field
The Jackson mapper is configured to serialize `Instant` and `byte` arrays properly (required for `io.vertx.core.json.JsonObject#mapFrom` method).

But it's not configured to deserialize properly (in method `io.vertx.core.json.JsonObject#mapTo`).

tsegismont:
Fixes #2753

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -13,16 +13,18 @@ package io.vertx.core.json;
 
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.exc.InvalidFormatException;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import io.netty.buffer.ByteBufInputStream;
 import io.vertx.core.buffer.Buffer;
 
-import java.io.DataInput;
 import java.io.IOException;
 import java.io.InputStream;
 import java.math.BigDecimal;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.util.Base64;
 import java.util.Iterator;
@@ -54,7 +56,9 @@ public class Json {
     module.addSerializer(JsonArray.class, new JsonArraySerializer());
     // he have 2 extensions: RFC-7493
     module.addSerializer(Instant.class, new InstantSerializer());
+    module.addDeserializer(Instant.class, new InstantDeserializer());
     module.addSerializer(byte[].class, new ByteArraySerializer());
+    module.addDeserializer(byte[].class, new ByteArrayDeserializer());
 
     mapper.registerModule(module);
     prettyMapper.registerModule(module);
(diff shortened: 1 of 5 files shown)


pmlopes:
@tsegismont sorry, but I think this is more or less the same as https://github.com/eclipse-vertx/vert.x/pull/2507 can you verify?

tsegismont:
@pmlopes it looks similar indeed. Do you want me to close this one and merge the work in the other PR?

I'm not sure why the other one was not merged. I read concerns about deserializing to `JsonObject`, but that wouldn't happen, as the deserializers will only be used when trying to map to a custom class, not `JsonObject` or `JsonArray`.

tsegismont:
I've incorporated @EmadAlblueshi 's work (thanks again buddy!). So we have now something consistent for both `Instant` and byte arrays.

@pmlopes Can you please review? Thanks!

src/main/java/io/vertx/core/json/Json.java
```diff
@@ -239,12 +243,36 @@ public void serialize(Instant value, JsonGenerator jgen, SerializerProvider prov
...
+
+  private static class ByteArrayDeserializer extends JsonDeserializer<byte[]> {
+
+    @Override
+    public byte[] deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+      String text = p.getText();
+      try {
+        return Base64.getDecoder().decode(text);
+      } catch (IllegalArgumentException e) {
+        throw new InvalidFormatException(p, "Expected a base64 encoded byte array", text, Instant.class);
```
EmadAlblueshi: Why Instant.class ?

EmadAlblueshi:
Thank you @tsegismont I really appreciate it :)

src/main/java/io/vertx/core/json/Json.java
tsegismont: This is what the Jackson `InvalidFormatException` expects. It helps give the user better failure messages.

tsegismont:
@pmlopes fyi, the PR build passed twice (as I was trying to keep @EmadAlblueshi authorship in commits without success and pushed again and again). But just failed for another reason.

src/main/java/io/vertx/core/json/Json.java
```diff
@@ -239,12 +243,36 @@ public void serialize(Instant value, JsonGenerator jgen, SerializerProvider prov
...
+    }
+  }
+
   private static class ByteArraySerializer extends JsonSerializer<byte[]> {
-    private final Base64.Encoder BASE64 = Base64.getEncoder();
 
     @Override
     public void serialize(byte[] value, JsonGenerator jgen, SerializerProvider provider) throws IOException {
-      jgen.writeString(BASE64.encodeToString(value));
+      jgen.writeString(Base64.getEncoder().encodeToString(value));
```
pmlopes: This is wrong according to: https://tools.ietf.org/html/rfc7493#section-4.4 it should be: https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html#getUrlEncoder--

src/main/java/io/vertx/core/json/Json.java
```diff
@@ -239,12 +243,36 @@ public void serialize(Instant value, JsonGenerator jgen, SerializerProvider prov
...
+    }
+  }
+
+  private static class ByteArrayDeserializer extends JsonDeserializer<byte[]> {
+
+    @Override
+    public byte[] deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+      String text = p.getText();
+      try {
+        return Base64.getDecoder().decode(text);
```
pmlopes: Same here it should be the url/filename decoder.

tsegismont:
As discussed off-list with @pmlopes , I am going to merge this PR.

It is true that the RFC recommends to use `base64url` encoding, but this is a breaking change and we want this code to be backported to 3.6.

As a consequence, after this PR merged and the code backported to 3.6 branch, I will create new issues in `core` and `codegen` to switch to `base64url` encoding. I will also update the breaking changes page.

EmadAlblueshi:
Thanks 👍

tsegismont:
You're welcome!

Issue #2756, opened 2018-12-07
Write buffer text message


YongGang:
In `WebSocketImplBase`, the `writeTextMessage` method accept a String as input, then convert it to `Buffer` internally.
I'd like to add a method which accept `Buffer` data but send out as `FrameType.TEXT`
Something like the following:
```
  @Override
  public S writeTextMessage(Buffer textBuffer) {
    synchronized (conn) {
      checkClosed();
      writePartialMessage(FrameType.TEXT, textBuffer, 0);
      return (S) this;
    }
  }
```
It's for the performance, in our case we already have a `Buffer`, it will be wasted to convert it to String then Vertx convert the String to Buffer again before sending out.

vietj:
feel free to make a PR with a test @YongGang

PR #2758, opened 2018-12-10
Write buffer text message #2756

YongGang:
It's for performance. In our case we already have a `Buffer`, it will be wasted to convert it to String then Vertx convert the String to Buffer again before sending out.

### DIFF
--- a/src/main/java/io/vertx/core/http/WebSocketBase.java
+++ b/src/main/java/io/vertx/core/http/WebSocketBase.java
@@ -138,6 +138,16 @@ public interface WebSocketBase extends ReadStream<Buffer>, WriteStream<Buffer> {
   WebSocketBase writeTextMessage(String text);
 
   /**
+   * Writes a (potentially large) piece of binary text data to the connection.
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -137,6 +137,16 @@
...
 
+  /**
+   * Writes a (potentially large) piece of binary text data to the connection.
+   * This data might be written as multiple frames if it exceeds the maximum WebSocket frame size.
+   *
+   * @param textBuffer  the data to write
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  WebSocketBase writeBinaryTextMessage(Buffer textBuffer);
```
vietj: it's not clear whether it writes binary or test message, can you clarify ? it should either be called writeTextMessage or writeBinaryMessage

YongGang: updated

Issue #2278, opened 2018-01-16
Clarify that classloading isolation not compatible with Java 9 in documentation


Marssmart:
https://github.com/eclipse/vert.x/blob/5eb55e8478fbe9ca1cb14aaef783a40857fee602/src/main/java/io/vertx/core/impl/DeploymentManager.java#L384

This line cause incompatibility with Java 9 , when combined with attempt to use isolation group.

Java 9
![image](https://user-images.githubusercontent.com/12251937/34984971-e4eb61be-fab2-11e7-8974-64d5c4256d8c.png)

Java 8
![image](https://user-images.githubusercontent.com/12251937/34984942-c985b118-fab2-11e7-987b-026bff8fedd2.png)

As can bee seen, 8 v 9 uses different classloader while launching jar. Java 8 one is extending URLClassloader, Java 9 one is **not**

Can be tested by following project

[vertx-test.zip](https://github.com/eclipse/vert.x/files/1634735/vertx-test.zip)

vietj:
yes it's expected and we have no way go fix it, so it should be documented I think in the method.

vietj:
can you take care of this @cescoffier ?

cescoffier:
I will @vietj

cescoffier:
Comment added in the master branch.

nabil-mansouri:
Hi guys!
Do you plan to fix class loader isolation on Java 9+?
@cescoffier  could you reopen this issue? 
Because we could not imagine change our architecture because of that. It seems that spring boot had a similar problem (get urls from classloader to reload an application) and they have fixed it:
https://github.com/spring-projects/spring-boot/blob/c1f7aa801a6c66b38c8b644ce50f05322980a4d2/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/DefaultRestartInitializer.java

It would be very nice if this bug could be fixed. 

Thanks

vietj:
@nabil-mansouri we actually have less classloading in vertx core itself

Isolation groups should be deprecated and moved out from core in 4.0, in a separate module that would be compatible with Java 8 and perhaps might also include Java > 8 hacks / fixes.

PR #2764, opened 2018-12-13
Enable isolation groups for Java 9+

pk-work:
The current implementation of isolation groups [1] requires the
instantiation of a new IsolatingClassLoader, which contains all JARs
from the class path of the SystemClassLoader. Before Java 9, this was
solved by casting the SystemClassLoader to an URLClassLoader to extract
JARs on the class path of the SystemClassLoader.

Due to the fact, that the SystemClassLoader is not longer an
URLClassLoader in Java 9, this approach will not longer work.

This patch implements a way to extract the class path of the
SystemClassLoader which does not require the SystemClassLoader to be an
instance of URLClassLoader.

WARNING: This approach will not work for modules, which were introduces
in Java 9, because modules are living in the "runtime image" [2]. Due to
this, it is not possible to pass the module classes into the current
implementation of IsolatingClassLoader.

[1] https://github.com/eclipse-vertx/vert.x/blob/4bbced9e8ac173ab73c4b4d1ef0bee99c6322a89/src/main/asciidoc/index.adoc#verticle-isolation-groups
[2] https://openjdk.java.net/jeps/220

Signed-off-by: Pascal Krause <pascal.krause@sap.com>

### DIFF
--- a/src/main/java/io/vertx/core/impl/DeploymentManager.java
+++ b/src/main/java/io/vertx/core/impl/DeploymentManager.java
@@ -25,6 +25,7 @@ import io.vertx.core.spi.VerticleFactory;
 import io.vertx.core.spi.metrics.VertxMetrics;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
@@ -32,11 +33,13 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.WeakHashMap;
@@ -380,10 +383,73 @@ public class DeploymentManager {
     return factoryList;
   }
 
+  private static URL mapToURL(String path) {
+    try {
+      File file = new File(path);
+      return file.toURI().toURL();
+    } catch (MalformedURLException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  private static List<URL> extractCPFromProperty() {
+    List<URL> classpathURLs = new ArrayList<>();
+    String classpath = System.getProperty("java.class.path");
+    if (Objects.nonNull(classpath)) {
+      for (String path : classpath.split(File.pathSeparator)) {
+        classpathURLs.add(mapToURL(path));
+      }
+    }
+    return classpathURLs;
+  }
+
+  // VisibleForTesting
+  static List<URL> extractCPByManifest(ClassLoader current) {
+    List<URL> classpathURLs = new ArrayList<>();
+    String searchFile = "META-INF/MANIFEST.MF";
+    Enumeration<URL> urls;
+    try {
+      urls = current.getResources(searchFile);
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    }
+    for (URL url : Collections.list(urls)) {
+      String urlString = url.toExternalForm();
+      if ("jar".equals(url.getProtocol().toLowerCase())) {
+        String prefix = "jar:";
+        String suffix = "!/" + searchFile;
+        urlString = urlString.replace(prefix, "").replace(suffix, "").trim();
+      }
+      // handle exploded jars
+      urlString = urlString.replace(searchFile, "").trim();
+      try {
+        classpathURLs.add(new URL(urlString));
+      } catch (MalformedURLException e) {
+        throw new IllegalStateException(e);
+      }
+    }
+    return classpathURLs;
+  }
+
+  // VisibleForTesting
+  static List<URL> extractClasspath(ClassLoader current) {
+    if (current instanceof URLClassLoader) {
+      URLClassLoader urlc = (URLClassLoader)current;
+      return Arrays.asList(urlc.getURLs());
+    } else {
+      List<URL> classpathURLs = extractCPFromProperty();
+      for (URL url : extractCPByManifest(current)) {
+        if (!classpathURLs.contains(url)) {
+          classpathURLs.add(url); 
+        }
+      }
+      return classpathURLs;
+    }
+  }
+
   /**
-   * <strong>IMPORTANT</strong> - Isolation groups are not supported on Java 9+ because the application classloader is not
-   * an URLClassLoader anymore. Thus we can't extract the list of jars to configure the IsolatedClassLoader.
-   *
+   * <strong>IMPORTANT</strong> - Isolation groups are not supported if modules are used. Because modules are living
+   * in the runtime image and there is no way, to extract these modules.
    */
   private ClassLoader getClassLoader(DeploymentOptions options) {
     String isolationGroup = options.getIsolationGroup();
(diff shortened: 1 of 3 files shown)


pk-work:
This is a patch for issue #2278

pk-work:
When I run the tests on my machine they failed, even without my change. I will try to get the test running on my local machine. Could someone else test this change with Java 9+, too? Or could we configure Travis, that Travis runs this test with Java 9+?

vietj:
@pk-work what does fail for you ?

vietj:
in addition this PR should be verified on Java 11 by the person merging as the CI does not run the job for java > 8

pk-work:
After I deleted and cloned the repository again I managed to build and run this change with Java 8. Maybe some remnant of my research.

After that I changed in the pom.xml the following lines:
```
     <profile>
-      <id>java-8</id>
+      <id>java-11</id>
     <properties>
```

```  
   <activation>
-        <jdk>1.8</jdk>
+        <jdk>11</jdk>
    </activation>
```
When I ran the tests again, I got the following error:

```
Failed tests: 
  ClusteredEventBusStartFailureTest.testCallbackInvokedOnFailure:47->AsyncTestBase.awaitLatch:596->AsyncTestBase.assertTrue:376 null

Tests run: 3723, Failures: 1, Errors: 0, Skipped: 12
```

vietj:
that test is independant from your contribution, some tests are badly written and sometimes fail (and we need to improve them)

pk-work:
I ran the tests now several times with different Java versions, this is the outcome:

Java  9 - 2 runs - 2 succeeded
Java 10 + 3 runs - 2 failed, 1 succeeded

```
Tests in error: 
  ClusteredEventBusWithSSLTest>AsyncTestBase.after:81->VertxTestBase.tearDown:85->AsyncTestBase.close:653 » UnsupportedOperation
  Http1xTest.testCloseTheConnectionAfterResetBeforeKeepAliveResponseReceived:3479->testCloseTheConnectionAfterResetBeforeResponseReceived:3569->AsyncTestBase.await:119->AsyncTestBase.await:131 » IllegalState
  Http1xTest>AsyncTestBase.after:81->HttpTestBase.tearDown:65->VertxTestBase.tearDown:85->AsyncTestBase.close:653 » UnsupportedOperation
  ClusteredSharedCounterTest>AsyncTestBase.after:81->VertxTestBase.tearDown:85->AsyncTestBase.close:653 » UnsupportedOperation

Tests run: 3723, Failures: 0, Errors: 4, Skipped: 12
```

java 11 - 2 runs - 1 failed, 1 succeeded

```
Tests in error: 
  Http1xTest.testCloseTheConnectionAfterResetBeforeKeepAliveResponseReceived:3479->testCloseTheConnectionAfterResetBeforeResponseReceived:3569->AsyncTestBase.await:119->AsyncTestBase.await:131 » IllegalState

Tests run: 3723, Failures: 0, Errors: 1, Skipped: 12
```

My observation was: If you look at the console all the time during the tests, they run through. Maybe they just need a little attention to work :)

The only downside of this patch is, that folders on class path (e.g. <pathToMyEclipseWorkSpace>/Test/bin/), will be ignored. Maybe this should be documented that with Java 9+ only classes in JAR files will be recognized.

vietj:
the UnsupportedOperation come from Netty itself.

https://github.com/netty/netty/issues/8475

pk-work:
Should I update also the documentation? Or is this limitation okay for you?

pk-work:
I just found a bug in the change .. Which shows that a test is missing.

I will provide a test case for this.

pk-work:
retest this please

pk-work:
Close / Reopen to retest

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -374,19 +376,68 @@ public void unregisterVerticleFactory(VerticleFactory factory) {
...
+      return file.toURI().toURL();
+    } catch (MalformedURLException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  private static List<URL> extractCPFromProperty() {
+    String classpath = System.getProperty("java.class.path");
+    List<URL> classpathURLs = new ArrayList<>();
+    for (String path : classpath.split(":")) {
```
cescoffier: Use path separator, this can break depending on OS.

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -374,19 +376,68 @@ public void unregisterVerticleFactory(VerticleFactory factory) {
...
+    try {
+      File file = new File(path);
+      return file.toURI().toURL();
+    } catch (MalformedURLException e) {
+      throw new IllegalStateException(e);
+    }
+  }
+
+  private static List<URL> extractCPFromProperty() {
+    String classpath = System.getProperty("java.class.path");
```
cescoffier: In some rare circumstances, this can be `null`

src/test/java/io/vertx/core/impl/DeploymentManagerTest.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.junit.Assert;
+
+public class DeploymentManagerTest {
+
+  @Test
+  public void extractClasspathTest() throws MalformedURLException, URISyntaxException {
+    for(URL cpURL : DeploymentManager.extractClasspath(getClass().getClassLoader())) {
+      Assert.assertTrue(cpURL + " is not a valid classpath entry", new File(cpURL.toURI()).exists());
+    }
```
cescoffier: Add a test with an exploded classpath, it may behave differently.

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -374,19 +376,68 @@ public void unregisterVerticleFactory(VerticleFactory factory) {
...
+    String suffix = "!/" + searchFile;
+    String prefix = "file:";
+    Enumeration<URL> urls;
+    try {
+      urls = DeploymentManager.class.getClassLoader().getResources(searchFile);
+    } catch (IOException e) {
+      throw new IllegalStateException(e);
+    }
+    for (URL url : Collections.list(urls)) {
+      String path = url.getFile().replace(prefix, "").replace(suffix, "").trim();
```
cescoffier: This is specific, typically in OSGi there are not `file:`.

src/main/java/io/vertx/core/impl/DeploymentManager.java
pk-work: Done

src/main/java/io/vertx/core/impl/DeploymentManager.java
pk-work: Done

src/test/java/io/vertx/core/impl/DeploymentManagerTest.java
pk-work: Done, good catch, this resulted really in a problem :)

src/main/java/io/vertx/core/impl/DeploymentManager.java
pk-work: But if there is not "file:", it will just not replace it. Or do you mean, there could be something else and  I should delete everything from beginning to the first colon?

pk-work:
@cescoffier  Thank you so much for your valuable feedback. I think I provided a fix for your requests. Except the last comment was not 100% clear to me. Maybe you can have a look at my question.

src/main/java/io/vertx/core/impl/DeploymentManager.java
pk-work: I now fixed it, but not sure if this is a proper solution.

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -374,19 +377,80 @@ public void unregisterVerticleFactory(VerticleFactory factory) {
     return factoryList;
   }
 
+  private static URL mapToURL(String path) {
+    try {
+      if (path.startsWith("/")) {
```
pk-work: This will not work for windows.
Test for this regexp ^(/|[a-zA-Z]\\:\\\\).* will work for absolute paths but not for relative paths.

But how to check if the string starts with a protocol or an relative path, because at least in linux a colon is a valid char in in the path ....

Any ideas?

pk-work: Okay, I found a way to solve this problem:

The only place where this method is called with strings containing an URL is in extractCPByManifest(). I will not longer call mapToURL from this place, then I can expect always a String which represents an absolute file path, which can be handled by new File()...

vietj:
@cescoffier is this good to merge ?

cescoffier:
should be good right now.

pk-work:
Hi, we now really uses this patch now in our internal fork and I found a stupid bug. I will upload a new patch soon. I simply forgot to remove the check if the SystemClassLoader is a URLClassLoader.

I'm so happy that this is not already merged. Sorry for that stupid mistake.

amaceing:
Do you know when/if this will get merged and published?

vietj:
the PR should be done on the 3.7 branch, not master

vietj:
@pk-work can you redo the PR on 3.7 branch ?

pk-work:
Done. Don't know why there are still conflicting files, I rebased on 3.7. Should be outdated.

vietj:
can you make a new clean PR ?

pk-work:
Done #2872

PR #2766, opened 2018-12-14
 Use base64url when converting to/from JSON

### LINKED ISSUE #2765: Use base64url when converting to/from JSON
When converting data objects to/from JSON, byte arrays are encoded/decoded to/from `base64`.

But the [I-JSON Message Format RFC](https://tools.ietf.org/html/rfc7493#section-4.4) actually recommends to use `base64url`.

> When it is required that an I-JSON protocol element contain arbitrary binary data, it is RECOMMENDED that this data be encoded in a string value in base64url; see Section 5 of [RFC4648].

tsegismont:
Fixes #2765

### DIFF
--- a/src/main/generated/io/vertx/core/dns/AddressResolverOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/dns/AddressResolverOptionsConverter.java
@@ -36,7 +36,7 @@ import java.time.format.DateTimeFormatter;
           break;
         case "hostsValue":
           if (member.getValue() instanceof String) {
-            obj.setHostsValue(io.vertx.core.buffer.Buffer.buffer(java.util.Base64.getDecoder().decode((String)member.getValue())));
+            obj.setHostsValue(base64Decode((String)member.getValue()));
           }
           break;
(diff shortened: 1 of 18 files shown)


bckfnn:
IIUC, this will break config files where binary data its stored and read with JsonObject.getBinary(), and where the data contains '+' or '/' like:

```
x509cert = "MIIF+/aa..."
```

And it will break cases where JsonObject.getBinary is used to extract 3.x version json data stored in databases.

The upgrade path is ofcourse to identify my uses of JsonObject.get/setBinary and use the java.util.Base64 encoders instead (or upgrade the config files). 

However I would hope that the decoding could be done in a backward compatible manner so the fields in databases can still be read with getBinary().

vietj:
it sounds fair in case of failure to try the other way I think

vietj:
so we always encode in the new format but allow decoding in the previous format in case it fails

vietj:
I think we dropped this feature.

tsegismont:
Correct -though I'd call it a fix rather than a feature :wink:

PR #2775, opened 2018-12-21
Deprecate logging API

### LINKED ISSUE #2774: Deprecate logging API
The Vert.x logging API is public and advertised in the docs. Besides, it is broken because it offers parameterized logging methods without standardizing on a message format (it delegates this to the logging backend so messages are not portable).

Therefore, we regularly get questions/bug reports about this, which increases the maintenance effort. Also, it's not good to provide a broken feature only to cause users' disappointment.

Considering it's not a programming toolking responsibility to provide yet another logging API, the Vert.x loggers shall be deprecated.

tsegismont:
Fixes #2774


vietj:
is there a need to deprecate implementations ? I believe only deprecate the API is enough

tsegismont:
Just in case. It doesn't hurt anyway.

vietj:
only the interface (which is what people interact with) needs this, no need to repaint the whole hierarchy of class/method with the deprecated annotation.

tsegismont:
@vietj updated as per-review

src/main/java/io/vertx/core/spi/logging/package-info.java
```diff
@@ -0,0 +1,16 @@
...
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+/**
+ * @deprecated see https://github.com/eclipse-vertx/vert.x/issues/2774
+ */
+@Deprecated
```
vietj: I don't think we need to deprecate this as we should continue using an internal logger after v4

tsegismont: Right, we will continue to have an internal logging API. But I believe it would be better to move it to an `impl` or `internal` package. We can do the changes in the module stack. People still using this API  after Vert.x 4 will have be warned.

vietj: this is an SPI package, SPI are not intended to be used directly, so it's fine where it stands.

tsegismont: Right, sorry, I overlooked your comment and thought this was about the other package.

tsegismont: I've updated the PR and rebased on master

PR #2788, opened 2019-01-02
Http client pool improvements

### LINKED ISSUE #2787: HttpClient pool improvements
A few improvements done to the HttpClient and the pool:

- the client suffers from races when a connection is closed while being delivered to the client request (see https://github.com/eclipse-vertx/vert.x/issues/2773)
- TCP idle timeout now is not handled anymore with pool events but instead when a TCP idle event happens, it is ignored when the connection is in the pool
- the activate/passivate connection has been removed (since it's not used anymore for TCP idle)
- the coupling of the pool to channel has been removed as it was not necessary
- 

The main change is that now the pool methods update the pool state and the corresponding events (deliver a connection to a waiter, connect to an endpoint, keep the waiter list bounded) are handled on the event-loop thread to avoid races. To accomodate this, now the pool has a context instead of having a context per connection.

vietj:
Fixes #2787

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
+++ b/src/main/java/io/vertx/core/http/impl/ConnectionManager.java
@@ -119,15 +119,15 @@ class ConnectionManager {
         int maxPoolSize = Math.max(client.getOptions().getMaxPoolSize(), client.getOptions().getHttp2MaxPoolSize());
         Object metric = metrics != null ? metrics.createEndpoint(host, port, maxPoolSize) : null;
         HttpChannelConnector connector = new HttpChannelConnector(client, metric, version, ssl, peerHost, host, port);
-        Pool<HttpClientConnection> pool = new Pool<>(connector, maxWaitQueueSize, connector.weight(), maxSize,
+        Pool<HttpClientConnection> pool = new Pool<>(ctx, connector, maxWaitQueueSize, connector.weight(), maxSize,
           v -> {
             if (metrics != null) {
               metrics.closeEndpoint(host, port, metric);
             }
             endpointMap.remove(key);
           },
-          connectionMap::put,
-          connectionMap::remove,
+          conn -> connectionMap.put(conn.channel(), conn),
+          conn -> connectionMap.remove(conn.channel(), conn),
           false);
         return new Endpoint(pool, metric);
       });
@@ -138,7 +138,7 @@ class ConnectionManager {
         metric = null;
       }
 
-      if (endpoint.pool.getConnection(ctx, ar -> {
+      if (endpoint.pool.getConnection(ar -> {
         if (ar.succeeded()) {
 
           HttpClientConnection conn = ar.result();
--- a/src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
@@ -59,15 +59,21 @@ class Http2ClientConnection extends Http2ConnectionBase implements HttpClientCon
   }
 
   @Override
-  synchronized void onGoAwaySent(int lastStreamId, long errorCode, ByteBuf debugData) {
-    listener.onDiscard();
-    super.onGoAwaySent(lastStreamId, errorCode, debugData);
+  synchronized boolean onGoAwaySent(int lastStreamId, long errorCode, ByteBuf debugData) {
+    boolean goneAway = super.onGoAwaySent(lastStreamId, errorCode, debugData);
+    if (goneAway) {
+      listener.onEvict();
+    }
+    return goneAway;
   }
 
   @Override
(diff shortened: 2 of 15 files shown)


vietj:
note : the PR should be merged with a squashed commit but I've let the original commits to ease the review

vietj:
@tsegismont can you review this PR ?

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -25,43 +25,58 @@
...
  * keep alive timeout.
  *
- * When a waiter asks for a connection, it is either added to the queue (when it's not empty) or attempted to be
- * served (from the pool or by creating a new connection) or failed. The {@link #waitersCount} is the number
- * of total waiters (the waiters in {@link #waitersQueue} but also the inflight) so we know if we can close the pool
- * or not. The {@link #waitersCount} is incremented when a waiter wants to acquire a connection succesfully (i.e
- * it is either added to the queue or served from the pool) and decremented when the it gets a reply (either with
- * a connection or with a failure).
+ * <h3>Acquiring a connection</h3>
+ * When a waiter wants to acquire a connection it is either added to the {@link #waitersQueue} and the request
```
tsegismont: You can remove "is either". Now it's always added to the waiter queue.

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -25,43 +25,58 @@
...
+ * will be handled by the pool asynchronously.
+ * <ul>
+ *   <li>when there is an available pooled connection, this connection is delivered to the waiter and it is removed
+ *   from the queue</li>
+ *   <li>when there is no available connection a connection is created when {@link #weight}{@code <}{@link #maxWeight}</li>
+ *   <li>when the max number of waiters is reached, the request is failed</li>
+ *   <li>otherwise the waiter remains in the queue until progress can be done (i.e a connection is recycled, etc...)</li>
+ * </ul>
+ * When a waiter is satisfied, to avoid races the waiter is guaranteed to read the correct connection state.
```
tsegismont: Could you rephrase this one? It's not clear what you mean.

src/main/java/io/vertx/core/http/impl/pool/Pool.java
```diff
@@ -346,6 +350,7 @@ private synchronized void setConcurrency(Holder holder, long concurrency) {
       throw new IllegalArgumentException("Cannot set a negative concurrency value");
     }
     if (holder.removed) {
+      assert false : "Cannot recycle removed holder";
```
tsegismont: Do we want to keep this assertion? We don't use them afaik

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
```diff
@@ -124,16 +124,16 @@ void getConnection(ContextInternal ctx, String peerHost, boolean ssl, int port,
...
-          connectionMap::remove,
+          conn -> connectionMap.put(conn.channel(), conn),
+          conn -> connectionMap.remove(conn.channel(), conn),
           false);
         return new Endpoint(pool, metric);
       });
       Object metric;
       if (metrics != null) {
-        metric = metrics.enqueueRequest(endpoint.metric); // Metrics might be enqueued several times if getConnection returns false
+        metric = metrics.enqueueRequest(endpoint.metric);
```
tsegismont: The comment was added in one the first commits. Was it before the logic was refactored and now the risk has been mitigated? I'm asking because I'm not sure why this comment was added and now removed without the surrounding code being modified.

tsegismont:
@vietj looks good to me. A few minor concerns.

src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
```diff
@@ -466,7 +466,15 @@ public NetSocket createNetSocket() {
       return conn.toNetSocket(this);
     }
   }
-  
+
+  @Override
+  protected void handleIdle() {
+    if (streams.isEmpty()) {
```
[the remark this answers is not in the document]
vietj: I think it should indeed, good catch!

src/main/java/io/vertx/core/http/impl/pool/Pool.java
vietj: it is actually used in tests (enabled when running tests) which is very useful

src/main/java/io/vertx/core/http/impl/ConnectionManager.java
vietj: it was before

src/main/java/io/vertx/core/http/impl/pool/Pool.java
vietj: I mean that it can catch more bugs

vietj: (and it did :-) )

Issue #2790, opened 2019-01-03
JsonParser should not resume or pause the source wrapped read stream when it is ended


abelhegedus:
We have implemented streaming Json from an AsyncFile as shown in [this example](https://github.com/vert-x3/vertx-examples/blob/master/core-examples/src/main/java/io/vertx/example/core/jsonstreaming/JsonStreamingExample.java). This has worked nicely before 3.6.0 but now results in a `java.lang.IllegalStateException: File handle is closed` in 3.6.2.

Based on some debugging, the following happens:
1. `io.vertx.core.parsetools.impl.JsonParserImpl#end` [calls](https://github.com/eclipse-vertx/vert.x/blob/3.6.2/src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java#L210) `io.vertx.core.parsetools.impl.JsonParserImpl#checkPending` which will [call](https://github.com/eclipse-vertx/vert.x/blob/3.6.2/src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java#L225) the `endHandler`.
2. `endHandler` [closes](https://github.com/vert-x3/vertx-examples/blob/3.6.1/core-examples/src/main/java/io/vertx/example/core/jsonstreaming/JsonStreamingExample.java#L60) the AsyncFile
3. `io.vertx.core.streams.ReadStream#resume` is [called](https://github.com/eclipse-vertx/vert.x/blob/3.6.2/src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java#L248) (on the already closed stream)
4. the `exceptionHandler` is [called](https://github.com/eclipse-vertx/vert.x/blob/3.6.2/src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java#L259) with the above exception

As far as I can see, issue #2584 and PR #2583 seems to be related.

Is the example outdated? Should we handle file closing in some other way?

vietj:
can you provide a reproducer for the issue ?

vietj:
what is actually calling resume ? it is not clear

abelhegedus:
> what is actually calling resume ? it is not clear

I have added stable links to the original comments.

> can you provide a reproducer for the issue ?

I can do that, though I think the linked example should reproduce the issue as well.

vietj:
ok, we will check the example

KowalczykBartek:
[...]

PR #2796, opened 2019-01-09
provide fix and test for #2790

KowalczykBartek:
Hi, 
that is initial bug-fix for #2790.

I have one problem with test - because of closing file is asynchronous I had to handle exception in Vert.x exception handler 
```java
 vertx.exceptionHandler(exception -> fail());
```
and also schedule completion to allow vert.x first catch exception and fail test.
```java
vertx.setTimer(10, val -> complete());
```

do you maybe have other idea how to test that case with file-closed case (my seems really ... bad)  ?

### DIFF
--- a/src/test/java/io/vertx/core/parsetools/FakeStream.java
+++ b/src/test/java/io/vertx/core/parsetools/FakeStream.java
@@ -24,6 +24,8 @@ class FakeStream implements ReadStream<Buffer> {
   private Handler<Buffer> eventHandler;
   private Handler<Void> endHandler;
   private Handler<Throwable> exceptionHandler;
+  private volatile int pauseCount;
+  private volatile int resumeCount;
 
   @Override
   public ReadStream<Buffer> exceptionHandler(Handler<Throwable> handler) {
@@ -50,11 +52,13 @@ class FakeStream implements ReadStream<Buffer> {
   @Override
   public ReadStream<Buffer> pause() {
     demand = 0L;
+    pauseCount++;
     return this;
   }
 
   @Override
   public ReadStream<Buffer> resume() {
+    resumeCount++;
     return fetch(Long.MAX_VALUE);
   }
 
@@ -88,5 +92,13 @@ class FakeStream implements ReadStream<Buffer> {
 
   void end() {
     endHandler.handle(null);
+  }
+
+  public int pauseCount() {
+    return pauseCount;
+  }
+
+  public int resumeCount() {
+    return resumeCount;
   }
 }
--- a/src/test/java/io/vertx/core/parsetools/JsonParserTest.java
+++ b/src/test/java/io/vertx/core/parsetools/JsonParserTest.java
@@ -789,6 +789,9 @@ public class JsonParserTest {
     stream.end();
     assertEquals(0, events.size());
     assertEquals(1, ended.get());
+    //regression check for #2790 - ensure that by accident resume method is not called.
+    assertEquals(0, stream.pauseCount());
+    assertEquals(0, stream.resumeCount());
   }
 
   @Test
--- a/src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java
+++ b/src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java
@@ -224,6 +224,7 @@ public class JsonParserImpl implements JsonParser {
             if (endHandler != null) {
               endHandler.handle(null);
             }
+            return;
           }
           break;
         } else {


vietj:
looks good, but can we have the same test using the `FakeStream` instead that reproduces the same asynchrony than `AsyncFile` ?

KowalczykBartek:
I will try to  replace async-file with FakeStream

vietj:
thanks

src/test/java/io/vertx/core/parsetools/JsonParserTest.java
```diff
@@ -780,7 +781,17 @@ public void testStreamFetchInHandler() {
 
   @Test
   public void testStreamEnd() {
-    FakeStream stream = new FakeStream();
+    FakeStream stream = new FakeStream(){
+      //regression check for #2790 - ensure that by accident resume method is not called.
+      @Override
```
vietj: rather add on FakeStream a pauseCount / resumeCount that are incremented when those methods are called and check the counters are == 0 at the end of the test

KowalczykBartek: But that is done in FakeReadStream - should I use it over FakeStream ?

KowalczykBartek: done

KowalczykBartek:
```
Tests in error: 
  Http1xTest.testHttpServerWithIdleTimeoutSendChunkedFile Âť Vertx Connection was...
```
failed - but I think it it not related to my change.

vietj:
@KowalczykBartek yes some tests fails due to races sometimes

src/test/java/io/vertx/core/parsetools/FakeStream.java
```diff
@@ -24,6 +24,8 @@
   private Handler<Buffer> eventHandler;
   private Handler<Void> endHandler;
   private Handler<Throwable> exceptionHandler;
+  int pauseCount;
```
vietj: make it private and add a pauseCount() accessor

KowalczykBartek: done

src/test/java/io/vertx/core/parsetools/FakeStream.java
```diff
@@ -89,4 +93,12 @@ void fail(Throwable err) {
   void end() {
     endHandler.handle(null);
   }
+
+  public int getPauseCount() {
```
vietj: rather `pauseCount()` and `resumeCount()`

KowalczykBartek: done

src/test/java/io/vertx/core/parsetools/FakeStream.java
```diff
@@ -89,4 +93,12 @@ void fail(Throwable err) {
   void end() {
     endHandler.handle(null);
   }
+
+  public int pauseCount() {
```
vietj: make both synchronized and I think it will be fine. (this class can be used in multithreaded tests)

KowalczykBartek: in such case, instead of int - AtomicInteger should be provided, because 
a. There is race during value incrementation
b. there is no happens-before relationship between adding and reading.

vietj: I think it's fine, we are not trying to remove race conditions so we don't need happens-before, just ensure safe publication of the value, these values are likely to be used at the end of the test.

KowalczykBartek: synchronized keyword do not ensure safe publication, this would be done with volatile on keyword - or do I miss some part of JMM ?

vietj: It does when the value was stored under synchronization, so we just need to ensure that we use the same monitor and synchronize when we store the value.

KowalczykBartek: ok, that makes sense :)

KowalczykBartek: @vietj where is corresponding synchronize on writing data (on the same monitor (so instance of FakeStream)) ?

KowalczykBartek: I added **volatile** to both **pauseCount** and **resumeCount** - this for sure will ensure visibility correctness between any thread reading and writing (of course do not prevent against races). I could not find example that uses FakeStream in multithreaded environment, so, I could not verify if synchronization keyword could help.
Is that ok ?

vietj: @KowalczykBartek it is used in rx-java module I haven't checked whether tests are multi-threaded though but there are chances there are.

KowalczykBartek: in rx-java module I see that 
```
io.vertx.test.fakestream.Fakestream
```
and that is true, that this class have synchronization in place. But we are talking about 
```
io.vertx.core.parsetools.FakeStream
```

vietj: ah you are right, so just leave it as is and in the future we'll try to use a single FakeStream class in vertx test classes

vietj:
thanks for the contribution @KowalczykBartek

PR #2812, opened 2019-01-29
Establish a relationship between the vertx thread name and the vertx name - fixes #2811

### LINKED ISSUE #2811: how to customize eventloop thread name and calculate how many eventloopContext created
currently:
* we create multiple vertx instances in one process
* sometimes our code is not so good, that caused create unnecessary eventloopContext

so we extend VertxImpl in vertx 3.5.3 and previous versions (https://github.com/apache/servicecomb-java-chassis/blob/master/foundations/foundation-vertx/src/main/java/io/vertx/core/impl/VertxImplEx.java):
* give vertx a name, and make eventloop thread name prefix relate to vertx name
* calculate how many eventloopContext created
image

but when i try this in 3.6.3-SNAPSHOT, i fount that constructor changed from package to private, so i can not do what we want.
someone help me to resolve this? thanks.

wujimin:
fixes https://github.com/eclipse-vertx/vert.x/issues/2811

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -2050,6 +2050,9 @@ Set the time unit of <code>maxWorkerExecuteTime</code>.
 |[[metricsOptions]]`@metricsOptions`|`link:dataobjects.html#MetricsOptions[MetricsOptions]`|+++
 Set the metrics options
 +++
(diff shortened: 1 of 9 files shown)


okou19900722:
the name is too simple, how about `threadNamePrefix`

PR #2814, opened 2019-01-29
MessageConsumer delivers more items than requested in fetch mode

### LINKED ISSUE #2813: MessageConsumer delivers more items than requested in fetch mode
Originally filed as an Rxified consumer issue (https://github.com/vert-x3/vertx-rx/issues/179) but it turns out the `MessageConsumer` implementation itself is the root cause.

tsegismont:
Fixes #2813


tsegismont:
@vietj sounds good to me

PR #2841, opened 2019-02-20
fix logging to system out and err

zyclonite:
there were several places where e.printStackTrace() was used, which messed up the logging for docker containers (e.g. if you change the logging pattern with the used logging framework)

### DIFF
--- a/src/main/java/io/vertx/core/dns/impl/decoder/RecordDecoder.java
+++ b/src/main/java/io/vertx/core/dns/impl/decoder/RecordDecoder.java
@@ -19,6 +19,8 @@ import io.netty.handler.codec.dns.DnsRecordType;
 import io.netty.util.CharsetUtil;
 import io.vertx.core.dns.impl.MxRecordImpl;
 import io.vertx.core.dns.impl.SrvRecordImpl;
+import io.vertx.core.logging.Logger;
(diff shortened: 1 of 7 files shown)


src/main/java/io/vertx/core/dns/impl/decoder/RecordDecoder.java
```diff
@@ -223,7 +223,7 @@ static String getName(ByteBuf buf, int offset) {
         try {
             result = (T) decoder.apply(record);
         } catch (Exception e) {
-            e.printStackTrace();
```
vietj: we should log here and not throw an ISE

zyclonite: ok, i will fix it

zyclonite:
@vietj do need something else?

vietj:
let me run the CI again @zyclonite to check

vietj:
can you make a PR for the 3.7 branch as well @zyclonite ?

zyclonite:
sure, will do

zyclonite:
@vietj if you like the fix for 3.7.1, the pr is still there https://github.com/eclipse-vertx/vert.x/pull/2854

PR #2843, opened 2019-02-20
Context tracing - fixes #2842

### LINKED ISSUE #2842: Context tracing
Support context tracing in a Vert.x runtime.

See also https://github.com/vert-x3/wiki/wiki/RFC:-context-tracing

### Goals

- trace an activity or flow attached to the current Vert.x context (which implies it can be statically retrieved using a thread local using the Vert.x context)
- hooks for transport context propagation (HTTP client/server, event-bus consumer/producer)
tracing management, i.e the ability to start/end an activity for third-party integration

### Non goals

- generic framework for tracing arbitrary continuations
- thread local like registries

### Changes

- allow to shallow duplicate and attach local data to a `io.vertx.core.Context`, this creates a new context which shares the same state than its creating context (same event-loop, same attribute map, same concurrency, etc...).
- add a local map storage on the context, that is used to store state to distinguish contexts (i.e it can store a span id, etc...)
- add an event-driven tracer SPI (that can be configured or discovered) for integration with external tracing systems (tracing clients such as Zipkin, OpenTracing, etc...)
- modify server implementations so that contexts are created when necessary
  - per HTTP server request
  - per event-bus message received
  - etc...

### Not covered by this issue (i.e todo)

- integration of other Vert.x middleware with the tracer SPI so their activity can be traced
- integration with tracing clients
- per verticle tracing configuration (not planned)

vietj:
See #2842

### DIFF
--- a/src/main/java/io/vertx/core/Context.java
+++ b/src/main/java/io/vertx/core/Context.java
@@ -205,6 +205,33 @@ public interface Context {
   boolean remove(String key);
 
   /**
+   * Get some local data from the context.
+   *
+   * @param key  the key of the data
+   * @param <T>  the type of the data
+   * @return the data
+   */
+  <T> T getLocal(String key);
+
+  /**
+   * Put some local data in the context.
+   * <p>
+   * This can be used to share data between different handlers that share a context
+   *
+   * @param key  the key of the data
+   * @param value  the data
+   */
+  void putLocal(String key, Object value);
+
+  /**
+   * Remove some local data from the context.
+   *
+   * @param key  the key to remove
+   * @return true if removed successfully, false otherwise
+   */
+  boolean removeLocal(String key);
+
+  /**
    * @return The Vertx instance that created the context
    */
   Vertx owner();
--- a/src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
@@ -12,6 +12,7 @@
 package io.vertx.core.eventbus.impl;
 
 import io.vertx.core.*;
+import io.vertx.core.eventbus.DeliveryContext;
 import io.vertx.core.eventbus.Message;
 import io.vertx.core.eventbus.MessageConsumer;
 import io.vertx.core.eventbus.impl.clustered.ClusteredMessage;
@@ -222,33 +223,12 @@ public class HandlerRegistration<T> implements MessageConsumer<T>, Handler<Messa
   private void deliver(Handler<Message<T>> theHandler, Message<T> message, ContextInternal context) {
     // Handle the message outside the sync block
     // https://bugs.eclipse.org/bugs/show_bug.cgi?id=473714
-    boolean local = true;
-    if (message instanceof ClusteredMessage) {
-      // A bit hacky
-      ClusteredMessage cmsg = (ClusteredMessage)message;
-      if (cmsg.isFromWire()) {
-        local = false;
-      }
-    }
     String creditsAddress = message.headers().get(MessageProducerImpl.CREDIT_ADDRESS_HEADER_NAME);
     if (creditsAddress != null) {
       eventBus.send(creditsAddress, 1);
     }
-    try {
-      if (metrics != null) {
-        metrics.beginHandleMessage(metric, local);
-      }
-      theHandler.handle(message);
-      if (metrics != null) {
-        metrics.endHandleMessage(metric, null);
-      }
-    } catch (Exception e) {
-      log.error("Failed to handleMessage. address: " + message.address(), e);
-      if (metrics != null) {
-        metrics.endHandleMessage(metric, e);
-      }
-      context.reportException(e);
-    }
+    InboundDeliveryContext deliveryCtx = new InboundDeliveryContext((MessageImpl<?, T>) message, theHandler, context);
+    deliveryCtx.next();
     checkNextTick();
   }
 
@@ -352,6 +332,76 @@ public class HandlerRegistration<T> implements MessageConsumer<T>, Handler<Messa
 
   public Object getMetric() {
     return metric;
+  }
+
+  protected class InboundDeliveryContext implements DeliveryContext<T> {
+
+    private final MessageImpl<?, T> message;
+    private final Iterator<Handler<DeliveryContext>> iter;
+    private final Handler<Message<T>> handler;
+    private final ContextInternal context;
+
+    private InboundDeliveryContext(MessageImpl<?, T> message, Handler<Message<T>> handler, ContextInternal context) {
+      this.message = message;
+      this.handler = handler;
+      this.iter = eventBus.receiveInterceptors();
+      this.context = context;
+    }
+
+    @Override
+    public Message<T> message() {
+      return message;
+    }
+
+    @Override
+    public void next() {
+      if (iter.hasNext()) {
+        try {
+          Handler<DeliveryContext> handler = iter.next();
+          if (handler != null) {
+            handler.handle(this);
+          } else {
+            next();
+          }
+        } catch (Throwable t) {
+          log.error("Failure in interceptor", t);
+        }
+      } else {
+        boolean local = true;
+        if (message instanceof ClusteredMessage) {
+          // A bit hacky
+          ClusteredMessage cmsg = (ClusteredMessage)message;
+          if (cmsg.isFromWire()) {
+            local = false;
+          }
+        }
+        try {
+          if (metrics != null) {
+            metrics.beginHandleMessage(metric, local);
+          }
+          handler.handle(message);
+          if (metrics != null) {
+            metrics.endHandleMessage(metric, null);
+          }
+        } catch (Exception e) {
+          log.error("Failed to handleMessage. address: " + message.address(), e);
+          if (metrics != null) {
+            metrics.endHandleMessage(metric, e);
+          }
+          context.reportException(e);
+        }
+      }
+    }
+
+    @Override
+    public boolean send() {
+      return message.isSend();
+    }
+
+    @Override
+    public Object body() {
+      return message.receivedBody;
+    }
   }
 
 }
--- a/src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
@@ -11,14 +11,10 @@
 
 package io.vertx.core.eventbus.impl;
 
-import io.vertx.codegen.annotations.Nullable;
 import io.vertx.core.*;
 import io.vertx.core.eventbus.Message;
 import io.vertx.core.eventbus.MessageConsumer;
-import io.vertx.core.eventbus.ReplyException;
-import io.vertx.core.eventbus.ReplyFailure;
 import io.vertx.core.eventbus.impl.clustered.ClusteredMessage;
-import io.vertx.core.http.CaseInsensitiveHeaders;
 import io.vertx.core.impl.Arguments;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.logging.Logger;
@@ -44,10 +40,9 @@ public class HandlerRegistration<T> implements MessageConsumer<T>, Handler<Messa
   private final Vertx vertx;
   private final EventBusMetrics metrics;
   private final EventBusImpl eventBus;
-  private final String address;
-  private final String repliedAddress;
+  final String address;
+  final String repliedAddress;
   private final boolean localOnly;
-  private long timeoutID = -1;
   private HandlerHolder<T> registered;
   private Handler<Message<T>> handler;
   private ContextInternal handlerContext;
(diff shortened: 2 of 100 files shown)


vietj:
This PR breaks a few internal API, so the stack needs to be checked after merge

src/main/java/io/vertx/core/Context.java
```diff
@@ -204,6 +204,33 @@ static boolean isOnVertxThread() {
    */
   boolean remove(String key);
 
+  /**
+   * Get some local data from the context.
```
adriancole: define local? (in the javadoc)

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
```diff
@@ -377,4 +343,82 @@ public Object getMetric() {
...
+            local = false;
+          }
+        }
+        try {
+          if (metrics != null) {
+            metrics.beginHandleMessage(metric, local);
+          }
+          VertxTracer tracer = handlerContext.tracer();
+          if (tracer != null && !src) {
+            List<Map.Entry<String, String>> tags = Collections.singletonList(new AbstractMap.SimpleEntry<>("peer.service", message.address));
```
adriancole: imho (and in zipkin) the remote address is important enough to be a type. It is easier to change a type to a named tag vs the other way around. Also the name "service" is usually about the friendly name, not the IP.

https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/Span.java#L168

adriancole: further, it would be interesting to know if this type of property (whether escalated to its own callback or not) is always there or only on remotely originated requests.

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -389,6 +395,14 @@ public void beginRequest(HttpClientRequestImpl req) {
           Object reqMetric = conn.metrics.requestBegin(conn.endpointMetric, conn.metric(), conn.localAddress(), conn.remoteAddress(), request);
           request.metric(reqMetric);
         }
+        VertxTracer tracer = context.tracer();
+        if (tracer != null) {
+          List<Map.Entry<String, String>> tags = new ArrayList<>();
+          tags.add(new AbstractMap.SimpleEntry<>("http.url", req.absoluteURI()));
+          tags.add(new AbstractMap.SimpleEntry<>("http.method", req.method().name()));
+          BiConsumer<String, String> headers = (key, val) -> req.headers().add(key, val);
+          trace = tracer.sendRequest(request.stream.getContext(), request, request.method.name(), headers, tags);
```
adriancole: this is presumptive about the data that will be parsed. Why not allow the impl to decide vs partially parsing here?

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -479,15 +493,20 @@ private HttpClientResponseImpl beginResponse(HttpResponse resp) {
...
           Object reqMetric = req.metric();
           if (req.exceptionOccurred != null) {
             conn.metrics.requestReset(reqMetric);
           } else {
             conn.metrics.responseEnd(reqMetric, response);
           }
         }
+        VertxTracer tracer = context.tracer();
+        if (tracer != null) {
+          List<Map.Entry<String, String>> tags = Collections.singletonList(new AbstractMap.SimpleEntry<>("http.status_code", "" + response.statusCode()));
```
adriancole: similarly here, not just data policy, but also we are allocating objects which may never be used. the tracer can decide what to do instead (certainly brave would)

src/main/java/io/vertx/core/spi/tracing/VertxTracer.java
```diff
@@ -0,0 +1,80 @@
...
+
+import io.vertx.core.Context;
+
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+public interface VertxTracer<I, O> {
```
adriancole: I don't think usually client and server policy is specified in the same interface as the library impls usually differ. I would recommend splitting these out as client and server.

Also, notice how even outside brave, the data policy opinions are not on the callback interface. Here's another example https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/ClientStreamTracer.java

src/main/java/io/vertx/core/tracing/TracingOptions.java
```diff
@@ -0,0 +1,129 @@
...
+   * @param json the JsonObject to create it from
+   */
+  public TracingOptions(JsonObject json) {
+    this();
+    TracingOptionsConverter.fromJson(json, this);
+    this.json = json.copy();
+  }
+
+  /**
+   * Will tracing be enabled on the Vert.x instance?
```
adriancole: define tracing and the impact. For example, does tracing imply header processing? If disabled will incoming headers propagated to outgoing requests? Some get confused by this.

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
vietj: you mean the `http.url`, etc... names ?

adriancole: not just the names.. that they are recorded at all. if you have access to req at the tracer impl, you can decide what to record.

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
vietj: we're assuming here allocated objects are going to used by the tracer

src/main/java/io/vertx/core/tracing/TracingOptions.java
vietj: I think we can say that when tracing is enabled then header propagation would always happen and the reporting policy will depend on the implementation and its configuration, how does that sound ?

vietj:
thanks @adriancole for the review, can you tell me what you name _data policy_ ? as I'm a mere tracing user, I don't know all the tracing vocabulary yet :-)

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
vietj: @adriancole so that was choice we made initially, however that forces the tracer to be cast the request object to `HttpClientRequest` which is fine.

However we can to have the `VertxTracer` to be used by other parts of the Vert.x stack and it means then that the tracer implementation would depend (at the artifact level) on too many artifacts and have specific code for each of them. This choice allows to decouple the tracing implementation from the activity being traced.

src/main/java/io/vertx/core/spi/tracing/VertxTracer.java
vietj: do you mean this interface should be split in 2 interfaces ?

src/main/java/io/vertx/core/Context.java
vietj: will do, thanks.

vietj:
also keep in mind that this PR is a first milestone for tracing focusing on the foundation and there will be another one later after we get initial feedback from the community

adriancole:
- we're assuming here allocated objects are going to used by the tracer

yeah they wouldnt be in brave and probably also census. there are other
tracers as well who may not use the same formats here. I think it is better
to pass the request/response/error objects with optional socket details if
available

vietj:
I will see how we can avoid this in the api design

vietj:
@adriancole is there a common franca lingua in tracers for tags ? that would be useful

adriancole:
@vietj my point was less about the tag names and more about the idea of parsing them at this layer of abstraction. zipkin's original list was adopted and expanded by opentracing and also opencensus. However, to get to my point, even if don't care about the extra overhead, policy around choosing to record "http.url" is questionable. There's often secrets inside. Also, to match with metrics need cleaning up the data. You can avoid this whole topic by just passing the objects I mentioned.

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
maheshpec: "peer.address" may be generic enough for this message address. Here's a reference to semantic conventions used in open tracing - https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table

vietj:
@adriancole the client request is actually passed, so the tracer can chose when it's an HttpClientRequest to use it instead of the tags (which is what the brave-tracing does for the moment). If there is a concern about security, we can make passing the `http.url` configurable in the tracing options for this particular case.

adriancole:
anyway I think you probably get my concern which is that the trace
instrumentation will be making data policy decisions and doing that also
here mixes config at different layers that which is containable below. it
is up to you what to do with this feedback

vietj:
beside HTTP client, I would like to have more examples about other data policy decisions made in practice (i.e which other libraries and which parts) ?

adriancole:
another example. there are several of these
https://github.com/openzipkin/brave/blob/master/instrumentation/grpc/src/main/java/brave/grpc/GrpcParser.java

vietj:
I see clearer now how per client/server how this works.

I think that for now we shall merge the PR as it is now and make a new branch for to support such feature in the near future.

vietj:
Some thoughts about this: the proposal would be to change:

```
  I receiveRequest(Context context, Object request, String operation, Iterable<Map.Entry<String, String>> headers, Iterable<Map.Entry<String, String>> tags);
```

to

```
  <R> I receiveRequest(Context context, R request, String operation, Iterable<Map.Entry<String, String>> headers, Iterable<Map.Entry<String, String>> tags, TagExtractor<R> extractor);
```

with:

```
// Stateless extractor
public interface TagExtractor<T> {
  int len(T obj);
  String tagName(T obj, int pos);
  String tagValue(T obj, int pos);
}
```

and give the choice to the tracer to either directly use the request object or use the extractor if it does not have predefined parser.

adriancole:
it still looks like you are moving features that should exist in the tracer
up.

even if used for metrics you'll notice what is going on is mixing semi
opinionated parsing higher up. ex micrometer has a tag parser as well.

if OT is the reason consider that they are considering stopping dev at
v0.32 for a merger with another library census. do you want to stain this
api for a long time based on this?

vietj:
I'm not sure to understand what you mean.

Vert.x has provided since 3.0 an SPI for metrics (implemented for Dropwizard and Micrometer).

The goals of this SPI are:

- decouple vertx-core and libraries
- has no overhead when disabled
- event driven style (e.g the SPI provides events and does not has the concept of counters, gauge, etc...)

This has been very successful.

This SPI is just the same concept applied to tracing, so indeed there are similarities but they are not the same from my perspective.

Julien

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
vietj: in this case, the message.address is the name under which the event-bus consumer was registered, i.e `eventBus.consumer("foo", handler);` then `peer.address` is `foo`, how would it be translated in Zipkin ?

adriancole:
I think I dont have another way to express that partially pre parsing in
the way you are doing is not ideal from choosing not to and allowing the
impl to do that. it is probably more flexible and more efficient without
reducing quality of the result.

I think you really must like this design anyway for a reason unrelated..
maybe you should just do it like you have disregarding feedback.

adriancole:
mobile spell check is maddening..

I will write another way as it changed the meaning by substituting two
words it thought wrong.

you are unusually defensive about what is not a great api design. it seems
to be catering to opentracing as brave obviously doesnt need this.

* partially pre parsing arbitrarily chosen data the stuffing that in lists
of map.entry makes no sense from a perf pov and also data policy pov.
* even if it did, the arity of the callbacks is awkward and long vs just
passing request and response objects

however I have said feedback before a bunch of times..  no idea why but you
really like the approach you chose. it isnt obvious why it is good and I
have found many problems with it.

you want to merge like you have it.. I dont want to know why you feel so
strong about it.. likely it is OT and that will just make this feedback
even more useless.

just do what you want to do

vietj:
I think you misunderstand me :-)

The issue it aims to solve is that actually we intent to have the VertxTracer to be used elsewhere in Vert.x stack:

- vertx kafka client
- vertx sql clients
- vertx redis client
- vertx cassandra client

etc...

Without pre-parsed tags, it means that vertx-tracing must integrate and depend on all these project.

Pre-parsing tags allows to decouple both and have the stack project provide enough information for tracers.

So when a tracer implements the SPI it has the choice to either:

1/ use the pre-parsed tags (or the extractor)
2/ parse itself using the request object (which is what the vertx-brave implementation does using Brave parsers and casting the request to HttpServerRequest/HttpClientRequest)

I hope it makes more sense to you

codepitbull:
I still have concerns with integrating this directly into Vert.x. 
While I completely agree with making things more tracer friendly and making them easier accessible I still think we should go more into the direction of Sleuth or Micronaut, keeping actual tracing integration outside the core framework by at most adding annotations and having letting tracers hook in via instrumentation.
This also makes us way more flexible for future changes.
That's how Sleuth and Micronaut are doing it.

Another thing, but that's probably a minor concern, is that adding things to our hot paths without being able to opt out might incur unexpected costs.

I'd rather be able to decide from the outside what parts I actually want to see, even being able to ignore whole package hierarchies than have it everywhere.

vietj:
@codepitbull it's not clear to which concern and whom you are actually replying to

codepitbull:
@vietj sorry, was going through my github backlog and was automatically commenting here. Should have done it on the mailinglist and will move it there.

PR #2854, opened 2019-02-28
fix logging to system out and err

zyclonite:
same as #2841 but for 3.7 branch


zyclonite:
please restart ci build

PR #2866, opened 2019-03-04
JSON decoding without knowledge of the content

### LINKED ISSUE #2864: JSON decoding without knowledge of the content
Currently, it's possible to decode content to a `JsonObject` or a `JsonArray`, if you know what the actual content is.

But sometimes you receive JSON without knowing the format beforehand, and it would be handy to have methods to decode to the right target object.

An example use case is a GraphQL request: some clients (Apollo) support request batching and then will send you an array of request instead of a single request object. Only by inspecting the content you can determine if the request is batched request or a regular one.

tsegismont:
Fixes #2864

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -126,6 +126,31 @@ public class Json {
   }
 
   /**
+   * Decode a given JSON string.
+   *
+   * @param str the JSON string.
+   *
+   * @return a JSON element which can be a {@link JsonArray}, {@link JsonObject}, {@link String}, ...etc if the content is an array, object, string, ...etc
+   * @throws DecodeException when there is a parsing or invalid mapping.
+   */
+  public static Object decodeValue(String str) throws DecodeException {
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/json/Json.java
```diff
@@ -125,6 +125,31 @@ public static String encodePrettily(Object obj) throws EncodeException {
     }
   }
 
+  /**
+   * Decode a given JSON string.
+   *
+   * @param str the JSON string.
+   *
+   * @return a {@link JsonArray}, {@link JsonObject}, {@link String}, ...etc if the content is an array, object, string, ...etc
```
vietj: `a JSON element which can be {@link JsonArray}, ....`

tsegismont:
Thanks for taking a look! I'll update the PR and merge.

src/main/java/io/vertx/core/json/Json.java
tsegismont: Done

PR #2868, opened 2019-03-05
HttpServer close handler not invoked if actual server is null

### LINKED ISSUE #2867: HttpServer close handler not invoked if actual server is null
If an exception occurs after `listening` is switched on but before `actualServer` is set, then the close method handler will not be invoked.

This prevents the `deploy` method to receive the failure (the verticle is undeployed correctly but the HTTP server hook never completes). Here's an example:

```java
public class HTTPSVerticle extends AbstractVerticle {

  @Override
    PfxOptions pfxOptions = new PfxOptions().setValue(Buffer.buffer("path")).setPassword("pwd");

    HttpServerOptions httpServerOptions = new HttpServerOptions();
    httpServerOptions
      .setSsl(true)
      .setPfxTrustOptions(pfxOptions);

    vertx.createHttpServer(httpServerOptions).requestHandler(req -> {
      req.response().end("Hello World!");
    }).listen(8443, ar -> {
      if (ar.succeeded()) {
        startFuture.complete();
      } else {
        startFuture.fail(ar.cause());
      }
    });
  }
}
```

If the PFX path is wrong, `listen` throws an exception.

```java
vertx.deploy(new HTTPSVerticle(), handler);
```

But the handler will not be invoked

tsegismont:
Fixes #2867

### DIFF
--- a/src/test/java/io/vertx/core/http/HttpServerCloseHookTest.java
+++ b/src/test/java/io/vertx/core/http/HttpServerCloseHookTest.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.http;
+
+import io.vertx.core.AbstractVerticle;
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.net.PfxOptions;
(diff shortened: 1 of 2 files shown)


tsegismont:
@vietj can you please review? I set the issue milestone to 3.7, if you're fine with backporting.

src/test/java/io/vertx/core/http/HttpServerCloseHookTest.java
```diff
@@ -0,0 +1,56 @@
...
+      complete();
+    }));
+    await();
+  }
+
+  private static class TestVerticle extends AbstractVerticle {
+
+    @Override
+    public void start(Future<Void> startFuture) {
+      PfxOptions pfxOptions = new PfxOptions().setValue(Buffer.buffer("boom"));
```
vietj: rename to _invalidOptions_ so it's clearer it's not correct (took some time to figure this it)

tsegismont: I thought `boom` was clear enough :stuck_out_tongue_winking_eye: 

But right, I'll update and merge. Thanks for reviewing.

vietj:
I missed it because I overlooked the content value

PR #2871, opened 2019-03-08
Set file cache dir to tmp

pmlopes:
As discussed on the thread: https://groups.google.com/forum/#!topic/vertx-dev/Gj_it5N7m34

Less pollution of working directory and make vert.x applications more container friendly where writing is usually not allowed, except on system temp dir.

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -578,6 +578,11 @@ Set the ALPN usage.
 When vert.x cannot find the file on the filesystem it tries to resolve the
  file from the class path when this is set to <code>true</code>.
 +++
+|[[fileCacheDir]]`@fileCacheDir`|`String`|+++
+When vert.x reads a file that is packaged with the application it gets
+ extracted to this directory first and subsequent reads will use the extracted
+ file to get better IO performance.
++++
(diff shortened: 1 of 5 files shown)


vietj:
cc @ludoch - while contributing the GAE Vert.x samples I needed to use a specific entry point to set the Vert.x cache base directory to `/tmp`. If we move this to `java.io.tmp` it believe this should work out of the box in GAE. Can you confirm ?

vietj:
also I'm wondering if we should not port this to 3.7 as well.

ludoch:
Hey,
On the Java11 GAE runtime I think we have the correct value for  java.io.tmpdir  to point  to  /tmp...

See some environment info in this live sample version:
https://gaeinfo-dot-openjdk11.appspot.com/

Now, remember that for us, /tmp is mapped to your memory, so it should not grow that much  over time, or you'll end up having Java OOM errors.

vietj:
@pmlopes can you perform the change I asked so we can merge this PR ? also the eclipse check does not seem to pass, can you make a new PR to avoid this ?

pmlopes:
This PR has been replaced by #2888 due to issues with eclipse validation

vietj:
I'll back-port to 3.7 in the minute

vietj:
@pmlopes can you add this to the breaking change wiki page for 3.7 ?

PR #2872, opened 2019-03-11
Enable isolation groups for Java 9+

pk-work:
The current implementation of isolation groups [1] requires the
instantiation of a new IsolatingClassLoader, which contains all JARs
from the class path of the SystemClassLoader. Before Java 9, this was
solved by casting the SystemClassLoader to an URLClassLoader to extract
JARs on the class path of the SystemClassLoader.

Due to the fact, that the SystemClassLoader is not longer an
URLClassLoader in Java 9, this approach will not longer work.

This patch implements a way to extract the class path of the
SystemClassLoader which does not require the SystemClassLoader to be an
instance of URLClassLoader.

WARNING: This approach will not work for modules, which were introduces
in Java 9, because modules are living in the "runtime image" [2]. Due to
this, it is not possible to pass the module classes into the current
implementation of IsolatingClassLoader.

[1] https://github.com/eclipse-vertx/vert.x/blob/4bbced9e8ac173ab73c4b4d1ef0bee99c6322a89/src/main/asciidoc/index.adoc#verticle-isolation-groups
[2] https://openjdk.java.net/jeps/220

Signed-off-by: Pascal Krause <pascal.krause@sap.com>

### DIFF
--- a/src/main/java/io/vertx/core/impl/DeploymentManager.java
+++ b/src/main/java/io/vertx/core/impl/DeploymentManager.java
@@ -25,6 +25,7 @@ import io.vertx.core.spi.VerticleFactory;
 import io.vertx.core.spi.metrics.VertxMetrics;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
@@ -32,11 +33,14 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.WeakHashMap;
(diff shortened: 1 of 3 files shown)


vietj:
as clarification, that would not work with malformed jars not having a manifest right ?

pk-work:
If the malformed jars part of the class path [1], it should work, shouldn't it? If they are not part of the classpath it will not work for malformed jars, because then they must be found with this trick [2].

[1]  String classpath = System.getProperty("java.class.path");
[2] String searchFile = "META-INF/MANIFEST.MF";

vietj:
@cescoffier can you make a final request and approve this PR ? you are our Java class-loading specialist :-)

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -380,43 +383,94 @@ public void unregisterVerticleFactory(VerticleFactory factory) {
     return factoryList;
   }
 
+  private static URL mapToURL(String path) {
+    try {
+      File file = new File(path);
```
cescoffier: Not sure if this works in all case. 

What about trying to build an url out of the string first and in case of failure try with File.

vietj: I agree, what does the new file brings ?

pk-work: @vietj Nothing, I can also write new File(path).toURI().toURL();
@cescoffier Do you mean something like this?

pk-work:
> However, I'm not sure about the usage of `File`. I've proposed a change for this part.
I assumed that there are only files on the Classpath, so using File.toURI().toURL() seemed the best way in my opinion.

> In addition, it needs to be tested in OSGi environment to be sure it works (Equinox and Felix at least).
Can someone assist here? I don't have such an environment.

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -380,43 +384,97 @@ public void unregisterVerticleFactory(VerticleFactory factory) {
...
   }
 
+  private static URL mapToURL(String path) {
+    try {
+        return new URL(path); 
+    } catch (MalformedURLException e) {
+      try {
+        return new File(path).toURI().toURL();
+      } catch (MalformedURLException e1) {
+        throw new IllegalStateException(e1);
```
vietj: new IllegalArgumentException is more appropriate

pk-work:
@cescoffier Can you check, if the latest changes fit to your request?

PR #2898, opened 2019-04-05
Json Pointer implementation

slinkydeveloper:
Implementation of RFC6901 - Json Pointer
Revisited #2330 

Signed-off-by: slinkydeveloper <francescoguard@gmail.com>

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -757,6 +757,8 @@ include::datagrams.adoc[]
 
 include::dns.adoc[]
 
+include::json-pointers.adoc[]
+
 [[streams]]
 include::streams.adoc[]
 
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -743,6 +743,8 @@ include::eventbus.adoc[]
 
 include::override/json.adoc[]
 
+include::json-pointers.adoc[]
+
 include::buffers.adoc[]
 
 include::net.adoc[]
@@ -756,8 +758,6 @@ include::filesystem.adoc[]
 include::datagrams.adoc[]
 
 include::dns.adoc[]
-
-include::json-pointers.adoc[]
 
 [[streams]]
 include::streams.adoc[]
--- a/src/main/asciidoc/json-pointers.adoc
+++ b/src/main/asciidoc/json-pointers.adoc
@@ -1,0 +1,22 @@
+== Json Pointers
+
+Vert.x provides an implementation of <a href="https://tools.ietf.org/html/rfc6901">Json Pointers from RFC6901</a>.
+You can use pointers both for querying and for writing. You can build your {@link io.vertx.core.json.pointer.JsonPointer} using
+a string, a URI or manually appending paths:
+
+[source,java]
+----
+{@link examples.JsonPointerExamples#example1Pointers}
+----
+
+After instantiating your pointer, you can use method {@link io.vertx.core.json.pointer.JsonPointer#queryJson(java.lang.Object)} to query
+a JsonObject/JsonArray. If you want to write inside a JsonObject/JsonArray use {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
+
+[source,java]
+----
+{@link examples.JsonPointerExamples#example2Pointers}
+----
+
+Vert.x Json Pointer algorithm abstracts from the queried data structure. That means that using {@link io.vertx.core.json.pointer.JsonPointer}
+you can query your own data structures. To define the methods to access to these, you must implement {@link io.vertx.core.json.pointer.JsonPointerIterator}
+and provide an instance of it to query/write methods
--- a/src/main/asciidoc/json-pointers.adoc
+++ b/src/main/asciidoc/json-pointers.adoc
@@ -9,14 +9,12 @@ a string, a URI or manually appending paths:
 {@link examples.JsonPointerExamples#example1Pointers}
 ----
 
-After instantiating your pointer, you can use method {@link io.vertx.core.json.pointer.JsonPointer#queryJson(java.lang.Object)} to query
-a JsonObject/JsonArray. If you want to write inside a JsonObject/JsonArray use {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
+After instantiating your pointer, use {@link io.vertx.core.json.pointer.JsonPointer#queryJson(java.lang.Object)} to query
+a JSON value. If can update a Json Value using {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
 
 [source,java]
 ----
 {@link examples.JsonPointerExamples#example2Pointers}
 ----
 
-Vert.x Json Pointer algorithm abstracts from the queried data structure. That means that using {@link io.vertx.core.json.pointer.JsonPointer}
-you can query your own data structures. To define the methods to access to these, you must implement {@link io.vertx.core.json.pointer.JsonPointerIterator}
-and provide an instance of it to query/write methods
+You can reuse Vert.x Json Pointer with custom object models by providing a custom implementation of {@link io.vertx.core.json.pointer.JsonPointerIterator}
--- a/src/main/asciidoc/json-pointers.adoc
+++ b/src/main/asciidoc/json-pointers.adoc
@@ -10,11 +10,11 @@ a string, a URI or manually appending paths:
 ----
 
 After instantiating your pointer, use {@link io.vertx.core.json.pointer.JsonPointer#queryJson(java.lang.Object)} to query
-a JSON value. If can update a Json Value using {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
+a JSON value. You can update a Json Value using {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
 
 [source,java]
 ----
 {@link examples.JsonPointerExamples#example2Pointers}
 ----
 
-You can reuse Vert.x Json Pointer with custom object models by providing a custom implementation of {@link io.vertx.core.json.pointer.JsonPointerIterator}
+You can use Vert.x Json Pointer with any object model by providing a custom implementation of {@link io.vertx.core.json.pointer.JsonPointerIterator}
--- a/src/main/asciidoc/override/json.adoc
+++ b/src/main/asciidoc/override/json.adoc
@@ -116,27 +116,3 @@ You get values from a JSON array using the `getXXX` methods, for example:
 ==== Encoding a JSON array to a String
 
 You use {@link io.vertx.core.json.JsonArray#encode} to encode the array to a String form.
-
-
-=== Json Pointers
-
-Vert.x provides an implementation of <a href="https://tools.ietf.org/html/rfc6901">Json Pointers from RFC6901</a>.
-You can use pointers both for querying and for writing. You can build your {@link io.vertx.core.json.pointer.JsonPointer} using
-a string, a URI or manually appending paths:
-
-[source,java]
-----
-{@link docoverride.json.Examples#example1Pointers}
-----
-
-After instantiating your pointer, you can use method {@link io.vertx.core.json.pointer.JsonPointer#queryJson(java.lang.Object)} to query
-a JsonObject/JsonArray. If you want to write inside a JsonObject/JsonArray use {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
-
-[source,java]
-----
-{@link docoverride.json.Examples#example2Pointers}
-----
-
-Vert.x Json Pointer algorithm abstracts from the queried data structure. That means that using {@link io.vertx.core.json.pointer.JsonPointer}
-you can query your own data structures. To define the methods to access to these, you must implement {@link io.vertx.core.json.pointer.JsonPointerIterator}
-and provide an instance of it to query/write methods
--- a/src/main/asciidoc/override/json.adoc
+++ b/src/main/asciidoc/override/json.adoc
@@ -121,7 +121,7 @@ You use {@link io.vertx.core.json.JsonArray#encode} to encode the array to a Str
 === Json Pointers
 
 Vert.x provides an implementation of <a href="https://tools.ietf.org/html/rfc6901">Json Pointers from RFC6901</a>.
-You can use pointers both for querying and for writing. You can build your {@link io.vertx.core.json.JsonPointer} using
+You can use pointers both for querying and for writing. You can build your {@link io.vertx.core.json.pointer.JsonPointer} using
 a string, a URI or manually appending paths:
 
 [source,java]
@@ -129,14 +129,14 @@ a string, a URI or manually appending paths:
 {@link docoverride.json.Examples#example1Pointers}
 ----
 
-After instantiating your pointer, you can use method {@link io.vertx.core.json.JsonPointer#queryJson(java.lang.Object)} to query
-a JsonObject/JsonArray. If you want to write inside a JsonObject/JsonArray use {@link io.vertx.core.json.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
+After instantiating your pointer, you can use method {@link io.vertx.core.json.pointer.JsonPointer#queryJson(java.lang.Object)} to query
+a JsonObject/JsonArray. If you want to write inside a JsonObject/JsonArray use {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
 
 [source,java]
 ----
 {@link docoverride.json.Examples#example2Pointers}
 ----
 
-Vert.x Json Pointer algorithm abstracts from the queried data structure. That means that using {@link io.vertx.core.json.JsonPointer}
-you can query your own data structures. To define the methods to access to these, you must implement {@link io.vertx.core.json.JsonPointerIterator}
+Vert.x Json Pointer algorithm abstracts from the queried data structure. That means that using {@link io.vertx.core.json.pointer.JsonPointer}
+you can query your own data structures. To define the methods to access to these, you must implement {@link io.vertx.core.json.pointer.JsonPointerIterator}
 and provide an instance of it to query/write methods
--- a/src/main/asciidoc/override/json.adoc
+++ b/src/main/asciidoc/override/json.adoc
@@ -116,3 +116,27 @@ You get values from a JSON array using the `getXXX` methods, for example:
 ==== Encoding a JSON array to a String
 
 You use {@link io.vertx.core.json.JsonArray#encode} to encode the array to a String form.
+
+
+=== Json Pointers
+
+Vert.x provides an implementation of <a href="https://tools.ietf.org/html/rfc6901">Json Pointers from RFC6901</a>.
+You can use pointers both for querying and for writing. You can build your {@link io.vertx.core.json.JsonPointer} using
+a string, a URI or manually appending paths:
+
+[source,java]
+----
+{@link docoverride.json.Examples#example1Pointers}
+----
+
+After instantiating your pointer, you can use method {@link io.vertx.core.json.JsonPointer#queryJson(java.lang.Object)} to query
+a JsonObject/JsonArray. If you want to write inside a JsonObject/JsonArray use {@link io.vertx.core.json.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
+
+[source,java]
+----
+{@link docoverride.json.Examples#example2Pointers}
+----
+
+Vert.x Json Pointer algorithm abstracts from the queried data structure. That means that using {@link io.vertx.core.json.JsonPointer}
+you can query your own data structures. To define the methods to access to these, you must implement {@link io.vertx.core.json.JsonPointerIterator}
+and provide an instance of it to query/write methods
--- a/src/main/java/examples/JsonPointerExamples.java
+++ b/src/main/java/examples/JsonPointerExamples.java
@@ -1,0 +1,33 @@
+package examples;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.json.pointer.JsonPointer;
+
+import java.net.URI;
+
+public class JsonPointerExamples {
+
+  public void example1Pointers() {
+    // Build a pointer from a string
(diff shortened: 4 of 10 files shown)


src/main/asciidoc/override/json.adoc
```diff
@@ -116,3 +116,27 @@ You get values from a JSON array using the `getXXX` methods, for example:
 ==== Encoding a JSON array to a String
 
 You use {@link io.vertx.core.json.JsonArray#encode} to encode the array to a String form.
+
+
+=== Json Pointers
```
vietj: this documentation will only be for Java (override).

I think we need to have a new doc for pointers so it is generated in all languages.

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+  boolean isLocalPointer();
+
+  boolean isParent(JsonPointer child);
+
+  /**
+   * Alias for toString()
+   *
+   * @return
+   */
+  String build();
```
vietj: why is this needed ?

vietj: because of polyglot ?

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   *
+   * @param objectToQuery the object to query
+   * @param iterator the json pointer iterator that provides the logic to access to the objectToQuery
+   * @param defaultValue default value if query result is null
+   * @return null if pointer points to not existing value, otherwise the requested value
+   */
+  Object queryOrDefault(Object objectToQuery, JsonPointerIterator iterator, Object defaultValue);
+
+  /**
+   * Query the provided object. <br/>
```
vietj: I think here we should refer to "json element" instead of "object"

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   * Note: if this pointer is a root pointer, this function returns the provided object
+   *
+   * @param object the object to queryJson
```
vietj: should be called _element_ instead

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Query the provided object tracing each element walked during the query, including the first and the result (if any).<br/>
+   * The first element of the stream is objectToQuery and the last is the result, or the element before the first null was encountered
+   *
+   * @param objectToQuery the object to query
+   * @param iterator the json pointer iterator that provides the logic to access to the objectToQuery
+   * @return the stream of walked elements
+   */
+  @GenIgnore
+  Stream<Object> tracedQuery(Object objectToQuery, JsonPointerIterator iterator);
```
vietj: why a Stream and not a list ?

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   * @return a reference to json if the write was completed, a reference to newElement if the pointer is a root pointer, null if the write failed
+   */
+  default Object writeJson(Object json, Object newElement, boolean createOnMissing) {
+    return write(json, JsonPointerIterator.JSON_ITERATOR, newElement, createOnMissing);
+  }
+
+  /**
+   * Copy a JsonPointer
+   *
+   * @return
```
vietj: missing javadoc

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  JsonPointer copy();
+
+  /**
+   * Build an empty JsonPointer
+   *
+   * @return a new empty JsonPointer
+   */
+  static JsonPointer create() {
+    return new JsonPointerImpl();
```
vietj: why is it useful ?

src/main/java/io/vertx/core/json/pointer/JsonPointerIterator.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @author Francesco Guardiani <a href="https://slinkydeveloper.github.io/">@slinkydeveloper</a>
+ *
+ */
+@VertxGen
+public interface JsonPointerIterator {
+
+  /**
+   * @param currentValue
+   * @return true if the current value is a queryable object
```
vietj: should be {@code true}

src/main/java/io/vertx/core/json/pointer/JsonPointerIterator.java
```diff
@@ -0,0 +1,100 @@
...
+   * @param currentValue
+   * @param value
+   * @return true if the operation is successful
+   */
+  boolean appendArrayElement(@Nullable Object currentValue, @Nullable Object value);
+
+  /**
+   * Instance of a JsonPointerIterator to query Vert.x Json structures
+   */
+  JsonPointerIterator JSON_ITERATOR = new JsonPointerIteratorImpl();
```
vietj: call it _INSTANCE_ instead

vietj:
I think we should try to enforce type safety, i.e have `JsonPointerIterator<T>` and have in `JsonPointer` generic methods declarations `<T> Object query(T objectToQuery, JsonPointerIterator<T> iterator) { return queryOrDefault(objectToQuery, iterator, null)`

and provide a `JsonPointerIterator` less method that works for Vert.x JSON package

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
slinkydeveloper: `build()` builds only the pointer https://tools.ietf.org/html/rfc6901#section-5
`buildUri()` builds the complete URI, prepending the starting URI if any https://tools.ietf.org/html/rfc6901#section-6

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
slinkydeveloper: It's useful to create a new empty pointer

slinkydeveloper:
> I think we should try to enforce type safety, i.e have `JsonPointerIterator<T>` and have in `JsonPointer` generic methods declarations `<T> Object query(T objectToQuery, JsonPointerIterator<T> iterator) { return queryOrDefault(objectToQuery, iterator, null)`
> 
> and provide a `JsonPointerIterator` less method that works for Vert.x JSON package

I think it's not a good idea because query method accepts any Vert.x Json type, included strings and boxed primitives: https://github.com/eclipse-vertx/vert.x/pull/2898/files#diff-383e4e3994bf5edbbcf185bfb46e2e50R190
Vert.x Json types doesn't have a common parent, so that T will always be Object.

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
vietj: it seems like `toString()`, perhaps we should remove this and rename `buildUri()` to `toUri()`

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
vietj: why are empty pointer useful ? I thought they are immutable and created from a string parsed

vietj:
@slinkydeveloper my point is that actually for Vert.x json object, we have an overloaded method that does not need to specify the `JsonPointerIterator` instance

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
slinkydeveloper: No, JsonPointer is mutable and offers the `append` methods to build the pointer programmatically without caring about encoding. I would avoid making it immutable because it's a waste of memory, in particular in some phases of json validator

slinkydeveloper: If we want to make JsonPointer immutable we can create a JsonPointerBuilder, but it seems a little bit overkilling to me. wdyt?

slinkydeveloper:
In other usages of `JsonPointer` i think the users still needs to query some dynamic data structure. If we make something like `JsonPointerIterator<OBJECT_TYPE, ARRAY_TYPE, VALUE_TYPE>` then we need to do:

```
<X, Y, Z> Object queryObject(X o, JsonPointerIterator<X, Y, Z> iterator)
<X, Y, Z> Object queryArray(Y a, JsonPointerIterator<X, Y, Z> iterator)
<X, Y, Z> Object queryValue(Z v, JsonPointerIterator<X, Y, Z> iterator)
```

And it's a thing that we should avoid, because of the dynamic nature of Json and JsonPointer

src/main/asciidoc/json-pointers.adoc
```diff
@@ -0,0 +1,22 @@
...
+Vert.x provides an implementation of <a href="https://tools.ietf.org/html/rfc6901">Json Pointers from RFC6901</a>.
+You can use pointers both for querying and for writing. You can build your {@link io.vertx.core.json.pointer.JsonPointer} using
+a string, a URI or manually appending paths:
+
+[source,java]
+----
+{@link examples.JsonPointerExamples#example1Pointers}
+----
+
+After instantiating your pointer, you can use method {@link io.vertx.core.json.pointer.JsonPointer#queryJson(java.lang.Object)} to query
```
vietj: simply "use can use" instead of "you can use method"

src/main/asciidoc/json-pointers.adoc
```diff
@@ -0,0 +1,22 @@
...
+a JsonObject/JsonArray. If you want to write inside a JsonObject/JsonArray use {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
```
vietj: JSON object or array : in JS JsonObject does not make sense

vietj: _You can update a JSON object or array thanks to_

src/main/asciidoc/json-pointers.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+[source,java]
+----
+{@link examples.JsonPointerExamples#example2Pointers}
+----
+
+Vert.x Json Pointer algorithm abstracts from the queried data structure. That means that using {@link io.vertx.core.json.pointer.JsonPointer}
```
vietj: _You can reuse Vert.x Json Pointer with custom object models too by providing a custom implementation of {@link io.vertx.core.json.pointer.JsonPointerIterator}._

src/main/asciidoc/index.adoc
```diff
@@ -757,6 +757,8 @@ include::datagrams.adoc[]
 
 include::dns.adoc[]
 
+include::json-pointers.adoc[]
```
vietj: I think this should be after the JSON doc

src/main/asciidoc/json-pointers.adoc
slinkydeveloper: maybe is better saying a JSON value, since this function manages JSON objects, arrays but also primitives

slinkydeveloper: If you have a root pointer and you pass a primitive, let's say a string, to the query function it returns the string (it's a valid operation for json pointers)

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
lukehutch: The intent of having a separate `JsonPointer` class (rather than building a json pointer query API directly into `JsonObject` / `JsonArray`, with the query string re-interpreted for each new lookup) seems to be so that a `JsonPointer` is created once and reused wherever possible. Maybe then the immutable model makes more sense, since the creation costs can in many cases be amortized over multiple uses?

One thing I would like to see is the ability to pre-build a query with named parameter slots, and then be able to pass a `Map` of concrete arguments into the `query` functions. When a named parameter slot is reached while following the query path, the value of the parameter will be looked up in the map. If this query form were supported, it would dramatically reduce the number of `JsonPointer` objects you would need to create, since the vast majority of queries could be built just once on startup, rather than having to be re-build for each new query. This would allow `JsonPointer` to act sort of like a prepared statement in SQL.

lukehutch: That said, for many uses, users will probably simply prefer never having to deal with `JsonPointer` directly, but rather just be able to do something like `jsonObject.query("/users/" + userId.replace("~", "~0").replace("/", "~1") + "/age")`, or maybe something like the following could be supported that supports automatic escaping: `jsonObject.query("/users", userId, "age")`. (It's a real shame the json pointer RFC escaping syntax is so weird...)

Except for the most performance-critical applications, direct querying of this form will simpler and more natural than having to work through a separate `JsonPointer` class.

slinkydeveloper: `JsonPointer` needs to exist because it can be used for querying also non Vert.x data structures (look at `JsonPointerIterator`) and the notation `JsonObject.query()` would not work because of polyglot (json objects in other languages supported by Vert.x are not instances of Vert.x `JsonObject`)

I don't like the argument of making immutable because we need some kind of builder (that adds an extra step to usage of the api) and it's not consistent with Vert.x Json APIs, where both JsonArray and JsonObject are mutable.

For the parameters, seems like a cool idea but I would like to keep it out of the scope of this PR. Since Json pointers are very simple to create, you can just create methods to generate the pointers you need, like:
```
public void createAgePointer(String userId) {
  return JsonPointer.create().append("users").append(userId).append("age")
}
```
And there is another problem: "{userId}" is a valid object key for json pointers, so that means that we should find some weird escaping too for it or we can allow it only using `appendParameter()`, making it harder to use/understand for people

lukehutch: @slinkydeveloper All good points. You're right about escaping, so I guess `appendParameter()` is the only option. I still think it would be a good idea to be able to parameterize these. Users that don't want the complexity can ignore that part of the API.

I just submitted a PR, #2931, with the simple `query()` syntax I was suggesting. I didn't know that other languages don't use `JsonObject` / `JsonArray`. However, for JVM languages that have those classes, this makes the API of those classes much simpler to use for deep object graphs, so I still think it's worth it.

src/main/asciidoc/json-pointers.adoc
```diff
@@ -0,0 +1,20 @@
...
+You can use pointers both for querying and for writing. You can build your {@link io.vertx.core.json.pointer.JsonPointer} using
+a string, a URI or manually appending paths:
+
+[source,java]
+----
+{@link examples.JsonPointerExamples#example1Pointers}
+----
+
+After instantiating your pointer, use {@link io.vertx.core.json.pointer.JsonPointer#queryJson(java.lang.Object)} to query
+a JSON value. If can update a Json Value using {@link io.vertx.core.json.pointer.JsonPointer#writeJson(java.lang.Object, java.lang.Object)}:
```
vietj: typo : You can update

src/main/asciidoc/json-pointers.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+[source,java]
+----
+{@link examples.JsonPointerExamples#example2Pointers}
+----
+
+You can reuse Vert.x Json Pointer with custom object models by providing a custom implementation of {@link io.vertx.core.json.pointer.JsonPointerIterator}
```
vietj: "You can use Vert.x Json Pointer with any object model by providing a custom implementation ... "

src/main/java/examples/JsonPointerExamples.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.net.URI;
+
+public class JsonPointerExamples {
+
+  public void example1Pointers() {
+    // Build a pointer from a string
+    JsonPointer pointer1 = JsonPointer.from("/hello/world");
+    // Build a pointer from an URI
+    JsonPointer pointer2 = JsonPointer.fromURI(URI.create("#/hello/world"));
```
vietj: don't use URI here as it will not be translated to other languages.

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+  boolean isRootPointer();
+
+  /**
+   * Return true if the pointer is local (URI with only fragment)
+   *
+   * @return
+   */
+  boolean isLocalPointer();
+
+  boolean isParent(JsonPointer child);
```
vietj: missing javadoc

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+   * Return true if the pointer is a root pointer
+   *
+   * @return
+   */
+  boolean isRootPointer();
+
+  /**
+   * Return true if the pointer is local (URI with only fragment)
```
vietj: Instead :

"@return {@code true} if the pointer is local (URI with only fragment)"

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   * @return
+   */
+  boolean isLocalPointer();
+
+  boolean isParent(JsonPointer child);
+
+  /**
+   * Build a <a href="https://tools.ietf.org/html/rfc6901#section-5">string representation</a> of the JSON Pointer
+   *
+   * @return
```
vietj: either remove `@return` or add a description.

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  @Override
+  String toString();
+
+  /**
+   * Build a <a href="https://tools.ietf.org/html/rfc6901#section-6">URI representation</a> of the JSON Pointer
+   *
+   * @return
```
vietj: either remove `@return`tag or add a description.

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  URI toURI();
+
+  /**
+   * Returns the underlying URI without the fragment
+   *
+   * @return
```
vietj: either remove `@return`tag or add a description.

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  URI getURIWithoutFragment();
+
+  /**
+   * Append unescaped path to JsonPointer. <br/>
```
vietj: rather "Append an unescaped {@code path} to this pointer"

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   * Note: If you provide escaped path the behaviour is undefined
+   *
+   * @param path unescaped path
```
vietj: the unescaped path value

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   * @return
```
vietj: complete return tag

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+  JsonPointer append(String path);
+
+  /**
+   * Append index path to JsonPointer
+   *
+   * @param i index
+   * @return
+   */
+  @Fluent
+  JsonPointer append(int i);
```
vietj: name it `index` instead

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   * Note: If you provide escaped path the behaviour is undefined
+   *
+   * @param path unescaped path
+   * @return
+   */
+  @Fluent
+  JsonPointer append(String path);
+
+  /**
+   * Append index path to JsonPointer
```
vietj: `append the {@code index} to this pointer`

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   *
+   * @param i index
+   * @return
+   */
+  @Fluent
+  JsonPointer append(int i);
+
+  /**
+   * Append unescaped list of paths to JsonPointer <br/>
```
vietj: of {@code paths}

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   * Note: If you provide escaped paths the behaviour is undefined
+   *
+   * @param paths unescaped paths
+   * @return
```
vietj: complete or remote return tag

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  @Fluent
+  JsonPointer append(List<String> paths);
+
+  /**
+   * Remove last append
```
vietj: Not clear what this does.

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   *
+   * @return
```
vietj: complete or remote return tag

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  @Fluent
```
vietj: is it really fluent ?

src/main/java/io/vertx/core/json/pointer/JsonPointer.java
```diff
@@ -0,0 +1,236 @@
...
+  JsonPointer parent();
+
+  /**
+   * Query the objectToQuery using the provided JsonPointerIterator. <br/>
```
vietj: {@code objectToQuery}

vietj: {@link JsonPointerIterator}

PR #2918, opened 2019-04-17
HTTP event handler should not be called when holding a lock and signa…

### LINKED ISSUE #2915: Signal end of stream using a sentinel
Currently streams implementations in core relies on a boolean flag to determine when a stream has ended and the delivery (possibly delayed) of the corresponding event. 

This can be simplified by using an _end_ sentinel that is the last event of the stream removing some complexity in the handling of the delayed stream end.

### LINKED ISSUE #2916: HTTP event handler should not be called when holding a lock
We still have event handlers that are called while holding a lock which can lead to deadlocks.

We should remove them.

vietj:
…l end of stream using a sentinel. Fixes #2915 - Fixes #2916

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -198,12 +198,12 @@ class Http1xClientConnection extends Http1xConnectionBase<WebSocketImpl> impleme
     private final Http1xClientConnection conn;
     private final ContextInternal context;
     private final Future<HttpClientStream> fut;
+    private final InboundBuffer<Object> queue;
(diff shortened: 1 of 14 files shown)


sfitts:
@vietj any ETA on 3.7.1?  One of these fixes also fixed a break in `FlowableReadStream` when a downstream operator calls `request` in the middle of processing an `onNext`.  When it came to the end of the stream it would fail to recognized that it still had data in the pending queue and instead would immediately call `onComplete`.  I test `3.7.1-SNAPSHOT` and the break is gone (thankfully, was not looking forward to coming up with a clean reproduction for that one ;)).

vietj:
3.7.1 should be release this month

PR #2919, opened 2019-04-22
Expose isClosed() method in WebSockets

gabixdev:


vietj:
See https://github.com/eclipse-vertx/vert.x/issues/3020

PR #2925, opened 2019-04-25
provide empty object/array instance

MaiCw4J:
To reduce JsonObject/JsonArray  instance creation
In some cases  we just want to return an 'empty object'. at present I had to create a new object/array 
**example:**

>   private JsonObject getJsonObjectByName(String name) {
    if (condition) {
      JsonObject entries = new JsonObject();
      // do something
      return entries;
    } else {
      return Json.emptyObject(); // reduce creation
    }
  }

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -30,6 +30,7 @@ import java.util.Base64;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Collections;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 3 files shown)


tsegismont:
That will make the empty object/array immutable and this is not the usual behavior.

MaiCw4J:
> That will make the empty object/array immutable and this is not the usual behavior.

;) Because it's singleton and If it can be modified It is no longer 'empty' object/array.

pmlopes:
@tsegismont and @MaiCw4J I do see some value on this contribution. For example in `auth` and `web` I need to declare in many places the *emtpyJsonObject*, *emptyJsonArray* as a static final field as a repetition of this PR.

The most useful use case is on JWT, for example, I need to check if an object hierarchy has a given property, for example:

```
root.claims.permissions.print
```

Claims and permissions are `JsonObjects`s.

Given the nature of json and JWTs there is no guarantee that this hierarchy will exist at runtime so I do:

```java
if (root.getJsonObject("claims", EMPTY_JSON).getJsonObject("permissions", EMPTY_JSON).containsKey("print") {
...
```

This pattern saves me the work of checking for existence or nulls all the way, sure if will not short circuit, but it won't allocate garbage by creating a mutable `new JsonObject()` all the time.

I also see the case @MaiCw4J wrote, specially on web, where in say, REST APIs where a create operation returns an empty object just to be sure older user-agents or clients will not break if no content is returned.

My 2cts, is that this is definitely a nice to have improvement, **BUT** it should be clearly documented, just like the JDK Collections counter parts that the `emptyXYZ()` methods are immutable and singleton.

PR #2931, opened 2019-05-03
Add `query(Object... path)` method for simple JSON querying

lukehutch:
This PR adds `JsonObject::query(Object... path)` and `JsonArray::query(Object... path)` methods for simple query of `JsonObject` / `JsonArray` structures, e.g.:

```
Object age = myDoc.query("users", userId, "age");
```


lukehutch:
See #449.

PR #2951, opened 2019-05-17
Include graal native-image metadata so users can build native images …

pmlopes:
…without issues.

This PR is overlapping with https://github.com/netty/netty/pull/9118 once that PR gets merged on netty and core gets updated to use that specific version, we can drop half of the config bellow.

The PR does not include any substitution code, for the same reason as above, it would be duplicated over the 2 repos.


pmlopes:
This PR should only be merged after netty accepts the changes in the referred pr.

This can however be used to try out that the upstream changes will work out of the box.

PR #2964, opened 2019-05-22
New message codec for service proxy feature

eutkin:
I would like to add new message codec for vert.x.

Vert.x is an actor model based system in which actors interact via databus. 
A message is serialized (only cluster vert.x) before transporting into databus and deserialized upon receiving. 
Default implementation allows using primitive types and json as well as custom message codecs.

What I propose new message codec for Object[]. 
By default, vert.x would’ve used Json Array, serializing every element first into 
String then into bytes. My codec will iterate through the array of objects, 
finds existing system and default codecs for each element and use them for serialization. 
It provides straightforward binary serialization and uses flexibility of other custom codecs.

This feature will allow you to implement another feature in the [vertx-service-proxy project](https://github.com/vert-x3/vertx-service-proxy/issues/17)
, namely "Pluggable marshallers".

This codec is needed to serialize the array of method arguments not into JsonObject, and then into bytes, but immediately into bytes. This will remove some limitations in the Service Proxy and improve performance.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/CodecManager.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/CodecManager.java
@@ -21,6 +21,8 @@ import io.vertx.core.json.JsonObject;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.function.BiFunction;
+import java.util.stream.Stream;
 
 /**
  * @author <a href="http://tfox.org">Tim Fox</a>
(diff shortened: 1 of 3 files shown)


eutkin:
Example:

```java
public class VerticleOnNode1 extends AbstractVerticle {
    public void start() {
        vertx.eventBus()
             .registerDefaultCodec(SampleDataObject.class, new SampleDataObjectMessageCodec());
        vertx.eventBus().send("address", new Object[] { "textValue", new SampleDataObject() });
    }
}

public class VerticleOnNode2 extends AbstractVerticle {
    public void start() {
        vertx.eventBus()
             .registerDefaultCodec(SampleDataObject.class, new SampleDataObjectMessageCodec());
        vertx.eventBus().consumer("address")
             .handler(mes -> {
                  String textValue = mes.body()[0];
                  SampleDataObject sample = mes.body()[1];
             });
    }
}
```

Service Proxy:

```java
public class SomeDatabaseServiceVertxEBProxy implements SomeDatabaseService {
  @Override
  public  void save(String collection, JsonObject document, Handler<AsyncResult<Void>> result){
    if (closed) {
      result.handle(Future.failedFuture(new IllegalStateException("Proxy is closed")));
      return;
    }
   /* Serialize data in text format
    JsonObject args = new JsonObject();
    args .put("collection", collection);
    args .put("document", document);
   */
  // now, serialize data in binary format, supporting user's codecs.
  // and DataObject may not have Constructor(JsonObject json) and toJson method.
   Object[] args = new Object[] {collection, document};

    DeliveryOptions _deliveryOptions = (_options != null) ? new DeliveryOptions(_options) : new DeliveryOptions();
    _deliveryOptions.addHeader("action", "save");
    _vertx.eventBus().<Void>send(_address, args, _deliveryOptions, res -> {
      if (res.failed()) {
        result.handle(Future.failedFuture(res.cause()));
      } else {
        result.handle(Future.succeededFuture(res.result().body()));
      }
    });
  }
}
```

eutkin:
Signed-off-by: utkinev2014@gmail.com

eutkin:
i will recreate pull request

PR #2965, opened 2019-05-22
Fixes #2963

### LINKED ISSUE #2963: Cannot close file if one of the write operations fails
Version :  
Vert.x : 3.7.0
Vert.x - Web : 3.7.0

I test thing with vert.x and I discover that :
-> If file transfer start , create file & fill it.
-> If disk size becomes full.
-> BodyHandler does not trigger an Http response.
-> The guilty seems to be the wich call close on the file but because it remains byte to write, closed is defered & never called, callback never called. 
https://github.com/eclipse-vertx/vert.x/blob/ba47fac59a4b78ff858b3858e7971083a2edce97/src/main/java/io/vertx/core/http/impl/HttpServerFileUploadImpl.java#L166-L180

because close will defer to :
https://github.com/eclipse-vertx/vert.x/blob/645f6d4d41e0301155501b1e287ac49d6e60573b/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java#L521-L531

and the because of writesOutstanding != 0 (some byte left because we cannot write the all things), close is defered and never called because the completion handler which is the only part who called the defered method is always in failure (no space on disk) : 
https://github.com/eclipse-vertx/vert.x/blob/645f6d4d41e0301155501b1e287ac49d6e60573b/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java#L163-L189
 
Reproducer : 
I have create a tempory partition on an unbuntu desktop (sudo mount -t tmpfs -o size=90M tmpfs /media/testUpload/) and try to download a file greater that 90Mo.

aygalinc:
Signed-off-by: aygalinc colin.aygalinc@gmail.com

### DIFF
--- a/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java
+++ b/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java
@@ -115,6 +115,34 @@ public class AsyncFileImpl implements AsyncFile {
     });
   }
 
+  public AsyncFileImpl(VertxInternal vertx,  AsynchronousFileChannel fileChannel, boolean isAppend, ContextInternal context) {
+    this.vertx = vertx;
+    this.ch = fileChannel;
+    this.closed = ch.isOpen();
+
+    try {
+      if (isAppend) {
+        writePos = ch.size();
+
+      }
+    } catch (IOException e) {
+      throw new FileSystemException(e);
+    }
+
+    this.context = context;
+    this.queue = new InboundBuffer<>(context, 0);
+    queue.handler(buff -> {
+      if (buff.length() > 0) {
+        handleBuffer(buff);
+      } else {
+        handleEnd();
+      }
+    });
+    queue.drainHandler(v -> {
+      doRead();
+    });
+  }
+
   @Override
   public void close() {
     closeInternal(null);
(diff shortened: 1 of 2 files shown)


vietj:
can you add a test for this ?

aygalinc:
I don't know how to create a unit test with a folder which is bound in size. 
If you have any input, maybe I can do it.

vietj:
I think a way to test would be to provide an alternative constructor for AsyncFileImpl that uses an `AsynchronousFileChannel` for which we can provide an implementation that behaves like if the disk were full.

aygalinc:
Ok. 
And where test must be located ? 
As soon that I have time i try to have hand on it.

vietj:
in vertx test suite there is a FileSystemTest class

aygalinc:
Test case is implemented. 
I'm just aking why it is so important to keep the AsyncFileImpl thread safe ?

tsegismont:
@vietj instead of a refactoring `AsyncFileImpl`, we could write the test so that it saves the uploaded file to [`/dev/full`](https://en.wikipedia.org/wiki//dev/full). This would only work on Linux but with jUnit assumptions the test can be skipped.
It would run on CI anyway.

tsegismont:
This test reproduces the issue on `3.7` and `master` branches:

```java
  @Test(timeout = 5000)
  public void testWriteOnFullFileSystem() {
    Buffer buff = TestUtils.randomBuffer(512);
    AsyncFile asyncFile = vertx.fileSystem().openBlocking("/dev/full", new OpenOptions());

    asyncFile.write(buff, onFailure(write -> {
      asyncFile.close(onSuccess(close -> testComplete()));
    }));

    await();
  }
```

tsegismont:
Superseded by #3413

PR #2969, opened 2019-05-28
RecordParser delimited parsing does not read the last file line

### LINKED ISSUE #2967: RecordParser delimited parsing does not read the last file line

tsegismont:
Fixes #2967

### DIFF
--- a/src/main/java/io/vertx/core/parsetools/impl/RecordParserImpl.java
+++ b/src/main/java/io/vertx/core/parsetools/impl/RecordParserImpl.java
@@ -42,6 +42,8 @@ public class RecordParserImpl implements RecordParser {
   private Handler<Buffer> eventHandler;
   private Handler<Void> endHandler;
   private Handler<Throwable> exceptionHandler;
+  private boolean parsing;
+  private boolean streamEnded;
 
   private final ReadStream<Buffer> stream;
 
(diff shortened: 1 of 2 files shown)


tsegismont:
@vietj this fixes the issue without breaking compatibility so it's fine for 3.8

Yet there are a couple of problems with `RecordParser`:
- it's only safe in verticle code (not truly thread safe)
- back-pressure is managed with the internal buffer state, whereas it could be simplified with an `InboundBuffer`

If that's fine with you, I can file an issue for 4.0.

src/main/java/io/vertx/core/parsetools/impl/RecordParserImpl.java
```diff
@@ -172,6 +174,10 @@ public RecordParser maxRecordSize(int size) {
   }
 
   private void handleParsing() {
+    if (parsing) {
```
vietj: why is this needed ?

vietj: ah I see

src/main/java/io/vertx/core/parsetools/impl/RecordParserImpl.java
```diff
@@ -210,9 +216,25 @@ private void handleParsing() {
       buff = EMPTY_BUFFER;
     } else {
       buff = buff.getBuffer(start, len);
+      if (streamEnded) {
```
vietj: we should try to move this logic upper in the code block where there is:

```
        if (delimited) {
          next = parseDelimited();
        } else {
          next = parseFixed();
        }
```

so basically it would be something like:

```
        int next;
        if (delimited) {
          next = parseDelimited();
        } else {
          next = parseFixed();
        }
        Buffer event;
        if (next == -1) {
          if (streamEnded) {
            // ...
          } else {
            ReadStream<Buffer> s = stream;
            if (s != null) {
              s.resume();
            }
            break;
          }
        } else {
          event = buff.getBuffer(start, next);
        }
```

src/main/java/io/vertx/core/parsetools/impl/RecordParserImpl.java
```diff
@@ -210,9 +216,25 @@ private void handleParsing() {
...
+          handler.handle(next == -1 ? buff : buff.getBuffer(0, next));
+        }
+      }
     }
     pos -= start;
     start = 0;
+    if (streamEnded) {
+      end();
+    }
+    parsing = false;
```
vietj: should use try/finally construct

src/main/java/io/vertx/core/parsetools/impl/RecordParserImpl.java
tsegismont: What's wrong with the code? I would prefer to avoid a refactoring and simply patch 3.8. Then, as I said in comments earlier, rework the impl for 4.0.

vietj: I believe we should not rework implemnetation for Vert.x 4 in order to make 3.x and 4 code similar to ease patch ports

tsegismont: See my [initial comment](https://github.com/eclipse-vertx/vert.x/pull/2969#issuecomment-496454962)

If we fix this right, we will break compatibility

vietj: actually we can't use inbound queue for such case because inbound queue requires to have a Vert.x context attached to it

tsegismont: Right, this is why I proposed to leave the fix as-is for 3.x and revisit impl in 4.x

vietj: you said you want to revisit using InboundQueue which is not feasible. How about fixing it the right away and we have deal done without requiring to revisit code that works ?

tsegismont: Not feasible in 4.x?

tsegismont: @vietj I've update the PR as requested

tsegismont:
@vietj can you please take another look?

src/main/java/io/vertx/core/parsetools/impl/RecordParserImpl.java
```diff
@@ -42,6 +42,8 @@
   private Handler<Buffer> eventHandler;
   private Handler<Void> endHandler;
   private Handler<Throwable> exceptionHandler;
+  private boolean parsing;
```
vietj: this new flag that seems to be there to avoid reentrancy is actually not tested, you need to add a test for it

tsegismont: Done

vietj:
thanks, I'll backport it to 3.8

tsegismont:
Thanks for backporting ;-)

PR #2970, opened 2019-05-28
CLI updates

pmlopes:

Overall CLI changes as presented here: https://groups.google.com/forum/#!topic/vertx-dev/iLb51VWp6dQ

* override of commands is now allowed with a `priority` field to decide which one to accept
* `Default-Verticle-Factory` allows selecting a desired factory from the MANIFEST
* Less reflection (useful for native)


pmlopes:
Summary of the changes:

* This code is (to my knowledge) backward compatible
* Existing commands can be overriden by a new command with higher priority (default priority: 0)
* A new `Default-Verticle-Factory` flag is read from the `MANIFEST.MF` which will prefix all main verticles with `xyz:` IIF the main verticle has no specific factory already in the name.
* Converters were refactored to use less reflection
* Command factory was refactored to use less reflection

Use cases:

Commands can be overriden as:

```java
@Name("run", priority = 100)
public class MyRunCommand extends RunCommand {
  ...
}
```

Polyglot verticles can now be declared as:

```
Default-Verticle-Factory: js
Main-Verticle: index
```

Issue #2973, opened 2019-05-30
Inbound buffer silently ignores an exception


manzur:
Hey,

I have been playing with a simple TCP server, and it seems like exceptions thrown in `socket::handler` are silently ignored. None of the below exception handlers are passed to `InboundBuffer` as a result an exception neither being caught or logged:

```java
public class SimpleServer {
	public static void main(String[] args) {
		System.err.println("Application starting...");
		
		final Vertx vertx = Vertx.vertx();
		final FileSystem fs = vertx.fileSystem();

		final NetServer server = vertx.createNetServer();
		server.exceptionHandler(System.err::println);

		server.connectHandler(socket -> {
			socket.exceptionHandler(System.err::println);
			socket.handler($ -> fs.readDirBlocking("/WHERE_IS_MY_EXCEPTION").forEach(System.out::println));
		}).listen(3456);
	}
}
```

lfmunoz:
I'm having errors with parsing the input stream and it not reporting errors.

Seems like NetSocketImpl should have an exception handler for InboundBuffer. 
I have a set a breakpoint in InboundBuffer in handleEvent shows

io.vertx.ext.stomp.impl.FrameException: Incorrect header value - the header uses an illegal escaped character '7' (55)

and it is silently ignored.

Issue #2948, opened 2019-05-17
ServerWebSocket handler should not ignore exception


FulaiZhang-cn:
here is reproduce code:
--

import io.vertx.core.AbstractVerticle;
import io.vertx.core.Vertx;
import io.vertx.core.buffer.Buffer;
import io.vertx.core.http.HttpClient;
import io.vertx.core.http.HttpServer;
import io.vertx.core.http.ServerWebSocket;
import io.vertx.ext.web.Router;
import io.vertx.ext.web.RoutingContext;

public class ReproduceWebSocketException extends AbstractVerticle {

    public static void main(String[] args) {
        Vertx vertx = Vertx.vertx();
        vertx.deployVerticle(new ReproduceWebSocketException(), ar->{
            runClient(vertx);
        });
    }

    private static void runClient(Vertx vertx) {
        HttpClient client = vertx.createHttpClient();
        client.websocket(8088,"localhost", "/", ws->{
            for (int i = 0; i < 10; i++) {
                ws.write(Buffer.buffer("hello"));
            }
        });
    }

    @Override
    public void start() throws Exception {
        HttpServer server = vertx.createHttpServer();
        Router router = Router.router(vertx);
        router.route("/").handler(this::handlerWebsocket);
        server.requestHandler(router).listen(8088);

        System.out.println("server started");
    }

    private void handlerWebsocket(RoutingContext context) {
        ServerWebSocket socket = context.request().upgrade();
        socket.handler(buff->{
            System.out.println("got buffer:" + buff.toString());

            //this exception is ignored
            throw new RuntimeException("error");
        });
        socket.closeHandler(v->{
            System.out.println("close");
        });
        socket.exceptionHandler(th->{
            th.printStackTrace();
        });
    }
}

there have a exception throwed in :

socket.handler(buff->{
        System.out.println("got buffer:" + buff.toString());

        //this exception is ignored
        throw new RuntimeException("error");
    });

while I am debugging source code, found that `InboundBuffer<Buffer> pending` in `WebSocketImplBase` doesn't setup exceptionHandler.
so I think this is a bug.

doctorpangloss:
`exceptionHandler` is for exceptions thrown by the stream, i.e. a closed connection (`SocketException`), not for exceptions thrown by your user-supplied handler. If you'd like to handle those exceptions wrap your handler in a try-catch.

PR #2974, opened 2019-05-30
Inbound exception handler

manzur:
Fixes an issue reported in https://github.com/eclipse-vertx/vert.x/issues/2973 . Some of the other modules that depend on `InboundBuffer` still need to be fixed(the rest of them have `try/catch` in the corresponding handlers).

### DIFF
--- a/src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
+++ b/src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
@@ -246,7 +246,6 @@ public class InboundBuffer<E> {
     Handler<Throwable> handler;
     synchronized (this) {
       if ((handler = exceptionHandler) == null) {
-        log.error("Unhandled exception", err);
         return;
       }
     }
--- a/src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
+++ b/src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
@@ -14,6 +14,8 @@ import io.vertx.core.Context;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.impl.Arguments;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
 
 import java.util.ArrayDeque;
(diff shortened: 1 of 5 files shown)


vietj:
we need a test, thanks

src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
```diff
@@ -243,6 +246,7 @@ private void handleException(Throwable err) {
     Handler<Throwable> handler;
     synchronized (this) {
       if ((handler = exceptionHandler) == null) {
+        log.error("Unhandled exception", err);
```
vietj: I prefer to not have this here and instead let the `InboundBuffer` user handle logging like you did for `NetSocket`

manzur: Yep, I have added logging to `InboundBuffer` as well in my previous commits, but I think it's a good practice to at least log the error when there's no explicit handler set on each level.

Actually, I would even re-throw the exception even further when there's no handler as it will help to detect those kind of errors earlier, but , I assume, that wouldn't be really backward compatible at the moment.

manzur:
Sure, I will add tests for this change. I also found similar issue with WebSocket version [link](https://github.com/eclipse-vertx/vert.x/issues/2948), which I will fix along that. One thing that I forgot was signing ECA, which I will do with coming commits.

src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
vietj: please remove the logging from there, the class using it should be responsible handling logging.

manzur: push a commit that drops logging for unhandled exceptions.

vietj:
this still does not pass ECA

manzur:
> this still does not pass ECA

Opened a new PR with signed and squashed commits: [PR-3032](https://github.com/eclipse-vertx/vert.x/pull/3032). Closing this.

PR #2976, opened 2019-06-02
Check promise

vietj:
See https://github.com/vert-x3/wiki/wiki/RFC:-Future-API#asynchronous-type

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -359,7 +359,7 @@ This example uses:
 
 * {@link io.vertx.core.Future#compose(java.util.function.Function)}: when the current future completes,
 run the given function, that returns a future. When this returned future completes, it completes the composition.
-* {@link io.vertx.core.Future#compose(io.vertx.core.Handler,io.vertx.core.Future)}: when the current future
+* {@link io.vertx.core.Future#compose(io.vertx.core.Handler,io.vertx.core.Promise)}: when the current future
 completes, run the given handler that completes the given `future` (next).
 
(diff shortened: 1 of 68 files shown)


umeshsehgal:
> Hello Experts,


I have been trying to implement a scenario where the verticle may be
executing a long running call. For e.g. Opening a socket with external
application (MS Exchange) and retrieving records.

I tried implementing the logic using "executeBlocking" (see below:) and
inside that block I use "*Thread.sleep(long)"* to mimic long running call,
however I see that when I fire two concurrent requests the  second one gets
queued up and completes only after the first one is done.

Example code
----------------------------------------------------------------------------------------------------------
static int a=0;
  @Override

    vertx.createHttpServer().requestHandler(request -> {

     vertx.<String>executeBlocking(future -> {
a++;
String result = null;
                try {
 if(a%2==0) {
Thread.sleep(10000);
result = "armadillos!";
  }else {
  Thread.sleep(5000);
  result = "Amol!";
 }
} catch (InterruptedException e) {e.printStackTrace();}
        future.complete(result);
      }, res -> {
        if (res.succeeded()) {
          request.response().putHeader("content-type",
"text/plain").end(res.result());
        } else {
          res.cause().printStackTrace();
        }
      });
    }).listen(8080);
  }
--------------------------------------------------------------------------------------------------------------------

I'm not sure if this  is the right way to implement the required use case,
but would like to hear your thoughts on this.

Thanks,
Umesh

vietj:
it is not clear how thread sleep interacts with the socket you mentioned, so it is hard to comment.

PR #2983, opened 2019-06-07
Promise

vietj:


vietj:
@jponge well spotted, I renamed the lambda parameters to "promise" instead of "future"

PR #2990, opened 2019-06-14
EventBus request method for request-response pattern - this closes #2989

### LINKED ISSUE #2989: EventBus request method for request-response pattern
Currently the `EventBus` overloads the `send` method to implement one-way messaging and request-response messaging (using a reply handler). Although both methods share a lot they have a different semantic. This will cause an issue in Vert.x 4 asynchronous model based on futures: the future form of request-response pattern will have the same name than the one-way pattern.

We introduce a new method `request` for the request-response pattern that will always specify a reply handler.

```java
// Before
eventBus.send("addr", payload, reply -> { ... };

// After
eventBus.request("addr", payload, reply -> { ... };
```

The `send` methods having a reply handler are deprecated and shall be removed in Vert.x 4.

Likewise, the `Message` reply methods having an handler are deprecated in favour of `replyAndRequest` method that will carry on the conversation between the two points:

```java
// Before
eventBus.consumer("addr", msg -> {
  msg.reply(reply, response -> { ... });
});

// After
eventBus.consumer("addr", msg -> {
  msg.replyAndRequest(reply, response -> { ... });
});
```

vietj:


src/main/java/io/vertx/core/eventbus/EventBus.java
```diff
@@ -77,10 +81,44 @@
...
+   * <p>
+   * The message will be delivered to at most one of the handlers registered to the address.
+   *
+   * @param address  the address to send it to
+   * @param message  the message body, may be {@code null}
+   * @param replyHandler  reply handler will be called when any reply from the recipient is received
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  @SuppressWarnings("deprecation")
```
jponge: Funny that it dispatches to the deprecated method 😉

vietj: in master this will be different

PR #2999, opened 2019-06-21
Json RFC implementation

slinkydeveloper:
* Made jackson-databind optional
* Implemented encoding and decoding with jackson-core
* Reorganized methods of `Json` class


src/main/java/io/vertx/core/json/JacksonMapper.java
```diff
@@ -0,0 +1,110 @@
+package io.vertx.core.json;
```
vietj: missing file header

vietj:
for the new codecs (Byte array, Instant) : register them manually in the registry since they are always here we don't need them to be service loaded

src/main/java/io/vertx/core/json/EncodeException.java
```diff
@@ -11,19 +11,24 @@
...
  */
 public class EncodeException extends RuntimeException {
 
+  public EncodeException(Throwable e) {
+    this("Failed to encode: " + e.getMessage(), e);
+  }
+
   public EncodeException(String message) {
-    super(message);
+    super(String.format("Failed to encode: %s", message));
```
vietj: why format ?

src/main/java/io/vertx/core/json/JacksonMapper.java
```diff
@@ -0,0 +1,110 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+
+import java.io.IOException;
+import java.time.DateTimeException;
+import java.time.Instant;
+import java.util.Base64;
+
+import static java.time.format.DateTimeFormatter.ISO_INSTANT;
+
+public class JacksonMapper extends JsonCodecMapper {
```
vietj: make this class package private so it's not need outside of the json package.

src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java
```diff
@@ -319,61 +322,69 @@ public void handle(JsonToken event) {
...
+    public <T> T convert(Class<T> type) {
       try {
-        return Json.mapper.readValue(buffer.asParser(), type);
+        if (type == JsonObject.class || type == JsonArray.class) {
+          return (T) Json.decodeValue(buffer.toString());
+        } else {
+          return Json.decodeValue(buffer.toString(), type);
+        }
       } catch (Exception e) {
+        e.printStackTrace();
```
vietj: no print stacktrace

src/main/java/io/vertx/core/json/JsonMapper.java
```diff
@@ -0,0 +1,48 @@
...
+   * @param c
+   * @return
+   * @throws DecodeException If there was an error during decoding
+   * @throws IllegalStateException If this mapper instance can't handle the provided class
+   */
+  <T> T decode(Object json, Class<T> c) throws DecodeException, IllegalStateException;
+
+  <T> T decode(Object json, TypeReference<T> t) throws DecodeException, IllegalStateException;
+
+  static JsonMapper load() {
```
vietj: move this method to `Json` class as private as it's not part of the JsonMapper API

PR #3006, opened 2019-06-27
Add unit tests for io.vertx.core.impl.StringEscapeUtils

EricHetti:
I've analysed your codebase and noticed that `io.vertx.core.impl.StringEscapeUtils` is not fully tested.
I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important.

### DIFF
--- a/src/test/java/io/vertx/core/impl/StringEscapeUtilsTest.java
+++ b/src/test/java/io/vertx/core/impl/StringEscapeUtilsTest.java
@@ -1,0 +1,106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


vietj:
actually this class is only used in vertx-web and should be moved there instead of vertx-core where it's not used at all

Braavos96:
Hi @vietj Apologies for the late response. Is there another class you would like me to test?

julianladisch:
@EricHetti Does your pull request contain code generated by Diffblue? If yes what license has the Diffblue code? The EULA http://docs.diffblue.com/cover/user-manual/topics/eula/ is a "non-sublicensable, non-transferable, limited term licence" for "internal business purposes".

PR #3043, opened 2019-07-24
Support http status code 308 redirection

### LINKED ISSUE #3038: HttpClient doesn't handle 308 redirect
Is there a reason?

Seems like it's supposed to be handled like a 301.

shenzhou-6:
Fixes #3038

### DIFF
--- a/src/main/asciidoc/http.adoc
+++ b/src/main/asciidoc/http.adoc
@@ -1082,10 +1082,10 @@ Alternatively you can just parse the `Set-Cookie` headers yourself in the respon
 
 ==== 30x redirection handling
 
-The client can be configured to follow HTTP redirections provided by the Location response header when the client receives:
+The client can be configured to follow HTTP redirections provided by the `Location` response header when the client receives:
 
-* a 301, 302, 307 or 308 status code along with a HTTP GET or HEAD method
-* a 303 status code, in addition the directed request perform an HTTP GET methodn
+* a `301`, `302`, `307` or `308` status code along with a HTTP GET or HEAD method
+* a `303` status code, in addition the directed request perform an HTTP GET methodn
 
 Here's an example:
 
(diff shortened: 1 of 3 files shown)


src/main/asciidoc/http.adoc
```diff
@@ -1083,7 +1083,7 @@ Alternatively you can just parse the `Set-Cookie` headers yourself in the respon
 ==== 30x redirection handling
 
 The client can be configured to follow HTTP redirections: when the client receives an
-`301`, `302`, `303` or `307` status code, it follows the redirection provided by the `Location` response header
+`301`, `302`, `303` or `307`, `308` status code, it follows the redirection provided by the `Location` response header
```
vietj: I think actually this statement is incorrect (the original form), it should be:

The client can be configured to follow HTTP redirections provided by the `Location` response header when the client receives:

- a `301`, `302`, `307` or `308` status code along with a HTTP `GET` or `HEAD` method
- a `303` status code, in addition the directed request perform an HTTP `GET` method

src/main/asciidoc/http.adoc
```diff
@@ -1096,7 +1096,7 @@ Here's an example:
 The redirection policy is as follow
```
vietj: this section should be removed because it's incorrect and better explained in the documentation above (see the comment)

src/main/asciidoc/http.adoc
shenzhou-6: Thank you for your reply. I have changed it and submitted it.

src/main/asciidoc/http.adoc
```diff
@@ -1082,9 +1082,10 @@ Alternatively you can just parse the `Set-Cookie` headers yourself in the respon
 
 ==== 30x redirection handling
 
-The client can be configured to follow HTTP redirections: when the client receives an
-`301`, `302`, `303` or `307` status code, it follows the redirection provided by the `Location` response header
-and the response handler is passed the redirected response instead of the original response.
+The client can be configured to follow HTTP redirections provided by the Location response header when the client receives:
```
vietj: should be \`Location\` , \`301\`, etc...

shenzhou-6: Sorry, I was careless.

Issue #3047, opened 2019-07-28
Make available on WebSocketBase the close status code and reason


vietj:
Currently one can be aware of the close status code and reason received by a WebSocket by setting a frame handler and catching the close frame. We can also retain such info on the WebSocket and expose them as accessors so they can be retrieved easily.

vietj:
see https://github.com/eclipse-vertx/vert.x/issues/3047

PR #3048, opened 2019-07-28
Make available on WebSocketBase the close status code and reason - se…

vietj:
…e #3047


vietj:
merged directly with a minor change in the test

Issue #3056, opened 2019-08-04
Future.future(Handler<Promise<T>> handler) not handle exception.


cpdyj:
I see that issue #1568.
Why don't make the Future fail when the handler throws an Exception?
Should I create a pr resolve that?
Code: [Here!](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/Future.java#L37)

Expanded, should I make all `Future` auto fail when `Future.handler` throws an `Exception` ?
Thank you.

vietj:
that would be a behavior similar to what we do have in `executeBlocking` so +1 for this behavior. You are welcome to do a PR that we can review.

cpdyj:
Oh, no. I think I fall in wrong. The handler can't handle exception. But I can make `Future.future` auto fail when it throws an Exception.
:)

vietj:
I mean that when the handler passed to Future.future thrown an exception then it returns a failed future

cpdyj:
Ok, I commit a pull request. 
Thanks for review.

PR #3057, opened 2019-08-05
Future will fail when the handler which is passed to `Future.future` throws an Exception.

cpdyj:
I means Future can auto fail when handler which is passed to `Future.future` throws an Exception.
This is my first PR. 
See also: #3056 
Thank you.

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -36,10 +36,10 @@ public interface Future<T> extends AsyncResult<T> {
    */
   static <T> Future<T> future(Handler<Promise<T>> handler) {
     Promise<T> promise = Promise.promise();
-    try{
+    try {
       handler.handle(promise);
-    }catch (Exception e){
-      promise.fail(e);
+    } catch (Exception e){
+      promise.tryFail(e);
     }
     return promise.future();
   }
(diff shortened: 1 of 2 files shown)


cpdyj:
I modify the code, But it break the test case. The testcase need update. Should I update it? 
Thank you.

src/main/java/io/vertx/core/Future.java
```diff
@@ -36,7 +36,11 @@
    */
   static <T> Future<T> future(Handler<Promise<T>> handler) {
     Promise<T> promise = Promise.promise();
-    handler.handle(promise);
+    try{
+      handler.handle(promise);
+    }catch (Exception e){
```
vietj: missing whitespace

src/main/java/io/vertx/core/Future.java
```diff
@@ -36,7 +36,11 @@
...
+      promise.fail(e);
```
vietj: use tryFail instead

vietj:
the test should indeed be modified so that it does not fail after and check that the returned promise is actually failed with the same exception that is thrown.

in addition we need a test that check that if the promise is resolved (like a completion) before the exception is thrown then the outcome of the promise resolution is not modified

vietj:
you need to sign the Eclipse Contributor Agreeement and sign your commits as well

src/main/java/io/vertx/core/Future.java
cpdyj: I will reformat code.

cpdyj:
Ok, I will modify code. Thank you.

cpdyj:
@vietj Ok, I finished.

Issue #2844, opened 2019-02-20
CLI parser should be case-sensitive


MitchAman:
When using the [CLI API](https://vertx.io/docs/vertx-core/java/#_vert_x_command_line_interface_api) for parsing command-line parameters I noticed that the _shortName_ for Options are case-insensitive.  I wouldn't have expected this behavior by default, and didn't see this behavior documented anywhere.  I recall working with other command-line tools which accept single-character ("shortName") options in both uppercase and lowercase forms (i.e. `./someTool -a someValue -A someOtherValue`).

Here's a quick test case demonstrating this behavior (short names):
```
    @Test
    public void testCliOptionShortNamesAreCaseSensitive() {
        CLI cli = CLI.create("test")
                .addOption(new Option().setShortName("a").setLongName("lowercase"))
                .addOption(new Option().setShortName("A").setLongName("uppercase"));

        String lowercaseValue = "someValue";
        String uppercaseValue = "someOtherValue";

        CommandLine commandLine = cli.parse(Arrays.asList("-a", lowercaseValue,
                                                          "-A", uppercaseValue));

        Assert.assertEquals(lowercaseValue, commandLine.getOptionValue("a"));
        Assert.assertEquals(uppercaseValue, commandLine.getOptionValue("A"));
    }
```
```
org.junit.ComparisonFailure: 
Expected :someOtherValue
Actual   :someValue
```

And here's another, where I pass the long names into `commandLine.getOptionValue()`:
```
    @Test
    public void testCliOptionLongNamesAreCaseSensitive() {
        CLI cli = CLI.create("test")
                .addOption(new Option().setShortName("a").setLongName("lowercase"))
                .addOption(new Option().setShortName("A").setLongName("uppercase"));

        String lowercaseValue = "someValue";
        String uppercaseValue = "someOtherValue";

        CommandLine commandLine = cli.parse(Arrays.asList("-a", lowercaseValue,
                                                          "-A", uppercaseValue));

        Assert.assertEquals(lowercaseValue, commandLine.getOptionValue("lowercase"));
        Assert.assertEquals(uppercaseValue, commandLine.getOptionValue("uppercase"));
    }
```
```
java.lang.AssertionError: 
Expected :someOtherValue
Actual   :null
```

Interestingly enough, if I pass the full long name representations into `cli.parse()`, and use the full long names when retrieving the values via `commandLine.getOptionValue()`, then I get the expected behavior (test passes):
```
    @Test
[...]

PR #3067, opened 2019-08-14
Make CLI options case sensitive for short names (#2844).

MitchAman:
Finally got around to addressing #2844.  Let me know if there's any formatting issues or anything like that - first time contributing in this code base.

### DIFF
--- a/src/main/asciidoc/cli.adoc
+++ b/src/main/asciidoc/cli.adoc
@@ -41,8 +41,10 @@ contain more details. Each option and argument are also added on the `CLI` objec
 
 An {@link io.vertx.core.cli.Option} is a command line parameter identified by a _key_ present in the user command
 line. Options must have at least a long name or a short name. Long name are generally used using a `--` prefix,
-while short names are used with a single `-`. Options can get a description displayed in the usage (see below).
-Options can receive 0, 1 or several values. An option receiving 0 values is a `flag`, and must be declared using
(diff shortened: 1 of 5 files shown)


tsegismont:
@MitchAman thanks for sending a PR!

I just read the bug report and it says that currently the short names are compared ignoring case but not long names. But when I look at the code in master, both are compared ignoring case.

So to me, this doesn't look like a bug. But rather a request for changing behavior, that might impact users.

Also, in this PR, short names comparison has been made case-sensitive, while the long names one remained case-insensitive. If we change the current behavior, it should be changed consistently.

@MitchAman @vietj considering #2844 is not a bug but a change request, I believe we should decide if the benefits (ability to, for example, have `-a` and `-A` used for distinct options) outweigh the risks (users scripts with typos suddenly broken).

MitchAman:
@tsegismont One potential mitigation for the risk of breaking user scripts would be to fall back to the case-insensitive behavior if there are no case-sensitivity conflicts.  This way if there's only an option for -a, then both -a and -A would still work, until a new option for -A is implemented.  That would make this more of a feature addition, rather than behavior change.

+1 on keeping short names and long names case-sensitivity behavior the same.

vietj:
@tsegismont can you finish review this ?

tsegismont:
@vietj will do tomorrow

MitchAman:
@tsegismont Certainly, I'll try to get to that early next week.  Thanks for taking another look!

PR #3091, opened 2019-09-01
improve http parse url performance for #path #query #params

qxo:
less gc (reduce create unncessary object) for #params
(if queryString == null, no need QueryStringDecoder)

no need repeat parse path for #query and #params

Signed-off-by: qxo <qxodream@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerRequestImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerRequestImpl.java
@@ -344,7 +344,7 @@ public class Http2ServerRequestImpl extends Http2ServerStream implements HttpSer
   
   @Override
   public String path() {
-    if (path == null) {
+    if (path == null && uri != null) {
       synchronized (conn) {
           final Object o =  HttpUtils.parsePathAndQueryStartIf(uri);
           if (o instanceof Object[]) {
@@ -362,7 +362,7 @@ public class Http2ServerRequestImpl extends Http2ServerStream implements HttpSer
 
   @Override
   public String query() {
-    if (query == null) {
+    if (query == null && uri != null) {
           query = path() == uri || queryStart == -1 ? null : queryStart > 0 && uri.length() > queryStart
             ? uri.substring(queryStart + 1) : HttpUtils.parseQuery(uri);
     }
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerRequestImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerRequestImpl.java
@@ -68,10 +68,9 @@ public class Http2ServerRequestImpl extends Http2ServerStream implements HttpSer
   private final String serverOrigin;
   private final MultiMap headersMap;
   private final String scheme;
-  //add volatile for fix Double-Checked Locking
-  private volatile String path;
-  private volatile String query;
-  private volatile MultiMap params;
+  private String path;
+  private String query;
+  private MultiMap params;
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/http/impl/HttpUtils.java
```diff
@@ -424,7 +424,15 @@ public static URI resolveURIReference(URI base, String ref) throws URISyntaxExce
...
-  static String parsePath(String uri) {
+  static String parsePath(final String uri) {
+    final Object o = parsePathAndQueryStartIf(uri);
+    return (String) ( o instanceof Object[] ? ((Object[])o)[0] : o );
+  }
+
+  /**
+   * Extract the path out of the uri and queryStart
+   */
+  static Object parsePathAndQueryStartIf(final String uri) {
```
pmlopes: I think this is a bad name, I can't tell what this method does from it's name

src/main/java/io/vertx/core/http/impl/Http2ServerRequestImpl.java
```diff
@@ -338,20 +342,37 @@ public String uri() {
...
 
   @Override
   public String query() {
-    synchronized (conn) {
-      this.query = uri != null ? HttpUtils.parseQuery(uri) : null;
-      return query;
+    if (query == null) {
+      synchronized (conn) {
+        if (query == null) {
+          query = path() == uri || queryStart == -1 ? null : queryStart > 0 && uri.length() > queryStart
```
pmlopes: This becomes unreadable with so many chain of logic operations and ternary blocks. I believe proper if statements would be preferred and more readable

src/main/java/io/vertx/core/http/impl/HttpUtils.java
qxo: Any suggesion  for name ?

src/main/java/io/vertx/core/http/impl/Http2ServerRequestImpl.java
qxo: if path already init( path != null ), should we "synchronized (conn) " for #path ?

src/main/java/io/vertx/core/http/impl/Http2ServerRequestImpl.java
```diff
@@ -414,9 +409,7 @@ public String getHeader(CharSequence headerName) {
   public MultiMap params() {
```
vietj: synchronized (conn) should be wrap the whole block we don't really need double checked locking here

vietj:
what's the point of this PR ? has this been spotted as a performance issue in an application ? I don't think this kind of code (called once per request) is on the hot path of an application. For sure it might be faster but it makes code harder to reason about for a performance gain that is not proven to be useful

qxo:
To be a bottleneck , or not ? -- It's depends on  concurrency(c)
if just one request, sure it's useless
But if concurrency over 400, it's pr will have some improve ...
I hope vert.x  will be extreme fast :)

pmlopes:
> But if concurrency over 400, it's pr will have some improve ...

I understand the idea but I find it a bit hard to imagine that in an application the methods: `path()` `query()` or `params()` get called 400 times. Do you have examples of these cases?

vietj:
@qxo this method is usually called from event loop and biased locking is used in practice. What I mean is that we have no evidence that this is an issue in actual applications. In your PR we are trading simplicity for potential performance. Unless it is proven to be a real problem we don't want to make this trade

PR #3107, opened 2019-09-13
Add websocket client response headers to completion handler

swinterberger:
Clear headers after handler to prevent resource useage when the user does not care about the headers. 

Got side tracked for a year sorry.  Old PR turned into a disaster since it was a year behind 3.8 branch and it had my other github account on it so I rebased the entire thing.


src/main/java/io/vertx/core/http/WebSocketBase.java
```diff
@@ -104,6 +105,15 @@
...
   String closeReason();
 
+  /**
+   *  Returns the HTTP response headers during the websocket connection handler.
+   *  <p/>
+   *  After the completion handler callback has completed the response headers will be {@code null}
+   *
+   * @return the response headers
+   */
+  MultiMap headers();
```
vietj: we should add this to `WebSocket` instead of `WebSocketBase`

seanwinterberger: > 
> 
> we should add this to `WebSocket` instead of `WebSocketBase`

instead of or both?  If removed form `WebSocketBase`  `WebsocketImplBase` won't have the impl since it is implementing `WebSocketBase` and I don't believe it would compile.

seanwinterberger:
@vietj is this closer to what you wanted?  

I opened this PR against the 3.8 branch hoping to get this on 3.8.2 since 4.0 is master and still not official.  Let me know if that was the wrong choice and if I need to migrate this branch to 4.0.

src/main/java/io/vertx/core/http/WebSocket.java
```diff
@@ -89,4 +90,13 @@
...
   WebSocket frameHandler(Handler<WebSocketFrame> handler);
+
+  /**
+   *  Returns the HTTP response headers during the websocket connection handler.
+   *  <p/>
+   *  After the completion handler callback has completed the response headers will be {@code null}
+   *
+   * @return the response headers
+   */
+  MultiMap headers();
```
vietj: should be `@CacheReturn`

vietj:
Currently `ServerWebSocketImpl` supports headers, the code should be moved in the common `WebSocketImplBase` class

vietj:
I'll merge it and finish it for 3.8.2

vietj:
I cherry picked the commits and made further required adjustments

vietj:
thanks for the contribution

PR #3120, opened 2019-09-28
Fixes #3119

### LINKED ISSUE #3119: Websocket compression seems broken
I am using the latest vertx milestone 4.0.0-milestone3 and it seems that websocket compression is not working.

ChannelRead of WebSocketServerExtensionHandler responsible for deciding whether to enable compression is not being called (I assume its because its added late to the netty pipeline) 

Most probably this happened due to this commit https://github.com/eclipse-vertx/vert.x/commit/713c1a834dfcfc943f482d628f5c9d0e6474b1e8 that moved initializeWebsocketExtensions inside HttpHandlers which is applied via executeFromIO (Hence my assumption that its added late to the pipeline)

zenios:
1.websocketExtensionHandler should be added before last "handler". 
2.Added a test to verify compression is enabled


vietj:
thanks for the fix @zenios

PR #3123, opened 2019-10-03
Breaking: Fix JSON types to correctly follow RFC7493

pmlopes:

This is a breaking change, which properly follows the RFC7493 when dealing with base64 (binary data) in json.

The fix also adds a utility function to `Json` class to convert from the new format (base64url) to the old format (plain base64).

The utility has 2 functions:

* help users handle integration between vert.x 3 and vert.x 4 projects
* help the interop with other systems that expect base64 strings

### DIFF
--- a/src/main/java/io/vertx/core/json/Json.java
+++ b/src/main/java/io/vertx/core/json/Json.java
@@ -122,4 +122,38 @@ public class Json {
   public static <T> T decodeValue(Buffer buf, Class<T> clazz) throws DecodeException {
     return CODEC.fromBuffer(buf, clazz);
   }
+
+  /**
+   * Convert a base64url encoded string to a base64 encoded string.
+   *
+   * Example:
+   * <code>
+   *     toBase64("qL8R4QIcQ_ZsRqOAbeRfcZhilN_MksRtDaErMA");
+   *     "qL8R4QIcQ/ZsRqOAbeRfcZhilN/MksRtDaErMA=="
+   * </code>
+   *
+   * JsonObject and JsonArray implement the RFC-7493 which defines how binary data is encoded. However
+   * the implementation on Vert.x 3.x was incorrect and did not use the correct encoding <code>base64url</code>.
+   *
+   * In order to interact with older Vert.x applications or other systems that use plain Base64 encoding this
+   * utility method will convert base64url strings to base64.
+   *
+   * @param base64url base64 url without padding string.
+   * @return base64 with padding encoded string.
+   */
+  public static String toBase64(String base64url) {
+    // padding is at most 3 '=' characters
+    final String padding = "===";
+    int stringLength = base64url.length();
+    int diff = stringLength % 4;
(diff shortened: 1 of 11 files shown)


dominik-cnx:
Hello

We have migrated our project from vert.x 3 to vert.x.4 and currently use the legacy mode:

System.setProperty("vertx.json.base64", "legacy");

This is because we have a high number of clients that we cannot migrate at once. Even if we could, we would still have to deal with data in persistent queues that is encoded in the old mode for some time (probably a few weeks in our case).

Now, to avoid issues in the future (the legacy mode is possibly dropped?) we aim to switch to the new format (base64url), we are not in a hurry though. 

It seems like the Apache implementation

https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html

supports both modes for decoding

--
The URL-safe parameter is only applied to encode operations. Decoding seamlessly handles both modes.
--

So, theoretically in

https://github.com/eclipse-vertx/vert.x/blame/118ebe7026c7a087a08bcfa923482cc6e61f6d09/src/main/java/io/vertx/core/json/impl/JsonUtil.java

we could change from

---
 if ("legacy".equalsIgnoreCase(System.getProperty("vertx.json.base64"))) {
      BASE64_ENCODER = Base64.getEncoder();
      BASE64_DECODER = Base64.getDecoder();
    } else {
      BASE64_ENCODER = Base64.getUrlEncoder().withoutPadding();
      BASE64_DECODER = Base64.getUrlDecoder();
    }
---

to something like

---
 BASE64_DECODER = new org.apache.commons.codec.binary.Base64(); 
 if ("legacy".equalsIgnoreCase(System.getProperty("vertx.json.base64"))) {
      BASE64_ENCODER = Base64.getEncoder();
    } else {
      BASE64_ENCODER = Base64.getUrlEncoder().withoutPadding();
    }
---

and then it would be possible to switch to the base64url format, once all components are on vert.x 4 that support decoding of both modes AND be able to process data that is encoded with the "old" mode. I think this would allow us to make the transition.

Thoughts?

pmlopes:
There are a few concerns, I'm not sure if we would like to add an extra dependency for a feature that the JDK already provides. Second, the Apache implementation is quite opinionated. While it can decode both base64 and base64url, it needs explicit configuration on the encoding and padding.

This means that it would be a "universal" decoder, but still an opinionated encoder.

dominik-cnx:
Yep, the extra dependency is probably not ideal. I already have it in my project where I encountered the issue but vert.x might not want it. Maybe we could make the exact implementation of the encoder / decoder configurable, then there would be no change for most of the users and only those, that have migration issues need to add a dependency to the apache.commons version (until the migration is through).

Otherwise, we could stay with the JDK provided encoder if we don't like the explicit configuration of the apache.commons version.

dominik-cnx:
Should I simply create a PR? Or is it unlikely to be considered / merged?

PR #3132, opened 2019-10-08
Clustering spi redesign

tsegismont:
See https://github.com/vert-x3/wiki/wiki/RFC:-clustering-SPI-revisited


vietj:
I see there is some sanitization of deprecated methods, can we split this PR in two and to the deprecation removals in another one ?

tsegismont:
It's going to take some time to extract those commits and fix the build. So
what would the benefit be?

vietj:
it splits the concerns and makes easier to track the breaking changes for the community and the wiki page.

tsegismont:
Is removal of deprecations considered a breaking change?

This PR is going to break a few things configuration-wise anyway (for the
good).

stream-iori:
Is there any progress?

tsegismont:
@stream-iori yes, we've discussed the proposal internally in December. I've been busy on other things recently but I have started to rebase this PR on `master`. You should see new commits in the next couple of weeks.

stream-iori:
OK，I would implements new cluster interface after commits be merged

tsegismont:
Thank you!

karesti:
@tsegismont do you need anything from our side at this point ?

tsegismont:
@karesti I have a couple questions regarding multimap that I was going to send to you :wink:
Thanks for offering your help!

vietj:
@tsegismont there is a conflict due to a commit that fixed a bug in master, can you rebase please ?

tsegismont:
@vietj done

tsegismont:
@stream-iori hi, I have updated the Infinispan and Hazelcast implementations. Do you have some bandwith in the near future for the Zookeeper implementation update?

stream-iori:
Well，I will take some time to have look.
end of April maybe make a first commit.

src/test/java/io/vertx/core/spi/cluster/ClusterManagerDelegate.java
```diff
@@ -0,0 +1,119 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.shareddata.AsyncMap;
+import io.vertx.core.shareddata.Counter;
+import io.vertx.core.shareddata.Lock;
+
+import java.util.List;
+import java.util.Map;
+
+public class ClusterManagerDelegate implements ClusterManager {
```
vietj, line 23: this class is unused and does not do much beside implementing the delegation pattern

src/test/java/io/vertx/test/fakecluster/FakeClusterManager.java
```diff
@@ -207,107 +203,162 @@ public void leave(Handler<AsyncResult<Void>> resultHandler) {
...
+  @Override
+  public Future<Void> register(String address, RegistrationInfo registrationInfo) {
+    registrations.computeIfAbsent(address, k -> Collections.synchronizedList(new ArrayList<>()))
+      .add(registrationInfo);
+    return vertx.getOrCreateContext().succeededFuture();
+  }
+
+  @Override
+  public Future<Void> unregister(String address, RegistrationInfo registrationInfo) {
+    Future<Void> result;
```
vietj, line 223: unused variable `result`

src/test/java/io/vertx/test/fakecluster/FakeClusterManager.java
```diff
@@ -207,107 +203,162 @@ public void leave(Handler<AsyncResult<Void>> resultHandler) {
...
+    List<RegistrationInfo> infos = registrations.get(address);
```
vietj, line 224: instead use `ConcurrentMap#compute` where you can return `null` to remove the entry. The current implementation is not race free, i.e you can have concurrent removal and addition that will let an orphan registration list in the map or a non empty registration list to be not registered

src/test/java/io/vertx/test/fakecluster/FakeClusterManager.java
```diff
@@ -207,107 +203,162 @@ public void leave(Handler<AsyncResult<Void>> resultHandler) {
...
   }
 
   @Override
   public boolean isActive() {
     return nodeID != null;
   }
 
+  @Override
+  public Future<Void> register(String address, RegistrationInfo registrationInfo) {
+    registrations.computeIfAbsent(address, k -> Collections.synchronizedList(new ArrayList<>()))
```
vietj, line 216: instead use a non plain unmodifiable list in the registrations map that is modified using compute. Otherwise there can be races with concurrent unregistration, see the comment below.

src/main/java/io/vertx/core/spi/cluster/impl/DefaultDeliveryStrategy.java
```diff
@@ -0,0 +1,197 @@
...
+          removeWaiters(context, address);
+          return;
+        }
+      } else {
+        waiter = peeked;
+        break;
+      }
+    }
+
+    clusterManager.registrationListener(address)
```
vietj, line 129: instead use a single `onComplete(...)` operation, using `onFailure`/`onSuccess` leads to create un-necessary objects (it will create in practice two `onComplete` listeners)

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -206,31 +193,18 @@ public MessageImpl createMessage(boolean send, String address, MultiMap headers,
...
-        sendToSubs(serverIDs, sendContext);
-      } else {
-        super.sendOrPub(sendContext);
-      }
-    } else {
-      log.error("Failed to send message", asyncResult.cause());
-      sendContext.written(asyncResult.cause());
+      sendContext.ctx.succeededFuture(sendContext.message)
+        .flatMap(deliveryStrategy::chooseNodes)
+        .onFailure(t -> {
```
vietj, line 201: avoid usage of `onFailure`/`onSuccess` for efficiency purpose

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -206,31 +193,18 @@ public MessageImpl createMessage(boolean send, String address, MultiMap headers,
...
-      ChoosableIterable<ClusterNodeInfo> serverIDs = asyncResult.result();
-      if (serverIDs != null && !serverIDs.isEmpty()) {
-        sendToSubs(serverIDs, sendContext);
-      } else {
-        super.sendOrPub(sendContext);
-      }
-    } else {
-      log.error("Failed to send message", asyncResult.cause());
-      sendContext.written(asyncResult.cause());
+      sendContext.ctx.succeededFuture(sendContext.message)
```
vietj, line 199: avoid using `flatMap`, instead simply call `deliveryStrategy.chooseNodes(sendContext.message)`. The current pattern does not do much and consumes allocation.

If we need to use future for proper message delivery then instead, change the contract of `chooseNodes` to use a `Promise<>List<String>` and let the caller pass a promise that defines the context on which the callback is done.

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredHandlerHolder.java
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.eventbus.impl.clustered;
+
+import io.vertx.core.eventbus.impl.HandlerHolder;
+import io.vertx.core.eventbus.impl.HandlerRegistration;
+import io.vertx.core.impl.ContextInternal;
+
+public class ClusteredHandlerHolder<T> extends HandlerHolder<T> {
```
vietj, line 18: is this class really useful ?

src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
```diff
@@ -61,32 +62,35 @@
...
-        log.warn("Connecting to server " + serverID + " failed", res.cause());
-        close(res.cause());
+  void connect() {
+    synchronized (this) {
+      if (connected) {
+        throw new IllegalStateException("Already connected");
       }
-    });
+    }
+    eventBus.vertx().getClusterManager().getNodeInfo(remoteNodeId)
```
vietj, line 71: looking at this, shouldn't `getNodeInfo` actually fails the future instead of completing the future will null ? that would be more clean and avoid an unecessary flatMap operation only for this purpose

src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
```diff
@@ -61,32 +62,35 @@
...
+      .flatMap(nodeInfo -> nodeInfo != null ? Future.succeededFuture(nodeInfo) : Future.failedFuture("Not a member of the cluster"))
+      .map(NodeInfo::getAddress)
```
vietj, line 73: avoid using map operation where you can simply do that in the next flatMap operation

src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
```diff
@@ -61,32 +62,35 @@
...
+      .flatMap(address -> client.connect(address.getPort(), address.getHost()))
+      .onSuccess(this::connected)
```
vietj, line 75: use `onComplete` instead of `onSuccess`/`onFailure`

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -486,14 +470,16 @@ public ClusterManager getClusterManager() {
 
   private void closeClusterManager(Handler<AsyncResult<Void>> completionHandler) {
```
vietj, line 471: it would be more clean here to have `closeClusterManager` to consume a `Promise<Void>` so that the implementation can simply call complete() or fail() and not require a `runOnContext` operation

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
```diff
@@ -68,7 +60,7 @@
   /**
    * Return the unique node ID for this node
    */
-  String getNodeID();
+  String getNodeId();
```
vietj, line 63: why changing case ?

src/test/java/io/vertx/core/spi/cluster/ClusterManagerDelegate.java
tsegismont: It's a leftover from previous commits: the delegate was used in tests to override specific methods (e.g. to make sure a registration never completes). I'll remove it.

src/test/java/io/vertx/test/fakecluster/FakeClusterManager.java
tsegismont: Removed

src/test/java/io/vertx/test/fakecluster/FakeClusterManager.java
tsegismont: I'm not sure to understand, could you give an example to illustrate the issue?

src/main/java/io/vertx/core/spi/cluster/impl/DefaultDeliveryStrategy.java
tsegismont: Done

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
tsegismont: Done

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
tsegismont: The current pattern allows to invoke the delivery strategy on the `sendContext.ctx`.

If we remove the creation of the `succeededFuture`, we'll still have to invoke `runOnContext` somewhere. This will imply allocation too, won't it?

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredHandlerHolder.java
tsegismont: It allows to identify each `handlerHolder`. In the end this identifier goes to the cluster manager. In standalone mode the identifier is not useful. We could use a single class though (value would always be zero in standalone mode)

src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
tsegismont: Done

src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
tsegismont: Done

src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
tsegismont: I had `nodeId` vars here and there, so just for consistency. I can revert if it's a biggie.

src/main/java/io/vertx/core/impl/VertxImpl.java
tsegismont: I can do that, but I haven't actually touched these parts. How about futurizing the whole close process?

tsegismont:
@vietj I've rebased on master and addressed or replied to your concerns

tsegismont:
@vietj can we go ahead and merge this one?

vietj:
no I haven't yet finished the review, I only reviewed the surface and I haven't looked yet at the important changes

src/main/java/io/vertx/core/eventbus/EventBusOptions.java
```diff
@@ -33,37 +33,38 @@
...
    * The default value of cluster ping reply interval = 20000 ms.
    */
-  public static final long DEFAULT_CLUSTER_PING_REPLY_INTERVAL = VertxOptions.DEFAULT_CLUSTER_PING_REPLY_INTERVAL;
+  public static final long DEFAULT_CLUSTER_PING_REPLY_INTERVAL = TimeUnit.SECONDS.toMillis(20);
 
   private String clusterPublicHost = DEFAULT_CLUSTER_PUBLIC_HOST;
   private int clusterPublicPort = DEFAULT_CLUSTER_PUBLIC_PORT;
   private long clusterPingInterval = DEFAULT_CLUSTER_PING_INTERVAL;
   private long clusterPingReplyInterval = DEFAULT_CLUSTER_PING_REPLY_INTERVAL;
+  private JsonObject nodeMetaData;
```
vietj, line 67: typo `nodeMetadata`, metadata is a closed compound word

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -238,37 +235,24 @@ public MessageImpl createMessage(boolean send, String address, MultiMap headers,
...
-    LocalRegistrationResult<T> result = addLocalRegistration(address, registration, replyHandler, localOnly);
-    addRegistration(result.newAddress, result.holder, completionHandler);
-    return result.holder;
+    HandlerHolder<T> holder = addLocalRegistration(address, registration, replyHandler, localOnly);
+    onLocalRegistration(holder, completionHandler);
+    return holder;
   }
 
-  protected <T> void addRegistration(boolean newAddress, HandlerHolder<T> holder, Handler<AsyncResult<Void>> completionHandler) {
+  protected <T> void onLocalRegistration(HandlerHolder<T> handlerHolder, Handler<AsyncResult<Void>> completionHandler) {
```
vietj, line 245: `onLocalUnregistration` has a `Promise<Void>` and this has an `Handler<AsyncResult<Void>>`, it should use also `Promise<Void>` (which actually is simpler to succeed or fail)

src/main/java/io/vertx/core/eventbus/impl/EventBusInternal.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package io.vertx.core.eventbus.impl;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.eventbus.EventBus;
+
+public interface EventBusInternal extends EventBus {
```
vietj, line 19: internally we should use `start(Promise<Void>)` and `close(Promise<Void>)`

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -111,41 +99,23 @@ static void setTrustOptions(TCPSSLOptions sslOptions, TrustOptions options) {
...
-      } else {
-        resultHandler.handle(Future.failedFuture(ar1.cause()));
-      }
-    });
+  public Future<Void> start() {
+    deliveryStrategy.setVertx(vertx);
+    server = vertx.createNetServer(getServerOptions());
+    server.connectHandler(getServerHandler());
+    return server.listen()
+      .<Void>mapEmpty()
```
vietj, line 107: mapEmpty() is unnecessary

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
vietj: why is it necessary to call the delivery strategy on a particular context ? I feel it should not depend on a particular context as the eventual synchronisation of the delivery strategy is not related to vertx concurrency, or am I missing something here ?

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
vietj: no it's not.

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
```diff
@@ -83,20 +75,56 @@
    */
   void nodeListener(NodeListener listener);
 
+  /**
+   * Store the details about this clustered node.
+   */
+  Future<Void> setNodeInfo(NodeInfo nodeInfo);
```
vietj, line 81: instead pass `Promise<Void>` so the `ClusterManager` does not need to care about which context to use

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
```diff
@@ -83,20 +75,56 @@
...
+   * Get details about this clustered node.
+   */
+  NodeInfo getNodeInfo();
+
+  /**
+   * Get details about a clustered Vert.x node.
+   *
+   * @param nodeId the clustered node id
+   */
+  Future<NodeInfo> getNodeInfo(String nodeId);
```
vietj, line 93: instead pass `Promise<NodeInfo>` so the `ClusterManager` does not need to care about which context to use

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
```diff
@@ -83,20 +75,56 @@
...
    *
-   * @return  true if active, false otherwise
+   * @return true if active, false otherwise
    */
   boolean isActive();
+
+  /**
+   * Share a new messaging handler registration with other nodes in the cluster.
+   */
+  Future<Void> register(String address, RegistrationInfo registrationInfo);
```
vietj, line 115: instead pass `Promise<Void>` so the `ClusterManager` does not need to care about which context to use

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
```diff
@@ -83,20 +75,56 @@
...
+
+  /**
+   * Signal removal of a messaging handler registration to other nodes in the cluster.
+   */
+  Future<Void> unregister(String address, RegistrationInfo registrationInfo);
```
vietj, line 120: instead pass `Promise<Void>` so the `ClusterManager` does not need to care about which context to use

src/main/java/io/vertx/core/spi/cluster/DeliveryStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+   * Invoked after the {@link io.vertx.core.eventbus.EventBus} has started
+   */
+  void eventBusStarted();
+
+  /**
+   * Choose nodes the given {@code message} should be delivered to.
+   * <p>
+   * The result must not be null and must hold distinct values.
+   */
+  Future<List<String>> chooseNodes(Message<?> message);
```
vietj, line 50: instead pass a `Promise<List<String>>` so the delivery strategy does not need to care about the context and only cares about completing the given promise independently of the context.

src/main/java/io/vertx/core/eventbus/EventBusOptions.java
tsegismont: Done

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
tsegismont: Done

src/main/java/io/vertx/core/eventbus/impl/EventBusInternal.java
tsegismont: Why? The startup process has been futurized in this PR and it is easier to follow. If we have to make start/close consistent, I would rather futurize the shutdown process too.

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
tsegismont: done

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
tsegismont: done

src/main/java/io/vertx/core/eventbus/impl/EventBusInternal.java
vietj: internally, I'm keen on using promises because it provides control over the context and implementations don't need to care about the context to use, as usually the most outer layer will decide which context to use.

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
tsegismont: done

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
tsegismont: done

src/main/java/io/vertx/core/eventbus/impl/EventBusInternal.java
tsegismont: ok

src/main/java/io/vertx/core/spi/cluster/DeliveryStrategy.java
tsegismont: The context is needed anyway because the default delivery strategy guarantees message delivery order on context.
Other than that ok, I'll remove futures.

vietj:
I think the delivery strategy should do this check and be passed the current active context as method argument.

this way it can take the appropriate decision based on the implementation.

some implementations might rely on the fact to be running on a specific thread, some might want to use atomic / CAS operations instead and could run on any thread.

tsegismont:
Yes that's what I'm doing right now. I'm also removing all usages of future in ClusterManager, EventBus and HAManager.

tsegismont:
@vietj I believe I've addressed all your latest comments/concerns.

vietj:
thanks @tsegismont I'll have a new look soon

vietj:
It seems to me that the ordering guarantee on the same context/address should always be provided by the event-bus and thus I think we can move the code in the delivery strategy in the event-bus itself.

In V3 it is implied that the cluster manager should provide this guarantee because the way it is originally built. You moved this concern in the delivery strategy which provides this guarantee independently of the cluster manager.

I think we should go one step beyond and try moving thiit in the clustered event bus so the guarantee is provided by the clustered event-bus itself whatsoever the delivery strategy is.

tsegismont:
Ok. I'll work on it next week but don't expect an update very soon because
it might take some time to think about the solution. I'll keep you
informed.

Le sam. 25 avr. 2020 à 09:30, Julien Viet <notifications@github.com> a
écrit :

vietj:
it's fine.

tsegismont:
@vietj PTAL, I've addressed your comments/concerns.

tsegismont:
@vietj the PR is rebased on `master` and I've added notes about the weighted round robin implementation.

src/main/java/io/vertx/core/spi/cluster/impl/selectors/Index.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private final int max;
+  private final AtomicInteger idx = new AtomicInteger(0);
+
+  Index(int max) {
+    this.max = max;
+  }
+
+  int nextVal() {
+    return idx.getAndUpdate(this::update);
```
vietj, line 29: instead have `Index` implement `IntUnaryOperator` and pass this to avoid unnecessary object creation with lambda creation.

src/main/java/io/vertx/core/eventbus/impl/ReplyHandler.java
```diff
@@ -70,7 +70,7 @@ protected void doReceive(Message<T> reply) {
   }
 
   void register() {
-    register(repliedAddress, true, ar -> {});
+    register(repliedAddress, true, Promise.promise());
```
vietj, line 73: we should pass `null` and allow the downstream `EventBusImpl#addRegistration` to be passed `null`, to avoid creating a `Promise` when it is not needed.

src/main/java/io/vertx/core/spi/cluster/DeliveryStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Invoked after the cluster manager has started.
+   */
+  void setVertx(Vertx vertx);
+
+  /**
+   * Invoked after the {@link io.vertx.core.eventbus.EventBus} has started
+   */
+  void eventBusStarted();
```
vietj, line 41: we can definitely merge this method and the above `setVertx` method in a single method called `init(Vertx vertx, ClusterManager clusterManager)` that will init the strategy state with what it needs, since the clustered event bus will call this method and has created the `ClusterManager` just before.

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
```diff
@@ -0,0 +1,40 @@
...
+import io.vertx.core.json.JsonObject;
+
+import java.util.Objects;
+
+/**
+ * Details about a clustered Vert.x node.
+ *
+ * @author Thomas Segismont
+ */
+public final class NodeInfo {
```
vietj, line 23: can we have this extends `NodeAddress` ? it will be more efficient

src/main/java/io/vertx/core/spi/cluster/RegistrationUpdateEvent.java
```diff
@@ -0,0 +1,49 @@
...
+
+  private final String address;
+  private final List<RegistrationInfo> registrations;
+
+  public RegistrationUpdateEvent(String address, List<RegistrationInfo> registrations) {
+    this.address = address;
+    if (registrations == null || registrations.isEmpty()) {
+      this.registrations = Collections.emptyList();
+    } else {
+      this.registrations = Collections.unmodifiableList(registrations);
```
vietj, line 32: can we avoid using `unmodifiableList` and have the cluster manager trust the list will be not modified ?

src/main/java/io/vertx/core/spi/cluster/impl/selectors/EmptySelector.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package io.vertx.core.spi.cluster.impl.selectors;
+
+import java.util.Collections;
+
+/**
+ * @author Thomas Segismont
+ */
+enum EmptySelector implements Selector {
```
vietj, line 19: rather `NullSelector`

src/main/java/io/vertx/core/spi/cluster/impl/selectors/Index.java
tsegismont: done

src/main/java/io/vertx/core/eventbus/impl/ReplyHandler.java
tsegismont: done

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
tsegismont: Actually we can have everything in `NodeInfo`. Done

src/main/java/io/vertx/core/spi/cluster/RegistrationUpdateEvent.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/impl/selectors/EmptySelector.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/DeliveryStrategy.java
tsegismont: Done

src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
```diff
@@ -516,18 +516,12 @@ public void testCustomDeliveryStrategy() throws Exception {
...
 
     @Override
     public void setVertx(Vertx vertx) {
-      this.vertx = (VertxInternal) vertx;
-    }
-
-    @Override
-    public void eventBusStarted() {
-      clusterManager = vertx.getClusterManager();
+      clusterManager = ((VertxInternal) vertx).getClusterManager();
```
vietj, line 524: I actually meant that the delivery strategy has an init method passing Vertx and the ClusterManager which avoid the delivery strategy go pull the cluster manager as it is provided in init method.

src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
```diff
@@ -44,31 +41,26 @@
  */
 public interface ClusterManager {
 
-  void setVertx(Vertx vertx);
-
-  /**
-   * Return an async multi-map for the given name
-   */
-  <K, V> void getAsyncMultiMap(String name, Handler<AsyncResult<AsyncMultiMap<K, V>>> resultHandler);
+  void init(Vertx vertx, DeliveryStrategy deliveryStrategy);
```
vietj, line 44: I actually meant that the delivery strategy has an init method passing Vertx and the ClusterManager which avoid the delivery strategy go pull the cluster manager as it is provided in init method.

tsegismont: Yes, see the latest commits, `DeliveryStrategy has been updated too`

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -111,63 +104,50 @@ static void setTrustOptions(TCPSSLOptions sslOptions, TrustOptions options) {
...
-                resultHandler.handle(Future.succeededFuture());
-              } else {
-                resultHandler.handle(Future.failedFuture(ar2.cause()));
-              }
-            });
+  public void start(Promise<Void> promise) {
+    server = vertx.createNetServer(getServerOptions());
+    server.connectHandler(getServerHandler());
+    server.listen(s -> {
+      if (s.succeeded()) {
```
vietj, line 111: I think here we can benefit from flatmap for better readability

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -111,63 +104,50 @@ static void setTrustOptions(TCPSSLOptions sslOptions, TrustOptions options) {
...
     });
   }
 
   @Override
-  public void close(Handler<AsyncResult<Void>> completionHandler) {
-    super.close(ar1 -> {
+  public void close(Promise<Void> promise) {
+    Promise<Void> parentClose = Promise.promise();
+    super.close(parentClose);
+    parentClose.future().onComplete(ar -> {
```
vietj, line 137: I think here we can benefit from flatmap for better readability

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -206,32 +204,41 @@ public MessageImpl createMessage(boolean send, String address, MultiMap headers,
...
+      (strategy, message, promise) -> strategy.chooseForSend(message, promise),
+      (sc, res) -> sendToNode(sc, res),
+      (sc, throwable) -> sendOrPublishFailed(sc, throwable)
+    );
+  }
+
+  private <T> void publish(OutboundDeliveryContext<T> sendContext, Serializer serializer) {
+    serializer.<Iterable<String>>queue(
+      sendContext,
+      (strategy, message, promise) -> strategy.chooseForPublish(message, promise),
```
vietj, line 231: make this a field of the `ClusteredEventBus` so it is cached and will not instantiate a lambda on every call

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -206,32 +204,41 @@ public MessageImpl createMessage(boolean send, String address, MultiMap headers,
...
+      (sc, res) -> sendToNodes(sc, res),
```
vietj, line 232: make this a field of the ClusteredEventBus so it is cached and will not instantiate a lambda on every call

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -206,32 +204,41 @@ public MessageImpl createMessage(boolean send, String address, MultiMap headers,
...
+      (sc, throwable) -> sendOrPublishFailed(sc, throwable)
```
vietj, line 233: make this a field of the ClusteredEventBus so it is cached and will not instantiate a lambda on every call

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
```diff
@@ -117,15 +116,14 @@ public Buffer encodeToWire() {
...
+    buffer.appendByte(send ? (byte) 0 : (byte) 1);
     writeString(buffer, address);
     if (replyAddress != null) {
       writeString(buffer, replyAddress);
     } else {
       buffer.appendInt(0);
     }
-    buffer.appendInt(sender.port);
-    writeString(buffer, sender.host);
+    writeString(buffer, sender);
```
vietj, line 126: this seems to mean that the wire protocol has changed between V3 and V4, is that right ?

src/main/java/io/vertx/core/spi/cluster/DeliveryStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+import io.vertx.core.Vertx;
+import io.vertx.core.eventbus.Message;
+
+/**
+ * {@link io.vertx.core.eventbus.EventBus Clustered EventBus (CEB)} delivery strategy.
+ * <p>
+ * This strategy is skipped by the <em>CEB</em> only when the user raises the {@link io.vertx.core.eventbus.DeliveryOptions#setLocalOnly(boolean)} flag.
+ * Consequently, implementations must be aware of local {@link io.vertx.core.eventbus.EventBus} registrations.
+ */
+public interface DeliveryStrategy {
```
vietj, line 24: Overall I'm not comfortable with the naming `DeliveryStrategy` that is very generic. I can see that the main responsibility of this interface is to choose a node and in fact the default implementation delegates to the `Selectors` class.

I believe we should simply rename this interface `NodeSelector`.

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
tsegismont: this lambda doesn't capture anything, so it isn't be necessary, is it?

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
tsegismont: Done

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
tsegismont: Not sure how it would help here since we want to perform the server close whether or not the parent closed successfully.

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
tsegismont: Yes

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
vietj: right, it is more a safety net

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
vietj: I think we should revert this change and keep the protocol stable, it does not seem to be a valid enough reason to break it.

vietj: if we really have to break it then we should discuss it

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
tsegismont: ok. I've reworked the `Serializer` to minimize the number of lambda instances created

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
tsegismont: The reason it changed is that nodes are no longer identified in the clustered eventbus by host/port. This solves a few issues (for example when a crashed nodes is restarted before all nodes in the cluster could notice).

Anyway compatibility is already broken because:

- the storage implementation is going to change in cluster managers
- we are going to upgrade HZ, ISPN and Ignite and they are not necessarily compatible with previous versions

src/main/java/io/vertx/core/spi/cluster/DeliveryStrategy.java
tsegismont: The default implementation in vertx-core is intentionally "simple": it reproduces the behaviour that exists in Vert.x 3.

However the goal of this extension point it to allow users with specific needs to implement advanced scenarios:

- sending to nodes in the same rack
- avoiding nodes which are overloaded
- partitioning (based on message headers)

The name strategy sounded appropriate to me to describe this but if that's a blocker I can change it.

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
vietj: ok, then we should proceed in a way that

1/ we version the protocol
2/ we have a way that a V4 clustering node detects a V3 node and take an appropriate measure to handle it (logging, close connection, etc...) and vice versa

src/main/java/io/vertx/core/spi/cluster/DeliveryStrategy.java
vietj: I am not keen on the precise `NodeSelector` name, it was a proposal (inspired from the `Selectors` class). Delivery strategy is not a great fit here: this SPI allows to have custom implementations of choosing a list of node(s) provided by a cluster manager to send a given message to.

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
tsegismont: Could you please file an issue for this that we'd address in a later milestone? It doesn't sound like a blocker for this one to be merged and cluster managers to be updated.

src/main/java/io/vertx/core/spi/cluster/DeliveryStrategy.java
tsegismont: To me it really sounds like a strategy: ours (the default) is round robin, users could implement any other they see fit.

tsegismont: I've renamed it to `NodeSelector`.

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
vietj: yes it is not blocker, we need to find a way to implement this

vietj: https://github.com/eclipse-vertx/vert.x/issues/3406

vietj:
I'll do another (hopefully final) pass today.

src/main/java/io/vertx/core/eventbus/EventBusOptions.java
```diff
@@ -33,37 +32,38 @@
...
    * The default value of cluster ping reply interval = 20000 ms.
    */
-  public static final long DEFAULT_CLUSTER_PING_REPLY_INTERVAL = VertxOptions.DEFAULT_CLUSTER_PING_REPLY_INTERVAL;
+  public static final long DEFAULT_CLUSTER_PING_REPLY_INTERVAL = TimeUnit.SECONDS.toMillis(20);
 
   private String clusterPublicHost = DEFAULT_CLUSTER_PUBLIC_HOST;
   private int clusterPublicPort = DEFAULT_CLUSTER_PUBLIC_PORT;
   private long clusterPingInterval = DEFAULT_CLUSTER_PING_INTERVAL;
   private long clusterPingReplyInterval = DEFAULT_CLUSTER_PING_REPLY_INTERVAL;
+  private JsonObject nodeMetadata;
```
vietj, line 66: should we rename this `clusterNodeMetada` instead ?

src/main/java/io/vertx/core/impl/VertxFactory.java
```diff
@@ -36,6 +35,7 @@
   private VertxOptions options;
   private Transport transport;
   private ClusterManager clusterManager;
+  private NodeSelector nodeSelector;
```
vietj, line 38: rename it `clusterNodeSelector`

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
```diff
@@ -0,0 +1,58 @@
...
+  public JsonObject getMetadata() {
+    return metadata;
+  }
+
+  @Override
+  public String toString() {
+    return "NodeInfo{" +
+      "host='" + host + '\'' +
+      ", port=" + port +
+      ", metadata=" + metadata +
```
vietj, line 55: metadata might contain sensitive information ?

src/main/java/io/vertx/core/spi/cluster/RegistrationInfo.java
```diff
@@ -0,0 +1,74 @@
...
+  private final boolean localOnly;
+
+  public RegistrationInfo(String nodeId, long seq, boolean localOnly) {
+    Objects.requireNonNull(nodeId, "nodeId is null");
+    this.nodeId = nodeId;
+    this.seq = seq;
+    this.localOnly = localOnly;
+  }
+
+  public String getNodeId() {
```
vietj, line 34: let's use same convention and name it `nodeId()`

src/main/java/io/vertx/core/spi/cluster/RegistrationInfo.java
```diff
@@ -0,0 +1,74 @@
...
+    return nodeId;
+  }
+
+  public long getSeq() {
```
vietj, line 38: let's use same convention and name it `seq()`

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public NodeInfo(String host, int port, JsonObject metadata) {
+    this.host = Objects.requireNonNull(host, "host is null");
+    Arguments.requireInRange(port, 1, 65535, "Not an actual port: " + port);
+    this.port = port;
+    this.metadata = metadata;
+  }
+
+
+  public String getHost() {
```
vietj, line 38: let's use same convention and name it `host()`

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
```diff
@@ -0,0 +1,58 @@
...
+    return host;
+  }
+
+  public int getPort() {
```
vietj, line 42: let's use same convention and name it `port()`

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
```diff
@@ -0,0 +1,58 @@
...
+    return port;
+  }
+
+  public JsonObject getMetadata() {
```
vietj, line 46: let's use same convention and name it `metadata()`

src/main/java/io/vertx/core/spi/cluster/RegistrationUpdateEvent.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public RegistrationUpdateEvent(String address, List<RegistrationInfo> registrations) {
+    this.address = address;
+    this.registrations = registrations == null ? Collections.emptyList() : registrations;
+  }
+
+  /**
+   * @return the address related to this event
+   */
+  public String getAddress() {
```
vietj, line 35: let's use same convention and name it `address()`

src/main/java/io/vertx/core/spi/cluster/RegistrationUpdateEvent.java
```diff
@@ -0,0 +1,45 @@
...
+    return address;
+  }
+
+  /**
+   * @return the new state of messaging handler registrations
+   */
+  public List<RegistrationInfo> getRegistrations() {
```
vietj, line 42: let's use same convention and name it `registrations()`

src/main/java/io/vertx/core/eventbus/EventBusOptions.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
tsegismont: No, it is meant to contain details such as your rack number, or the set of keys this node is handling if you are implementing a partitioning strategy.

src/main/java/io/vertx/core/spi/cluster/RegistrationInfo.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/RegistrationUpdateEvent.java
tsegismont: Done

src/main/java/io/vertx/core/spi/cluster/RegistrationUpdateEvent.java
tsegismont: Done

src/main/java/io/vertx/core/impl/VertxFactory.java
vietj: node and the setter

tsegismont: Do you mean field and the setter?

src/main/java/io/vertx/core/spi/cluster/NodeInfo.java
vietj: thanks

src/main/java/io/vertx/core/impl/VertxFactory.java
tsegismont: Done

tsegismont:
@vietj can you please take a look to the last updates? Thanks

Issue #3116, opened 2019-09-26
null value header getting into request and validateHeaderValue throwing NullPointerException


lbroman:
Vert.x version 3.8.1

I have a problem with a proxied request with additional headers which somehow allows a header with null value to get into the request and subsequently the validateHeaderValue will throw a nullpointerexception which is only handled in the vertx exception handler.

Some code
`
proxyRequest.putHeaders(clientRequest.headers());
proxyRequest.putHeader("X-Test",null);

clientRequest.bodyHandler(bb -> {
	proxyRequest.sendBuffer(bb, ar -> {
`

The proxy is mapping headers from clientRequest to the proxyRequest and adding an null header (our fault). Vertx will later trigger an exception.

`
java.lang.NullPointerException
        at io.vertx.core.http.impl.HttpUtils.validateHeaderValue(HttpUtils.java:685)
        at io.vertx.core.http.impl.HttpUtils.validateHeader(HttpUtils.java:673)
        at io.vertx.core.http.impl.headers.VertxHttpHeaders.add0(VertxHttpHeaders.java:522)
        at io.vertx.core.http.impl.headers.VertxHttpHeaders.add(VertxHttpHeaders.java:68)
        at io.vertx.core.http.impl.headers.VertxHttpHeaders.add(VertxHttpHeaders.java:84)
        at io.vertx.core.http.impl.headers.VertxHttpHeaders.addAll(VertxHttpHeaders.java:114)
        at io.vertx.core.http.impl.headers.VertxHttpHeaders.addAll(VertxHttpHeaders.java:104)
        at io.vertx.ext.web.client.impl.HttpContext.handlePrepareRequest(HttpContext.java:351)
        at io.vertx.ext.web.client.impl.HttpContext.execute(HttpContext.java:283)
        at io.vertx.ext.web.client.impl.HttpContext.next(HttpContext.java:270)
        at io.vertx.ext.web.client.impl.predicate.PredicateInterceptor.handle(PredicateInterceptor.java:69)
        at io.vertx.ext.web.client.impl.predicate.PredicateInterceptor.handle(PredicateInterceptor.java:32)
        at io.vertx.ext.web.client.impl.HttpContext.next(HttpContext.java:267)
        at io.vertx.ext.web.client.impl.HttpContext.fire(HttpContext.java:277)
        at io.vertx.ext.web.client.impl.HttpContext.prepareRequest(HttpContext.java:166)
[...]

PR #3133, opened 2019-10-08
#3116 - Strengthen null guards in MultiMap

slgriff:


vietj:
can you add tests for this ? also the commit should be signed along with the Contributor Agreement.

PR #3141, opened 2019-10-14
2934 make compressed http2 request possible

KowalczykBartek:
Signed-off-by: Kowalczyk <bkowalczyyk@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xUpgradeToH2CHandler.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xUpgradeToH2CHandler.java
@@ -8,6 +8,7 @@ import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelPipeline;
 import io.netty.handler.codec.http.*;
 import io.netty.handler.codec.http2.*;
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/http/impl/Http1xUpgradeToH2CHandler.java
```diff
@@ -26,10 +27,12 @@
   private final HttpServerChannelInitializer initializer;
   private final HandlerHolder<? extends Handler<HttpServerConnection>> holder;
   private VertxHttp2ConnectionHandler<Http2ServerConnection> handler;
+  private HttpServerOptions httpServerOptions;
```
vietj: rather use 2 booleans for compression/decompression supported than the full options

vietj:
looks good, minor change asked

src/main/java/io/vertx/core/http/impl/Http1xUpgradeToH2CHandler.java
KowalczykBartek: fixed

vietj:
thanks for the contribution

PR #3153, opened 2019-10-22
Catch Throwable in Future.future

ProgMaq:
Change "Exception" to "Throwable" as in the rest of the class.


vietj:
done manually, eclipse contrib agreement was not signed

PR #3176, opened 2019-11-04
Multi futures

vietj:


src/main/java/io/vertx/core/impl/FutureImpl.java
```diff
@@ -78,29 +80,48 @@ public synchronized boolean isComplete() {
...
   }
 
-  /**
-   * Set a handler for the result. It will get called when it's complete
-   */
-  public Future<T> setHandler(Handler<AsyncResult<T>> handler) {
-    boolean callHandler;
+  @Override
+  public Future<T> onComplete(Handler<AsyncResult<T>> h) {
+    if (h == null) {
```
jponge: You might as well use https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String-

src/main/java/io/vertx/core/impl/FutureImpl.java
```diff
@@ -78,29 +80,48 @@ public synchronized boolean isComplete() {
...
+    if (handler instanceof Handlers) {
+      for (Handler<AsyncResult<T>> h : (Handlers<T>)handler) {
+        dispatch0(h);
+      }
+    } else {
+      dispatch0(handler);
+    }
+  }
+
+  private void dispatch0(Handler<AsyncResult<T>> handler) {
```
jponge: Any more meaningful name?

src/main/java/io/vertx/core/impl/FutureImpl.java
```diff
@@ -78,29 +80,48 @@ public synchronized boolean isComplete() {
...
+        } else {
+          addHandler(h);
         }
+        return this;
       }
     }
-    if (callHandler) {
-      dispatch(handler);
-    }
+    dispatch(h);
```
jponge: Isn't there a chance for a concurrent modification exception when iterating through the handlers? The list addition is in a `synchronized` block, but all for loops may throw a `java.util.ConcurrentModificationException` if someone added a handler in-between.

I reckon this is highly unlikely to happen, but I'd like to have another opinion 😉

src/main/java/io/vertx/core/impl/FutureImpl.java
vietj: no :-)

vietj: the goal is to avoid recursive dispatch that would not be jitted

src/main/java/io/vertx/core/impl/FutureImpl.java
vietj: I think it is not possible because the `tryComplete` / `tryFail` method will set the `handler` field to `null` in the synchronized block. Further calls to `onComplete` after will call `dispatch` and not create or set the `handler` field using the same synchronized monitor.

src/main/java/io/vertx/core/impl/FutureImpl.java
vietj: is requireNonNull intrisified ?

src/main/java/io/vertx/core/impl/FutureImpl.java
vietj: I can rename it to `doDispatch` ?

jponge: Yep

src/main/java/io/vertx/core/impl/FutureImpl.java
jponge: https://stackoverflow.com/questions/31005009/objects-requirenonnullt-obj-instead-of-null-checks-and-manually-thrown-illegal#comment50039504_31005009

At the very least the the JIT will easily inline it.

src/main/java/io/vertx/core/impl/FutureImpl.java
vietj: Any further comment on this ?

jponge: Nope

PR #3182, opened 2019-11-07
Configurable WorkerExecutorFactory

rvansa:
* Enables custom implementation of the executor service that handles blocking invocations as the default ThreadPoolExecutor using LinkedBlockingQueue does not scale well in case of very short (non-blocking) tasks.
* Alternative implementation is *not* provided in this commit.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
     <slf4j.version>1.7.21</slf4j.version>
     <log4j2.version>2.8.2</log4j2.version>
     <junit.version>4.12</junit.version>
+    <jctools.version>2.1.2</jctools.version>
     <assertj.version>3.4.1</assertj.version>
     <apacheds-protocol-dns.version>1.5.7</apacheds-protocol-dns.version>
     <generated.dir>${project.basedir}/src/main/generated</generated.dir>
@@ -155,6 +156,14 @@
       <artifactId>slf4j-api</artifactId>
       <version>${slf4j.version}</version>
       <scope>provided</scope>
+    </dependency>
+
+    <!-- JCTools -->
+    <dependency>
+      <groupId>org.jctools</groupId>
+      <artifactId>jctools-core</artifactId>
+      <version>${jctools.version}</version>
+      <scope>compile</scope>
     </dependency>
 
     <!-- Generators -->
@@ -550,6 +559,32 @@
       </plugin>
 
       <plugin>
+        <artifactId>maven-shade-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <artifactSet>
+                <includes>
+                  <include>org.jctools</include>
+                </includes>
+              </artifactSet>
+              <relocations>
+                <relocation>
+                  <pattern>org.jctools.</pattern>
+                  <shadedPattern>io.vertx.core.impl.shaded.org.jctools.</shadedPattern>
+                </relocation>
+              </relocations>
+              <minimizeJar>true</minimizeJar>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
         <artifactId>maven-assembly-plugin</artifactId>
         <executions>
           <execution>
@@ -614,6 +649,7 @@
                 <Import-Package>org.slf4j*;resolution:=optional,
                   org.apache.log4j;resolution:=optional,
                   org.apache.logging.log4*;resolution:=optional,
+                  !org.jctools.*,
                   *
                 </Import-Package>
               </instructions>
--- a/src/main/java/io/vertx/core/VertxOptions.java
+++ b/src/main/java/io/vertx/core/VertxOptions.java
@@ -18,9 +18,11 @@ import io.vertx.core.file.FileSystemOptions;
 import io.vertx.core.impl.cpu.CpuCoreSensor;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.metrics.MetricsOptions;
+import io.vertx.core.spi.WorkerExecutorFactory;
 import io.vertx.core.spi.cluster.ClusterManager;
 
 import java.util.Objects;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import static io.vertx.core.file.FileSystemOptions.DEFAULT_FILE_CACHING_ENABLED;
@@ -42,6 +44,12 @@ public class VertxOptions {
    * The default number of threads in the worker pool = 20
    */
   public static final int DEFAULT_WORKER_POOL_SIZE = 20;
+
+  /**
+   * The default factory that creates a thread pool for workers.
+   */
+  public static final WorkerExecutorFactory DEFAULT_WORKER_EXECUTOR_FACTORY =
+    (options, threadFactory) -> Executors.newFixedThreadPool(options.getWorkerPoolSize(), threadFactory);
 
   /**
    * The default number of threads in the internal blocking  pool (used by some internal operations) = 20
@@ -168,6 +176,7 @@ public class VertxOptions {
 
   private int eventLoopPoolSize = DEFAULT_EVENT_LOOP_POOL_SIZE;
   private int workerPoolSize = DEFAULT_WORKER_POOL_SIZE;
+  private WorkerExecutorFactory workerExecutorFactory = DEFAULT_WORKER_EXECUTOR_FACTORY;
   private int internalBlockingPoolSize = DEFAULT_INTERNAL_BLOCKING_POOL_SIZE;
   private long blockedThreadCheckInterval = DEFAULT_BLOCKED_THREAD_CHECK_INTERVAL;
   private long maxEventLoopExecuteTime = DEFAULT_MAX_EVENT_LOOP_EXECUTE_TIME;
@@ -201,6 +210,7 @@ public class VertxOptions {
   public VertxOptions(VertxOptions other) {
     this.eventLoopPoolSize = other.getEventLoopPoolSize();
     this.workerPoolSize = other.getWorkerPoolSize();
+    this.workerExecutorFactory = other.getWorkerExecutorFactory();
     this.blockedThreadCheckInterval = other.getBlockedThreadCheckInterval();
     this.maxEventLoopExecuteTime = other.getMaxEventLoopExecuteTime();
     this.maxWorkerExecuteTime = other.getMaxWorkerExecuteTime();
@@ -275,6 +285,24 @@ public class VertxOptions {
       throw new IllegalArgumentException("workerPoolSize must be > 0");
     }
     this.workerPoolSize = workerPoolSize;
+    return this;
+  }
+
+  /**
+   * @return Factory creating executor service for worker threads.
+   */
+  public WorkerExecutorFactory getWorkerExecutorFactory() {
+    return workerExecutorFactory;
+  }
+
+  /**
+   * Set the factory creating executor service for worker threads.
+   *
+   * @param workerExecutorFactory Factory creating executor service for worker threads.
+   * @return a reference to this, so the API can be used fluently
+   */
+  public VertxOptions setWorkerExecutorFactory(WorkerExecutorFactory workerExecutorFactory) {
+    this.workerExecutorFactory = Objects.requireNonNull(workerExecutorFactory);
     return this;
   }
 
@@ -904,6 +932,7 @@ public class VertxOptions {
 
     if (eventLoopPoolSize != that.eventLoopPoolSize) return false;
     if (workerPoolSize != that.workerPoolSize) return false;
+    if (!Objects.equals(workerExecutorFactory, that.workerExecutorFactory)) return false;
     if (internalBlockingPoolSize != that.internalBlockingPoolSize) return false;
     if (blockedThreadCheckInterval != that.blockedThreadCheckInterval) return false;
     if (blockedThreadCheckIntervalUnit != that.blockedThreadCheckIntervalUnit) return false;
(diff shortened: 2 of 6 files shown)


rvansa:
I can create another PR for master of course.
My implementation that works much better in Quarkus for REST calls handling is based on [JCTools](https://github.com/JCTools/JCTools) and I did not want to add a dependency to those; therefore I'll add that dependency rather to the Quarkus extension.

vietj:
@rvansa Netty bundles (via repackaging) some of the JCTools classes in Netty, can you check if the one you need is not bundled ? can you check ?

Other than that I think instead of having this as part of `VertxOptions` to be a `META-INF/services` provider instead.

rvansa:
@vietj Netty has only MPSC queues, while we need MPMC. (they're shading it with `<minimizeJar>` so the other queues are stripped off). Naturally I can do the same in Vert.x if you're ok with that.

Even if we register this in `META-INF/services` there must be a way to select implementation. Should I just name the implementations and let user decide through system property?

vietj:
@rvansa can you write a JMH benchmarks that shows the difference between the two implementations that will help making a decision here ?

hutchig:
I need to learn a lot more about Vert.x before I can say anything sensible
but I have been looking at this PR and postulating into the eventLoopThreadFactory and the acceptorEventLoopThreadFactory use (in fact all Thread creation) to try to understand how
the pattern used here might be extended to encapsulate all the Thread factories in order
to work when Vert.x is embedded into a JEE environment.

pmlopes:
I don't follow the pull request, sorry. From the original issue a performance issue is identified but this pull request doesn't seem to fix it, it appears to be just opening for extension but (i assume the "fix" will be not here, so where is it?)

Because the default implementation is the one that was identified as problematic...

vietj:
@pmlopes I believe that we want indeed to use a better implementation in Vert.x itself, however it means having a new dependency or shading a class, in both case this requires to open a CQ and get it analyse by the IP team.

That being said, this problem mostly happen with very intensive usage of the worker pool like what Quarkus is doing at the moment and this is something that people should avoid when using Vert.x so it is not so much important to have a better implementation in Vert.x

There is a trade-off and this is a choice we need to do.

rvansa:
After a while I was able to prepare a [JMH benchmark](https://github.com/rvansa/benchmarks/tree/master/executors) that manifests the issue. It was not easy as it does not manifest if the producers (eventloop) or consumer (worker threads) are not doing any work; notice the `pwork=2000` and `cwork=2000` params to the test. If I set these to just `1` (doing hardly anything) the implementations perform very similar or the LinkedBlockingQueue (LBQ) performs better.

One dubious aspect of this test is using the `Semaphore` to limit concurrency in the same way as the quarkus benchmark would do (through limited concurrency in `wrk`), but I haven't figured out a better way. From my tests it seems that 84 threads are capable of running about 5M acquire/s and 5M release/s (in this test on 2 semaphores), so the effect is probably significant but not dominating. 

See results for 1, 4, 16 and 64 producer threads (on 32-core machine):

* `NONE` = ImmediateEventExecutor
* `LBQ` = LinkedBlockingQueue
* `LTQ` = LinkedTransferQueue
* `BUSY` = MpmcArrayQueue with busy spinning
* `MPMC` = MpmcArrayQueue with custom await impl (not) offloading to LBQ
* `XADD` = MpmcProgressiveUnboundedQueue with custom await impl

```
producer threads: 1
Benchmark                   (capacity)  (cwork)  (impl)  (pwork)  (workers)   Mode  Cnt       Score      Error  Units
ExecutorBenchmark.producer        1000     2000    NONE     2000         20  thrpt   25  103064.888 ±  702.007  ops/s
ExecutorBenchmark.producer        1000     2000     LBQ     2000         20  thrpt   25  114749.764 ± 2188.676  ops/s
ExecutorBenchmark.producer        1000     2000     LTQ     2000         20  thrpt   25  144382.841 ± 1345.987  ops/s
ExecutorBenchmark.producer        1000     2000    BUSY     2000         20  thrpt   25  185224.551 ± 1119.644  ops/s
ExecutorBenchmark.producer        1000     2000    MPMC     2000         20  thrpt   25  149886.512 ± 2790.551  ops/s
ExecutorBenchmark.producer        1000     2000    XADD     2000         20  thrpt   25  151295.116 ± 1780.780  ops/s


Producer threads: 4
Benchmark                   (capacity)  (cwork)  (impl)  (pwork)  (workers)   Mode  Cnt       Score       Error  Units
ExecutorBenchmark.producer        1000     2000    NONE     2000         20  thrpt   25  403392.027 ±   719.229  ops/s
ExecutorBenchmark.producer        1000     2000     LBQ     2000         20  thrpt   25  595198.736 ± 19735.090  ops/s
ExecutorBenchmark.producer        1000     2000     LTQ     2000         20  thrpt   25  555250.077 ±  1677.761  ops/s
ExecutorBenchmark.producer        1000     2000    BUSY     2000         20  thrpt   25  724137.418 ±  8481.848  ops/s
ExecutorBenchmark.producer        1000     2000    MPMC     2000         20  thrpt   25  571376.094 ±  1956.449  ops/s
ExecutorBenchmark.producer        1000     2000    XADD     2000         20  thrpt   25  565412.850 ±  2048.931  ops/s


Producer threads: 16
Benchmark                   (capacity)  (cwork)  (impl)  (pwork)  (workers)   Mode  Cnt        Score       Error  Units
ExecutorBenchmark.producer        1000     2000    NONE     2000         20  thrpt   25  1577664.252 ± 10584.738  ops/s
ExecutorBenchmark.producer        1000     2000     LBQ     2000         20  thrpt   25   632060.072 ± 19946.471  ops/s
ExecutorBenchmark.producer        1000     2000     LTQ     2000         20  thrpt   25  1995513.151 ± 33896.155  ops/s
ExecutorBenchmark.producer        1000     2000    BUSY     2000         20  thrpt   25  2029422.421 ± 50796.813  ops/s
ExecutorBenchmark.producer        1000     2000    MPMC     2000         20  thrpt   25  2345557.845 ± 41764.691  ops/s
ExecutorBenchmark.producer        1000     2000    XADD     2000         20  thrpt   25  2302541.985 ± 42415.361  ops/s


Producer threads: 64
Benchmark                   (capacity)  (cwork)  (impl)  (pwork)  (workers)   Mode  Cnt        Score       Error  Units
ExecutorBenchmark.producer        1000     2000    NONE     2000         20  thrpt   25  2680256.634 ± 35576.514  ops/s
ExecutorBenchmark.producer        1000     2000     LBQ     2000         20  thrpt   25   626612.499 ± 22087.875  ops/s
ExecutorBenchmark.producer        1000     2000     LTQ     2000         20  thrpt   25  2029667.253 ± 24464.467  ops/s
ExecutorBenchmark.producer        1000     2000    BUSY     2000         20  thrpt   25  2271511.114 ± 44292.175  ops/s
ExecutorBenchmark.producer        1000     2000    MPMC     2000         20  thrpt   25  2374089.646 ± 34171.036  ops/s
ExecutorBenchmark.producer        1000     2000    XADD     2000         20  thrpt   25  2274267.573 ± 35661.893  ops/s
```
Note: to include XADD queue I had to use @franz1981 's [fork of JCTools](https://github.com/franz1981/JCTools/tree/mpmc_xadd_poll), but it shouldn't change anything about MPMC.

I am still working on understanding these results. If I compare with (not presented) results without the `pwork`/`cwork`, it seems that while the extra CPU work slows down LBQ, the effect on other queues is much less significant.

rvansa:
I've added a commit with the implementations, though given the results with LTQ I'll open another PR with lower impact.

rvansa:
One more observation regarding the Quarkus benchmark: since I could execute the HTTP requests against multiple endpoints, I've tested scaling the app to 2 or 3 instances (during the early assesment). There was a slight improvement, but not anywhere to the expected linear scaling. That leads me to hypothesis that the contention observed with LBQ is not on the queue itself (as the chunk of memory or lock protecting that) but maybe its behaviour created a problems somewhere else (on the memory bus, in caches, lack of imps whipping electrons...). This is somewhat aligned to the observation that the queue can sustain throughput about 2M tasks/s if there's no other work.

franz1981:
I leave the results by running the same tests on a 48 cores machine (2 NUMA node), but using the latest version of the XADD q on https://github.com/franz1981/JCTools/tree/mpmc_xadd_poll :
```
producer threads: 1
--------------------------
Benchmark                   (capacity)  (cwork)  (impl)  (pwork)  (workerNamePattern)  (workers)   Mode  Cnt       Score      Error  Units
ExecutorBenchmark.producer        1000     2000    NONE     2000                p%04d         20  thrpt   25  102334.374 ±   42.460  ops/s
ExecutorBenchmark.producer        1000     2000     LBQ     2000                p%04d         20  thrpt   25   35356.967 ± 3128.730  ops/s
ExecutorBenchmark.producer        1000     2000     LTQ     2000                p%04d         20  thrpt   25  127423.575 ± 2677.574  ops/s
ExecutorBenchmark.producer        1000     2000    BUSY     2000                p%04d         20  thrpt   25  164218.570 ± 2509.620  ops/s
ExecutorBenchmark.producer        1000     2000    MPMC     2000                p%04d         20  thrpt   25  134965.951 ± 4614.786  ops/s
ExecutorBenchmark.producer        1000     2000    XADD     2000                p%04d         20  thrpt   25  134400.313 ± 4706.191  ops/s
--------------------------
producer threads: 4
--------------------------
Benchmark                   (capacity)  (cwork)  (impl)  (pwork)  (workerNamePattern)  (workers)   Mode  Cnt       Score       Error  Units
ExecutorBenchmark.producer        1000     2000    NONE     2000                p%04d         20  thrpt   25  392703.457 ±  1757.190  ops/s
ExecutorBenchmark.producer        1000     2000     LBQ     2000                p%04d         20  thrpt   25  474983.335 ± 32190.323  ops/s
ExecutorBenchmark.producer        1000     2000     LTQ     2000                p%04d         20  thrpt   25  510333.101 ±  1782.394  ops/s
ExecutorBenchmark.producer        1000     2000    BUSY     2000                p%04d         20  thrpt   25  657729.669 ±  4898.845  ops/s
ExecutorBenchmark.producer        1000     2000    MPMC     2000                p%04d         20  thrpt   25  512765.035 ±  2206.303  ops/s
ExecutorBenchmark.producer        1000     2000    XADD     2000                p%04d         20  thrpt   25  509250.474 ±  1689.264  ops/s
---------------------------
producer threads: 16
---------------------------
Benchmark                   (capacity)  (cwork)  (impl)  (pwork)  (workerNamePattern)  (workers)   Mode  Cnt        Score       Error  Units
ExecutorBenchmark.producer        1000     2000    NONE     2000                p%04d         20  thrpt   25  1393929.015 ±  2003.301  ops/s
ExecutorBenchmark.producer        1000     2000     LBQ     2000                p%04d         20  thrpt   25   611563.064 ± 18780.967  ops/s
ExecutorBenchmark.producer        1000     2000     LTQ     2000                p%04d         20  thrpt   25  1473687.508 ± 43343.100  ops/s
ExecutorBenchmark.producer        1000     2000    BUSY     2000                p%04d         20  thrpt   25  2238837.585 ±  9419.288  ops/s
ExecutorBenchmark.producer        1000     2000    MPMC     2000                p%04d         20  thrpt   25  2208809.886 ± 20905.331  ops/s
ExecutorBenchmark.producer        1000     2000    XADD     2000                p%04d         20  thrpt   25  2084232.054 ± 18616.841  ops/s
----------------------------
producer threads: 64
----------------------------
Benchmark                   (capacity)  (cwork)  (impl)  (pwork)  (workerNamePattern)  (workers)   Mode  Cnt        Score        Error  Units
ExecutorBenchmark.producer        1000     2000    NONE     2000                p%04d         20  thrpt   25  3762733.463 ±  50719.334  ops/s
ExecutorBenchmark.producer        1000     2000     LBQ     2000                p%04d         20  thrpt   25   617748.328 ±  21554.965  ops/s
ExecutorBenchmark.producer        1000     2000     LTQ     2000                p%04d         20  thrpt   25  1785981.964 ±  70177.224  ops/s
ExecutorBenchmark.producer        1000     2000    BUSY     2000                p%04d         20  thrpt   25  2507162.075 ±  94656.831  ops/s
ExecutorBenchmark.producer        1000     2000    MPMC     2000                p%04d         20  thrpt   25  2514403.206 ± 136024.350  ops/s
ExecutorBenchmark.producer        1000     2000    XADD     2000                p%04d         20  thrpt   25  2951176.040 ±  15095.641  ops/s
```

rvansa:
@vietj So what's the future of this PR? I'll have to update it ofc after #3193 but would Vert.x welcome the option to swap the executor implementation? Is there a configuration style more preferred to the one included in this PR?

vietj:
we need to spend time studying the question and see how best quarkus should handle it too, so for now it's on hold as the other PR you did provides a good trade off

hutchig:
I understand that the main original motivation for this PR is handled by Radim's other PR.

I am looking into what quarkus does to embedd vert.x but I imagined that having configurable thread executor factories would help towards making Vert.x more easily embeddable into other servers' services.

If configurable worker executor factories is on hold and with vertx-jca being pruned for V4 in https://github.com/vert-x3/vertx-jca/issues/15
is there any ongoing recommendation, or better thing to build, for embedding Vert.x 4+ into other server environments?

Vert.x in Open Liberty (why does Frankenstein come to mind?)
image

vietj:
nothing wrong having such SPI, however we are now switching to Vert.x 4 being the main place where development happens, so we would like this to be rather in Vert.x 4 rather than introducing a mechanism that will not be present in Vert.x 4 .

Given that there are a few SPI for proper embedding (`Transport`, `ClusterManager` at least), I believe we should perhaps go toward a _builder_ style for building a proper Vert.x instance instead of having an Java service loader / a field on `VertxOptions`.

Of course doing `Vertx.vertx(options)` would still be the primary way to do it and would use the build with the default settings.

rvansa:
Ok, I agree that it would be better to file this against master.

franz1981:
@rvansa @vietj Finally JCTools is on the third version and both the XADD queues (MPSC and MPMC) are tested,cleaner and available to be used: I remember they were already delivering a nice boost, if we care getting good numbers of Techmpowered benchs ;)

PR #3186, opened 2019-11-07
Fixes incomplete promise when a child verticle could not be deployed …

zenios:
…because its parent has been undeployed

Signed-off-by: zenios <dimitris.zenios@gmail.com>

Not sure if its better to wait for underemployment to complete and then fail the promise.

### DIFF
--- a/src/main/java/io/vertx/core/impl/DeploymentManager.java
+++ b/src/main/java/io/vertx/core/impl/DeploymentManager.java
@@ -463,8 +463,7 @@ public class DeploymentManager {
                   deployment.child = true;
                 } else {
                   // Orphan
-                  deployment.undeploy(event -> {});
-                  promise.fail("Verticle deployment failed.Could not be added as child of parent verticle");
+                  deployment.undeploy(event -> promise.fail("Verticle deployment failed.Could not be added as child of parent verticle"));
                   return;
                 }
               }
--- a/src/main/java/io/vertx/core/impl/DeploymentManager.java
+++ b/src/main/java/io/vertx/core/impl/DeploymentManager.java
@@ -463,7 +463,8 @@ public class DeploymentManager {
                   deployment.child = true;
                 } else {
                   // Orphan
-                  deployment.undeploy(null);
+                  deployment.undeploy(event -> {});
+                  promise.fail("Verticle deployment failed.Could not be added as child of parent verticle");
                   return;
                 }
               }
--- a/src/test/java/io/vertx/core/DeploymentTest.java
+++ b/src/test/java/io/vertx/core/DeploymentTest.java
@@ -1235,6 +1235,28 @@ public class DeploymentTest extends VertxTestBase {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


vietj:
can you write a test for this ?

zenios:
I will try

As for the promise do you think its better to wait for the undeployment to
complete or its ok like this?

vietj:
let's write a test for this first :-)

vietj:
that will help having a decision

zenios:
Tests added

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -463,7 +463,8 @@ private ClassLoader getCurrentClassLoader() {
                   deployment.child = true;
                 } else {
                   // Orphan
-                  deployment.undeploy(null);
+                  deployment.undeploy(event -> {});
```
vietj: I think we should fail the promise in the undeploy callback

zenios:
Ok let me change it

zenios:
Done

src/test/java/io/vertx/core/DeploymentTest.java
```diff
@@ -1234,6 +1234,28 @@ public void testUndeployAll() throws Exception {
...
+        vertx.deployVerticle(ChildVerticle.class.getName())
+          .<Void>mapEmpty()
+          .setHandler(stopPromise);
+      }
+    }
+
+    CountDownLatch latch = new CountDownLatch(1);
+    vertx.deployVerticle(new ParentVerticle(), ar1 -> {
+      assertTrue(ar1.succeeded());
+      vertx.undeploy(ar1.result(), ar2 -> {
```
vietj: you can use the helper method onSuccess / onFailure to make the test shorter and easier to read

zenios: Done

src/test/java/io/vertx/core/DeploymentTest.java
```diff
@@ -1234,6 +1234,23 @@ public void testUndeployAll() throws Exception {
...
+        vertx.deployVerticle(ChildVerticle.class.getName())
+          .<Void>mapEmpty()
+          .setHandler(stopPromise);
+      }
+    }
+
+    CountDownLatch latch = new CountDownLatch(1);
+    vertx.deployVerticle(new ParentVerticle(), onSuccess(id ->
+      vertx.undeploy(id, onFailure(u -> latch.countDown()))));
+    awaitLatch(latch);
```
vietj: instead use `await()` and `testComplete()` to signal test completion

vietj:
another comment and I think we are good

src/test/java/io/vertx/core/DeploymentTest.java
zenios: Done.Do you want me to squash to single commit?

vietj: thanks, no I'll do it on merge

PR #3197, opened 2019-11-15
Code changed to solve the JSON RFCs

### LINKED ISSUE #3171: Feature Request: JsonArray.add(int index, Object value)
this is easy to add and is available in most json libs
PS: and for JsonArray.set(int index, Object value)
methods add and set have the same semantics as in java's List

pmlopes:

This Pull request addresses the RFC: https://github.com/vert-x3/wiki/wiki/RFC:JSON-improvements

* `JSON{Object|Array}` allow any type to be stored which allows future POJO mappings
* No serialization is done for non JSON types at insert (less GC)
* No deserialization is done for `Temporal|Binary` data if not in `String` format (*ISODate* or *Base64*)
* No Copy on Read (however copy is still present and will behave like before)
* `Base64` encoding is fixed to properly use the right encoding, however it can be disabled for now using a system property to fall back to the 3.x format (Fixes #3123)
* Missing `JsonArray#add(int, Object)` added (Fixes #3171)

Important takeaways:

1. API is 100% backwards compatible (no breaking changes)
2. Behavior is also backwards compatible

For most users this change will not be visible, except the fix of the base64 encoder which can be set to act as before if there's a need to interact with older systems.

### DIFF
--- a/src/main/java/io/vertx/core/impl/ConversionHelper.java
+++ b/src/main/java/io/vertx/core/impl/ConversionHelper.java
@@ -15,11 +15,11 @@ import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.time.Instant;
+import java.util.*;
+
+import static io.vertx.core.json.impl.JsonUtil.BASE64_ENCODER;
+import static java.time.format.DateTimeFormatter.ISO_INSTANT;
 
 /**
  * An internal conversion helper, later it could be merged with JsonObject mapFrom/mapTo and moved in Json class
@@ -47,7 +47,7 @@ public class ConversionHelper {
     } else if (obj instanceof CharSequence) {
       return obj.toString();
     } else if (obj instanceof Buffer) {
-      return Base64.getEncoder().encodeToString(((Buffer)obj).getBytes());
+      return Base64.getEncoder().encodeToString(((Buffer) obj).getBytes());
     }
     return obj;
   }
@@ -66,7 +66,7 @@ public class ConversionHelper {
       return null;
     }
     list = new ArrayList<>(list);
-    for (int i = 0;i < list.size();i++) {
+    for (int i = 0; i < list.size(); i++) {
       list.set(i, toJsonElement(list.get(i)));
     }
     return new JsonArray(list);
@@ -75,11 +75,18 @@ public class ConversionHelper {
   @SuppressWarnings("unchecked")
   public static <T> T fromObject(Object obj) {
     if (obj instanceof JsonObject) {
-      return (T)fromJsonObject((JsonObject)obj);
+      return (T) fromJsonObject((JsonObject) obj);
     } else if (obj instanceof JsonArray) {
-      return (T)fromJsonArray((JsonArray)obj);
+      return (T) fromJsonArray((JsonArray) obj);
+    } else if (obj instanceof Instant) {
+      return (T) ISO_INSTANT.format((Instant) obj);
+    } else if (obj instanceof byte[]) {
+      return (T) BASE64_ENCODER.encodeToString((byte[]) obj);
+    } else if (obj instanceof Enum) {
+      return (T) ((Enum) obj).name();
     }
-    return (T)obj;
+
+    return (T) obj;
   }
 
   public static Map<String, Object> fromJsonObject(JsonObject json) {
@@ -98,7 +105,7 @@ public class ConversionHelper {
       return null;
     }
     List<Object> list = new ArrayList<>(json.getList());
-    for (int i = 0;i < list.size();i++) {
+    for (int i = 0; i < list.size(); i++) {
       list.set(i, fromObject(list.get(i)));
     }
     return list;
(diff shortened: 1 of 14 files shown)


src/main/java/io/vertx/core/impl/ConversionHelper.java
```diff
@@ -47,7 +47,7 @@ private static Object toJsonElement(Object obj) {
     } else if (obj instanceof CharSequence) {
       return obj.toString();
     } else if (obj instanceof Buffer) {
-      return Base64.getEncoder().encodeToString(((Buffer)obj).getBytes());
+      return Base64.getEncoder().encodeToString(((Buffer) obj).getBytes());
```
pmlopes: Should use the constant encoder

src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -660,7 +394,7 @@ public boolean contains(Object value) {
    * Remove the specified value from the JSON array. This method will scan the entire array until it finds a value
    * or reaches the end.
    *
-   * @param value  the value to remove
+   * @param value the value to remove
    * @return true if it removed it, false if not found
    */
   public boolean remove(Object value) {
```
pmlopes: This one is tricky, do we allow unwrapped values? To keep behavior consistency we can rely on the storage but need to custom implement it.

vietj: can you elaborate ? I'm not sure to understand

pmlopes: I'm thinking that `getValue()` returns a wrapped value, but remove just delegates to the list so equality will not match the item we want to remove if the value was wrapped. I believe this issue already applies to the current implementation.

src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -93,28 +94,43 @@ public JsonArray(Buffer buf) {
...
+      return val.toString();
+    } else if (val instanceof Instant) {
+      return ISO_INSTANT.format((Instant) val);
+    } else if (val instanceof byte[]) {
+      return BASE64_ENCODER.encodeToString((byte[]) val);
+    } else if (val instanceof Enum) {
+      return ((Enum) val).name();
+    }
+
+    throw new ClassCastException("class " + val.getClass().getName() + " cannot be cast to class java.lang.String");
```
vietj: how about instead using `toString()` ?

vietj:
can you avoid un-necessary code reformatting to make the review easier to read ?

src/main/java/io/vertx/core/json/JsonArray.java
pmlopes: That is also a possibility and if I'm not mistaken how javascript does it but it will change the behavior from the older version that would throw a CCE if the value isn't a CharSequence. I'm open to it, if you prefer

vietj: I believe that being lenient here is good specially if we use this for simple object mapping with SQL client as the intent is to pass a Map<String, Object> from JsonObject to wrap the row.

src/main/java/io/vertx/core/json/impl/JsonUtil.java
```diff
@@ -0,0 +1,114 @@
...
+      val = val.toString();
+    } else if (val instanceof JsonObject) {
+      val = ((JsonObject) val).copy();
+    } else if (val instanceof JsonArray) {
+      val = ((JsonArray) val).copy();
+    } else if (val instanceof Map) {
+      val = (new JsonObject((Map) val)).copy();
+    } else if (val instanceof List) {
+      val = (new JsonArray((List) val)).copy();
+    } else if (val instanceof byte[]) {
```
vietj: we don't want to encode here and instead let the encoder do it

src/main/java/io/vertx/core/json/impl/JsonUtil.java
```diff
@@ -0,0 +1,114 @@
...
+      val = BASE64_ENCODER.encodeToString((byte[]) val);
+    } else if (val instanceof Instant) {
+      val = ISO_INSTANT.format((Instant) val);
```
vietj: we don't want to encode here and instead let the encoder do it

src/main/java/io/vertx/core/json/impl/JsonUtil.java
```diff
@@ -0,0 +1,114 @@
...
+    } else if (val instanceof Enum) {
```
vietj: we don't want to encode here and instead let the encoder do it

vietj:
two comments in the review:

1. we don't test that if you do object.getMap().put("foo", someInstant) then doing object.getInstant("foo") returns an instant value (the same?) and also object.getValue("foo")
2. when an instant is stored it is serialised to String and instead we would like to keep the actual instance . The `JsonCodec` should be responsible to encode the value toString instead (using databind or without when it's not available)

same for `byte[]` type

src/main/java/io/vertx/core/json/impl/JsonUtil.java
```diff
@@ -0,0 +1,114 @@
...
+    } else if (val instanceof List) {
+      val = (new JsonArray((List) val)).copy();
+    } else if (val instanceof byte[]) {
+      // OK
+    } else if (val instanceof Instant) {
+      // OK
+    } else if (val instanceof Enum) {
+      // OK
+    } else {
+      throw new IllegalStateException("Illegal type in Json: " + val.getClass());
```
pmlopes: @vietj with the goal of being lenient, should we relax this rule?

This code is executed during a copy operation e.g.:

```java
JsonObject other = json.copy()
```

If the goal is to allow everything we might want to allow other types during a copy and only fail if the object is encoded using the codec. wdyt?

This will be a significant change I think that this rule is verified on several tests, but nothing really visible for the end user.

pmlopes:
2 tests added to verify that `byte[]` and instants are kept as is (same identity) when put on the underlying map/list.

PR #3201, opened 2019-11-18
Fixes #3200

### LINKED ISSUE #3200: deleteRecursive does not need a recursive parameter
https://github.com/eclipse-vertx/vert.x/blob/6b13bdc3d5ae96da76149cd7d5b732f9e54ffeaa/src/main/java/io/vertx/core/file/FileSystem.java#L418

Since we are changing the API i am not sure if it should be considered a breaking change

Do you want me to make a PR?

zenios:
Signed-off-by: zenios <dimitris.zenios@gmail.com>

Also added missing @Override annotations


vietj:
indeed this is a breaking change can you modify the PR so that it adds new methods without recursive without removing existing ones ?

PR #3202, opened 2019-11-19
Proposal for same site cookie value

pmlopes:

This PR is a proposal to support same site policy for cookies (server side only). There is no parsing involved (as that is a feature only applicable to browsers).

The Same Site policy is supported by all major browsers and a security feature to prevent cross site request forgery.

Although this isn't present in netty (yet) discussions seem to show that this will only land on the 5.x branch which means vertx can't offer the current owasp recommendation on cookies https://www.owasp.org/index.php/SameSite 

As a workaround this PR allows users to build secure cookie based applications and once support lands upstream we can easily drop this and fully support it.

Also note that the chrome team will be "strict" with this flag:

https://www.chromium.org/updates/same-site

### DIFF
--- a/src/main/java/io/vertx/core/http/Cookie.java
+++ b/src/main/java/io/vertx/core/http/Cookie.java
@@ -121,12 +121,11 @@ public interface Cookie {
   /**
    * Sets the same site of this cookie.
    *
-   * @param policy The policy should be one of {@code Lax}, {@code Strict} or {@code None}
+   * @param policy The policy should be one of {@link CookieSameSite}.
    * @return a reference to this, so the API can be used fluently
    */
   @Fluent
-  Cookie setSameSite(String policy);
-
+  Cookie setSameSite(CookieSameSite policy);
 
   /**
    * Encode the cookie to a string. This is what is used in the Set-Cookie header
--- a/src/main/java/io/vertx/core/http/Cookie.java
+++ b/src/main/java/io/vertx/core/http/Cookie.java
@@ -119,6 +119,16 @@ public interface Cookie {
   Cookie setHttpOnly(boolean httpOnly);
 
   /**
+   * Sets the same site of this cookie.
+   *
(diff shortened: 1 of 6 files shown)


pmlopes:
Following discussion on netty: https://github.com/netty/netty/issues/8530

src/main/java/io/vertx/core/http/Cookie.java
```diff
@@ -118,6 +118,16 @@ static Cookie cookie(String name, String value) {
...
   Cookie setHttpOnly(boolean httpOnly);
 
+  /**
+   * Sets the same site of this cookie.
+   *
+   * @param policy The policy should be one of {@code Lax}, {@code Strict} or {@code None}
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  Cookie setSameSite(String policy);
```
vietj: I think we should use an enum here instead of a string

pmlopes:
@vietj PR updated using an `Enum` now.

vietj:
I think this should be documented in .adoc files

ddjones-fss:
Any progress on this? We will soon need this or something similar very soon due to the upcoming Chrome releases.

ddjones-fss:
Can there also be a modification to SessionHandler to add a method of "setSessionCookieSameSite(CookieSameSite.???)" so that we can set this value when instantiating a session?

pmlopes:
@ddjones-fss the session handler should be added once this PR is merged, can you create an issue on vert-web and link to this one so we can track it?

ddjones-fss:
Issue https://github.com/vert-x3/vertx-web/issues/1530 has been created as requested.

pmlopes:
Will need to be backported to 3.9

marcban:
oups we needing the 3.9.0 version urgently , Chrome implemented the change on SameSite cookies by default, and I don't see a way of implementing it...

vietj:
3.9.0 will be released soon @marcban

marcban:
thanks!

PR #3206, opened 2019-11-22
Avoid eventbus npe on verticle factories

pmlopes:
Verticle factories will see an NPE on `vertx.eventBus().xxx()` because the object is only initialized later. Changing the order will avoid it, and allow factories to access the eventbus for doing stuff like registering custom codecs, etc...


pmlopes:
Same applies to shared data.

If a factory would attempt to store any data (think a Singleton as an example) it would NPE too.

vietj:
can you add a test for this ?

PR #3223, opened 2019-12-10
Allow Shareable on JSON copy + Autocast getString()

pmlopes:
This is the follow up of: https://groups.google.com/forum/#!topic/vertx-dev/SNx0XXHO3Pg

* To ease the use of JSON, `getString(key)` will autocast if a value is not of type `String`.
* The `copy()` method will allow besides JSON primitives `Shareable` objects which also implement the `copy()` method making it easy to share such POJOs over the eventbus or shared data.

### DIFF
--- a/src/test/java/io/vertx/core/json/JsonObjectTest.java
+++ b/src/test/java/io/vertx/core/json/JsonObjectTest.java
@@ -331,9 +331,9 @@ public class JsonObjectTest {
     jsonObject.put("bar", 123);
     try {
       jsonObject.getString("bar");
-      fail();
-    } catch (ClassCastException e) {
       // Ok
+    } catch (ClassCastException e) {
+      fail();
     }
 
     // Null and absent values
@@ -357,10 +357,10 @@ public class JsonObjectTest {
     assertEquals("bar", jsonObject.getString("foo", null));
     jsonObject.put("bar", 123);
     try {
-      jsonObject.getString("bar", "wibble");
-      fail();
+      assertEquals("123", jsonObject.getString("bar", "wibble"));
+      // OK, non string types are casted to string using .toString()
     } catch (ClassCastException e) {
-      // Ok
+      fail();
     }
 
     // Null and absent values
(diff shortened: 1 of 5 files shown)


src/test/java/io/vertx/core/json/JsonObjectTest.java
```diff
@@ -1734,6 +1735,44 @@ public void testNoEncode() {
...
+    assertEquals("{\"bd1\":124567890.0987654321}", json.encode());
+  }
+
+  @Test
+  public void testShareable() {
+
+    Shareable myShareable = new Shareable() {
+      @Override
+      public Shareable copy() {
+        return this;
```
vietj: can you add a counter increment to check copy is called in the test ?

src/test/java/io/vertx/core/json/JsonObjectTest.java
```diff
@@ -1734,6 +1735,44 @@ public void testNoEncode() {
...
+
+    // storing Shareable should not be an issue
+    JsonObject json = new JsonObject();
+    json.put("0", myShareable);
+    assertEquals(myShareable, json.getValue("0"));
+    assertSame(myShareable, json.getValue("0"));
+
+    // copy() should allow it too.
+    JsonObject json2 = json.copy();
+  }
```
vietj: you are testing copy() operation but you don't make any assertion after like before, I think you should unwrap the underlying map and check that the map contains the value

src/test/java/io/vertx/core/json/JsonObjectTest.java
```diff
@@ -331,9 +332,9 @@ public void testGetString() {
     jsonObject.put("bar", 123);
     try {
       jsonObject.getString("bar");
-      fail();
-    } catch (ClassCastException e) {
       // Ok
+    } catch (ClassCastException e) {
```
vietj: you can simply remove the catch block instead

src/test/java/io/vertx/core/json/JsonObjectTest.java
```diff
@@ -357,10 +358,10 @@ public void testGetStringDefault() {
     assertEquals("bar", jsonObject.getString("foo", null));
     jsonObject.put("bar", 123);
     try {
-      jsonObject.getString("bar", "wibble");
-      fail();
+      assertEquals("123", jsonObject.getString("bar", "wibble"));
+      // OK, non string types are casted to string using .toString()
     } catch (ClassCastException e) {
-      // Ok
+      fail();
```
vietj: You can simply remove the catch block

PR #3231, opened 2019-12-16
Update HttpMethod to include all known HTTP verbs listed on RFCs

pmlopes:
Currently `HttpMethod` includes the verbs defined in 2 RFCs:

* RFC 2616
* RFC 5789

This pull request adds a few more scattered on other RFCs

* RFC 2518
* RFC 4918
* RFC 4791
* RFC 3253
* RFC 3648
* RFC 3744
* RFC 5323

This allows users to build more than "traditional" web servers with vertx, but for example implement `dav` servers, etc...


Stwissel:
I highly appreciate the additional keyword support. It allows us to implement http based round-trip file editing (webDAV - the same approach Sharepoint takes) and support for calDAV to have read/write support for calendar entries.

src/main/java/io/vertx/core/http/HttpMethod.java
```diff
@@ -13,12 +13,98 @@
...
+
+  private static final Map<String, HttpMethod> METHOD_MAP = new HashMap<>();
+
+  static {
+    for (HttpMethod method : HttpMethod.values()) {
+      METHOD_MAP.put(method.toString(), method);
+    }
+  }
+
+  public static HttpMethod fromHTTPProtocol(String name) {
```
vietj: missing javadoc

vietj:
I'm wondering if instead we should not make this enum a plain Java interface instead

vietj:
code would not break except when people used it as a switch in an enum, that they could replace with a String switch. That would allow getting rid of HttpMethod.OTHER and allow to plug any HTTP method.

pmlopes:
See #3252

PR #3239, opened 2020-01-01
Added default value for query parameters.

hradecek:
Return default value instead of null if query param is not present. 
 - removed unused imports

Improvement based on: [vertx-web#1473](https://github.com/vert-x3/vertx-web/issues/1473)

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServerRequest.java
+++ b/src/main/java/io/vertx/core/http/HttpServerRequest.java
@@ -29,6 +29,7 @@ import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
 import javax.security.cert.X509Certificate;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * Represents a server-side HTTP request.
@@ -203,6 +204,19 @@ public interface HttpServerRequest extends ReadStream<Buffer> {
   @Nullable
   default String getParam(String paramName) {
     return params().get(paramName);
+  }
+
+  /**
+   * Return the first param value with the specified name or {@code defaultValue} when the query param is not present
+   *
+   * @param paramName    the param name
+   * @param defaultValue the default value, must be non-null
+   * @return the param value or {@code defaultValue} when not present
+   */
+  default String getParam(String paramName, String defaultValue) {
+    Objects.requireNonNull(defaultValue, "defaultValue");
+    final String paramValue = params().get(paramName);
+    return paramValue != null ? paramValue : defaultValue;
   }
 
   /**
--- a/src/test/java/io/vertx/core/http/HttpTest.java
+++ b/src/test/java/io/vertx/core/http/HttpTest.java
@@ -857,6 +857,32 @@ public abstract class HttpTest extends HttpTestBase {
   }
 
   @Test
+  public void testGetParamDefaultValue() {
+    String paramName1 = "foo";
+    String paramName1Value = "bar";
+    String paramName2 = "notPresentParam";
+    String paramName2DefaultValue = "defaultValue";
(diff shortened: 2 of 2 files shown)


src/main/java/io/vertx/core/http/HttpServerRequest.java
```diff
@@ -170,6 +169,18 @@ default String getParam(String paramName) {
     return params().get(paramName);
   }
 
+  /**
+   * Return the first param value or defaultValue if the query param is not provided.
```
vietj: "Return the first param value with the specified name or `@code{defaultValue}` when the query param is not present."

src/main/java/io/vertx/core/http/HttpServerRequest.java
```diff
@@ -170,6 +169,18 @@ default String getParam(String paramName) {
...
+   *
+   * @param paramName    the param name
+   * @param defaultValue the default value, must be non-null
+   * @return the param value or default value
```
vietj: "the param value or `@code{defaultValue}` when not present"

vietj:
Can you add a simple test for this in `HttpTest` ?

src/main/java/io/vertx/core/http/HttpServerRequest.java
```diff
@@ -170,6 +169,18 @@ default String getParam(String paramName) {
...
+   */
+  default String getParam(String paramName, String defaultValue) {
+    Objects.requireNonNull(defaultValue, "defaultValue");
+    return params().contains(paramName) ? params().get(paramName) : defaultValue;
```
petarov: This method is a good idea! 😎  I know I'm nitpicking, but may I suggest to skip the `contains()` call and just do a `get()` and if `null`, then return `defaultValue`?  This will skip at least one MultiMap call.

src/main/java/io/vertx/core/http/HttpServerRequest.java
hradecek: Done.

src/main/java/io/vertx/core/http/HttpServerRequest.java
hradecek: Done.

src/main/java/io/vertx/core/http/HttpServerRequest.java
hradecek: > I know I'm nitpicking

Noo, I was actually thinking about this way of implementation too, but I found `contains()` then `get()` more general and interface-friendlier.

But you're right as definition of `MultiMap#get()` is: _Something or **null** if there is no such entry_. 
Re-implemented as you have suggested, thanks.

hradecek:
Added `HttpTest`, commit amended.

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -782,6 +782,29 @@ public void testNoParams() {
...
+
+    server.requestHandler(req -> {
+      assertTrue(req.params().contains(paramName1));
+      assertEquals(paramName1Value, req.getParam(paramName1, defaultValue));
+      assertFalse(req.params().contains(paramName2));
+      assertEquals(defaultValue, req.getParam(paramName2, defaultValue));
+      req.response().end();
+    });
+
+    server.listen(onSuccess(server -> {
```
vietj: please start server like in other tests (using testAddress) so the test can run with other config such as native transport.

hradecek:
Shame on me I have left this opened for so a long time! 

I've rebased to `master` and resolved all comments.

vietj:
can you rebase this PR ?

vietj:
thanks

PR #3244, opened 2020-01-06
Expose setTimer and setPeriodic on ContextInternal

tsegismont:
This can simplify code when a component is bound to a context and we have to schedule async tasks


src/main/java/io/vertx/core/impl/ContextInternal.java
```diff
@@ -255,4 +250,8 @@ static ContextInternal current() {
    */
   ContextInternal duplicate();
 
+  long setPeriodic(long delay, Handler<Long> handler);
```
vietj: javadoc :-)

tsegismont: Done

PR #3247, opened 2020-01-07
HttpMethod is an interface

### LINKED ISSUE #3246: HttpMethod is an interface
In Vert.x 2 and 3, the `HttpMethod` is declared as an enum. This leashes the extensibility of HTTP and prevents modelling other HTTP methods with this type unless using `HttpMethod#OTHER` in addition with the `rawMethod` attribute on `HttpServerRequest` and `HttpClientRequest`.

We can turn this enum into an interface in Vert.x 4 with a few benefits:

1. non predefined HTTP methods are now handled like others without special handling
2. one can define its own HTTP method constant for other protocols (like 

And at the expanse of 2 breakages:

1. the enum cannot be used as is dispatching a call with a `switch` construct. Instead the switch must be done on the `HttpMethod#name()` string value.
2. the `rawMethod` property is removed because now the `HttpMethod` can model the raw value.

vietj:
closes #3246

### DIFF
--- a/src/test/java/io/vertx/core/http/HttpTest.java
+++ b/src/test/java/io/vertx/core/http/HttpTest.java
@@ -3495,25 +3495,14 @@ public abstract class HttpTest extends HttpTestBase {
   }
 
   @Test
-  public void testOtherMethodWithRawMethod() {
-    try {
-      client.request(HttpMethod.OTHER, testAddress, DEFAULT_HTTP_PORT, DEFAULT_HTTP_HOST, "/somepath", resp -> {
-      }).end();
-      fail();
-    } catch (IllegalStateException expected) {
-    }
-  }
-
-  @Test
   public void testOtherMethodRequest() {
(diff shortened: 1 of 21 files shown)


src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -3494,26 +3494,16 @@ public void testDumpManyRequestsOnQueue() throws Exception {
...
-      fail();
-    } catch (IllegalStateException expected) {
-    }
-  }
-
   @Test
   public void testOtherMethodRequest() {
     server.requestHandler(r -> {
-      assertEquals(HttpMethod.OTHER, r.method());
+//      assertEquals(HttpMethod.OTHER, r.method());
```
vietj: oups :-)

vietj:
after this is merged we should check whether it breaks other projects in the stack

Issue #3126, opened 2019-10-04
Clustered Lock is not released if more locks are requested than available worker threads


naveen140990:
I am facing this issue while migrating from vertx 3.5.4 to 3.7.x.

If I am requesting more locks than available workers, then lock is not released in vertx versions 3.6.x and 3.7.x. Please find the reproducer below:

https://github.com/naveen140990/ClusteredLockReproducer/blob/master/src/test/java/EventbusTest.java

Above same code works if I change vertx version to 3.5.4.
Also if I reduce numTestEvents in dataprovider(for above reproducer code) to a number less than 20 then also everything seems to be working.

tsegismont:
@naveen140990 could you please share a simplified reproducer? Thanks

naveen140990:
@tsegismont I have refactored code to simplify reproducer. numTestEvents variable holds value of concurrent locks requested.  When I use vertx version 3.5.4, locks are acquired correctly. But when version is upgraded to 3.7.0, only first lock is acquired and after 10 seconds(default timeout I guess), rest all other efforts return "io.vertx.core.VertxException: Timed out waiting to get lock" error message.

naveen140990:
@tsegismont This issue is present in recently released 3.8.2 version as well. Have you a got a chance to look into this? Thanks !

tsegismont:
@naveen140990 in Vert.x 3.5, locks were acquired on the internal worker pool and released on the worker pool. Your test passes with 3.5 but this implementation had other problems (like preventing internal tasks to execute).

We could create a dedicated pool to release locks (a cached thread pool with core pool size set to zero).

naveen140990:
Hi @tsegismont , I am thinking of  experimenting with vert.x 4 using milestone versions for a new project, but for that I would need this implementation. Could you please help me by giving an idea if this would indeed be available in general release? Thanks !

tsegismont:
Not sure what you mean. The implementation in master is present in
milestone4.

Le jeu. 19 déc. 2019 à 12:43, Naveen Katiyar <notifications@github.com> a
écrit :

naveen140990:
Probably I was not clear enough . Actually, I was asking about the fix of reported issue(Clustered Lock is not released if more locks are requested than available worker threads) . I have checked in milestone 4, but issue still exists. So, I wanted to enquire if this fix would be eventually provided in subsequent milestone versions ?

tsegismont:
[...]

PR #3264, opened 2020-01-28
Added test for worker starvation when lock concurrency is greater than worker pool size

tsegismont:
See #3126


tsegismont:
@fabienpomerol can you please take a look as well? Thank you

vietj:
+1 for all CM too

PR #3278, opened 2020-02-06
Completed futures should operate on provided context

tsegismont:
SucceededFuture and FailedFuture, like other futures, take a context as constructor param.
If not null, this context should be used to dispatch the execution of future handlers.

### DIFF
--- a/src/test/java/io/vertx/core/FutureTest.java
+++ b/src/test/java/io/vertx/core/FutureTest.java
@@ -1575,11 +1575,16 @@ public class FutureTest extends VertxTestBase {
   }
 
   @Test
-  public void testCompletedFuturesContext() {
+  public void testCompletedFuturesContext() throws Exception {
     waitFor(4);
 
     Thread testThread = Thread.currentThread();
     ContextInternal context = (ContextInternal) vertx.getOrCreateContext();
+
+
+    CompletableFuture<Thread> cf = new CompletableFuture<>();
+    context.runOnContext(v -> cf.complete(Thread.currentThread()));
+    Thread contextThread = cf.get();
 
     Future.succeededFuture().onSuccess(v -> {
       assertSame(testThread, Thread.currentThread());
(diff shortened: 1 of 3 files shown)


src/test/java/io/vertx/core/FutureTest.java
```diff
@@ -1570,4 +1573,38 @@ public void testFromCompletionStageWithContext() {
...
+    ContextInternal context = (ContextInternal) vertx.getOrCreateContext();
+
+    Future.succeededFuture().onSuccess(v -> {
+      assertSame(testThread, Thread.currentThread());
+      assertNull(Vertx.currentContext());
+      complete();
+    });
+
+    context.succeededFuture().onSuccess(v -> {
+      assertNotSame(testThread, Thread.currentThread());
```
vietj: can you replace this check with event-loop thread instead of not being the junit thread ?

tsegismont: Do you mean this?

```java
    context.succeededFuture().onSuccess(v -> {
      assertNotSame(testThread, Thread.currentThread());
      assertThat(Thread.currentThread(), is(instanceOf(VertxThread.class)));
      assertSame(context, Vertx.currentContext());
      complete();
    });
```

vietj: or even better the thread of the context that is associated with the future (that you can latch before)

vietj: I think something like:

```java
ctx.runOnContext(v -> {
  Thread ctxThread = Thread.currentThread();
  new Thread(() -> {
    ctx.succeededFuture().onSuccess(v -> {
      // Checks here
    });
  }).start();
});
```

tsegismont: What do we want to verify here? That a standard context always operates on the same thread? I believe we could take this for granted here.

vietj: this is what should be checked, given that this implementation is different from `FutureImpl`. I was taking for granted that this impl was bug free 2 days ago :-)

tsegismont: @vietj I've added some checks. PTAL

Issue #2762, opened 2018-12-12
[Request] Add global MessageCodec


ufoscout:
It would be good to be able to register a global MessageCodec to be used when a specific codec for a message is not provided.
As of today (Vertx 3.6.0) when a message without a registered MessageCodec is sent through the event bus, a RuntimeError is thrown. Nevertheless, most of the time a silent fallback to a default codec would be enough.

For example, in our current application, we want that all messages are sent in Json format; we have tens of different messages and we are forced to register for each of them a specific codec.

vietj:
you mean a default codec ?

ufoscout:
@vietj I mean a single generic message codec able to handle messages of whatever class type.

In this situation:
```
// My app sends 100 different types of messages on the bus
class MyMessageType1 {...}
class MyMessageType2 {...}
...
class MyMessageType100 {...}

// To make it work I have to register 100 default MessageCodec(s)
var bus = vertx().eventbus();

bus.registerDefaultCodec(MyMessageType1.class, new MyMessageType1_codec());
bus.registerDefaultCodec(MyMessageType2.class, new MyMessageType2_codec());
...
bus.registerDefaultCodec(MyMessageType100.class, new MyMessageType100_codec());
```
However, in my case ALL of the MessageCodecs are exactly the same! they always serialize/deserialize using vertx.Json. So I ended up having a single MessageCodec that I have to manually register for each message type. This is the (kotlin) code of it:
```kotlin
class JsonMessageCodec<T>(private val clazz: Class<T>): MessageCodec<T, T> {

    override fun encodeToWire(buffer: Buffer, customMessage: T) {
        buffer.appendString(Json.encode(customMessage))
    }

    override fun decodeFromWire(position: Int, buffer: Buffer): T {
        val length = buffer.getInt(position)
        val end_position = position + length
        val jsonStr = buffer.getString(position , end_position)
        return Json.decodeValue(jsonStr, clazz)
    }
    ...
}
```

Using this generic JsonMessageCodec, the eventbus setup becomes a long list of boilerplate code (pseudo code):
```
// Register 100 the same MessageCodec
var bus = vertx().eventbus();

bus.registerDefaultCodec(MyMessageType1.class, new JsonMessageCodec(MyMessageType1.class));
bus.registerDefaultCodec(MyMessageType2.class, new JsonMessageCodec(MyMessageType2.class));
...
[...]

PR #3279, opened 2020-02-09
Use super classes to search MessageCodecs.

rodrigoSaladoAnaya:
Motivation:

Use a super class to implement a MessageCodec, it allow to send/receive; if you need, a generic type of class.

In my work we have a monolith and I want to break it into modular services and I will use the eventBus to send the updates of those modules, but since at this moment all the services extend from a base class, I need to implement a MessageCodec from that class base.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/CodecManager.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/CodecManager.java
@@ -92,6 +92,9 @@ public class CodecManager {
       codec = BYTE_MESSAGE_CODEC;
     } else if (body instanceof ReplyException) {
       codec = defaultCodecMap.get(body.getClass());
+      if(codec == null) {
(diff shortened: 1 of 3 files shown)


vietj:
I mentioned in this discussion (https://github.com/eclipse-vertx/vert.x/issues/2762#issuecomment-531591802) we should have such approach, however this one does not seem recursive / generic enough. Would you mind improving your contribution for this ?

rodrigoSaladoAnaya:
> I mentioned in this discussion ([#2762 (comment)](https://github.com/eclipse-vertx/vert.x/issues/2762#issuecomment-531591802)) we should have such approach, however this one does not seem recursive / generic enough. Would you mind improving your contribution for this ?

Yes of course, I keep you posted.

rodrigoSaladoAnaya:
Hi @vietj, can you see #3280, thanks.

PR #3283, opened 2020-02-12
Use a http server to hold the port busy before the real test.

rodrigoSaladoAnaya:
Signed-off-by: Rodrigo Salado Anaya <rodrigo.salado.anaya@gmail.com>

Motivation:
Improvement to the test `testListenInvalidPort`, currently when port 80 is available the test fails.

```java
mvn -Dtest=NetTest#testListenInvalidPort test
```

### DIFF
--- a/src/test/java/io/vertx/core/net/NetTest.java
+++ b/src/test/java/io/vertx/core/net/NetTest.java
@@ -904,21 +904,22 @@ public class NetTest extends VertxTestBase {
   @Test
   public void testListenInvalidPort() {
     final int port = 9090;
-    final HttpServer httpServer = vertx.createHttpServer();
-    httpServer.requestHandler(r -> {}).listen(port, httpResult -> {
-      server.close();
+    HttpServer httpServer = vertx.createHttpServer();
+    server.close();
+    final Runnable netServerTest = () -> {
       server = vertx.createNetServer(new NetServerOptions().setPort(port));
       server.connectHandler((netSocket) -> {
       }).listen(ar -> {
-        if(httpResult.succeeded()) {
-          httpServer.close();
-        }
         assertTrue(ar.failed());
         assertFalse(ar.succeeded());
         assertNotNull(ar.cause());
+        httpServer.close();
         testComplete();
       });
-    });
+    };
+    httpServer.requestHandler(ignore -> {}).listen(port)
+      .onSuccess(s -> netServerTest.run())
+      .onFailure(e -> netServerTest.run());
     await();
   }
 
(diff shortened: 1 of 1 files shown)


src/test/java/io/vertx/core/net/NetTest.java
```diff
@@ -903,16 +903,21 @@ public void testConnectInvalidConnectHandler() throws Exception {
...
-    server = vertx.createNetServer(new NetServerOptions().setPort(80));
-    server.connectHandler((netSocket) -> {
-    }).listen(ar -> {
-      assertTrue(ar.failed());
-      assertFalse(ar.succeeded());
-      assertNotNull(ar.cause());
-      testComplete();
+    final int port = 9090;
+    final HttpServer httpServer = vertx.createHttpServer();
+    httpServer.requestHandler(r -> {}).listen(port, httpResult -> {
```
vietj: can you use the onSuccess / onFailure lambdas to improve the test ?

vietj:
Actually I meant rather the one defined in AsyncTestBase, e.g

```java
server.listen(onSuccess(s ->
   // ...
});
```

rodrigoSaladoAnaya:
Ok, I will review other tests that use them to apply the change.

rodrigoSaladoAnaya:
What do you think @vietj ?

src/test/java/io/vertx/core/net/NetTest.java
```diff
@@ -903,18 +903,18 @@ public void testConnectInvalidConnectHandler() throws Exception {
...
-    }).listen(ar -> {
-      assertTrue(ar.failed());
-      assertFalse(ar.succeeded());
-      assertNotNull(ar.cause());
-      testComplete();
-    });
+    final int port = 9090;
+    final HttpServer httpServer = vertx.createHttpServer()
+      .requestHandler(ignore -> {})
+      .listen(port, httpResult ->
```
vietj: I think here we need an onSuccess(...) block

rodrigoSaladoAnaya:
I didn't use the method onSuccess, to take advantage of the fact that the port cannot be occupied and the test can still be valid.

Although it is also a good idea to narrow the test as much as possible, ensuring that the port was used before the actual test and released later.

I will apply the suggested change, just the indentation does not convince me.

rodrigoSaladoAnaya:
Hi @vietj, do you agree with the last change?

src/test/java/io/vertx/core/net/NetTest.java
```diff
@@ -903,18 +903,18 @@ public void testConnectInvalidConnectHandler() throws Exception {
...
+      .requestHandler(ignore -> {})
+      .listen(port, onSuccess(s ->
+        vertx.createNetServer()
+          .connectHandler(ignore -> {})
+          .listen(port, onFailure(error -> {
+            assertNotNull(error);
+            testComplete();
+          }))));
     await();
+    httpServer.close();
```
vietj: we need this close to be in a finally block

vietj:
one more comment @rodrigoSaladoAnaya

PR #3292, opened 2020-02-19
Removed deprecated Starter class

tsegismont:
For quite a few versions now the recommendation is to use the Launcher class.


tsegismont:
@vietj when this is merged I'll update the breaking changes page

PR #3294, opened 2020-02-20
Removed multithreaded worker verticles from the doc

tsegismont:
We had questions on the forum and gitter about MT verticles again.

When we deprecated them, the agreement was to document an alternative instead of removing them from the doc completely.

I believe it is better now if new users avoid to loose time investigating something that goes away in 4.0.

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -391,13 +391,11 @@ include::override/verticles.adoc[]
 
 === Verticle Types
 
-There are three different types of verticles:
+There are two different types of verticles:
 
 Standard Verticles:: These are the most common and useful type - they are always executed using an event loop thread.
 We'll discuss this more in the next section.
 Worker Verticles:: These run using a thread from the worker pool. An instance is never executed concurrently by more
-than one thread.
-Multi-threaded worker verticles:: These run using a thread from the worker pool. An instance can be executed concurrently by more
 than one thread.
 
 === Standard verticles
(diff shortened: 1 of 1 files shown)


tsegismont:
@vietj if you're fine with this, I'll upgrade the `3.9` branch too. Then I can update the website with an ovverride.

vietj:
that's fine for this, however for 3.9 we need to still mention they exist but are deprecated and give ways to achieve the same feature (basically executeBlocking + ordered = false)

tsegismont:
@vietj this is a PR for `3.8`. So I guess you thought it was for `master`, right?

vietj:
oh right, but you need to do this PR for 3.9 then

tsegismont:
IIUC, you want to keep the documentation about MT worker verticles, just add a warning about deprecation?

vietj:
it should remain documented because it is part of 3.x, however it should be discouraged to use it and show how it can be achieved differently.

tsegismont:
@vietj for logging, we stop documenting the deprecated package a few versions ago and it seemed it worked (no more questions about it).

I though we could do the same for MT workers given:

- they are deprecated since 3.6
- the workaround has been documented in 3.6
- we still get questions now

What's the benefit for keeping them?

vietj:
@tsegismont yes you are right

vietj:
however this PR should be moved for 3.9 branch and not 3.8 as it is

tsegismont:
I had in mind to cherry-pick to 3.9 after merging and then fix the website with an override. Then the website would be fixed whether or not we release a 3.8.6.

tsegismont:
Closing, will rebase on `3.9`

PR #3297, opened 2020-02-24
Fixes #3289 : Renames toJsonXXX() methods from Buffer to mapToJsonXXX()

### LINKED ISSUE #3289: Rename Buffer toJsonXXX() to mapToJsonXXX()
The `Buffer` interface defines these methods:

- `toJsonObject()`: parse the buffer content as a JSON object
- `toJsonArray()`: parse the buffer content as a JSON array
- `toJson()`: parse the buffer content as a JSON element or value

These methods are actually mapping methods a JSON element or value according whose type depends on the content of the buffer. The `toJson()` method convention is actually used to encode data objects to their JSON representation. e.g for a buffer it would actually encode a base 64 representation of the content like `JsonObject` does. The `toJson()` method will likely be changed in Vert.x 4 to return an encoded base 64 String.

jlengrand:
* Renames toJsonObject to mapToJsonObject and deprecates the method
* Renames toJsonArray to mapToJsonArray and deprecates the method
* Renames toJson to mapToJson and deprecates the method

Signed-off-by: Julien Lengrand-Lambert <julien@lengrand.fr>

Motivation:

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/buffer/Buffer.java
+++ b/src/main/java/io/vertx/core/buffer/Buffer.java
@@ -133,20 +133,47 @@ public interface Buffer extends ClusterSerializable, Shareable {
 
   /**
    * Returns a Json object representation of the Buffer.
+   *
+   * @deprecated use {@link #mapToJsonObject()} instead.
    */
+  @Deprecated
   JsonObject toJsonObject();
 
   /**
    * Returns a Json array representation of the Buffer.
+   *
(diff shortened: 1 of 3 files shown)


vietj:
we decided to not go forward with this change

vietj:
thanks for the contribution and effort, that's much appreciated

jlengrand:
My pleasure!

PR #3320, opened 2020-03-02
avoid rawtype warning

CaoYusong:
Motivation:

Conformance:


vietj:
this code cannot currently compile because the code generator does not support wildcards, so we cannot have such PR until the code generator supports. We are planning to try having this for Vert.x 4.0, for now i'll just close the PR because if we do support wildcard, the only use case we know in codebase is this one for now.

PR #3343, opened 2020-03-25
Bug when unregistering a message consumer and registering a handler again

jponge:
Fixes a bug when unregistering a message consumer then registering it again where the new handler would not receive any message.

Spotted in https://github.com/vert-x3/vertx-rx/issues/221 and fixes the corresponding issue on my machine.


jponge:
You can safely ignore the ECA check failure. I signed off my commits and GPG-signed them with my Red Hat address, so... we're covered.

PR #3347, opened 2020-03-28
feat: add keepAliveTTL to evict connection even if active

srijan02420:
Signed-off-by: Srijan Gupta <srijan02420@gmail.com>

Motivation:

If you are making http request to a remote-service with `keep-alive: true` the connection will never recycle if the request rate is high. In this scenario if a deployment took place for the remote service with blue-green configuration. Our client will still hit requests to old stack.
To solve this keeping a global TTL for active connections makes sense.
Similar to node's: https://www.npmjs.com/package/agentkeepalive#new-agentoptions


vietj:
@srijan02420 there is no test for this feature

vietj:
my point is that there is a new test for the connection pool, but no code is changed in connection pool (so at best it is a new test), the feature is added for HTTP/1.1 protocol and there is no test for with the HTTP client using HTTP/1

srijan02420:
> @srijan02420 there is no test for this feature

@vietj I will add a tests for this in `testClientOptions`. Suggest where else I need to add tests.

srijan02420:
@vietj I have added `testActiveConnectionTTL` with HttpClient using this option. Let me know if this works.

srijan02420:
@vietj any update on this?

vietj:
not yet, I need to understand the issue and see if that's the right thing to do

srijan02420:
> not yet, I need to understand the issue and see if that's the right thing to do

Okay. Let me know if you need more info about the use case.

vietj:
a review / analysis is set for 3.9.1

vietj:
I have had a look at it and it seems an interesting feature, however while it seems to work in this case the current implementation is not correct because it will check the validity after the connection has been used which means that it needs to be used before closing the connection.

I believe this should be a feature of the pool itself and not the layer using the pool, so the pool would close such connection like it does today for expired connections.

vietj:
Actually I believe it is correctly done :-), sorry for the incorrect analysis. Indeed the pool will take care of this. Allow me to do further analysis.

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -87,6 +87,7 @@
   private int keepAliveTimeout;
   private int seq = 1;
   private long bytesRead;
+  private long initialTimestamp = System.currentTimeMillis();
```
vietj, line 90: we should call this only when the option is active, now need to call it otherwise

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -898,7 +899,13 @@ public void createStream(ContextInternal context, HttpClientRequestImpl req, Pro
   }
 
   private void recycle() {
-    long expiration = keepAliveTimeout == 0 ? 0L : System.currentTimeMillis() + keepAliveTimeout * 1000;
+    long now = System.currentTimeMillis();
```
vietj, line 902: we should compute this only when the option is active

src/test/java/io/vertx/core/net/ConnectionPoolTest.java
```diff
@@ -487,6 +487,26 @@ public void testDiscardExpiredConnections() {
     // assertWaitUntil(mgr::closed);
   }
 
+  @Test
```
vietj, line 490: why is this test necessary ? is it a new test to ensure a behavior that was not tested previously ?

src/test/java/io/vertx/core/net/ConnectionPoolTest.java
```diff
@@ -923,6 +944,14 @@ synchronized long recycle(long timestamp) {
       return inflight;
     }
 
+    synchronized long recycle(long keepAliveTimeout, long activeConnectionTTL, long currentTimestamp) {
```
vietj, line 947: why is this necessary ?

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -196,6 +206,8 @@
   private int maxPoolSize;
   private boolean keepAlive;
   private int keepAliveTimeout;
+  private boolean isActiveConnectionTTL;
```
vietj, line 209: I think those should be named differently, i.e in relation with the keep-alive feature, I think that keepAliveTTL is a better choice. Also these should be a single value that is either which means disabled or a positive value that is the TTL value.

srijan02420: @vietj I will change `isActiveConnectionTTL` to `keepAliveTTL`.
I kept it as single variable initially and set default value as -1 because int in java cannot be null.
It felt it is sort of confusing.
Please suggest if I should revert back to that or you have something else in mind.

src/test/java/io/vertx/core/net/ConnectionPoolTest.java
srijan02420: It's actually not necessary. I got confused before. I am removing it.

src/test/java/io/vertx/core/net/ConnectionPoolTest.java
srijan02420: Yes not necessary. Removing it.

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -680,6 +688,35 @@ public HttpClientOptions setKeepAliveTimeout(int keepAliveTimeout) {
     return this;
   }
 
+  /**
+   * Is active connection ttl enabled on the client?
+   *
+   * @return {@code true} if enabled
+   */
+  public boolean isKeepAliveTTLEnabled() {
```
vietj, line 696: can you annotate with `@GenIgnore` to not confuse the code generator ?

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -87,6 +87,7 @@
   private int keepAliveTimeout;
   private int seq = 1;
   private long bytesRead;
+  private long initialTimestamp;
```
vietj, line 90: make it final

srijan02420: @vietj if I will make it final then will have to initialise it.
Then will not be able to do 
```
if (options.isKeepAliveTTLEnabled())
      initialTimestamp = System.currentTimeMillis();
```

vietj: you can init it with `0L`

srijan02420: Yes but then cannot reassign it in constructor when options.isKeepAliveTTLEnabled() is true as 
```
initialTimestamp = System.currentTimeMillis();
```

srijan02420:
@vietj done with the changes.
Any reason why `src/main/asciidoc/dataobjects.html` is not in .gitignore.

vietj:
@srijan02420 I believe the work here is good but incomplete because validity cannot be checked when the connection is in the pool. I will improve the current pool so that now `recycle` is just a simple operation with no expiration timeout and the pool will make the actual validity check.

vietj:
@srijan02420 can you try redo your contribution onto https://github.com/eclipse-vertx/vert.x/tree/pool-connection-validity-check-rework to check it works ?

srijan02420:
@vietj sure. Doing that. Thanks.

srijan02420:
@vietj check now.

vietj:
I will finish this feature in master and then backport to 3.9 then you can update your PR for 3.9 and create a new one for master

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -897,8 +900,19 @@ public void createStream(ContextInternal context, HttpClientRequestImpl req, Pro
...
 
+  @Override
+  public boolean isValid() {
+    return expirationTimestamp > 0 && System.currentTimeMillis() <= expirationTimestamp;
+  }
+
   private void recycle() {
-    long expiration = keepAliveTimeout == 0 ? 0L : System.currentTimeMillis() + keepAliveTimeout * 1000;
-    listener.onRecycle(expiration);
+    if(!options.isKeepAliveTTLEnabled())
```
vietj, line 909: this should actually be moved to `isValid` and when initialTimestamp > keepAliveTTL then return false

vietj:
we also need a new test that check that the connection is closed from the pool when the connection is not active, so doing something like:

1/ make a client request when response comes, get the `HttpConnection` and set a `closeHandler` on the connection
2/ wait until the pool close the connection itself due to keepAliveTTL without an interaction with the pool

srijan02420:
> we also need a new test that check that the connection is closed from the pool when the connection is not active, so doing something like:
> 
> 1/ make a client request when response comes, get the `HttpConnection` and set a `closeHandler` on the connection
> 2/ wait until the pool close the connection itself due to keepAliveTTL without an interaction with the pool

@vietj isn't `testKeepAliveTTL()` is doing this only?

vietj:
I think there are two missing cases that are not covered:

1/ we need to test connections are evicted when keepAliveTimeout = 0 , in this case the vertx timer to evict connection from the pool is not started, so no check will be done

2/ HTTP/2 protocol is not implemented nor tested

srijan02420:
You are right. I will write test for 1 and implement for H2 as well. Initially I was thinking because H2 uses only 1 connection and multiplex on that so there is nothing like pool. But I think it will make sense for that as well because it will help for blue/green deployments when H2 as wll.

vietj:
It is still pooled actually (you pool a single connection) and handled in the same pool, the difference is that the concurrent for HTTP/2 is not the same.

srijan02420:
@vietj I added test for `keepAliveTimeout = 0`. Please check if this was something you had in mind, I am still little bit confused if I understood that.
Meanwhile I am working for H2 implementation and test.

vietj:
@srijan02420 actually I said something wrong, I thought that having keep alive timeout = 0 , disables the timer (see `HttpClientImpl` constructor) and it is actually the cleaner period.

I need to have a look again whether the timer is necessary when keep alive timeout = 0. The reason is that if timeout = 0 (which is never the case) I'm wondering if it's needed to close idle connections.

srijan02420:
> @srijan02420 actually I said something wrong, I thought that having keep alive timeout = 0 , disables the timer (see `HttpClientImpl` constructor) and it is actually the cleaner period.
> 
> I need to have a look again whether the timer is necessary when keep alive timeout = 0. The reason is that if timeout = 0 (which is never the case) I'm wondering if it's needed to close idle connections.

Okay. Let me know.

vietj:
@srijan02420 branch has been updated with new commits, I'll backport that to 3.9 branch

srijan02420:
@vietj once you do that I will rebase with 3.9 branch and update this PR to that branch. And raise a separate PR for master.

vietj:
@srijan02420 3.9 has been updated

srijan02420:
@vietj rebased and updated.
If you want to add this to 3.9 as well. Will cheery-pick and create PR for that separately.

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -76,6 +77,11 @@
    */
   public static final int DEFAULT_KEEP_ALIVE_TIMEOUT = 60;
 
+  /**
+   *  Default keep alive time to live. -1 represents that it is disabled. Connection will never drop if active.
```
vietj, line 81: _{@code -1}_

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -900,7 +902,8 @@ public void createStream(ContextInternal context, HttpClientRequestImpl req, Pro
 
   @Override
   public boolean isValid() {
-    return expirationTimestamp == 0 || System.currentTimeMillis() <= expirationTimestamp;
+    long now = System.currentTimeMillis();
+    return expirationTimestamp == 0 || now <= expirationTimestamp && (!options.isKeepAliveTTLEnabled() || options.getKeepAliveTTL() * 1000 > now - initialTimestamp);
```
vietj, line 906: can you surround with parenthesis the two expressions right and left of `&&` to make it more clear ?

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5048,6 +5048,19 @@ public void testKeepAliveTimeout() throws Exception {
     testKeepAliveTimeout(options, 1);
   }
 
+  @Test
+  public void testKeepAliveTTL() throws Exception {
+    server.requestHandler(req -> {
+      req.response().end();
+    });
+    HttpClientOptions options = createBaseClientOptions()
```
vietj, line 5056: I believe we should use longer keepAliveTimeout values here like 3 seconds

srijan02420: Sorry didn't understand you here.
I used 300 sec as keepAliveTimeout and 3 sec as keepAliveTTL. So connection should close after 3 sec.

vietj:
out of my mind @srijan02420 , what is a good TTL value for blue-green deployments when using this feature ?

srijan02420:
> out of my mind @srijan02420 , what is a good TTL value for blue-green deployments when using this feature ?
@vietj 5 mins i.e. 300 sec seems to be a fine number. Now I understood you want to keep that value in the test. They only problem is the test will have to wait that long.

vietj:
@srijan02420 a small value is fine for tests, I was asking in case somebody asks me what is a good value to use in practice :-)

srijan02420:
@vietj should we merge it. My team desperately need this feature. :)

vietj:
@srijan02420 sure it will be merged, don't worry

src/test/java/io/vertx/core/http/HttpTest.java
vietj: ok sorry

vietj:
@srijan02420 I am actually not sure the current approach we chose for this feature is correct. I think we should try this test:

1/ define a pool with a max connection size = 1 and keepAliveTTL = 1 (1 second)
2/ make two HTTP requests
3/ the HTTP server will wait for more than one second (1.5 second)
4/ the pool should close the first connection before serving the 2nd request still waiting in the pool

I am not sure this is currently satisfied, can you add a test for this case ?

srijan02420:
> @srijan02420 I am actually not sure the current approach we chose for this feature is correct. I think we should try this test:
> 
> 1/ define a pool with a max connection size = 1 and keepAliveTTL = 1 (1 second)
> 2/ make two HTTP requests
> 3/ the HTTP server will wait for more than one second (1.5 second)
> 4/ the pool should close the first connection before serving the 2nd request still waiting in the pool
> 
> I am not sure this is currently satisfied, can you add a test for this case ?

Sure. Adding.

vietj:
thanks

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5061,6 +5061,49 @@ public void testKeepAliveTTL() throws Exception {
     testKeepAliveTimeout(options, 1);
   }
 
+  @Test
+  public void testKeepAliveTTLWith2Req() throws Exception {
```
srijan02420, line 5065: @vietj this is something you had in mind right?

vietj: the test needs to check that each connection obtained for the two responses are different instances (i.e not the same physical connections)

src/main/java/io/vertx/core/net/impl/clientconnection/Pool.java
```diff
@@ -310,6 +291,26 @@ private Runnable nextTask() {
         };
       }
     }
+    if (waitersQueue.size() > 0) {
```
srijan02420, line 294: @vietj we will have to check if connections are valid and expire them before assigning the connection.

src/test/java/io/vertx/core/http/HttpTest.java
srijan02420: Added test for same.

vietj:
@srijan02420 I see your suggestion about validity check I think it is valid but only in the case of an HTTP/1 keep alive connection. It will not work for HTTP/2 and the connection would remain in the pool.

So it seems for this case we would need to do it differently using a shutdown phase, i.e the connection is still in the pool but will not accept any new connection until it is closed.

vietj:
The other reason I think it should use shutdown is that connection max life time can be achieved anytime and has no relationship with the pool itself, whereas keep alive timeout is actually has an intrinsic relationship with the pool.

So I believe that the way to go would be to implement `HttpConnection#shutdown()` for HTTP/1 (it is for HTTP/2) to have a proper eviction of the connection from the pool.

This way the connection max life time can be achieved simply by setting a timer and call shutdown when the timer fires.

srijan02420:
> The other reason I think it should use shutdown is that connection max life time can be achieved anytime and has no relationship with the pool itself, whereas keep alive timeout is actually has an intrinsic relationship with the pool.
> 
> So I believe that the way to go would be to implement `HttpConnection#shutdown()` for HTTP/1 (it is for HTTP/2) to have a proper eviction of the connection from the pool.
> 
> This way the connection max life time can be achieved simply by setting a timer and call shutdown when the timer fires.

Understood. I will try and do it.

srijan02420:
@vietj whenever I am doing something like
```
vertx.setTimer(client.getOptions().getKeepAliveTTL(), a -> {
      if (handler.connection().numActiveStreams() ==  0)
                shutdown(0L);
}
```
I am getting `io.vertx.core.VertxException: Connection was closed` ie I am trying to shutdown connection before `responseEnded = true;`. But I am not able to figure where should I call graceful shutdown.

vietj:
@srijan02420 the shutdown needs to be implemented for HTTP/1

vietj:
I will have a look at it

srijan02420:
> @srijan02420 the shutdown needs to be implemented for HTTP/1

Yes above was for H2 only. Shutdown some is not waiting for response to write for my test apparently.
Its not shutting down gracefully.

srijan02420:
@vietj I think connection max life time is also related to pool as keepAliveTmeout is because as pool understand that this connection sat idle for long and now should be evicted. Pool should understand the connection hit its life time limit and should be eveicted.
Other way to look at it is that option keep-alive can be used with pool only so keepAliveTTL should be used with pool only. Without keep-alive true keepAliveTTL doesn't make sense.

vietj:
@srijan02420 for now I am quite cautious because it seemed like an easy feature to implement but in fact it is not and I don't want to bring a new feature with bugs. For now I will push a branch that implements shutdown for HTTP/1.1 connection and you can use it for your case by using a timer to shutdown the connection after a while (it is trivial).

vietj:
@srijan02420 can you have a look at https://github.com/eclipse-vertx/vert.x/tree/http-1-shutdown

srijan02420:
> @srijan02420 can you have a look at https://github.com/eclipse-vertx/vert.x/tree/http-1-shutdown

@vietj I checked.

1/ Using this should I do something like
```
vertx.setTimer(client.getOptions().getKeepAliveTTL(), a -> {
      shutdown(0L);
}
```
inside `Http1xClientConnection` constructor?

2/ History of `http-1-shutdown` is different from history of `master`. Showing a lot of conflicts while merging.

vietj:
@srijan02420 for now we would not implement this feature until shutdown becomes more robust, so in your application you should manage yourself the timers and the shutdown calls

srijan02420:
Oh okay. Let me know if I can help in improving shutdown.

vietj:
you can test the feature extensively and report feedback that will be useful

srijan02420:
with shutdown you mean?
Will start using once its merged.

vietj:
it needs some work on documentation and also port to master branch

vietj:
The branch has been merged @srijan02420 https://github.com/eclipse-vertx/vert.x/pull/3384

srijan02420:
@vietj thanks. I will start using it and share results. Let me know whenever you think we can start re-work for this feature.

vietj:
for now I will close this PR to clean up things

vietj:
thanks for your time @srijan02420 and please give feedback on the shutdown once you have tested it

Issue #3230, opened 2019-12-16
Remove clustered boolean option


vietj:
Currently the options to build Vertx provide a `clustered` boolean with the following behavior:

- calling `vertx(options)` with `clustered` set to `false` builds a non clustered instance
- calling `vertx(options)` with `clustered` set to `true` throw an `IllegalArgumentException`
- calling `clusteredVertx(options)` with `clustered` set to `true` builds a clustered instance
- calling `clusteredVertx(options)` with `clustered` set to `false` does set `clustered` to `true` and then builds a clustered instance

Given this behaviour, the usefulness of this setting seems very low and instead it could be removed in Vert.x 4 with no actual impact other than not throwing an `IllegalStateException` when `clustered` is set to `true` and the `vertx(options)` method is called.

wowselim:
should `vertx(options)` in this case set `clustered` to `false`?

tsegismont:
@wowselim the idea is to remove it completely. So:

- if `vertx(options)` is called, a standalone instance is created
- if `clusteredVertx(options)` is called, a clustered instance is created

wowselim:
I have drafted #3348 with the change discussed here

tsegismont:
Closed via #3350

PR #3350, opened 2020-03-31
remove clustered flag in favor of createClusteredVertx

wowselim:
Signed-off-by: Selim Dincer <wowselim@live.de>

Motivation: This PR implements the change discussed in #3230. I have removed the clustered flag entirely and adjusted the tests accordingly.

There is one test where I don't know whether it should still exist or not:
`LauncherExtensibilityTest#testThatCustomLauncherCanCustomizeTheClusteredOption`


tsegismont:
> There is one test where I don't know whether it should still exist or not:
LauncherExtensibilityTest#testThatCustomLauncherCanCustomizeTheClusteredOption

To me it's fine if a custom launcher  cannot customize this option. If this level of flexibility is needed by the user, then creating a custom main class is the alternative.

PR #3354, opened 2020-04-01
Deprecate the start/stop Future-based methods in AbstractVerticle

jponge:
Otherwise IDEs do not report these methods as deprecated, even if the Verticle interface has them deprecated.

### DIFF
--- a/src/main/java/io/vertx/core/AbstractVerticle.java
+++ b/src/main/java/io/vertx/core/AbstractVerticle.java
@@ -100,7 +100,7 @@ public abstract class AbstractVerticle implements Verticle {
    * and call the startFuture some time later when start up is complete.
    * @param startFuture  a future which should be called when verticle start-up is complete.
    * @throws Exception
-   * @deprecated Use {@link #start(Promise)}
+   * @deprecated Override {@link AbstractVerticle#start(Promise)} instead.
    */
   @Override
   @Deprecated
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/AbstractVerticle.java
```diff
@@ -100,8 +100,10 @@ public JsonObject config() {
    * and call the startFuture some time later when start up is complete.
    * @param startFuture  a future which should be called when verticle start-up is complete.
    * @throws Exception
+   * @deprecated Use {@link Verticle#start(Promise)}
```
vietj, line 103: I think it should be `AbstractVerticle#start(Promise)` as link

vietj: I mean this is clearly a message for people extending this interface and nobody is calling it (beside the vertx deployment), so the message rather should be _instead override {@link AbstractVerticle#start(Promise)}_

jponge: Done

vietj: the wording should be changed too to _instead override {@link AbstractVerticle#start(Promise)}_

jponge: There you go

jponge:
Note: build fails thanks to SonaType OSS being down.

vietj:
thanks @jponge , Sonatype down, how come ?

PR #3357, opened 2020-04-02
Add Future#compose(Supplier<U> supplier) method

aguibert:
Motivation:

When chaining operations with a Future object, sometimes operations only depend on the completion of the previous operation, and don't care about the result value of the previous operation.

For example, with vertx-sql-client to perform a simple "insert then select" secenario, we have to do this:
```java
DB2Pool client = DB2Pool.pool(connectOptions, poolOptions);

// Note the "junk" variables
client.query("INSERT INTO Fortune (message) VALUES ('some data')").execute()
  .flatMap(junk -> client.query("SELECT * FROM Fortune WHERE id=23").execute())
  .onComplete(this::printResults)
  .flatMap(junk -> client.query("SELECT * FROM Fortune WHERE id=24").execute())
  .onComplete(this::printResults)
  .onComplete(junk -> {
    System.out.println("All done");
    client.close();
  });
```

After this new compose function is added, the above code reads a bit nicer:
```java
DB2Pool client = DB2Pool.pool(connectOptions, poolOptions);

client.query("INSERT INTO Fortune (message) VALUES ('some data')").execute()
  .flatMap(() -> client.query("SELECT * FROM Fortune WHERE id=23").execute())
  .onComplete(this::printResults)
  .flatMap(() -> client.query("SELECT * FROM Fortune WHERE id=24").execute())
  .onComplete(this::printResults)
  .onComplete(() -> {
    System.out.println("All done");
    client.close();
  });
```

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -20,6 +20,7 @@ import io.vertx.core.spi.FutureFactory;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * Represents the result of an action that may, or may not, have occurred yet.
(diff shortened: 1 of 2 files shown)


vietj:
I think we need support of Supplier in codegen and other languages before so it can be translated.

PR #3358, opened 2020-04-05
HAProxy protocol support

zenios:
Signed-off-by: zenios dimitris.zenios@gmail.com

**Notes:**
1. Hardcoded netty version (4.1.48.Final) for netty-codec-haproxy in order to avoid creating a pull request for vertx-dependencies

2. Tests were copied from #1271. That PR can be closed if this one gets merged

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,6 +111,11 @@
     </dependency>
     <dependency>
       <groupId>io.netty</groupId>
+      <artifactId>netty-codec-haproxy</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
       <artifactId>netty-transport-native-epoll</artifactId>
       <optional>true</optional>
     </dependency>
@@ -537,6 +542,25 @@
               <classpathDependencyExcludes>
                 <classpathDependencyExclude>com.fasterxml.jackson.core:jackson-core</classpathDependencyExclude>
                 <classpathDependencyExclude>com.fasterxml.jackson.core:jackson-databind</classpathDependencyExclude>
+              </classpathDependencyExcludes>
+            </configuration>
+          </execution>
+          <execution>
+            <id>no-haproxy-codec</id>
+            <goals>
+              <goal>integration-test</goal>
+              <goal>verify</goal>
+            </goals>
+            <configuration>
+              <includes>
+                <include>io/vertx/it/HAProxyTest.java</include>
+              </includes>
+              <systemProperties>
+                <vertx-test-alpn-jdk>false</vertx-test-alpn-jdk>
+                <vertx-test-alpn-openssl>false</vertx-test-alpn-openssl>
+              </systemProperties>
+              <classpathDependencyExcludes>
+                <classpathDependencyExclude>io.netty:netty-codec-haproxy</classpathDependencyExclude>
               </classpathDependencyExcludes>
             </configuration>
           </execution>
--- a/src/main/asciidoc/http.adoc
+++ b/src/main/asciidoc/http.adoc
@@ -1803,6 +1803,29 @@ use the full URL specified in the request URI:
 {@link examples.HTTPExamples#example60}
 ----
 
+=== Using HA PROXY protocol
+
+https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt[HA PROXY protocol] provides a convenient way to safely transport connection
+information such as a client's address across multiple layers of NAT or TCP
+proxies.
+
+HA PROXY protocol can be enabled by setting the option {@link io.vertx.core.http.HttpServerOptions#setUseProxyProtocol(boolean)}
+and adding the following dependency in your classpath:
+
+[source,xml]
+----
+<dependency>
+  <groupId>io.netty</groupId>
+  <artifactId>netty-codec-haproxy</artifactId>
+  <!--<version>Should align with netty version that Vert.x uses</version>-->
+</dependency>
+----
+
+[source,$lang]
+----
+{@link examples.HTTPExamples#example61}
+----
+
 === Automatic clean-up in verticles
 
 If you're creating http servers and clients from inside verticles, those servers and clients will be automatically closed
--- a/src/main/asciidoc/net.adoc
+++ b/src/main/asciidoc/net.adoc
@@ -770,3 +770,27 @@ Here's an example:
 
 The DNS resolution is always done on the proxy server, to achieve the functionality of a SOCKS4 client, it is necessary
 to resolve the DNS address locally.
+
+
+=== Using HA PROXY protocol
+
+https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt[HA PROXY protocol] provides a convenient way to safely transport connection
+information such as a client's address across multiple layers of NAT or TCP
+proxies.
+
+HA PROXY protocol can be enabled by setting the option {@link io.vertx.core.net.NetServerOptions#setUseProxyProtocol(boolean)}
+and adding the following dependency in your classpath:
+
+[source,xml]
+----
+<dependency>
+  <groupId>io.netty</groupId>
+  <artifactId>netty-codec-haproxy</artifactId>
+  <!--<version>Should align with netty version that Vert.x uses</version>-->
+</dependency>
+----
+
+[source,$lang]
+----
+{@link examples.NetExamples#example51}
+----
--- a/src/main/java/examples/HTTPExamples.java
+++ b/src/main/java/examples/HTTPExamples.java
@@ -918,6 +918,21 @@ public class HTTPExamples {
 
   }
 
+  public void example61(Vertx vertx) {
+
+    HttpServerOptions options = new HttpServerOptions()
+      .setUseProxyProtocol(true);
+
+    HttpServer server = vertx.createHttpServer(options);
+    server.requestHandler(request -> {
+      // Print the actual client address provided by the HA proxy protocol instead of the proxy address
+      System.out.println(request.remoteAddress());
+
+      // Print the address of the proxy
+      System.out.println(request.localAddress());
+    });
+  }
+
   public void serversharing(Vertx vertx) {
     vertx.createHttpServer().requestHandler(request -> {
       request.response().end("Hello from server " + this);
--- a/src/main/java/examples/NetExamples.java
+++ b/src/main/java/examples/NetExamples.java
@@ -598,6 +598,18 @@ public class NetExamples {
       .listen(8080);
   }
 
+  public void example51(Vertx vertx) {
+    NetServerOptions options = new NetServerOptions().setUseProxyProtocol(true);
+    NetServer server = vertx.createNetServer(options);
+    server.connectHandler(so -> {
+      // Print the actual client address provided by the HA proxy protocol instead of the proxy address
+      System.out.println(so.remoteAddress());
+
+      // Print the address of the proxy
+      System.out.println(so.localAddress());
+    });
+  }
+
   public void configureSNIServer(Vertx vertx) {
     JksOptions keyCertOptions = new JksOptions().setPath("keystore.jks").setPassword("wibble");
 
--- a/src/main/java/io/vertx/core/http/impl/HttpServerChannelInitializer.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerChannelInitializer.java
@@ -17,6 +17,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.EventLoop;
+import io.netty.handler.codec.haproxy.HAProxyMessageDecoder;
 import io.netty.handler.codec.http.HttpContentDecompressor;
 import io.netty.handler.logging.LoggingHandler;
 import io.netty.handler.ssl.SniHandler;
@@ -25,6 +26,8 @@ import io.netty.handler.stream.ChunkedWriteHandler;
 import io.netty.handler.timeout.IdleState;
 import io.netty.handler.timeout.IdleStateEvent;
 import io.netty.handler.timeout.IdleStateHandler;
+import io.netty.util.concurrent.Future;
+import io.netty.util.concurrent.GenericFutureListener;
 import io.vertx.core.Handler;
 import io.vertx.core.http.HttpServerOptions;
 import io.vertx.core.http.impl.cgbystrom.FlashPolicyHandler;
@@ -34,6 +37,7 @@ import io.vertx.core.net.impl.HandlerHolder;
 import io.vertx.core.net.impl.SSLHelper;
 import io.vertx.core.net.impl.SslHandshakeCompletionHandler;
 import io.vertx.core.net.impl.VertxHandler;
+import io.vertx.core.net.impl.HAProxyMessageCompletionHandler;
 import io.vertx.core.spi.metrics.HttpServerMetrics;
 
 import java.nio.charset.StandardCharsets;
@@ -77,9 +81,44 @@ import java.util.function.Function;
 
   @Override
   protected void initChannel(Channel ch) {
+    if (HAProxyMessageCompletionHandler.canUseProxyProtocol(options.isUseProxyProtocol())) {
+      IdleStateHandler idle;
+      io.netty.util.concurrent.Promise<Channel> p = ch.eventLoop().newPromise();
+      ch.pipeline().addLast(new HAProxyMessageDecoder());
+      if (options.getIdleTimeout() > 0) {
+        ch.pipeline().addLast("idle", idle = new IdleStateHandler(0, 0, options.getIdleTimeout(), options.getIdleTimeoutUnit()));
+      } else {
+        idle = null;
+      }
+      ch.pipeline().addLast(new HAProxyMessageCompletionHandler(p));
+      p.addListener((GenericFutureListener<Future<Channel>>) future -> {
+        if (future.isSuccess()) {
+          if (idle != null) {
+            ch.pipeline().remove(idle);
+          }
+          configurePipeline(future.getNow());
+        } else {
+          //No need to close the channel.HAProxyMessageDecoder already did
+          handleException(ch, future.cause());
+        }
+      });
+    } else {
+      configurePipeline(ch);
+    }
+  }
+
+
+  private void configurePipeline(Channel ch) {
     ChannelPipeline pipeline = ch.pipeline();
     if (sslHelper.isSSL()) {
-      ch.pipeline().addFirst("handshaker", new SslHandshakeCompletionHandler(ar -> {
+      if (options.isSni()) {
+        SniHandler sniHandler = new SniHandler(sslHelper.serverNameMapper(vertx));
+        pipeline.addLast(sniHandler);
+      } else {
+        SslHandler handler = new SslHandler(sslHelper.createEngine(vertx));
+        pipeline.addLast("ssl", handler);
+      }
+      pipeline.addLast("handshaker", new SslHandshakeCompletionHandler(ar -> {
         if (ar.succeeded()) {
           if (options.isUseAlpn()) {
             SslHandler sslHandler = pipeline.get(SslHandler.class);
@@ -96,13 +135,6 @@ import java.util.function.Function;
           handleException(ch, ar.cause());
         }
       }));
-      if (options.isSni()) {
-        SniHandler sniHandler = new SniHandler(sslHelper.serverNameMapper(vertx));
-        pipeline.addFirst(sniHandler);
-      } else {
-        SslHandler handler = new SslHandler(sslHelper.createEngine(vertx));
-        pipeline.addFirst("ssl", handler);
-      }
     } else {
       if (disableH2C) {
         handleHttp1(ch);
--- a/src/main/java/io/vertx/core/http/impl/HttpServerChannelInitializer.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerChannelInitializer.java
@@ -85,8 +85,8 @@ import java.util.function.Function;
       IdleStateHandler idle;
       io.netty.util.concurrent.Promise<Channel> p = ch.eventLoop().newPromise();
       ch.pipeline().addLast(new HAProxyMessageDecoder());
-      if (options.getIdleTimeout() > 0) {
-        ch.pipeline().addLast("idle", idle = new IdleStateHandler(0, 0, options.getIdleTimeout(), options.getIdleTimeoutUnit()));
+      if (options.getProxyProtocolTimeout() > 0) {
+        ch.pipeline().addLast("idle", idle = new IdleStateHandler(0, 0, options.getProxyProtocolTimeout(), options.getProxyProtocolTimeoutUnit()));
       } else {
         idle = null;
       }
--- a/src/main/java/io/vertx/core/net/NetServerOptions.java
+++ b/src/main/java/io/vertx/core/net/NetServerOptions.java
@@ -55,9 +55,19 @@ public class NetServerOptions extends TCPSSLOptions {
   public static final boolean DEFAULT_SNI = false;
 
   /**
-   * Default value of whether the server supports HA Proxy protocol = false
+   * Default value of whether the server supports HA PROXY protocol = false
    */
   public static final boolean DEFAULT_USE_PROXY_PROTOCOL = false;
+
+  /**
+   * The default value of HA PROXY protocol timeout = 10
+   */
+  public static final long DEFAULT_PROXY_PROTOCOL_TIMEOUT = 10L;
+
+  /**
+   * Default HA PROXY protocol time unit = SECONDS
+   */
+  public static final TimeUnit DEFAULT_PROXY_PROTOCOL_TIMEOUT_TIME_UNIT = TimeUnit.SECONDS;
 
   private int port;
   private String host;
@@ -65,6 +75,8 @@ public class NetServerOptions extends TCPSSLOptions {
   private ClientAuth clientAuth;
   private boolean sni;
   private boolean useProxyProtocol;
+  private long proxyProtocolTimeout;
+  private TimeUnit proxyProtocolTimeoutUnit;
 
   /**
    * Default constructor
@@ -87,6 +99,10 @@ public class NetServerOptions extends TCPSSLOptions {
     this.clientAuth = other.getClientAuth();
     this.sni = other.isSni();
     this.useProxyProtocol = other.isUseProxyProtocol();
+    this.proxyProtocolTimeout = other.proxyProtocolTimeout;
+    this.proxyProtocolTimeoutUnit = other.getProxyProtocolTimeoutUnit() != null ?
+      other.getProxyProtocolTimeoutUnit() :
+      DEFAULT_PROXY_PROTOCOL_TIMEOUT_TIME_UNIT;
   }
 
   /**
@@ -402,7 +418,6 @@ public class NetServerOptions extends TCPSSLOptions {
    */
   public boolean isUseProxyProtocol() { return useProxyProtocol; }
 
-
   /**
    * Set whether the server uses the HA Proxy protocol
    *
@@ -413,6 +428,45 @@ public class NetServerOptions extends TCPSSLOptions {
     return this;
   }
 
+  /**
+   * @return the Proxy protocol timeout, in time unit specified by {@link #getProxyProtocolTimeoutUnit()}.
+   */
+  public long getProxyProtocolTimeout() {
+    return proxyProtocolTimeout;
+  }
+
+  /**
+   * Set the Proxy protocol timeout, default time unit is seconds.
+   *
+   * @param proxyProtocolTimeout the Proxy protocol timeout to set
+   * @return a reference to this, so the API can be used fluently
+   */
+  public NetServerOptions setProxyProtocolTimeout(long proxyProtocolTimeout) {
+    if (proxyProtocolTimeout < 0) {
+      throw new IllegalArgumentException("proxyProtocolTimeout must be >= 0");
+    }
+    this.proxyProtocolTimeout = proxyProtocolTimeout;
+    return this;
+  }
+
+  /**
+   * Set the Proxy protocol timeout unit. If not specified, default is seconds.
+   *
+   * @param proxyProtocolTimeoutUnit specify time unit.
+   * @return a reference to this, so the API can be used fluently
+   */
+  public NetServerOptions setProxyProtocolTimeoutUnit(TimeUnit proxyProtocolTimeoutUnit) {
+    this.proxyProtocolTimeoutUnit = proxyProtocolTimeoutUnit;
+    return this;
+  }
+
+  /**
+   * @return the Proxy protocol timeout unit.
+   */
+  public TimeUnit getProxyProtocolTimeoutUnit() {
+    return proxyProtocolTimeoutUnit;
+  }
+
   private void init() {
     this.port = DEFAULT_PORT;
     this.host = DEFAULT_HOST;
@@ -420,5 +474,7 @@ public class NetServerOptions extends TCPSSLOptions {
     this.clientAuth = DEFAULT_CLIENT_AUTH;
     this.sni = DEFAULT_SNI;
     this.useProxyProtocol = DEFAULT_USE_PROXY_PROTOCOL;
+    this.proxyProtocolTimeout = DEFAULT_PROXY_PROTOCOL_TIMEOUT;
+    this.proxyProtocolTimeoutUnit = DEFAULT_PROXY_PROTOCOL_TIMEOUT_TIME_UNIT;
   }
 }
--- a/src/main/java/io/vertx/core/net/NetServerOptions.java
+++ b/src/main/java/io/vertx/core/net/NetServerOptions.java
@@ -54,11 +54,17 @@ public class NetServerOptions extends TCPSSLOptions {
    */
   public static final boolean DEFAULT_SNI = false;
 
+  /**
+   * Default value of whether the server supports HA Proxy protocol = false
+   */
+  public static final boolean DEFAULT_USE_PROXY_PROTOCOL = false;
+
   private int port;
   private String host;
   private int acceptBacklog;
   private ClientAuth clientAuth;
   private boolean sni;
+  private boolean useProxyProtocol;
 
   /**
    * Default constructor
@@ -80,6 +86,7 @@ public class NetServerOptions extends TCPSSLOptions {
     this.acceptBacklog = other.getAcceptBacklog();
     this.clientAuth = other.getClientAuth();
     this.sni = other.isSni();
+    this.useProxyProtocol = other.isUseProxyProtocol();
   }
 
   /**
@@ -390,11 +397,28 @@ public class NetServerOptions extends TCPSSLOptions {
     return this;
   }
 
+  /**
+   * @return whether the server uses the HA Proxy protocol
+   */
+  public boolean isUseProxyProtocol() { return useProxyProtocol; }
+
+
+  /**
+   * Set whether the server uses the HA Proxy protocol
+   *
+   * @return a reference to this, so the API can be used fluently
+   */
+  public NetServerOptions setUseProxyProtocol(boolean useProxyProtocol) {
+    this.useProxyProtocol = useProxyProtocol;
+    return this;
+  }
+
   private void init() {
     this.port = DEFAULT_PORT;
     this.host = DEFAULT_HOST;
     this.acceptBacklog = DEFAULT_ACCEPT_BACKLOG;
     this.clientAuth = DEFAULT_CLIENT_AUTH;
     this.sni = DEFAULT_SNI;
+    this.useProxyProtocol = DEFAULT_USE_PROXY_PROTOCOL;
   }
 }
--- a/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
+++ b/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
@@ -15,6 +15,7 @@ import io.netty.buffer.Unpooled;
 import io.netty.channel.*;
 import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.stream.ChunkedFile;
+import io.netty.util.AttributeKey;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.FutureListener;
 import io.vertx.core.*;
@@ -54,6 +55,8 @@ public abstract class ConnectionBase {
    * trace to not be misleading.
    */
   public static final VertxException CLOSED_EXCEPTION = new VertxException("Connection was closed", true);
+  public static final AttributeKey<java.net.SocketAddress> REMOTE_ADDRESS_OVERRIDE = AttributeKey.valueOf("RemoteAddressOverride");
+  public static final AttributeKey<java.net.SocketAddress> LOCAL_ADDRESS_OVERRIDE = AttributeKey.valueOf("LocalAddressOverride");
   private static final Logger log = LoggerFactory.getLogger(ConnectionBase.class);
   private static final int MAX_REGION_SIZE = 1024 * 1024;
 
@@ -479,7 +482,10 @@ public abstract class ConnectionBase {
   public SocketAddress remoteAddress() {
     SocketAddress address = remoteAddress;
     if (address == null) {
-      java.net.SocketAddress addr = chctx.channel().remoteAddress();
+      java.net.SocketAddress addr = chctx.channel().hasAttr(REMOTE_ADDRESS_OVERRIDE) ?
+        chctx.channel().attr(REMOTE_ADDRESS_OVERRIDE).getAndSet(null) :
+        chctx.channel().remoteAddress();
+
       if (addr != null) {
         address = vertx.transport().convert(addr);
         remoteAddress = address;
@@ -491,7 +497,10 @@ public abstract class ConnectionBase {
   public SocketAddress localAddress() {
     SocketAddress address = localAddress;
     if (address == null) {
-      java.net.SocketAddress addr = chctx.channel().localAddress();
+      java.net.SocketAddress addr = chctx.channel().hasAttr(LOCAL_ADDRESS_OVERRIDE) ?
+        chctx.channel().attr(LOCAL_ADDRESS_OVERRIDE).getAndSet(null) :
+        chctx.channel().localAddress();
+
       if (addr != null) {
         address = vertx.transport().convert(addr);
         localAddress = address;
--- a/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
+++ b/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
@@ -55,8 +55,8 @@ public abstract class ConnectionBase {
    * trace to not be misleading.
    */
   public static final VertxException CLOSED_EXCEPTION = new VertxException("Connection was closed", true);
-  public static final AttributeKey<java.net.SocketAddress> REMOTE_ADDRESS_OVERRIDE = AttributeKey.valueOf("RemoteAddressOverride");
-  public static final AttributeKey<java.net.SocketAddress> LOCAL_ADDRESS_OVERRIDE = AttributeKey.valueOf("LocalAddressOverride");
+  public static final AttributeKey<SocketAddress> REMOTE_ADDRESS_OVERRIDE = AttributeKey.valueOf("RemoteAddressOverride");
+  public static final AttributeKey<SocketAddress> LOCAL_ADDRESS_OVERRIDE = AttributeKey.valueOf("LocalAddressOverride");
   private static final Logger log = LoggerFactory.getLogger(ConnectionBase.class);
   private static final int MAX_REGION_SIZE = 1024 * 1024;
 
@@ -482,14 +482,17 @@ public abstract class ConnectionBase {
   public SocketAddress remoteAddress() {
     SocketAddress address = remoteAddress;
     if (address == null) {
-      java.net.SocketAddress addr = chctx.channel().hasAttr(REMOTE_ADDRESS_OVERRIDE) ?
-        chctx.channel().attr(REMOTE_ADDRESS_OVERRIDE).getAndSet(null) :
-        chctx.channel().remoteAddress();
-
-      if (addr != null) {
-        address = vertx.transport().convert(addr);
-        remoteAddress = address;
+      if (chctx.channel().hasAttr(REMOTE_ADDRESS_OVERRIDE)) {
+        address = chctx.channel().attr(REMOTE_ADDRESS_OVERRIDE).getAndSet(null);
+      } else {
+        java.net.SocketAddress addr = chctx.channel().remoteAddress();
+        if (addr != null) {
+          address = vertx.transport().convert(addr);
+        }
       }
+
+      if (address != null)
+        remoteAddress = address;
     }
     return address;
   }
@@ -497,14 +500,17 @@ public abstract class ConnectionBase {
   public SocketAddress localAddress() {
     SocketAddress address = localAddress;
     if (address == null) {
-      java.net.SocketAddress addr = chctx.channel().hasAttr(LOCAL_ADDRESS_OVERRIDE) ?
-        chctx.channel().attr(LOCAL_ADDRESS_OVERRIDE).getAndSet(null) :
-        chctx.channel().localAddress();
-
-      if (addr != null) {
-        address = vertx.transport().convert(addr);
-        localAddress = address;
+      if (chctx.channel().hasAttr(LOCAL_ADDRESS_OVERRIDE)) {
+        address = chctx.channel().attr(LOCAL_ADDRESS_OVERRIDE).getAndSet(null);
+      } else {
+        java.net.SocketAddress addr = chctx.channel().localAddress();
+        if (addr != null) {
+          address = vertx.transport().convert(addr);
+        }
       }
+
+      if (address != null)
+        localAddress = address;
     }
     return address;
   }
--- a/src/main/java/io/vertx/core/net/impl/HAProxyMessageCompletionHandler.java
+++ b/src/main/java/io/vertx/core/net/impl/HAProxyMessageCompletionHandler.java
@@ -4,18 +4,21 @@ import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.MessageToMessageDecoder;
 import io.netty.handler.codec.haproxy.HAProxyMessage;
+import io.netty.handler.codec.haproxy.HAProxyProxiedProtocol;
 import io.netty.handler.timeout.IdleState;
 import io.netty.handler.timeout.IdleStateEvent;
-import io.netty.util.NetUtil;
 import io.netty.util.concurrent.Promise;
 import io.vertx.core.impl.logging.Logger;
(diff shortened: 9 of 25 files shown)


pom.xml
```diff
@@ -101,6 +101,11 @@
       <groupId>io.netty</groupId>
       <artifactId>netty-codec-http2</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-codec-haproxy</artifactId>
+      <version>4.1.48.Final</version>
```
vietj, line 107: we should have this to optional=true

src/main/java/io/vertx/core/http/impl/HttpServerChannelInitializer.java
```diff
@@ -217,6 +218,7 @@ private void configureHttp1OrH2C(ChannelPipeline pipeline, HandlerHolder<? exten
     if (HttpServerImpl.USE_FLASH_POLICY_HANDLER) {
       pipeline.addLast("flashpolicy", new FlashPolicyHandler());
     }
+    pipeline.addLast("optionalHAProxyMessageDecoder", new OptionalHAProxyMessageDecoder());
```
vietj, line 221: we should only add this to the pipeline when the option is configured. pipeline handlers have effect on performance and we should avoid adding them when it is not necessary

vietj: so there should be an option for enabling / disabling it in server

vietj: also this will allow to run without the ha proxy class on the classpath by default

zenios: Have in mind that its autoremoved once the first 12 bytes are received (12 bytes are the number of bytes required to find out if its an haproxy protocol)

Instead of an option we could check if HaProxy is in classpatch and enable it

Tell me which one you prefer and will implement

Option or classpath?

vietj: good point for the removal, for now let's do classpath as I would like to avoid having the jar OOTB.

pom.xml
```diff
@@ -105,6 +105,7 @@
       <groupId>io.netty</groupId>
       <artifactId>netty-codec-haproxy</artifactId>
       <version>4.1.48.Final</version>
+      <optional>true</optional>
```
vietj, line 108: we need an integration test without the `netty-codec-haproxy` jar, have a look at failsafe tests where a jar can be removed from the classpath to run the test in pom.xml

src/main/java/io/vertx/core/http/impl/haproxy/OptionalHAProxyMessageDecoder.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+
+public class OptionalHAProxyMessageDecoder extends ByteToMessageDecoder {
+  private static final boolean hasProxyProtocolSupport;
+
+  static {
+    boolean proxyProtocolCheck = true;
+    try {
+      Class.forName("io.netty.handler.codec.haproxy.HAProxyMessageDecoder");
+    } catch (ClassNotFoundException ex) {
```
vietj, line 30: can you catch `Throwable` to avoid any class initialisation error

zenios:
@vietj i have a question in NetSupport

If we have a netserver that receives shorts messages under 12 bytes

then the message will not pass through since OptionalHAProxyMessageDecoder will wait for at least 12 bytes to decide if its an HAProxy message or not.

How shall we handle this cases?

vietj:
we should make the ha proxy protocol configured in all cases in the `NetServerOptions`, vertx should provide control over this anyway

zenios:
So an option in netserver.

Do you want me to add also one for http server?

vietj:
HttpServerOptions extends NetServerOptions so it will naturally be there

zenios:
@vietj  I have implemented all that was requested

The only caveat is with NetServer
You cannot get the ip on connectionHandler. A message handler should be added first and get the ip there

The reason is that connectionhandler is called directly when the socker opened and then the pipeline is configured.

If we want to have proxy ip available at connection handler then something like  SslHandshakeCompletionHandler should be implemented which will call connection once the header is parsed.

vietj:
you can get the IP from the NetSocket.

Julien

vietj:
I don't think it is complete yet

- I would like to understand what it takes to support `NetServer`
- still need to have an option to enable proxy protocol support (should not be active by default)
- there is no documentation and example

vietj:
Actually there is more than I thought :-)

So there is option for enabling/disabling and support in `NetServer` and I get that the server will get the relevant information only when data has been actually sent by the proxy.

zenios:
NetServer support is there
The same for the option

Maybe i have confused you because i did a squash and force push for the
latest changes

As for the documentation i will work on it yomorrow but to tell you the
truth not much is needed

options.proxySupport(true)
Add haproxy codec to pom
Goot to go :)

src/main/java/io/vertx/core/net/NetServerOptions.java
```diff
@@ -54,11 +54,17 @@
...
+   * Default value of whether the server supports HA Proxy protocol = false
+   */
+  public static final boolean DEFAULT_PROXY_SUPPORT = false;
+
   private int port;
   private String host;
   private int acceptBacklog;
   private ClientAuth clientAuth;
   private boolean sni;
+  private boolean proxySupport;
```
vietj, line 67: rename it `proxyProtocolSupport` instead, proxy is confusing given that the client supports a `proxyOptions`

zenios:
Exactly if we want to fix the last part which is a problem only for
netserver then we have to wait for at least 12 bytes and then emit
connection

Almost the same as SslHandshakeCompletionHandler

vietj:
how does proxy protocol works with SSL ?

I believe we need tests to check that too.

zenios:
Already did with haproxy.

It first sends the proxy protocol and then starts the ssl negotiation
process

This is the reason that proxy decoder has to be added first to the pipeline
(before ssl)

zenios:
As for the tests i am not sure how could i emulate using a socket client
sending proxy protocol following with ssl negotiation

vietj:
I see now more clearly, indeed we need to make sure that ha proxy is done before SSL as it is level 4 proxy.

As for testing I believe we can achieve with proxy server (using NetServer/NetClient) that will simply send data when connecting to the server before forwarding the traffic.

vietj:
I believe actually that if we build such proxy server for testing it can also be used for non TLS tests as well.

vietj:
Note that vertx tests have a package `io.vertx.test.proxy` with already similar proxies for testing client with Socks4 and Socks5

vietj:
here is what we need:

- create a proxy NetServer that listen on port XXX
- this server is configured with a buffer of arbitrary data
- when this proxy process a connection it pauses the NetSocket to stop any incoming traffic
- and then connect to the actual server that will receive the incoming proxy protocol
- when the proxy is connected it sends immediately the arbitrary data
- and then create a pipe between inbound and outbound connection and resume the inbound NetSocket

To test SSL, the client will setup SSL traffic as well as the actual server.

This proxy server can be reused to test anything related to proxy protocol.

zenios:
Ok got it.

Will do some refactorings in order to attach proxyhandler before all and afterwards will proceed with these proxy implementation

Thanks

zenios:
@vietj I did all the changes required in order to full support this feature. I have also fixed the case of the IP not available during the connection handler

The only thing that is missing is a test proxy implementation

Could you please do another round of review and let me know if everything looks ok?

Thanks

zenios:
> we should use a single Netty handlers and not two

I assume you are talking about this

```
ch.pipeline().addLast(new HAProxyMessageDecoder());
ch.pipeline().addLast(new HAProxyMessageCompletionHandler(ar -> {
 if (ar.succeeded()) {
  configurePipeline(ch);
 } else {
  handleException(ch, ar.cause());
 }
}));
```

The first one is provided by netty that does decoding of the header and pushes HAProxyMessage and the second one is ours which will capture the message setup the ip override and proceed configuring the remaining pipeline.

I dont see how they can be merged.

zenios:
Have in mind that both handlers will remove their selfs when completed

vietj:
so I think we need to ahve the tests with proxy and then we can do another review.

I believe one thing we need to test several extra cases and handle them:

- the channel is closed before the entire proxy line is closed, e.g the client connects and send `PROXY TCP4 192.168.0.1` and the close the TCP connection, we need to check that the channel closure is correctly handled
- the client send data that will make `HAProxyMessageDecoder` fail, there are several ways it can fail and according to the server (HttpServer or NetServer) we need to handle the various cases

zenios:
Proxy has been implemented and more tests added

vietj:
I will do a new review pass soon. For the NetServer not having the IP at connect time, I will handle it myself if we decide to implement a feature that delay the connect event after the proxy decoder did the job.

zenios:
> I will do a new review pass soon. For the NetServer not having the IP at connect time, I will handle it myself if we decide to implement a feature that delay the connect event after the proxy decoder did the job.

Not needed.It has been fixed by doing what you have mentioned.

src/test/java/io/vertx/core/http/Http1xTest.java
```diff
@@ -4865,4 +4866,80 @@ public void testUnsolicitedHttpResponse() throws Exception {
...
     await();
   }
+
+  @Test
+  public void testProxyProtocolIllegalHeader() throws Exception {
+    HAProxy proxy = new HAProxy(DEFAULT_HTTP_HOST, DEFAULT_HTTP_PORT, Buffer.buffer());
+    proxy.start(vertx);
+
+    server.close();
+    server = vertx.createHttpServer(createBaseServerOptions().setProxyProtocolSupport(true));
```
vietj, line 4876: we should set an `exceptionHandler` on the HTTP server and check it is called with the failure.

src/main/java/io/vertx/core/net/impl/ConnectionBase.java
```diff
@@ -468,18 +471,19 @@ public ChannelPromise channelFuture() {
   }
 
   public String remoteName() {
-    java.net.SocketAddress addr = chctx.channel().remoteAddress();
-    if (addr instanceof InetSocketAddress) {
-      // Use hostString that does not trigger a DNS resolution
-      return ((InetSocketAddress)addr).getHostString();
```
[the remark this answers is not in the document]
zenios: In case of ha proxy we need to use the overriden address and not chctx.channel().remoteAddress

I was going to make another PR later to remove this method or at least deprecate it since i think it does exactly the same work as remoteAddress().host()

Please corrent me if i am wrong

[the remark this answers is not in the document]
vietj: I mean the precise usage of `getHostString()` that has a purpose here to be not blocking

src/main/java/io/vertx/core/net/impl/HAProxyMessageCompletionHandler.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public static boolean hasProxyProtocolSupport(boolean requested) {
+    if (requested && !hasProxyProtocolSupport)
+      log.warn("Proxy protocol support could not be enabled.Make sure that netty-codec-haproxy is included in your classpath");
+    return hasProxyProtocolSupport && requested;
+  }
+
+
+  private final Handler<AsyncResult<Channel>> handler;
```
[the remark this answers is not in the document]
zenios: SslHandshakeCompletionHandler which is almost the same is using Handler<AsyncResult<Channel>> handler so i copied it

Will change it now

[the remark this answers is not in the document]
vietj: we might eventually change the `SslHandshakeCompletionHandler` as well later, thanks for noticing

src/main/java/io/vertx/core/net/impl/ConnectionBase.java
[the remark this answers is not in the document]
zenios: remoteAddress().host() is exactly the same think.

```
public SocketAddressImpl(InetSocketAddress address) {
    Arguments.requireInRange(address.getPort(), 0, 65535, "port p must be in range 0 <= p <= 65535");
    this.path = null;
    this.port= address.getPort();
    this.host = address.getHostString();
    if (address.isUnresolved()) {
      this.hostName = address.getHostName();
      this.ipAddress = null;
    } else {
      String host = address.getHostString();
      if (NetUtil.isValidIpV4Address(host) || NetUtil.isValidIpV6Address(host)) {
        host = null;
      }
      this.hostName = host;
      this.ipAddress = address.getAddress();
    }
  }
```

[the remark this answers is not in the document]
zenios: Now i get it. I assume that is the part that will block when using normal remoteAddress().host(). Right?
```
if (address.isUnresolved()) {
      this.hostName = address.getHostName();
      this.ipAddress = null;
} 
```

zenios:
I have made the changes requested and i have one question and one issue

**Issue**
Sometimes when haproxy and ssl are used i get _handshake timed out after 10000ms_ for ssl after haproxy has completed decoding.

Have in mind that all handlers including ssl are added only when haproxy decoder has completed.

Is there any chance that until we add the new handlers the data (ssl handshake) has already passed through and missed by ssl handler?

**Question**
How shall we handle idle connections and haproxy?

Someone opens a connection and never transmits anythings. Do we need to setup an idlehandler also?

**Thanks for your timer and your help**

vietj:
yes it is not the same API

vietj:
@zenios what you said sounds indeed like a race.

it should synchronously add the new handlers to the pipeline and then remove the other handlers from the pipeline, I think

to reproduce this issue, you can annotate the corresponding test method with @Repeat(times = 1000) for instance to repeat the test a lot of time and reproduce the issue if that's a race

zenios:
@vietj 

I have fixed the raise condition and added IdleStateHandler. I have also fixed everything that you have reviewed.

Please check it out and let me know if anything else is needed

Thanks

vietj:
why is IdleStateHandler required ?

zenios:
> why is IdleStateHandler required ?

In case a connection is established and no data is send or received.

Dont forget that if HAProxy is enabled the whole pipeline configuration (Including idleStateHandled)  is delayed until ha proxy decoder has finished

vietj:
yes but it should be able to work without it, i.e if the user sends a non proxy line then the pipeline should be reinstiated and the data should be passed through the pipeline.

This is what the `Http1xOrH2CHandler` does for instance

zenios:
Not exactly.if haproxy is enabled and the user sends a nonproxy line then
an exception is thrown and the socket is closed.

At the initial pr we had an optionalhaproxydecoder that worked only for
http requests. I removed it since now we have an option for the user to
enable it or not

Also idlestatehandler is mostly needed for the case that someone connects
and sends nothing

vietj:
> On 7 Apr 2020, at 19:31, zenios <notifications@github.com> wrote:
> 
> 
> Not exactly.if haproxy is enabled and the user sends a nonproxy line then
> an exception is thrown and the socket is closed.
> 
> At the initial pr we had an optionalhaproxydecoder that worked only for
> http requests. I removed it since now we have an option for the user to
> enable it or not

I think this is what we need pretty much like Http1xOrH2CHandler

> 
> Also idlestatehandler is mostly needed for the case that someone connects
> and sends nothing

that's not our problem to deal with, IdleStateHandler should be set when the option TCP idle timeout is set in NetServerOptions only.

> 
>

zenios:
Ok

Just to be clear

1. Bring back OptionalHAProxyDecoder on HTTP mode.
- If an invalid header is send an exception is thrown
- If no header is sent the pipeline should reinstiate to its default 
2. Leave TCP mode as is
- If an invalid header is send an exception is thrown
- If no header is send an exception is thrown

Right?

vietj:
it is not clear about the behaviour your describe because `OptionalHAProxyDecoder` makes thing it decodes if there is no proxy line sent and you say _If an invalid header is send an exception is thrown_

zenios:
> it is not clear about the behaviour your describe because `OptionalHAProxyDecoder` makes thing it decodes if there is no proxy line sent and you say _If an invalid header is send an exception is thrown_

My bad with OptionalHAProxyDecoder If an invalid header is send the pipeline should reinstiate to its default

zenios:
Let me restore that functionality back tomorrow morning and will continue from there

Thanks for your time

vietj:
agreed let's have a fresh start tomorrow :-)

zenios:
@vietj  I have done the changes.

Check them out when you have time

vietj:
I am wondering if we really need `HAProxyMessageDetector`. I believe that when the server is configured with proxy protocol then the server whether it is TCP or HTTP should always expect to get a proxy protocol line.

That would simplify the code and make it less magic.

zenios:
> I am wondering if we really need `HAProxyMessageDetector`. I believe that when the server is configured with proxy protocol then the server whether it is TCP or HTTP should always expect to get a proxy protocol line.
> 
> That would simplify the code and make it less magic.

This is what we had yesterday. If the server was configured with proxy protocol but a protocol line was not received / or not valid then it would close the connection and throw an exception

Changed it today in order to be optional. ie if no protocol line is send then it fallbacks to handling the request normally without throwing an exception.

If you want i can remove it again and revert to yesterdays functionality

vietj:
@zenios let me ask a friend that is using HA proxy the advice, I'm sorry for the confusion here

zenios:
I am more in favor of non optional as well. Let me do that change and will talk again.

Once i finish i will write a little to explain the differences of HA implementation between tcp and http

vietj:
ok @zenios

zenios:
@vietj I have applied the changes.
Below is a short description on the difference between net / http implementation


Both net and http implementations fail when proxy protocol is enabled and proxy protocol line is not received / or illegal

The major difference is on the construction of the pipeline

For net we first attach ha proxy decoders and only when it succeeds parsing we attach the remaining handlers (SSL / handshake etc). This is done in order to have the ha proxy ip available at connectionHandler

For this reason we also need to attach an IdleStateHandler (Along with ha proxy handler)  if idleTimeout > 0.

For http ha proxy decoders are attached from the beginning along with the remaining handlers  (SSL / handshake etc) 

In order to decode an ha proxy protocol you needs two handlers
1. HAProxyMessageDecoder
Which is provided by netty, decodes the header and pushes an HAProxyMessage to the pipeline. When an invalid header is received an exception is thrown and the connection is closed
2. HAProxyMessageReader
Which reads the message pushed by HAProxyMessageDecoder and sets the ip override attributes.


Hope now its more clear
Thanks

src/test/java/io/vertx/core/http/HttpTLSTest.java
```diff
@@ -921,13 +922,15 @@ public TrustOptions copy() {
     boolean serverOpenSSL;
     boolean serverUsesAlpn;
     boolean serverSSL = true;
+    boolean proxyProtocolSupport = false;
```
vietj, line 925: actually we should rename to `useProxyProtocol` like we do have `useAlpn` and `useProxyAuth` already

src/main/java/io/vertx/core/net/NetServerOptions.java
```diff
@@ -54,11 +54,17 @@
...
+   * Default value of whether the server supports HA Proxy protocol = false
+   */
+  public static final boolean DEFAULT_PROXY_PROTOCOL_SUPPORT = false;
+
   private int port;
   private String host;
   private int acceptBacklog;
   private ClientAuth clientAuth;
   private boolean sni;
+  private boolean proxyProtocolSupport;
```
vietj, line 67: we should rename to `useProxyProtocol` we already have `useAlpn` in options

src/main/java/io/vertx/core/net/NetServerOptions.java
```diff
@@ -400,16 +400,16 @@ public NetServerOptions setSni(boolean sni) {
...
 
 
   /**
    * Set whether the server supports HA Proxy protocol
    *
    * @return a reference to this, so the API can be used fluently
    */
-  public NetServerOptions setProxyProtocolSupport(boolean proxyProtocolSupport) {
-    this.proxyProtocolSupport = proxyProtocolSupport;
+  public NetServerOptions setUseProxyProtocol(boolean useProxyProtocol) {
```
vietj, line 411: rather than support we should say "Set whether the server uses the HA Proxy protocol" , the "supports" words indicate it can be optional and it is actually not

src/main/java/io/vertx/core/net/NetServerOptions.java
```diff
@@ -400,16 +400,16 @@ public NetServerOptions setSni(boolean sni) {
   /**
    * @return whether the server supports HA Proxy protocol
```
vietj, line 401: _whether the server uses the HA Proxy protocol_

src/main/java/io/vertx/core/net/impl/haproxy/HAProxyMessageReader.java
```diff
@@ -14,22 +14,22 @@
...
     }
-    hasProxyProtocolSupport = proxyProtocolCheck;
+    proxyProtocolCodecFound = proxyProtocolCodecCheck;
   }
 
-  public static boolean hasProxyProtocolSupport(boolean requested) {
-    if (requested && !hasProxyProtocolSupport)
+  public static boolean canUseProxyProtocol(boolean requested) {
+    if (requested && !proxyProtocolCodecFound)
       log.warn("Proxy protocol support could not be enabled.Make sure that netty-codec-haproxy is included in your classpath");
```
vietj, line 31: Extra space between the dot and upper case M in _.Make_

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
     }));
     await();
   }
+
+
+  @Test
+  public void testProxyProtocolIllegalHeader() throws Exception {
```
vietj, line 5920: here we should use `waitFor(2)` and have two condition to end the test

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
...
+    HAProxy proxy = new HAProxy(DEFAULT_HTTP_HOST, DEFAULT_HTTP_PORT, Buffer.buffer());
+    proxy.start(vertx);
+
+    server.close();
+    server = vertx.createHttpServer(createBaseServerOptions()
+      .setUseProxyProtocol(true))
+      .exceptionHandler(exception -> {
+        if(exception instanceof io.netty.handler.codec.haproxy.HAProxyProtocolException)
+          testComplete();
```
vietj, line 5929: use `complete()` here instead of `testComplete()`

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
...
+      });
+    server.requestHandler(req -> {
+      fail("Should not be called");
+    });
+
+    startServer();
+    client.request(HttpMethod.GET, proxy.getPort(), proxy.getHost(), DEFAULT_TEST_URI)
+      .onComplete(event -> {
+        assertTrue(event.failed());
```
vietj, line 5938: add a `complete()` statement after `assertTrue(event.failed())` to end the test and be sure that the assertion will be reported correctly to the test

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
...
+      })
+      .end();
+    await();
+    proxy.stop();
```
vietj, line 5942: call stop() that in a finally block

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
...
+    });
+    startServer();
+    client.request(HttpMethod.GET, proxy.getPort(), proxy.getHost(), DEFAULT_TEST_URI)
+      .onComplete(event -> {
+        assertTrue(event.succeeded());
+        testComplete();
+      })
+      .end();
+    await();
+    proxy.stop();
```
vietj, line 5969: call proxy.stop() in a finally block

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
...
+  }
+
+
+  @Test
+  public void testHAProxyProtocolVersion1() throws Exception {
+    Buffer version1ProtocolHeader = HAProxy.createVersion1ProtocolHeader("192.168.0.1", 56324, "192.168.0.11", 443);
```
vietj, line 5948: use `waitFor(2)`

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
...
+    server.requestHandler(req -> {
+      assertEquals(req.remoteAddress().hostAddress(), "192.168.0.1");
+      assertEquals(req.localAddress().hostAddress(), "192.168.0.11");
+      assertEquals(req.remoteAddress().port(), 56324);
+      assertEquals(req.localAddress().port(), 443);
+      req.response().end();
+    });
+    startServer();
+    client.request(HttpMethod.GET, proxy.getPort(), proxy.getHost(), DEFAULT_TEST_URI)
+      .onComplete(event -> {
```
vietj, line 5989: use `onSuccess(response -> { ... })` and avoid if/else blockg

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
...
```
vietj, line 5963: use `onSuccess(event -> }` block

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5913,4 +5914,88 @@ private void testClientRequestWithLargeBodyInSmallChunks(boolean chunked) throws
...
+        if(event.failed())
+          event.cause().printStackTrace();
+        else {
+          assertTrue(event.succeeded());
+          testComplete();
+        }
+      })
+      .end();
+    await();
+    proxy.stop();
```
vietj, line 5999: stop in a finally

src/test/java/io/vertx/core/net/NetTest.java
```diff
@@ -3682,4 +3689,106 @@ public void testNetSocketHandlerFailureReportedToContextExceptionHandler() throw
...
+      assertEquals(so.localAddress().port(), 443);
+      testComplete();
+    });
+    startServer();
+    client.connect(proxy.getPort(), proxy.getHost(), onSuccess(so -> {
+      so.write("ping");
+      so.close();
+    }));
+    await();
+    proxy.stop();
```
vietj, line 3713: finally :)

src/test/java/io/vertx/core/net/NetTest.java
```diff
@@ -3682,4 +3689,106 @@ public void testNetSocketHandlerFailureReportedToContextExceptionHandler() throw
...
```
vietj, line 3737: finally clause

src/test/java/io/vertx/core/net/NetTest.java
```diff
@@ -3682,4 +3689,106 @@ public void testNetSocketHandlerFailureReportedToContextExceptionHandler() throw
...
+    server.connectHandler(event -> {
+      assertEquals(event.remoteAddress().hostAddress(), "192.168.0.1");
+      assertEquals(event.localAddress().hostAddress(), "192.168.0.11");
+      assertEquals(event.remoteAddress().port(), 56324);
+      assertEquals(event.localAddress().port(), 443);
+    });
+
+    startServer();
+    client.connect(proxy.getPort(), proxy.getHost(), ar2 -> {
+      assertTrue(ar2.succeeded());
```
vietj, line 3769: use `onSuccess`

src/test/java/io/vertx/it/HAProxyTest.java
```diff
@@ -0,0 +1,47 @@
...
+  @Test
+  public void testHttpWithoutHAProxySupport() {
+    Vertx vertx = Vertx.vertx();
+    try {
+      vertx.createHttpServer().requestHandler(req -> {
+        req.response().end("hello");
+      }).listen(DEFAULT_HTTP_PORT, DEFAULT_HTTP_HOST, onSuccess(s -> {
+        HttpClient client = vertx.createHttpClient();
+        client.get(DEFAULT_HTTP_PORT, DEFAULT_HTTP_HOST, "/", onSuccess(resp -> {
+          resp.exceptionHandler(this::fail);
```
vietj, line 35: instead of setting an exception handler + body handler , use 

```
resp.body().onComplete(onSuccess(body -> ...));
```

src/test/java/io/vertx/core/http/HttpTest.java
zenios: @vietj 

I have fixed all the test cases except this one.The problem with this one is that it behave differently for http1 and http2

For http1 

We get two HAProxyDecoder Exception
```
io.netty.handler.codec.haproxy.HAProxyProtocolException: io.netty.handler.codec.haproxy.HAProxyProtocolException: unknown identifier: GET
io.netty.handler.codec.haproxy.HAProxyProtocolException: io.netty.handler.codec.haproxy.HAProxyProtocolException: unknown identifier: host:
```

And for http2 

We get one HAProxyDecoder Exception

```
io.netty.handler.codec.haproxy.HAProxyProtocolException: header length (over 269) exceeds the allowed maximum (108)
```

And one from SSLHandler
```
java.nio.channels.ClosedChannelException
```

So if using waitFor(2) assuming 1 is from complete of client  the other should countdown when exception instanceof HAProxyProtocolException

The we will have 3 countdowns for http1 and two countdowns for http2

How do you suggest we should handle this problem?

Split the test between http1 / http2 ?

vietj:
@zenios what about documentation in asciidoc format ?

zenios:
> @zenios what about documentation in asciidoc format ?

@vietj  Hi

I will complete it but lets finish the issue with the remaining test.Please see comment on your rweview above.

Thanks

vietj:
ok, we were in deadlock then :-)

src/main/java/io/vertx/core/net/impl/ConnectionBase.java
```diff
@@ -479,7 +482,10 @@ public String remoteName() {
   public SocketAddress remoteAddress() {
     SocketAddress address = remoteAddress;
     if (address == null) {
-      java.net.SocketAddress addr = chctx.channel().remoteAddress();
+      java.net.SocketAddress addr = chctx.channel().hasAttr(REMOTE_ADDRESS_OVERRIDE) ?
+        chctx.channel().attr(REMOTE_ADDRESS_OVERRIDE).getAndSet(null) :
```
vietj, line 486: why `getAndSet` here instead of `get()` ? clearing the value will not remove the attribue

src/test/java/io/vertx/core/http/HttpTest.java
vietj: that's fine for now

src/main/java/io/vertx/core/net/impl/ConnectionBase.java
zenios: Since we cache the address no need to keep the attribute around.

getAndRemove is deprecated with a note to use getAndSet with null value. 

https://netty.io/4.0/api/io/netty/util/Attribute.html

zenios:
Another issue is that for http1 it complains that testComplete was called
twice since we complete if exception instance of HAProxyDecoder exception.

Do you want me to set a flag and complete only once (To avoid that warning)
?

zenios:
> 1/ we need to modify the `remoteAddress()` method documentation to mention that it will return the address of the actual client when `useProxyProtocol` is set to `true` and not the proxy address. This method is on `NetSocket`, `HttpServerRequest`, `HttpConnection` and `WebSocketBase`.

Will do

> 2/ we need a test for idle timeout when using proxy protocol, i.e a simple client that only connects and we check the client is disconnected

 I assume this is only neede for net which is the only place we add an extra idleHandler. 
For http the original pipeline is already constructed wtih an idle timeout

vietj:
about 2/ : doesn't NetServer support idle timeout too in this case ?

zenios:
> about 2/ : doesn't NetServer support idle timeout too in this case ?

This is what i am saying.

HAProxyProtocolDecoder does not timeout by its self and this is the reason that for Net (only) we add an extra idle handler (Which should be tested)

For http we dont add an extra idle handler since we construct the full pipeline (Which already contains an idle handler) from the beginning

vietj:
ok thanks, for now we simply test it then

zenios:
@vietj 

While i was testing idleTimeout i found out that for http sslHandshakeTimeout option is not respected

The following line is missing from here
https://github.com/eclipse-vertx/vert.x/blob/78aea977851105148f56b6d66d7cbf010c2d6b8e/src/main/java/io/vertx/core/http/impl/HttpServerChannelInitializer.java#L103

```
sslHandler.setHandshakeTimeout(sslHelper.getSslHandshakeTimeout(), sslHelper.getSslHandshakeTimeoutUnit());
```

Is that intentional? If not do you want me to include the fix with this PR?

Thanks

vietj:
@zenios please open another issue and fix it another PR instead

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -5992,7 +6005,29 @@ public void testHaProxyProtocolVersion2() throws Exception {
...
+
+  @Test
+  public void testHAProxyProtocolIdleTimeout() throws Exception {
+    HAProxy proxy = new HAProxy(DEFAULT_HTTP_HOST, DEFAULT_HTTP_PORT, Buffer.buffer());
+    proxy.start(vertx);
+
+    server.close();
+    server = vertx.createHttpServer(createBaseServerOptions().
+      setIdleTimeout(2).
+      setSslHandshakeTimeout(2).
```
vietj, line 6021: why do we need `setSslHandshakeTimeout` here ? we don't need it for this test

src/main/asciidoc/http.adoc
```diff
@@ -1803,6 +1803,29 @@ use the full URL specified in the request URI:
...
+
+HA PROXY protocol can be enabled by setting the option {@link io.vertx.core.http.HttpServerOptions#useProxyProtocol(boolean)}
+and adding the following dependency in your classpath:
+
+[source,xml]
+----
+<dependency>
+  <groupId>io.netty</groupId>
+  <artifactId>netty-codec-haproxy</artifactId>
+  <version>4.1.18.Final</version>
```
vietj, line 1820: use `${maven.version}` instead

src/main/asciidoc/net.adoc
```diff
@@ -770,3 +770,27 @@ Here's an example:
...
+
+HA PROXY protocol can be enabled by setting the option {@link io.vertx.core.http.NetServerOptions#useProxyProtocol(boolean)}
+and adding the following dependency in your classpath:
+
+[source,xml]
+----
+<dependency>
+  <groupId>io.netty</groupId>
+  <artifactId>netty-codec-haproxy</artifactId>
+  <version>4.1.18.Final</version>
```
vietj, line 789: use `${maven.version}` instead

src/main/asciidoc/net.adoc
```diff
@@ -770,3 +770,27 @@ Here's an example:
...
+
+
+=== Using HA PROXY protocol
+
+https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt[HA PROXY protocol] provides a convenient way to safely transport connection
+information such as a client's address across multiple layers of NAT or TCP
+proxies.
+
+HA PROXY protocol can be enabled by setting the option {@link io.vertx.core.http.NetServerOptions#useProxyProtocol(boolean)}
+and adding the following dependency in your classpath:
```
vietj, line 782: _to_ your classpath ?

src/main/asciidoc/http.adoc
```diff
@@ -1803,6 +1803,29 @@ use the full URL specified in the request URI:
...
 ----
 
+=== Using HA PROXY protocol
+
+https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt[HA PROXY protocol] provides a convenient way to safely transport connection
+information such as a client's address across multiple layers of NAT or TCP
+proxies.
+
+HA PROXY protocol can be enabled by setting the option {@link io.vertx.core.http.HttpServerOptions#useProxyProtocol(boolean)}
+and adding the following dependency in your classpath:
```
vietj, line 1813: _to_ your classpath ?

src/test/java/io/vertx/core/http/HttpTest.java
zenios: Not exactly. In HTTP2 the connection is closed not because of IdleHandler (Which is not yet added to the pipeline) but because of timeout waiting for handshake from (SSLHandler) which is configured using setSslHandshakeTimeout. 

This is how i found that the setting is not respected

src/main/java/examples/HTTPExamples.java
```diff
@@ -918,6 +918,14 @@ public void example60(Vertx vertx) {
 
   }
 
+  public void example61(Vertx vertx) {
+
+    HttpServerOptions options = new HttpServerOptions()
+      .setUseProxyProtocol(true);
+
+    HttpServer server = vertx.createHttpServer(options);
```
vietj, line 926: I think we should add:

```java
server.requestHandler(request -> {
  // Print the actual client address provided by the HA proxy protocol instead of the proxy address
  System.out.println(request.remoteAddress());
});
```

src/main/java/examples/NetExamples.java
```diff
@@ -598,6 +598,11 @@ public void example50(Vertx vertx) throws CertificateException {
       .listen(8080);
   }
 
+  public void example51(Vertx vertx) {
+    NetServerOptions options = new NetServerOptions().setUseProxyProtocol(true);
+    NetServer server = vertx.createNetServer(options);
```
vietj, line 603: I think we should add:

```java
server.connectHandler(so -> {
  // Print the actual client address provided by the HA proxy protocol instead of the proxy address
  System.out.println(so.remoteAddress());
});
```

src/main/asciidoc/http.adoc
zenios: maven.version is used for vertx dependencies not for netty

src/main/asciidoc/net.adoc
zenios: maven.version is used for vertx dependencies not for netty

src/main/asciidoc/net.adoc
zenios: === Native Linux Transport

You need to add the following dependency in your classpath:

src/main/asciidoc/http.adoc
zenios: === Native Linux Transport

You need to add the following dependency in your classpath:

src/main/asciidoc/http.adoc
vietj: ah right, then we should either use a comment here like  or pass the netty version so we can resolve it using `${netty.version}`

src/main/asciidoc/net.adoc
vietj: that might be wrong too :-)

src/main/asciidoc/http.adoc
vietj: it seems that using "in" is correct

src/main/asciidoc/net.adoc
vietj: it seems that using "in" is correct

vietj: sorry for noise

src/main/asciidoc/http.adoc
zenios: I could not find a way to pass netty.version provided from vertx-dependencies down to maven-processor-plugin

We either
1. Declare a new property and we will have to remember to copy the version in the new property and inside vertx-dependencies
2. Remove the version and add a note
3. Any other idea?

vietj: let's go for 2, for now we can revisit later

zenios:
@vietj 

Before we complete this PR do you want me to make a PR to vertx-dependencies adding netty-codec-haproxy as optional dependency?

It should be done at one point so we can get rid of the hardcoded version in this PR

vietj:
yes @zenios for updating vertx-dependencies

I would still like to like resolve the issue with idle timeout in tests: testing idle timeout with the proxy not sending anything should not require SSL timeout, since the proxy send nothing then the SSL handshake handler should not yet be added to the pipeline

zenios:
> yes @zenios for updating vertx-dependencies
> 
> I would still like to like resolve the issue with idle timeout in tests: testing idle timeout with the proxy not sending anything should not require SSL timeout, since the proxy send nothing then the SSL handshake handler should not yet be added to the pipeline

In order to achieve this then we need to delay constructing the pipeline until ha proxy decoder completes. Just like net 

We will also need to add one more idle timeout handler and remove it once proxy decoder completes. Again just like net :)

Do you want me to proceed with that change?

vietj:
@zenios yes please I think it is cleaner this way to proceed in steps when updating the pipeline

zenios:
@vietj 

I have done the changes. If you are happy with it let me know so i can prepare the PR for vertx-dependencies and also squash the commits into one

Thanks

vietj:
sure thing, I'll do another review after that

zenios:
@vietj Ready

This is the relevant PR for deps
https://github.com/vert-x3/vertx-dependencies/pull/42

I have not removed the <version>4.1.48.Final</version> from this PR yet because if removed CI will fail

src/main/java/io/vertx/core/net/impl/ConnectionBase.java
```diff
@@ -491,7 +497,10 @@ public SocketAddress remoteAddress() {
   public SocketAddress localAddress() {
```
zenios, line 497: Do we need todocument localAddress as well? Just like we did for remote address?

vietj: yes, I forgot this

vietj: yes

zenios: Ok will do that now and remove the hardcoded version.Thanks

vietj:
I believe we forgot to implement support for non IPV4 address.

- in text mode the proxy can pass IPV4 or IPV6 addresses
- in binary mode the proxy can pass IPV4, IPV6 or unix domain sockets addresses

We should implement and test these cases for `io.netty.handler.codec.haproxy.HAProxyProxiedProtocol`:

- `HAProxyProxiedProtocol.TCP4` : accept
- `HAProxyProxiedProtocol.TCP6` : accept
- `HAProxyProxiedProtocol.UNIX_STREAM` : accept
- `HAProxyProxiedProtocol.UNKNOWN` : reject
- `HAProxyProxiedProtocol.UDP4` : reject
- `HAProxyProxiedProtocol.UDP6` : reject
- `HAProxyProxiedProtocol.UNIX_DGRAM` : reject

zenios:
> I believe we forgot to implement support for non IPV4 address.
> 
> * in text mode the proxy can pass IPV4 or IPV6 addresses
> * in binary mode the proxy can pass IPV4, IPV6 or unix domain sockets addresses
> 
> We should implement and test these cases for `io.netty.handler.codec.haproxy.HAProxyProxiedProtocol`:
> 
> * `HAProxyProxiedProtocol.TCP4` : accept
> * `HAProxyProxiedProtocol.TCP6` : accept
> * `HAProxyProxiedProtocol.UNIX_STREAM` : accept
> * `HAProxyProxiedProtocol.UNKNOWN` : reject
> * `HAProxyProxiedProtocol.UDP4` : reject
> * `HAProxyProxiedProtocol.UDP6` : reject
> * `HAProxyProxiedProtocol.UNIX_DGRAM` : reject

What exception shall we throw in rejected cases?

vietj:
you mean for `NetServer` and `HttpServer` exception handlers ?

zenios:
> you mean for `NetServer` and `HttpServer` exception handlers ?



> you mean for `NetServer` and `HttpServer` exception handlers ?

When the protocol is one of the following 

HAProxyProxiedProtocol.UNKNOWN : reject
HAProxyProxiedProtocol.UDP4 : reject
HAProxyProxiedProtocol.UDP6 : reject
HAProxyProxiedProtocol.UNIX_DGRAM : reject

we should reject.That means i will throw an exception (To be reported to the exception handlers) and close the connection

What exception shall i throw?

vietj:
I think for now a simple `IOException` with an appropriate message is enough. We will see later if we need something more accurate or reuse an existing exception. We should statically cache these exceptions and have a single instance for each case.

zenios:
> I think for now a simple `IOException` with an appropriate message is enough. We will see later if we need something more accurate or reuse an existing exception. We should statically cache these exceptions and have a single instance for each case.

Is it ok to create only one. For example 

```
new IOException("Unsupported protocol");
```

vietj:
very fine @zenios

zenios:
@Viet 

The default transport of vertx does not support DomainSocketAddress and when haproxy sends a unix socket protocol when we convert it to vertx address it returns null

Implementation of default Vertx transport
```
  public io.vertx.core.net.SocketAddress convert(SocketAddress address) {
    if (address instanceof InetSocketAddress) {
      return io.vertx.core.net.SocketAddress.inetSocketAddress((InetSocketAddress) address);
    } else {
      return null;
    }
  }
```

And this is where we convert 
```
  public SocketAddress remoteAddress() {
    SocketAddress address = remoteAddress;
    if (address == null) {
      java.net.SocketAddress addr = chctx.channel().hasAttr(REMOTE_ADDRESS_OVERRIDE) ?
        chctx.channel().attr(REMOTE_ADDRESS_OVERRIDE).getAndSet(null) :
        chctx.channel().remoteAddress();

      if (addr != null) {
        address = vertx.transport().convert(addr);
        remoteAddress = address;
      }
    }
    return address;
  }
```

Why do we convert it like that?

We could just use
```
  public SocketAddress remoteAddress() {
    SocketAddress address = remoteAddress;
    if (address == null) {
      java.net.SocketAddress addr = chctx.channel().hasAttr(REMOTE_ADDRESS_OVERRIDE) ?
        chctx.channel().attr(REMOTE_ADDRESS_OVERRIDE).getAndSet(null) :
        chctx.channel().remoteAddress();

      if (addr instanceof InetSocketAddress) {
        remoteAddress = SocketAddress.inetSocketAddress((InetSocketAddress) addr);
      } else if (addr instanceof DomainSocketAddress) {
        remoteAddress = SocketAddress.domainSocketAddress(((DomainSocketAddress) addr).path());
      } else
        remoteAddress = null;
    }
    return address;
  }
```

Thanks

zenios:
I checked were we use remoteAddress and localAddress inside vertx code and its used only for metrics  I that remoteAddress / localAddress should not be concerned with the transport used.

vietj:
the issue with this code:

```
public SocketAddress remoteAddress() {
    SocketAddress address = remoteAddress;
    if (address == null) {
      java.net.SocketAddress addr = chctx.channel().hasAttr(REMOTE_ADDRESS_OVERRIDE) ?
        chctx.channel().attr(REMOTE_ADDRESS_OVERRIDE).getAndSet(null) :
        chctx.channel().remoteAddress();

      if (addr instanceof InetSocketAddress) {
        remoteAddress = SocketAddress.inetSocketAddress((InetSocketAddress) addr);
      } else if (addr instanceof DomainSocketAddress) {
        remoteAddress = SocketAddress.domainSocketAddress(((DomainSocketAddress) addr).path());
      } else
        remoteAddress = null;
    }
    return address;
  }
```

is that it forces to have `netty-transport-native-unix-common` on the classpath by default and it is a dependency declared with `<optional>true</optional>`, so that's not an option.

to avoid loosing information due to the `java.net.SocketAddress` we need actually to create the Vert.x `SocketAddress` within the handler directly when the information whether we know the type of the address.

vietj:
i.e Vert.x `SocketAddress` should be created in `HAProxyMessageCompletionHandler` instead of `java.net.SocketAddress`

zenios:
> the issue with this code:
> 
> ```
> public SocketAddress remoteAddress() {
>     SocketAddress address = remoteAddress;
>     if (address == null) {
>       java.net.SocketAddress addr = chctx.channel().hasAttr(REMOTE_ADDRESS_OVERRIDE) ?
>         chctx.channel().attr(REMOTE_ADDRESS_OVERRIDE).getAndSet(null) :
>         chctx.channel().remoteAddress();
> 
>       if (addr instanceof InetSocketAddress) {
>         remoteAddress = SocketAddress.inetSocketAddress((InetSocketAddress) addr);
>       } else if (addr instanceof DomainSocketAddress) {
>         remoteAddress = SocketAddress.domainSocketAddress(((DomainSocketAddress) addr).path());
>       } else
>         remoteAddress = null;
>     }
>     return address;
>   }
> ```
> 
> is that it forces to have `netty-transport-native-unix-common` on the classpath by default and it is a dependency declared with `<optional>true</optional>`, so that's not an option.
> 

You are right.I though DomainSocketAddress was part of Java. My mistake

vietj:
so now I'm actually reading the spec :-)

for UNKNOWN actually we should have the behaviour of accepting the connection but we need to use the physical socket address

_If the announced transport protocol is "UNKNOWN", then the receiver knows that
the sender speaks the correct PROXY protocol with the appropriate version, and
SHOULD accept the connection and use the real connection's parameters as if
there were no PROXY protocol header on the wire._

vietj:
for v1 this is _UNKNOWN_ for v2 this is _LOCAL_ : we need to check both I think

vietj:
NOTE: for v2 when using _LOCAL_, it will send _ AF_UNSPEC_ type of socket but in this case we need to use the physical socket addresses so it should work fine

vietj:
that is the same for v2 : "UNSPEC" in this case we use the physical socket address as well

zenios:
Sure no problem.

vietj:
I'm wondering also that we should use a different timeout value than the usual TCP idle timeout, like we do have for SSL handshake timeout.

zenios:
> I'm wondering also that we should use a different timeout value than the usual TCP idle timeout, like we do have for SSL handshake timeout.

We will need two options. One for the timeout and one for the unit.

What should be the defaults?

vietj:
let's use 10 seconds like the SSL handshake.

normally the proxy information arrives in the first TCP packet in a single segment.

zenios:
@vietj Ready

vietj:
I'll do a final (hopefully) review now

zenios:
@viet  Thanks. You can also close https://github.com/eclipse-vertx/vert.x/pull/1271

PR #3388, opened 2020-04-23
Event-bus streaming WIP

vietj:
Prototype for streaming on top of event-bus.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/EventBusStream.java
+++ b/src/main/java/io/vertx/core/eventbus/EventBusStream.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.eventbus;
+
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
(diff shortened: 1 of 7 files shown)


src/main/java/io/vertx/core/eventbus/impl/ConsumerStream.java
```diff
@@ -0,0 +1,158 @@
...
+    EventBus bus = context.owner().eventBus();
+    bus.consumer(address, msg -> {
+      String remoteAddress = msg.headers().get(STREAM_ADDRESS);
+      if (remoteAddress == null) {
+        // No back pressure (not supported)
+        msg.fail(0, "No credits address");
+        return;
+      }
+      String localAddress = UUID.randomUUID().toString();
+      ConsumerStream<T> stream = new ConsumerStream<>(bus, context);
```
gaol, line 45: This would create new ConsumerStream on each message, should it be put before `bus.consumer()` ?

vietj:
this is expected, on each message on the specified address it creates a stream, this is similar to a server socket that creates socket

src/main/java/io/vertx/core/eventbus/EventBusStream.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
```
jponge, line 2: You can safely add 2020 :-)

src/main/java/io/vertx/core/eventbus/impl/ConsumerStream.java
```diff
@@ -0,0 +1,158 @@
...
+        } else {
+          refund = false;
+        }
+      }
+      if (handler != null) {
+        handler.handle(msg.body());
+      }
+      if (refund) {
+        // Refund 500 credits at once
+        bus.send(remoteAddress, null, new DeliveryOptions().addHeader(STREAM_CREDITS, "500"));
```
jponge, line 115: Quick question: why 500? (if that's just an early prototyping value I get it)

vietj:
no particular reason, in practice this will use rather a low water mark and high water mark instead than being 0 - 500 range

knotenpunkt:
@vietj what is the state of this pr?

vietj:
it is currently on hold

PR #3390, opened 2020-04-25
Message consumer is not anymore a write stream

### LINKED ISSUE #3389: Message consumer is not anymore a write stream
In Vert.x 3, the `MessageProducer` interface extends the `WriteStream` interface and provides limited support for message back-pressure. In practice credit leak might happen leading to a depletion of credits in the producer that won't allow to emit messages.

The `MessageConsumer` will continue to extend `ReadStream` and will drop messages when the consumer is paused and the pending message queue is full. This allows to continue supporting the nice integration with Rx generators for building message consuming pipelines.

Point to point message streaming over the event-bus will be introduced in Vert.x 4 to provide a supportable solution to replace this.

vietj:
Fixes #3389


vietj:
@tsegismont before we merge this PR, I think you can spend some time looking at the new construct and review it / break it, I would feel more comfortable this way

Issue #524, opened 2013-03-04
2.0.0 SNAPSHOT classloader


phungleson:
Hi guys,

Just want to test out the classloader in Vertx master branch.

As in the wiki https://github.com/vert-x/vert.x/wiki/Vert.x-2.0-tasks

```
Classloader refactoring - DONE
Each module/verticle type has its own classloader, not a classloader per instance
See Vert.x 2.0 plan for more information
```

Since a Verticle has its own classloader (not per instance) I would think that I can create an singleton class (connection pool) and reuse it between instances. I wrote a very simple example here:

https://github.com/phungleson/vertx-classloader/tree/master/src/main/java/vertx/classloader

If my theory is correct when I run https://github.com/phungleson/vertx-classloader/blob/master/start with `-instances 2` it should display only one line of `Singleton constructor` but somehow the results are 2 lines, meaning `Singleton` class is loaded twice.

Not sure if my understand is correct or I need to enable something or it is the way it should react and no more changes will be done in terms of classloader in Vertx 2?

Regards,
Son.

PR #3401, opened 2020-05-06
Add automatic module name

afloarea:
Signed-off-by: afloarea <adrianfloarea07@yahoo.com>

Motivation:

Add automatic module name as per issue [#524](https://github.com/vert-x3/issues/issues/524)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,6 +56,7 @@
     <jmh.version>1.19</jmh.version>
     <vertx.testNativeTransport>false</vertx.testNativeTransport>
     <vertx.testDomainSockets>false</vertx.testDomainSockets>
+    <jar.manifest>${project.basedir}/src/main/resources/META-INF/MANIFEST.MF</jar.manifest>
   </properties>
 
   <dependencyManagement>
--- a/src/main/resources/META-INF/MANIFEST.MF
+++ b/src/main/resources/META-INF/MANIFEST.MF
@@ -1,0 +1,2 @@
+Automatic-Module-Name: io.vertx.core
+


vietj:
I woiuld like first to see if we can do something in the vertx-parent pom

vietj:
@afloarea please have a look at https://github.com/vert-x3/vertx-parent/commit/1384372552de56622b7455d79ce610b5a3b3c1a9

This change should allow to have a file named `MANIFEST.MF` that will contain the entry containing `Automatic-Module-Name: io.vertx.core` that will define the automatic module name.

src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,2 @@
+Automatic-Module-Name: io.vertx.core
+
```
afloarea, line 2: Note: The new line is required. Otherwise the maven plugin does not work for some reason.

vietj: shall we upgrade the plugin version ?

vietj: I mean would it solve the issue ?

afloarea: I tried with the newest I could find on maven central (maven jar plugin 3.2.0) and the issue is still present

afloarea: There is no error or anything, the jar is built but the Automatic-Module-Name property is not in the built jar.

pom.xml
```diff
@@ -56,6 +56,7 @@
     <jmh.version>1.19</jmh.version>
     <vertx.testNativeTransport>false</vertx.testNativeTransport>
     <vertx.testDomainSockets>false</vertx.testDomainSockets>
+    <jar.manifest>src/main/resources/META-INF/MANIFEST.MF</jar.manifest>
```
vietj, line 59: use rather `${project.basedir}/src/main/resources/META-INF/MANIFEST.MF`

src/main/resources/META-INF/MANIFEST.MF
vietj: ok then

pom.xml
afloarea: done

PR #3405, opened 2020-05-11
Implement file descriptor leak tests

zenios:
This is part of work on #3396

It establishes a junit annotation that when added on a test method it evaluates it for file descriptor leaks.

This way we can add it to existing tests and easily find leaks

CI fails since the test has detected the leak mentioned on #3396

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
@@ -601,7 +601,13 @@ public class Http2ServerResponseImpl implements HttpServerResponse {
           }
         }
         checkSendHeaders(false);
-        file.pipeTo(this, h);
+        file.pipeTo(this, ar1 -> file.close(ar2 -> {
+          Throwable failure = ar1.failed() ? ar1.cause() : ar2.failed() ? ar2.cause() : null;
+          if(failure == null)
+            h.handle(ar1);
+          else
+            h.handle(Future.failedFuture(failure));
+        }));
       } else {
         h.handle(ar.mapEmpty());
       }
--- a/src/test/java/io/vertx/test/core/FileDescriptorLeak.java
+++ b/src/test/java/io/vertx/test/core/FileDescriptorLeak.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.test.core;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({java.lang.annotation.ElementType.METHOD})
+public @interface FileDescriptorLeak {
+}
+
--- a/src/test/java/io/vertx/test/core/FileDescriptorLeakDetectorRule.java
+++ b/src/test/java/io/vertx/test/core/FileDescriptorLeakDetectorRule.java
@@ -17,21 +17,31 @@ import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
 import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
 import java.util.ArrayList;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class FileDescriptorLeakRule implements TestRule {
+public class FileDescriptorLeakDetectorRule implements TestRule {
   private static final int BASE_LINE_RUNS = 20;
   private static final int TEST_RUNS = 20;
 
-  private static final UnixOperatingSystemMXBean os = (UnixOperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+  private final UnixOperatingSystemMXBean os;
+
+  public FileDescriptorLeakDetectorRule() {
+    OperatingSystemMXBean os = ManagementFactory.getOperatingSystemMXBean();
+    if (!(os instanceof UnixOperatingSystemMXBean))
+      this.os = null;
+    else
+      this.os = (UnixOperatingSystemMXBean) os;
+
+  }
 
   @Override
   public Statement apply(Statement statement, Description description) {
-    FileDescriptorLeak fileDescriptorLeak = description.getAnnotation(FileDescriptorLeak.class);
-    if (fileDescriptorLeak == null) {
+    DetectFileDescriptorLeaks detectFileDescriptorsLeaks = description.getAnnotation(DetectFileDescriptorLeaks.class);
+    if (detectFileDescriptorsLeaks == null || os == null) {
       return statement;
     }
     return new Statement() {
--- a/src/test/java/io/vertx/test/core/FileDescriptorLeakRule.java
+++ b/src/test/java/io/vertx/test/core/FileDescriptorLeakRule.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ *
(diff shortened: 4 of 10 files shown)


src/test/java/io/vertx/test/core/FileDescriptorLeak.java
```diff
@@ -0,0 +1,22 @@
...
+
+package io.vertx.test.core;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({java.lang.annotation.ElementType.METHOD})
+public @interface FileDescriptorLeak {
```
vietj, line 20: What is this interface useful for ?

src/test/java/io/vertx/test/core/VertxTestBase.java
```diff
@@ -42,6 +42,9 @@
   @Rule
   public RepeatRule repeatRule = new RepeatRule();
 
+  @Rule
+  public FileDescriptorLeakRule fileDescriptorLeakRule = new FileDescriptorLeakRule();
```
vietj, line 46: I believe that the vertx test base should not annotated by default because it will cause some overhead in testing. Instead we should have an FD leak specific test that will test a few operations we have like starting HTTP server, using file system, etc... and we will have of course the HTTP/2 send file test.

src/test/java/io/vertx/test/core/FileDescriptorLeak.java
zenios: It is added on a method that we need to check for leaks.

Without this and only the rule it would run in all the tests of each file (I think)

zenios: In the PR you can see it added on the 

testSendFile in order to check for leaks both for http1 and http2

The same can be added on a net test / file system test and will instantly check for leaks

src/test/java/io/vertx/test/core/VertxTestBase.java
zenios: Overhead is added only when the annotation is present in one / many test methods

If we want to create separate tests just for this then for each test we have to setup 
-vertx
-http
-ssl if http2 
and so on

With these method everything is already there since its run on top of the existing tests

vietj: ah right, so it's good here.

I think it's fine like this then. We'll see later if we need to establish a specific test class.

src/test/java/io/vertx/test/core/FileDescriptorLeak.java
vietj: yes I missed it

vietj: Can you add a same test for other usage of send file in NetSocket and HTTP/1 ?

zenios: Since we annotated the sendFile of HttpTests class then it works for both http1 and http2

You can clearly see that http1 passes and http2 fails

vietj: indeed, good job

vietj:
So you only need to add a commit for fixing the HTTP/2 case ?

zenios:
I will add some more tests for net and fs and then a fix for the leak

One more question though
Do you know why

https://github.com/eclipse-vertx/vert.x/blob/00eb24827e7dc3ae64e9b2720c33b58830c46aa2/src/main/java/io/vertx/core/http/impl/VertxHttp2Stream.java#L288

is needed? Isnt fs.open supposed to file if file cannot be opened?

vietj:
I don't remember precisely.

zenios:
Ok will check it out and let you know. For 3.9.1 i will backport only the fix. Not the leak detection tests.

Is that ok?

vietj:
why can 't we backport the detection ?

zenios:
We can but is it needed? 

I was hoping that we are close to 4 :)

> why can 't we backport the detection ?

vietj:
testing is always useful

zenios:
@vietj Leas has been patched

Check it out. When it gets merge i will backport to 3.9.1

Thanks

src/main/java/io/vertx/core/http/impl/Http2ServerResponseImpl.java
```diff
@@ -601,7 +601,13 @@ public HttpServerResponse sendFile(String filename, long offset, long length, Ha
           }
         }
         checkSendHeaders(false);
-        file.pipeTo(this, h);
+        file.pipeTo(this, ar1 -> file.close(ar2 -> {
```
vietj, line 604: can you try instead : file.pipe().endOnComplete(true).to(h) this should have the same effect

src/test/java/io/vertx/test/core/FileDescriptorLeak.java
vietj: can you rename it `DetectFileDescriptorLeak`

src/test/java/io/vertx/test/core/FileDescriptorLeakRule.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class FileDescriptorLeakRule implements TestRule {
+  private static final int BASE_LINE_RUNS = 20;
+  private static final int TEST_RUNS = 20;
+
+  private static final UnixOperatingSystemMXBean os = (UnixOperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
```
vietj, line 29: We cannot assume that `UnixOperatingSystemMXBean` is always present.

We need to instead use the MBeanServer (`ManagementFactory.getPlatformMBeanServer()`)  and invoke the `getOpenFileDescriptorCount` as an MBean attribute via the `MBeanServer`

The rule should do nothing when such attribute does not exist (on Windows).

src/test/java/io/vertx/test/core/FileDescriptorLeakRule.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+import java.lang.management.ManagementFactory;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class FileDescriptorLeakRule implements TestRule {
```
vietj, line 25: rename it `FileDescriptorLeakDetetectorRule`

zenios:
@vietj changes applied

zenios:
@vietj ping

src/test/java/io/vertx/test/core/FileDescriptorLeakDetectorRule.java
```diff
@@ -0,0 +1,89 @@
...
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.test.core;
+
+import com.sun.management.UnixOperatingSystemMXBean;
```
vietj, line 14: we don't even want to import this class because it will not load in windows

vietj: we should use the platform `MBeanServer` and invoke this as an `MBean` using its `ObjectName` and do a `getAttribute` on `OpenFileDescriptorCount`.

before that we want to check the mbean metadata to see if the info exist or nor (on windows it will nto exist) using the `MBeanInfo` returned by the `MBeanServer`.

vietj: On windows this rule would then just do nothing.

vietj: I'll handle this and continue the PR @zenios

zenios: Really sorry @vietj  i got blocked by a an urgent project :(

vietj: N/p @zenios you did most of the hard work :-)

Here is the continuation PR https://github.com/eclipse-vertx/vert.x/pull/3408

PR #3413, opened 2020-05-14
Cannot close file if one of the write operations fails

### LINKED ISSUE #2963: Cannot close file if one of the write operations fails

tsegismont:
Fixes #2963

### DIFF
--- a/src/test/java/io/vertx/core/file/FileSystemTest.java
+++ b/src/test/java/io/vertx/core/file/FileSystemTest.java
@@ -13,12 +13,7 @@ package io.vertx.core.file;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
-import io.vertx.core.AsyncResult;
-import io.vertx.core.Context;
-import io.vertx.core.Future;
-import io.vertx.core.Handler;
-import io.vertx.core.Promise;
-import io.vertx.core.Vertx;
+import io.vertx.core.*;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.file.impl.AsyncFileImpl;
 import io.vertx.core.impl.Utils;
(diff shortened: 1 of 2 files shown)


src/test/java/io/vertx/core/file/FileSystemTest.java
```diff
@@ -1133,6 +1127,24 @@ public void testWriteAsync() {
     await();
   }
 
+  @Test(timeout = 5000)
```
vietj, line 1130: Did you check this timeout in Travis ? it sounds low to me and prone to false negatives

tsegismont: The build passed https://travis-ci.org/github/eclipse-vertx/vert.x/builds/687019097

Actually I added this timeout so that it failed fast while the bug was not fixed on my box. I will remove it

tsegismont: Done

PR #3425, opened 2020-05-25
The shutdown hook wasn't starting the right thread and leaking the ca…

pmlopes:
…cheDir to null while still in use

Motivation:

The delete cache dir shutdown hook isn't starting a thread properly, this means that the delete code set's the cache dir to null while still on the eventloop and may leak (like it happens on vertx-web unit tests). In this case cache files are written to the `CWD` when they shouldn't.

### DIFF
--- a/src/main/java/io/vertx/core/file/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/file/impl/FileResolver.java
@@ -112,7 +112,7 @@ public class FileResolver {
       cwd = null;
     }
     if (this.enableCpResolving) {
-      setupCacheDir();
+      setupCacheDir(UUID.randomUUID().toString());
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -377,31 +377,29 @@ private ClassLoader getClassLoader() {
...
+          deleteCacheDir();
+        } catch (IOException ignore) {
         }
-      });
-      Runtime.getRuntime().addShutdownHook(shutdownHook);
-    }
+        latch.countDown();
+      }).start();
+      try {
+        latch.await(10, TimeUnit.SECONDS);
```
vietj, line 397: instead of a Latch use thread.join on the thread that was started

Issue #3430, opened 2020-05-29
Context affinity for non Vert.x thread


vietj:
Vert.x 3 creates a new instance of `io.vertx.core.Context` each time it is called from a non Vert.x thread. In Vert.x 4 we use a future model and this model does not work well for code wrapping Vert.x objects and compose them, e.g

```java
public Future<Integer> doSomething() {
  Future<Integer> result1 = getSomeResult();
  Future<Integer> result2 = getOtherResult();
  return CompositeFuture.all(result1, result2).map(this::composteResults);
}
```

When this is executed from a non Vert.x thread, two contexts will be created and the final result might be on either context.

This feature associates non Vert.x threads with a context the first time a context is created, hence allowing:

```java
new Thread() -> {
  assertEquals(vertx.getOrCreateContext(), vertx.getOrCreateContext());
}).start();
```

This does is not a breaking change per se, since the `getOrCreateContext()` method either create or return an existing context. The only affected code will be code that relies on this implementation detail to create contexts instead of using a `Verticle`.

The implementation uses weak references to keep track of the context in the thread local to allow releasing contexts when the thread ends, although this is not a realistic use case.

Consequently, the trick in the event-bus that uses a specific contexts for ordering messages sent from a non Vert.x thread is not necessary anymore and it can simply use the context without worrying of this.

PR #3431, opened 2020-05-29
Event loop affinity

vietj:
Implementation of https://github.com/eclipse-vertx/vert.x/issues/3430

### DIFF
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -110,7 +110,7 @@ public interface Vertx extends Measured {
   /**
    * Gets the current context
    *
-   * @return The current context or @code null if there is no current context
+   * @return The current context or {@code null} if there is no current context
    */
   static @Nullable Context currentContext() {
     return ContextInternal.current();
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -110,7 +110,7 @@ public interface Vertx extends Measured {
   /**
    * Gets the current context
    *
-   * @return The current context or null if no current context
+   * @return The current context or @code null if there is no current context
    */
   static @Nullable Context currentContext() {
     return ContextInternal.current();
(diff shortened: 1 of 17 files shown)


src/main/java/io/vertx/core/Vertx.java
```diff
@@ -110,7 +110,7 @@ static void clusteredVertx(VertxOptions options, Handler<AsyncResult<Vertx>> res
   /**
    * Gets the current context
    *
-   * @return The current context or null if no current context
+   * @return The current context or @code null if there is no current context
```
jponge, line 113: `{@code null}`

src/test/java/io/vertx/core/ContextTest.java
```diff
@@ -809,4 +809,10 @@ public void testWorkerExecutor() {
     });
     await();
   }
+
+  @Test
+  public void testSticky() {
```
jponge, line 814: 👍

src/test/java/io/vertx/core/VertxTest.java
```diff
@@ -213,4 +214,42 @@ public void testFinalizeNetClient() throws Exception {
     }
     await();
   }
+
+
+  @Test
+  public void testStickContextFinalization() throws Exception {
+    VertxInternal vertx = (VertxInternal) Vertx.vertx();
```
jponge, line 221: Not sure you need to cast here.

jponge:
Overall LGTM, and a welcomed change. I'm just really wondering about the opportunity to move `createEventLoopContext` to the public APIs.

vietj:
@jponge for now I'd like to keep it internal, people should deploy verticle in practice to achieve this

jponge:
I agree that users should deploy verticles, I was relating to things like clients where it’d avoid having people cast to VertxInternal.

vietj:
@jponge a cast to VertxInternal is fine, it means you are crossing a boundary. We might take most `*Internal` interfaces at some point and have a `.internal.` package separate from `.impl.`

PR #3441, opened 2020-06-11
Prevent extra buffer copies when parsing did not extract anything

tcataldo:
Signed-off-by: Thomas Cataldo <thomas.cataldo@blue-mind.net>

Motivation:

My workload needs to parse a stream containing 10 chunks with 100MB of fixed sized items inside a bigger string. Performance is abysmal and yourkit pointed to this line.


vietj:
can you do a PR for 3.9 as well ?

Issue #3051, opened 2019-08-01
Test failures on Ubuntu 16.04 on X86 architecture


SantoshMagdum:
I am trying to build Eclipse Vert.x package master branch on Ubuntu 16.04 platform on X86 architecture. I could build the package but I am getting around 8 test case failures.

These are the steps that I followed -
	- Installed Java version: 1.8.0
	- Installed Apache Maven 3.6.0
	- Cloned the master copy of eclipse-vertx git code base.
	- Triggered the build

I am currently trying to look at the logs to investigate the failures, however any pointers on any additional dependencies / configuration required for these, which I may be missing would be of great help.

Appreciate your help in advance.

Below are the test cases which are failing. Please let me if I should share more specific details.

	io.vertx.core.DeploymentTest.txt:Tests run: 76, Failures: 4, Errors: 0, Skipped: 0, Time elapsed: 38.469 sec <<< FAILURE! - in io.vertx.core.DeploymentTest
	io.vertx.core.file.FileSystemTest.txt:Tests run: 101, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 11.874 sec <<< FAILURE! - in io.vertx.core.file.FileSystemTest
	io.vertx.core.http.Http1xTest.txt:Tests run: 404, Failures: 1, Errors: 1, Skipped: 1, Time elapsed: 214.646 sec <<< FAILURE! - in io.vertx.core.http.Http1xTest
	io.vertx.core.http.Http2Test.txt:Tests run: 293, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 214.5 sec <<< FAILURE! - in io.vertx.core.http.Http2Test
	io.vertx.core.net.NetTest.txt:Tests run: 125, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 31.268 sec <<< FAILURE! - in io.vertx.core.net.NetTest

vietj:
can you provide more details ?

SantoshMagdum:
We had also setup a travis job to build eclipse vert.x on ppc64le. We see similar test case failures on ppc64le as well. Please note, I had removed the errors related to open ssl on my local ppc64le build.
Locally I had built ‘Hawtnji’ and ‘netty-tcnative-master’ on ppc64le and still building dependent ‘netty’ package.

Below is the travis ppc64le build job log -
                https://travis-ci.org/SantoshMagdum/vert.x/jobs/566829629

[...]

PR #3443, opened 2020-06-12
testListenInvalidPort now valid for Windows and OSX

hutchig:
Motivation:

The test 'testListenInvalidPort' fails on OSX as it is possible to use port 7 in that env,
from reading the test it also looks like this is true in Windows too.
 
Its test's failure trips up 'mvn clean install' on OSX.

This PR changes the test to first occupy a valid port (by asking for port 0)
exclusively (need to set this on OSX) and then using that occupied port
for the collision. The socket is Closable so the try with resources should
close the port (after the await) meaning the test should not leak socket
resources.

This test is also mentioned as failing in another env in #3051 

Signed-off-by: Gordon Hutchison <Gordon.Hutchison@gmail.com>

### DIFF
--- a/src/test/java/io/vertx/core/http/HttpTest.java
+++ b/src/test/java/io/vertx/core/http/HttpTest.java
@@ -2706,11 +2706,18 @@ public abstract class HttpTest extends HttpTestBase {
   @Test
   public void testListenInvalidPort() throws Exception {
     server.close();
-    try(ServerSocket occupied = new ServerSocket(0)){
+    ServerSocket occupied = null;
+    try{
+      /* Ask to be given a usable port, then use it exclusively so Vert.x can't use the port number */
+      occupied = new ServerSocket(0);
       occupied.setReuseAddress(false);
       server = vertx.createHttpServer(new HttpServerOptions().setPort(occupied.getLocalPort()));
       server.requestHandler(noOpHandler()).listen(onFailure(server -> testComplete()));
       await();
+    }finally {
+      if( occupied != null ) {
+        occupied.close();
+      }
     }
   }
 
--- a/src/test/java/io/vertx/core/http/HttpTest.java
+++ b/src/test/java/io/vertx/core/http/HttpTest.java
@@ -36,6 +36,7 @@ import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
 import java.io.*;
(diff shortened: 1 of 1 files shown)


hutchig:
Fixes #3444

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -2704,12 +2705,13 @@ public void testGetAbsoluteURI() {
...
     server.close();
-    server = vertx.createHttpServer(new HttpServerOptions().setPort(7));
-    server.requestHandler(noOpHandler()).listen(onFailure(server -> testComplete()));
-    await();
+    try(ServerSocket occupied = new ServerSocket(0)){
+      occupied.setReuseAddress(false);
+      server = vertx.createHttpServer(new HttpServerOptions().setPort(occupied.getLocalPort()));
+      server.requestHandler(noOpHandler()).listen(onFailure(server -> testComplete()));
+      await();
+    }
```
vietj, line 2714: the occupied server socket should be closed in a finally block surrounding this

hutchig: Please correct me if I am wrong but:

- The `occupied` resource is opened in the [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)
- ServerSocket implements java.io.Closeable which extends AutoCloseable
- The `await()` will prevent the try-with-resources AutoClosing the socket before
the test has executed. 

I do see the code is not obvious if read quickly and I tried to talk to this above in ["The socket is Closable so the try with resources should close the port (after the await) "](https://github.com/eclipse-vertx/vert.x/pull/3443#issue-433582297). I apologise that I did not put a comment this effect in the code (and also for the use of port `0` result of assigning a free port) but the rest of the tests are not very wordy so I was trying to adopt the same style.

hutchig: ServerSocket is autoclosable and occupied is closed by the try-with-resources statement

vietj: indeed that's very correct, I am not using this pattern as we don't use it often with non blocking resources

vietj:
thanks, can you backport this to 3.9 branch @hutchig

PR #3445, opened 2020-06-14
MessageConsumer should be registered as close hook

vietj:
If we register a class we don't keep a reference, the close hook can be collected and lead to not closing the hook when the vertical is underplayed


vietj:
reported by @zenios

PR #3456, opened 2020-06-18
use rawPath for HttpUtils resolveURIReference(vertx-web#1637)

shenzhou-6:
Fix [Repeat redirects 301 same location](https://github.com/vert-x3/vertx-web/issues/1637)

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpUtils.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpUtils.java
@@ -379,12 +379,12 @@ public final class HttpUtils {
     if (_ref.getScheme() != null) {
       scheme = _ref.getScheme();
       authority = _ref.getAuthority();
-      path = removeDots(_ref.getPath());
+      path = removeDots(_ref.getRawPath());
       query = _ref.getRawQuery();
     } else {
       if (_ref.getAuthority() != null) {
         authority = _ref.getAuthority();
-        path = _ref.getPath();
+        path = _ref.getRawPath();
         query = _ref.getRawQuery();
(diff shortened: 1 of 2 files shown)


vietj:
can you give a proper explanation of what is changed and the reason

shenzhou-6:
@vietj  Because the getPath method obtains the original data without encoding, for example:
"http://exapmle/%E4%BD%A0%E5%A5%BD"

If  use getPath, the value  is chinese  "你好",

but use getRawPath,the value is same location "/%E4%BD%A0%E5%A5%BD"

shenzhou-6:
If the path is "你好", then it is illegal in HttpUrl

shenzhou-6:
in [https://www.ietf.org/rfc/rfc1738.txt](https://www.ietf.org/rfc/rfc1738.txt)

image

shenzhou-6:
The reason for the repeated redirect 301 is that its target server does not recognize the illegal path in the URL. So always Response 301

shenzhou-6:
ISSUES 3456 
Its target website address is : "https://apkcombo.com/천사티비-드라마-다시보기-tv/com.angelseries.angeltv/"

Replace "천사티비-드라마-다시보기" with other illegal characters, 
for example: "https://apkcombo.com/你好-tv/com.angelseries.angeltv/" 
It will still redirect to https://apkcombo.com/천사티비-드라마-다시보기-tv/com.angelseries.angeltv/

shenzhou-6:
Therefore, as long as we guarantee that the redirected URL is in the original point "Location", this will not happen

PR #3458, opened 2020-06-18
Support header size limit exceeded with http status 431

mihaico:
Motivation:
Vertx does not return `431 Request Header Fields Too Large` status for HTTP/1.1 (could not confirm how it works on HTTP/2.0), instead it falls back to the generic `400 Bad Message` (only handling explicitly `414 URI Too Long`).

Since Netty does handle both `414` and `431`, although with the same exception `TooLongFrameException`, the status message is different. I could not find any info whether omitting `431` was on purpose, but in any case if there was reason for sending the generic `400`, could anyone clarify?

In the absence of the whole picture, this my attempt to amend the current Vertx behaviour.

Conformance:
Signed Eclipse Contributor Agreement

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
@@ -477,7 +477,14 @@ public class Http1xServerConnection extends Http1xConnectionBase<ServerWebSocket
       } else {
         version = HttpVersion.HTTP_1_1;
       }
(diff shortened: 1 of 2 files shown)


vietj:
do you mind back port to 3.9 branch @mihaico ?

mihaico:
> do you mind back port to 3.9 branch @mihaico ?

sure, here it is https://github.com/eclipse-vertx/vert.x/pull/3459. Thanks for the quick reply @vietj

Issue #3460, opened 2020-06-19
Add Http PATCH support to HTTPClient


jbtrystram:
#### Describe the feature

Currently, HttpClient does support `PUT`, `POST` `DELETE` and `GET`. the `PATCH` method is not supported.

#### Use cases

Use the HTTPClient to submit patches through the JSON Patch format, using the corresponding HTTP verb.

#### Contribution

Pr incoming

PR #3461, opened 2020-06-19
[#3460] Add patch methods to HttpClient and implementation

jbtrystram:
The vertx Http router supports PATCH routes but the client does not have "shorcuts" to use the Http PATCH methods. 

I am working on a JSON Patch (RFC 6902) implementation for eclipse Hono and we use the Vert.x `HttpClientImpl` for the unit tests. Maybe there is a way I could make use of the `send` methods specifying the HTTP method, but those are not exposed in the `HttpClient` interface.

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpClient.java
+++ b/src/main/java/io/vertx/core/http/HttpClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2020 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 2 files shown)


vietj:
as PATCH is quite rare to use, we haven't included it yet by default.

You can use the generic `HttpClient#send(RequestOptions)` method and `RequestOptions` has an HTTP method on it.

vietj:
thanks for your contribution

vietj:
e.g `client.send(new RequestOptions().setMethod(HttpMethod.PATCH)..., )`

jbtrystram:
@vietj in Vertx 3.9 the `HttpClient` interface does not have `send` methods. Should I use `HttpClientRequest request(HttpMethod method, RequestOptions options, Handler<HttpClientResponse> responseHandler);` ?

vietj:
@jbtrystram since your PR was for master, I assumed that you were using a milestone release of Vert.x 4

You can proceed like this indeed : https://github.com/eclipse-vertx/vert.x/blob/3.9/src/test/java/io/vertx/core/http/HttpTest.java#L561

if you want something simpler to use, you can consider using Vert.x WebClient instead that provides an higher level interface than the HTTP client.

PR #3462, opened 2020-06-19
Reduced wrappers and copies on BufferImpl to reduce allocations and CPU usage

franz1981:
Buffer creation is creating unecessary Netty ByteBuf wrappers and performing avoidable copies while encoding into US_ASCII/UTF 8 and other charsets.

### DIFF
--- a/src/main/java/io/vertx/core/buffer/ByteBufUtils.java
+++ b/src/main/java/io/vertx/core/buffer/ByteBufUtils.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2011-2017 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.buffer;
+
+import io.netty.buffer.*;
+import io.netty.util.CharsetUtil;
+import io.netty.util.internal.PlatformDependent;
+
+import java.nio.charset.Charset;
+import java.util.Objects;
+
+import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
+
+public final class ByteBufUtils {
+  private ByteBufUtils() {
+
+  }
+
+  // this wrapper is used to save copying byte[]
+  private static final class UnpooledWrappedByteBuf extends UnpooledHeapByteBuf {
+
+    protected UnpooledWrappedByteBuf(ByteBufAllocator alloc, byte[] initialArray, int maxCapacity) {
+      super(alloc, initialArray, maxCapacity);
+    }
+  }
+
+  public static UnpooledHeapByteBuf unpooledNotInstrumentedHeapByteBuf(int initialCapacity, int maxCapacity) {
+    checkPositiveOrZero(initialCapacity, "initialCapacity");
+    // this save instrumented heap ByteBuf allocation that would save statistics collection
+    return PlatformDependent.hasUnsafe() ?
+      new UnpooledUnsafeHeapByteBuf(UnpooledByteBufAllocator.DEFAULT, initialCapacity, maxCapacity) :
+      new UnpooledHeapByteBuf(UnpooledByteBufAllocator.DEFAULT, initialCapacity, maxCapacity);
+  }
+
+  public static ByteBuf unpooledBufferOf(String str, String enc) {
+    final Charset charset = Charset.forName(Objects.requireNonNull(enc));
+    return unpooledBufferOf(str, charset);
+  }
+
+  public static ByteBuf unpooledBufferOf(String str, Charset charset) {
+    if (charset.equals(CharsetUtil.UTF_8)) {
+      return utf8UnpooledBufferOf(str);
+    }
+    if (charset.equals(CharsetUtil.US_ASCII) || charset.equals(CharsetUtil.ISO_8859_1)) {
+      return usAsciiUnpooledBufferOf(str);
+    }
+    final byte[] bytes = str.getBytes(charset);
+    return new UnpooledWrappedByteBuf(UnpooledByteBufAllocator.DEFAULT, bytes, Integer.MAX_VALUE);
+  }
+
+  public static ByteBuf utf8UnpooledBufferOf(String str) {
+    final int utf8Bytes = ByteBufUtil.utf8Bytes(str);
+    final UnpooledHeapByteBuf buffer = unpooledNotInstrumentedHeapByteBuf(utf8Bytes, Integer.MAX_VALUE);
+    ByteBufUtil.reserveAndWriteUtf8(buffer, str, utf8Bytes);
+    return buffer;
+  }
+
+  public static ByteBuf usAsciiUnpooledBufferOf(String str) {
+    final int asciiBytes = str.length();
+    final UnpooledHeapByteBuf buffer = unpooledNotInstrumentedHeapByteBuf(asciiBytes, Integer.MAX_VALUE);
+    ByteBufUtil.writeAscii(buffer, str);
+    return buffer;
+  }
+
+}
--- a/src/main/java/io/vertx/core/buffer/ByteBufUtils.java
+++ b/src/main/java/io/vertx/core/buffer/ByteBufUtils.java
@@ -12,6 +12,7 @@
 package io.vertx.core.buffer;
 
 import io.netty.buffer.*;
+import io.netty.util.AsciiString;
 import io.netty.util.CharsetUtil;
 import io.netty.util.internal.PlatformDependent;
 
@@ -19,6 +20,7 @@ import java.nio.charset.Charset;
 import java.util.Objects;
 
 import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
+import static io.netty.util.internal.StringUtil.isSurrogate;
 
 public final class ByteBufUtils {
   private ByteBufUtils() {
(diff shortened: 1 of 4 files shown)


franz1981:
Note; I've slightly changed the semantic of these 2 methods..
```java
  private Buffer append(String str, Charset charset) {
    buffer.writeCharSequence(str, charset);
    return this;
  }

  private Buffer setBytes(int pos, String str, Charset charset) {
    buffer.setCharSequence(pos, str, charset);
    return this;
  }
```
eg For UTF-8 Strings they estimate the length of the encoded buffer by using an O(1) method called `ByteBufUtil.utf8MaxBytes` while previously it was using a temporary `byte[]` of the exact size.
I can manually re-implement them to preserve the original behaviour by using a method I've introduced myself on Netty ie `ByteBufUtil.utf8Bytes` that is being used instead while creating `BufferImpl` instead. Let me know what works better for you :)

src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
```diff
@@ -367,20 +368,23 @@ public void end(Buffer chunk, Handler<AsyncResult<Void>> handler) {
   }
 
   private void end(Buffer chunk, ChannelPromise promise) {
+    end(chunk.getByteBuf(), promise);
+  }
+
+  private void end(ByteBuf chunk, ChannelPromise promise) {
```
vietj, line 374: can we avoid adding a new method ? that decrease the chances of inlining

vietj:
did you have the chance to measure the impact of this in a benchmark ?

src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
franz1981: This method should be trivially inlineable IIRC so it shouldn't reduce the inlining budget IIRC: I'll check better to be sure and will create a new whole method if needed

franz1981:
@vietj 
Let me check if I can find any or I will write one..do you have any non-micro in mind?

src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
franz1981: I'm still checking here http://hg.openjdk.java.net/jdk8/jdk8/hotspot/file/87ee5ee27509/src/share/vm/opto/bytecodeInfo.cpp#l331

franz1981:
@vietj Some initial microbench shows an advantage while creating `BufferImpl` UTF-8 encoded until 16 chars String, I should look better into it to undestand if there is some linear cost in place (other then `ByteBufUtil.utf8Bytes(str)`) 

```
Benchmark                                        (length)   Mode  Cnt      Score      Error   Units
BufferCreationBenchmark.utf8Buffer                      1  thrpt   10  18749.937 ± 7483.458  ops/ms
BufferCreationBenchmark.utf8Buffer                      4  thrpt   10  16416.950 ±  722.478  ops/ms
BufferCreationBenchmark.utf8Buffer                      8  thrpt   10  12288.932 ±  332.795  ops/ms
BufferCreationBenchmark.utf8Buffer                     16  thrpt   10   7501.509 ±  677.007  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         1  thrpt   10  17496.220 ± 4161.562  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         4  thrpt   10  13901.585 ±  616.267  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         8  thrpt   10  10196.233 ±  451.351  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer        16  thrpt   10   6342.962 ±  457.109  ops/ms

Benchmark                                        (length)   Mode  Cnt     Score      Error   Units
BufferCreationBenchmark.utf8Buffer                      1  thrpt   10  8963.888 ±  380.705  ops/ms
BufferCreationBenchmark.utf8Buffer                      4  thrpt   10  7645.414 ± 2518.473  ops/ms
BufferCreationBenchmark.utf8Buffer                      8  thrpt   10  8095.727 ±  190.704  ops/ms
BufferCreationBenchmark.utf8Buffer                     16  thrpt   10  8081.709 ±  612.367  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         1  thrpt   10  8740.260 ±  431.319  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         4  thrpt   10  8667.429 ±  520.054  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         8  thrpt   10  8560.394 ±  408.692  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer        16  thrpt   10  8036.151 ±  569.181  ops/ms
```

While the benchmark is
```java
/*
 * Copyright (c) 2011-2018 Contributors to the Eclipse Foundation
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License 2.0 which is available at
 * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
 * which is available at https://www.apache.org/licenses/LICENSE-2.0.
 *
 * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
 */

package io.vertx.benchmarks;

import io.netty.util.CharsetUtil;
import org.openjdk.jmh.annotations.*;

@State(Scope.Benchmark)
public class BufferCreationBenchmark extends BenchmarkBase {

  @Param({"1", "4", "8", "16"})
  private int length;
  private String endString;

  @Setup
  public void setup() {
    final StringBuilder builder = new StringBuilder(length);
    for (int i = 0; i < length; i++) {
      // single ASCII char
      builder.append('a');
    }
    endString = builder.toString();
  }

  @Benchmark
  public Buffer utf8Buffer() {
    return Buffer.buffer(endString);
  }

  @Benchmark
  public Buffer utf8CharsetLookupBuffer() {
    return Buffer.buffer(endString, CharsetUtil.UTF_8.name());
  }
}
```

franz1981:
I've had some fun by reading the ASM for this PR changes and I've spotted something weird: I see virtual calls for `ByteBuf's _setByte` and `String::charAt` so probably there is some weird/unexpected polymorphic behaviour 
```
....[Hottest Methods (after inlining)]..............................................................
 43.47%   36.98%         C2, level 4  java.lang.String::charAt, version 42 
 31.22%   35.27%         C2, level 4  io.vertx.benchmarks.generated.BufferCreationBenchmark_utf8CharsetLookupBuffer_jmhTest::utf8CharsetLookupBuffer_thrpt_jmhStub, version 650 
 14.16%   17.71%         C2, level 4  io.netty.buffer.UnpooledUnsafeHeapByteBuf::_setByte, version 547 
```

franz1981:
I've modified the bench adding
```java
  @Benchmark
  public ByteBuf utf8ByteBuffer() {
    return ByteBufUtils.utf8UnpooledBufferOf(endString);
  }
```
and I see that there is defenetly something weird going on with ByteBuf and String polymorphism:

```
Benchmark                                             (length)   Mode  Cnt      Score    Error   Units
BufferCreationBenchmark.utf8Buffer                          64  thrpt   20   2035.418 ± 20.380  ops/ms
BufferCreationBenchmark.utf8Buffer:·asm                     64  thrpt             NaN              ---
BufferCreationBenchmark.utf8ByteBuffer                      64  thrpt   20  11051.786 ± 47.059  ops/ms
BufferCreationBenchmark.utf8ByteBuffer:·asm                 64  thrpt             NaN              ---
BufferCreationBenchmark.utf8CharsetLookupBuffer             64  thrpt   20   1843.157 ± 27.796  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer:·asm        64  thrpt             NaN              ---
```

franz1981:
After some investigation it seems that we're hitting the -XX:MaxInlineLevel barrier for `String` and `ByteBuf` calls here: rasing `-XX:MaxInlineLevel` to `15` changes the numbers:
```
Benchmark                                             (length)   Mode  Cnt      Score     Error   Units
BufferCreationBenchmark.utf8Buffer                          64  thrpt   20   8963.965 ± 702.374  ops/ms
BufferCreationBenchmark.utf8Buffer:·asm                     64  thrpt             NaN               ---
BufferCreationBenchmark.utf8ByteBuffer                      64  thrpt   20  11189.376 ± 146.315  ops/ms
BufferCreationBenchmark.utf8ByteBuffer:·asm                 64  thrpt             NaN               ---
BufferCreationBenchmark.utf8CharsetLookupBuffer             64  thrpt   20  10457.911 ± 450.175  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer:·asm        64  thrpt             NaN               ---
```
Let's see if the code on `BufferImpl` can be changed to bring the level of performance to be the same of using `ByteBufUtils`...

franz1981:
I've pushed a further change (although I'm not proud of it) that's handrolling a `byte[]` version for both ASCII and UTF-8 encoding cases based on the `ByteBuf` Netty versions; results shows that it performs on par (or better) then https://github.com/eclipse-vertx/vert.x/pull/3462#issuecomment-647038310 with the bumped up `-XX:MaxInlineLevel ` but with no need to tune it:
```
Benchmark                                        (length)   Mode  Cnt      Score      Error   Units
BufferCreationBenchmark.utf8Buffer                      1  thrpt   10  26233.958 ±  128.062  ops/ms
BufferCreationBenchmark.utf8Buffer                      4  thrpt   10  18045.952 ± 3899.802  ops/ms
BufferCreationBenchmark.utf8Buffer                      8  thrpt   10  21226.808 ± 1256.438  ops/ms
BufferCreationBenchmark.utf8Buffer                     16  thrpt   10  18618.598 ±  319.346  ops/ms
BufferCreationBenchmark.utf8Buffer                     64  thrpt   10  11737.028 ±  304.477  ops/ms
BufferCreationBenchmark.utf8ByteBuffer                  1  thrpt   10  26906.380 ± 1138.791  ops/ms
BufferCreationBenchmark.utf8ByteBuffer                  4  thrpt   10  25164.225 ±  333.853  ops/ms
BufferCreationBenchmark.utf8ByteBuffer                  8  thrpt   10  22584.024 ± 1534.954  ops/ms
BufferCreationBenchmark.utf8ByteBuffer                 16  thrpt   10  19857.723 ±  589.767  ops/ms
BufferCreationBenchmark.utf8ByteBuffer                 64  thrpt   10  12783.522 ±   85.275  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         1  thrpt   10  20164.095 ±   71.150  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         4  thrpt   10  18436.163 ±   81.198  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer         8  thrpt   10  15541.589 ±   52.038  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer        16  thrpt   10  13653.665 ±   42.570  ops/ms
BufferCreationBenchmark.utf8CharsetLookupBuffer        64  thrpt   10   9764.406 ±  149.438  ops/ms
```

src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java
franz1981: After the last commits on this PR I can say that the inlining budget secret with trivially inlineable callee too :) good suggestion!

franz1981:
Probably I will experiment with this one a bit more on some Netty PRs..

src/main/java/io/vertx/core/buffer/impl/BufferImpl.java
```diff
@@ -469,7 +469,7 @@ public Buffer slice(int start, int end) {
   public ByteBuf getByteBuf() {
     // Return a duplicate so the Buffer can be written multiple times.
     // See #648
-    return buffer.duplicate();
+    return Unpooled.unreleasableBuffer(buffer.duplicate());
```
vietj, line 472: funny I just did that in master

src/main/java/io/vertx/core/buffer/ByteBufUtils.java
```diff
@@ -0,0 +1,136 @@
...
+  private static final class UnpooledWrappedByteBuf extends UnpooledHeapByteBuf {
+
+    protected UnpooledWrappedByteBuf(ByteBufAllocator alloc, byte[] initialArray, int maxCapacity) {
+      super(alloc, initialArray, maxCapacity);
+    }
+  }
+
+  public static UnpooledHeapByteBuf unpooledNotInstrumentedHeapByteBuf(int initialCapacity, int maxCapacity) {
+    checkPositiveOrZero(initialCapacity, "initialCapacity");
+    // this save instrumented heap ByteBuf allocation that would save statistics collection
```
vietj, line 40: why would we avoid statistics collection ?

src/main/java/io/vertx/core/buffer/ByteBufUtils.java
```diff
@@ -0,0 +1,136 @@
...
+    return PlatformDependent.hasUnsafe() ?
+      new UnpooledUnsafeHeapByteBuf(UnpooledByteBufAllocator.DEFAULT, initialCapacity, maxCapacity) :
+      new UnpooledHeapByteBuf(UnpooledByteBufAllocator.DEFAULT, initialCapacity, maxCapacity);
+  }
+
+  public static ByteBuf unpooledBufferOf(String str, String enc) {
+    final Charset charset = Charset.forName(Objects.requireNonNull(enc));
```
vietj, line 47: can we reuse this algorithm from another place ?

src/main/java/io/vertx/core/buffer/ByteBufUtils.java
```diff
@@ -0,0 +1,136 @@
...
+import io.netty.util.CharsetUtil;
+import io.netty.util.internal.PlatformDependent;
+
+import java.nio.charset.Charset;
+import java.util.Objects;
+
+import static io.netty.util.internal.ObjectUtil.checkPositiveOrZero;
+import static io.netty.util.internal.StringUtil.isSurrogate;
+
+public final class ByteBufUtils {
```
vietj, line 25: I can see a lot of interesting algorithms here, it would be better if we could reuse them from Netty directly

src/main/java/io/vertx/core/buffer/ByteBufUtils.java
franz1981: On massive scale benchs and many small sized temporary buffers, the statistics collection can cost more then the buffer allocation

PR #3475, opened 2020-07-02
Explain WriteStream.write internal queue and Handler/Future

julianladisch:
The javadoc of write(T, Handler) is unclear whether the Handler
is called after putting T into the internal queue or after
the actual write of T.

Solution:

Copy the javadoc of write(T) to write(T, Handler) to make it clear.

Also mention it in the Future `@return` and Handler `@param` javadoc.


vietj:
I did a general refresh of the documentation @julianladisch https://github.com/eclipse-vertx/vert.x/commit/6124eb7d4a3224ea1991c406fce4db5579d29c7c

Issue #3457, opened 2020-06-18
Consider failing when number read from JsonObject would overflow


gabor-noenv:
### Version

3.9.1

### Context

When reading a big number from a JsonObject the result may overflow.
I expect it to fail when it happens. Currently it succeeds silently with wrong value returned. 
I believe this is dangerous for the majority of the use-cases. Same applies for JsonArray.
 
### Do you have a reproducer?

```
var json = new JsonObject("{\"amount\": 21474836470}");
long a1 = json.getLong("amount");    // 21474836470
int  a2 = json.getInteger("amount"); // -10

var array = new JsonArray("[21474836470]");
int a3 = array.getInteger(0); // -10
```

vietj:
@pmlopes what's your take on this ?

pmlopes:
@gabor-noenv this is the JVM behavior. For example:

```java
    long l = 21474836470L;
    System.out.println(l);
    int i = (int) l;
    System.out.println(i);
```

Will also print:

```
21474836470
-10
```

Without any exception being thrown. Same goes with non primitive types, for example a `Number` instance for `21474836470` when executing:

```java
Number n = new Long(21474836470);
System.out.println(n.intValue());
```

Will again print `-10`. I am of the opinion that users should know what kind of data their input is, and pick the right getter for it, plus we follow what is _"normal"_ on the JVM.

vietj:
thanks @pmlopes that makes sense.

I think we could provide a `getNumber` method that returns a `Number` that does not make any assumption on the implementation allowing somebody to get the `Number` and be in charge of the conversion.

vietj:
Instead we provide accessors for `java.lang.Number`

PR #3480, opened 2020-07-06
Add getNumber() to JSON types

pmlopes:

Motivation:

Solve #3457

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -118,11 +118,11 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
   }
 
   /**
-   * Get the Integer at position {@code pos} in the array,
+   * Get the Number at position {@code pos} in the array,
    *
    * @param pos the position in the array
    * @return the Number, or null if a null value present
-   * @throws java.lang.ClassCastException if the value cannot be converted to Number
+   * @throws java.lang.ClassCastException if the value is not a Number
    */
   public Number getNumber(int pos) {
     return (Number) list.get(pos);
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -124,6 +124,17 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
    * @return the Integer, or null if a null value present
    * @throws java.lang.ClassCastException if the value cannot be converted to Integer
    */
+  public Number getNumber(int pos) {
+    return (Number) list.get(pos);
+  }
+
+  /**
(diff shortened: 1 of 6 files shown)


src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -117,6 +117,17 @@ public String getString(int pos) {
     }
   }
 
+  /**
+   * Get the Integer at position {@code pos} in the array,
+   *
+   * @param pos the position in the array
+   * @return the Integer, or null if a null value present
+   * @throws java.lang.ClassCastException if the value cannot be converted to Integer
```
julianladisch, line 125: Integer should be replaced by Number in the javadoc

src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -117,6 +117,17 @@ public String getString(int pos) {
     }
   }
 
+  /**
+   * Get the Integer at position {@code pos} in the array,
```
julianladisch, line 121:
```suggestion
   * Get the Number at position {@code pos} in the array,
```

src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -117,6 +117,17 @@ public String getString(int pos) {
...
+   *
+   * @param pos the position in the array
+   * @return the Number, or null if a null value present
+   * @throws java.lang.ClassCastException if the value cannot be converted to Number
```
julianladisch, line 125:
```suggestion
   * @throws java.lang.ClassCastException if the value is not a Number
```
There is no conversion, only casting.

src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -145,6 +145,18 @@ public String getString(String key) {
     }
   }
 
+  /**
+   * Get the Number value with the specified key
+   *
+   * @param key the key to return the value for
+   * @return the value or null if no value for that key
+   * @throws java.lang.ClassCastException if the value is not an Number
```
julianladisch, line 153:
```suggestion
   * @throws java.lang.ClassCastException if the value is not a Number
```

src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -394,6 +406,22 @@ public String getString(String key, String def) {
...
   }
 
+  /**
+   * Like {@link #getNumber(String)} but specifying a default value to return if there is no entry.
+   *
+   * @param key the key to lookup
+   * @param def the default value to use if the entry is not present
+   * @return the value or {@code def} if no entry present
+   */
+  public Number getInteger(String key, Number def) {
```
julianladisch, line 416:
```suggestion
  public Number getNumber(String key, Number def) {
```

PR #3491, opened 2020-07-08
ClusteredEventBus config improvements

### LINKED ISSUE #1529: Vertx eventbus cluster issue - version 3.x (cluster host listen on all interfaces)
Follow up to the thread https://groups.google.com/forum/#!topic/vertx/Lh7uc4s_pOg

Please note that i have gone through few of posts like
https://groups.google.com/forum/#!searchin/vertx/cluster$20eventbus/vertx/RisJt07-_IE/Cij3VeALCQAJ
and
https://groups.google.com/forum/#!searchin/vertx/cluster$20eventbus/vertx/UWS0WAwHO6k/NQfkErPVAwAJ

and few more. So this issue is not related to port being blocked by firewall...

Now coming to the issue if we use vertx as API and try to use it in cluster creating vertx object using
Vertx.clusteredVertx(options, handler)

Then if the cluster host is not set on the VertxOptions it defaults to "localhost" even if the cluster is enabled. Thus the only interface its listening on is localhost and two separate machines cannot send/publish messages to each other over eventbus. 

If we want to listen on all interfaces and set clusterhost as "0.0.0.0" then too the eventbus send/publish is not working. The reason being in "ClusteredEventBus"
the "serrverId" is created with combination of host, port. On both the machines it would end up with same values ("0.0.0.0" : <port>)

With this configuration when we do send/publish a message in ClusteredEventBus -> sendToSubs we have a check if the serverID is same as current nodes serverid 

if (!sid.equals(serverID)) {  //We don't send to this node

This is true for as all nodes which have been configured with cluster host as "0.0.0.0". So the message would never be propagated to the remote host/nodes

May be we can use a unique ID auto generated or some other mechanism to get the "serverid" to keep it unique across a cluster configured to listen on all interfaces.

### LINKED ISSUE #2696: Usability: clustered event bus defaults
Currently, there are different processes for configuring clustered eventbus.

- the cluster manager configuration file (eg `cluster.xml` for HZ)
- the `ClusteredEventBusOptions`
- the `Launcher` CLI args (`cluster-host`, `cluster-port`)

When running a Vert.x app from the CLI, or in an executable jar with the `Launcher`, we try to determine a host address if not provided via argument. Then the host/port are set in the `ClusteredEventBusOptions`.

When it starts the `ClusteredEventBus` uses values from options.

There are a number of issues with this process:
- the address chosen by the cluster manager is not always consistent with the address chosen by the CLI
- it's difficult to configure for beginners: users must update the cluster manager configuration file and provide CLI arguments or EB options
- EB options defaults to localhost and when embedding Vert.x, users often forget to set cluster host

For Vert.x 4 we must think about better processes, for a better user experience and less questions/issues posted of course :)

Not all issues need to be solved with code changes: a better documentation about this process can help.

tsegismont:
Resolves #2696, fixes #1529

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -449,7 +449,11 @@ Add an enabled cipher suite, appended to the ordered suites.
 Sets the list of enabled SSL/TLS protocols.
 +++
 |[[host]]`@host`|`String`|+++
-Sets the host.
+Sets the host. Defaults to <code>null</code>.
+ <p>
+ When the clustered eventbus starts, it tries to bind to the provided <code>host</code>.
+ If <code>host</code> is <code>null</code>, then it tries to bind to the same host as the underlying cluster manager.
+ As a last resort, an address will be picked among the available network interfaces.
 +++
 |[[idleTimeout]]`@idleTimeout`|`Number (int)`|+++
 Set the idle timeout, default time unit is seconds. Zero means don't timeout.
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -842,8 +842,9 @@ one instance. If omitted a single instance will be deployed.
  cluster communication with other Vert.x instances. Default is `0` - which means '_choose a free random port_'. You
  don't usually need to specify this parameter unless you really need to bind to a specific port.
  * `-cluster-host` - If the `cluster` option has also been specified then this determines which host address will be
- bound for cluster communication with other Vert.x instances. By default it will try and pick one from the available
- interfaces. If you have more than one interface and you want to use a specific one, specify it here.
+ bound for cluster communication with other Vert.x instances. If not set, the clustered eventbus tries to bind to the
+ same host as the underlying cluster manager. As a last resort, an address will be picked among the available network
+ interfaces.
  * `-cluster-public-port` - If the `cluster` option has also been specified then this determines which port will be advertised for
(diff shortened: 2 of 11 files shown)


tsegismont:
Do not merge yet, the PR still misses doc updates and new tests.

tsegismont:
@vietj @jponge can you please review? :bow: 

@jponge I'm particularly interested in how this could affect/improve your book code samples.

jponge:
I don't see any public API breaking change here

tsegismont:
@jponge a couple of duplicated `EventBusOptions` defaults have been removed https://github.com/eclipse-vertx/vert.x/pull/3491/files#diff-2212e239df92541d100a82bc3899cb48L70-L79

jponge:
Ok I need to check something for chapter 13 then

jponge:
Ok I tested and if snapshots resolution did not fail me it works. I did a Gradle build scan and it looks like I did use the `vertx-core` build I did locally with your PR.

tsegismont:
Thank you @jponge

tsegismont:
@vietj  test added and doc updated, the PR is ready for final review/merge. Thanks

jponge:
@tsegismont I'll re-run when merged and artefacts published to OSS snapshots just to make sure. I had to use Skaffold with Gradle to test and could not switch it to Maven for that example.

jponge:
Perhaps I forgot but when I tested it worked

tsegismont:
Merged 7201acd94

PR #3494, opened 2020-07-09
If port 0 is selected only select a single port

### LINKED ISSUE #3493: Wildcard port (i.e. port 0) should be shared when created by several instances of the same Verticle
The full details are in this Quarkus issue, but as far as I can see this is only vert.x code so I don't think the Quarkus part is relevant: https://github.com/quarkusio/quarkus/issues/9890

If you set the HTTP port to 0 the every IO thread will open its own HTTP port, but only the last opened one will be reported as the actual port. For HTTPS this causes other problems as the binding is relativity expensive and happens in a sync block, so if you have lots of IO threads they all wait at the sync block and the blocked thread detector prints a warning.

stuartwdouglas:
Current if a wildcard port is select each IO thread will
bind to its own port. This changes things so that all IO 
threads bind to a single port when deployed from the
same verticle. 

This means that wildcard binding outside a verticle will
work the way they currently do, however if a wildcard binding
is done from a verticle with numInstance >1 then all the
instances will share the wildcard port.

Fixes #3493

Signed-off-by: Stuart Douglas <sdouglas@redhat.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -19,15 +19,31 @@ import io.netty.channel.group.ChannelGroupFuture;
 import io.netty.channel.group.DefaultChannelGroup;
 import io.netty.util.concurrent.GenericFutureListener;
 import io.netty.util.concurrent.GlobalEventExecutor;
-import io.vertx.core.*;
+import io.vertx.core.AsyncResult;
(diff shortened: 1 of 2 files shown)


cescoffier:
This would need to be backported to the 3.9 branch.

stuartwdouglas:
I opened it against 3.9, it needs to be forward ported to master.

vietj:
I ported it to master

Issue #3487, opened 2020-07-07
Null pointer exception in MultiMap.set


adamdickmeiss:
While upgrading Okapi to latest Vert.x core we encounter null pointer exceptions in io.vertx.core.http.impl.headers.HeadersMultiMap$MapEntry.remove(HeadersMultiMap.java:494)

Using  commit 8377be65e4cfa25e809eab3860186fe2dc590127 (HEAD -> master, origin/master, origin/HEAD)
Author: Caleb Browne <caleb@ploceus.co.uk>
Date:   Tue Jul 7 00:02:22 2020 +0100

Here's a test case that demonstrates the issue .. can be added to VertxHttpHeadersTest.java.
```
  @Test
  public void test1() {
    MultiMap mmap = newMultiMap();
    mmap.set("k", "v");
    for (int i = 0; i < 500; i++) {
      mmap.add(String.valueOf(i), "foo");
    }
    mmap.remove("k");
    mmap.set("k", "w");
    assertEquals("w", mmap.get("k"));
  }
```

We get a null pointer exception in last mmap.set, due to MapEntry.before == null.
The for loop is added to get a hash collision for value of "k".

adamdickmeiss:
Looks like next handling is not really updated for the removal case..

Something like this works better for me:
```
private void remove0(int h, int i, CharSequence name) {
    HeadersMultiMap.MapEntry e = entries[i];
    MapEntry prev = null;
    while (e != null) {
      MapEntry next = e.next;
      CharSequence key = e.key;
      if (e.hash == h && (name == key || AsciiString.contentEqualsIgnoreCase(name, key))) {
        if (prev == null) {
          entries[i] = next;
        } else {
          prev.next = next;
        }
        e.remove();
      }
      prev = e;
      e = next;
    }
  }
```

PR #3496, opened 2020-07-09
Yet another issue related to #3487

adamdickmeiss:
The prev(ous) pointer stays the same when removing.

Signed-off-by: Adam Dickmeiss <adam@indexdata.dk>


vietj:
thanks, I rewrote the logic of this remove0 method recently (because it was complex) and I broke stuff that was not tested.

I did that of course only in master and would not have done that in 3.x

PR #3497, opened 2020-07-10
Context.put and .putLocal should remove if value is null

julianladisch:
Context.put(key, value) and Context.putLocal(key, value) are implemented
with a ConcurrentHashMap that throws NullPointerException if key or value
is null.

For usability remove the key when putting a null value. This is almost
always what the caller wants.

Add javadoc when put, get, remove, putLocal, getLocal, removeLocal
throw NullPointerException.

Signed-off-by: Julian Ladisch <eclipse.org-rtn@ladisch.de>

### DIFF
--- a/src/main/java/io/vertx/core/Context.java
+++ b/src/main/java/io/vertx/core/Context.java
@@ -195,6 +195,7 @@ public interface Context {
    * @param key  the key of the data
    * @param <T>  the type of the data
    * @return the data
+   * @throws NullPointerException  if the specified key is null
    */
   <T> T get(String key);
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/Context.java
```diff
@@ -204,7 +205,8 @@ static boolean isOnVertxThread() {
    * This can be used to share data between different handlers that share a context
    *
    * @param key  the key of the data
-   * @param value  the data
+   * @param value  the data, or null to remove the key
```
vietj, line 208: {@code null} instead

julianladisch: Done.

PR #3498, opened 2020-07-14
HTTP client improvements

### LINKED ISSUE #3500: HTTP client eager connect
The HTTP client design creates HTTP requests before they are actually connected which means they can be used before a request is connected or obtained from the pool. This design forces to buffer request data before leading to extra complexity in the code and race bugs.

Removing this might seem like removing some usability in the client, the impact is fine because most users seeking for usability are using the Web client that provides a different model where the stream is actually managed by the client.

vietj:
fixes #3500 and #3501

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientPush.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientPush.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2020 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
--- a/src/main/java/io/vertx/core/http/impl/HttpClientPush.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientPush.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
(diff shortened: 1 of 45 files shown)


vietj:
Note this PR will require to update a few projects in the stack because of API changes

src/main/java/io/vertx/core/http/impl/HttpClientPush.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
```
jponge, line 2: 2020 😉

src/test/java/io/vertx/core/spi/tracing/HttpTracerTestBase.java
```diff
@@ -156,8 +160,8 @@ public Object sendRequest(Context context, Object request, String operation, BiC
       public void receiveResponse(Context context, Object response, Object payload, Throwable failure, TagExtractor tagExtractor) {
         assertSame(val, context.getLocal(key));
         assertTrue(context.removeLocal(key));
-        assertNotNull(response);
-        assertTrue(response instanceof HttpClientResponse);
+//        assertNotNull(response);
```
tsegismont, line 163: Update or remove?

src/main/java/io/vertx/core/http/impl/VertxHttp2Stream.java
```diff
@@ -210,10 +212,20 @@ final void writeData(ByteBuf chunk, boolean end, Handler<AsyncResult<Void>> hand
...
+      chunk = Unpooled.EMPTY_BUFFER;
+    } else {
+      chunk = buf;
+    }
+    if (chunk != null) {
+      bytesWritten += chunk.readableBytes();
+      FutureListener<Void> promise = handler == null ? null : context.promise(handler);
+      conn.handler.writeData(stream, chunk, end, promise);
+    } else {
+      // todo ??????????
```
tsegismont, line 227: todo, fail ?

vietj:
I think this is good for merging now - I will merge this soon and update the rest of the stack with fixes.

PR #3506, opened 2020-07-20
Deprecate Message replyAddress method

### LINKED ISSUE #2704: Deprecate Message replyAddress method
The `replyAddress` cannot be used in both standalone and clustered mode. When clustered, the remote recipient cannot use the reply address because it's not registered in cluster manager.

Reply address is an internal detail of implementation.

tsegismont:


vietj:
while do we need to remove this method ?

tsegismont:
@vietj it used to be advertised in the docs as an alternative way to reply to a message instead of using the `Message#reply` method. It became broken when reply implementation was modified to no longer register reply handlers on the cluster manager and use the sender node id instead. The docs has been updated later (after some bug reports) to no longer promote usage of `replyAddress`.

In other words, this method now exposes an internal implementation detail and I can't see what users can do with it.

vietj:
In this case we should deprecate this in 3.x and remove it in 4.0 instead.

unless it breaks @jponge V.I.A

jponge:
Looks good to me

tsegismont:
Closing, will reopen the same PR for 3.9

PR #3511, opened 2020-07-22
Fix cache dir path, atomically create dirs and set ownership

### LINKED ISSUE #3510: Proper defaults for cache directory
The default cache directory has two issues regarding system integration on Linux. 

Issue 1: It falls back to /tmp/vertx-cache, which is absolutely in the correct directory for temp files, but being solely static binds the directory to the user being first on the system to start up any vert.x based Java. Every follow-up user is being denied write-access (as is policy and correct) and greeted with a cryptic message (which does not contain the directory having the issue). 
Issue 2: This even creates a persistent problem, since vert.x seems to not care to remove the directory when its process is shutting down.

My proposal is as follows:
a) using a dynamic & unique component to the name used, like /tmp/vertx-cache-$(md5($USER$PID$TIME))
b) specifying the directory in the error message, if there's an access issue and
c) removing it in the end, for /tmp is not a trashheap, even if it is cleaned during system boot

Why $USER, $PID and $TIME?
Systemd is able to start processes simultaneously, so only using $USER and $TIME is likely to produce hash collisions. $PID is guaranteed to be different, so in theory it would be sufficient to just add the PID, but then forgetting to delete the directory during termination (or being killed with -9 and not being able to) might result in future processes reusing the PID and have the problem again. Adding $USER would mitigate that or at least only lead to the new process finding ancient stuff, but since that might not be desireable, $TIME adds that extra safety clip on top. Hashing the whole shebang is just to not blatantly giving everyone the information which process is using the directory.

julianladisch:
On a multi-user system the first user that runs Vert.x
creates /tmp/vertx-cache/ and owns this directory and
prevents all other users from using it.

Solution:

Replace
/tmp/vertx-cache/file-cache-xxxxx/
by
/tmp/vertx-cache-xxxxx/

To prevent any race conditions use Files.createTempDirectory.

The "vertx.cacheDirBase" configuration option works as before: If
the directory doesn't exist it is created. Files.createDirectories
is used to prevent race conditions.

For security the directory permissions are restricted to the user
atomically when they are created.

Resolves #3510

### DIFF
--- a/src/main/java/io/vertx/core/file/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/file/impl/FileResolver.java
@@ -19,13 +19,17 @@ import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.net.URL;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.FileAttribute;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
 import java.util.Enumeration;
-import java.util.UUID;
+import java.util.Set;
 import java.util.function.IntPredicate;
 import java.util.regex.Pattern;
 import java.util.zip.ZipEntry;
@@ -92,7 +96,7 @@ public class FileResolver {
   private Thread shutdownHook;
   private final boolean enableCaching;
   private final boolean enableCpResolving;
-  private final String fileCacheDir;
+  private final String cacheDirBaseName;
 
   public FileResolver() {
     this(new FileSystemOptions());
@@ -101,7 +105,7 @@ public class FileResolver {
   public FileResolver(FileSystemOptions fileSystemOptions) {
     this.enableCaching = fileSystemOptions.isFileCachingEnabled();
     this.enableCpResolving = fileSystemOptions.isClassPathResolvingEnabled();
-    this.fileCacheDir = fileSystemOptions.getFileCacheDir();
+    this.cacheDirBaseName = fileSystemOptions.getFileCacheDir();
 
     String cwdOverride = System.getProperty("vertx.cwd");
     if (cwdOverride != null) {
@@ -110,7 +114,7 @@ public class FileResolver {
       cwd = null;
     }
     if (this.enableCpResolving) {
-      setupCacheDir(UUID.randomUUID().toString());
+      setupCacheDir();
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -375,11 +379,16 @@ private ClassLoader getClassLoader() {
...
-    cacheDir = new File(cacheDirName);
-    if (!cacheDir.mkdirs()) {
-      throw new IllegalStateException("Failed to create cache dir: " + cacheDirName);
+  private synchronized void setupCacheDir() {
+    FileAttribute<Set<PosixFilePermission>> ownerPermissions =
+        PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwx------"));
+
+    try {
+      Path cacheDirBase = new File(cacheDirBaseName).toPath();
+      Files.createDirectories(cacheDirBase, ownerPermissions);
```
tsegismont, line 388: I think we shouldn't try to create the `cacheDirBase` but only verify that it exists. Otherwise we'd fall in the same trap described in #3510

src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -375,11 +379,16 @@ private ClassLoader getClassLoader() {
...
+      cacheDir = Files.createTempDirectory(cacheDirBase, "vertx-cache-", ownerPermissions).toFile();
+    } catch (IOException e) {
+      throw new UncheckedIOException(e.getMessage() + ": " + cacheDirBaseName, e);
```
tsegismont, line 391: With this implementation it will no longer possible to run Vert.x on Windows if CP resolving is enabled.

Let's not deal with permissions here. If users need that level of confidentiality, there are a couple of options:

- use a dedicated `java.io.tmpdir` or `cacheDirBaseName`
- change the file creation mask before starting the JVM

julianladisch: Thanks for catching this. I've added a check to only use PosixFilePermissions if the underlying file system supports posix attributes so that this also works on Windows.
I think that the file creation mask should not be used for this. The main goal of some vertx application might be to do some file tasks that can be configured by the file creation mask.
The vertx-cache has a completely different purpose, many sysops don't even know that it exists. Therefore the vertx-cache should have security by default by restricting to owner permissions. This is best practice, for example [mktemp](https://www.gnu.org/software/coreutils/manual/html_node/mktemp-invocation.html): 

> When creating a file, the resulting file has read and write permissions for the current user, but no permissions for the group or others

src/main/java/io/vertx/core/file/impl/FileResolver.java
julianladisch: There are 3 use cases:
A) `vertx.cacheDirBase` is not set, then `java.io.tmpdir` is used, this is an existing directory, `Files.createDirectories` is a no-op, this works as before.
B) The sysop sets `vertx.cacheDirBase` to an existing directory. `Files.createDirectories` is a no-op, this works as before. It was and will be the sysop's task to set directory permissions if this is to be used by multiple users.
C) The sysop sets `vertx.cacheDirBase` to a non-existing directory. If we don't want to introduce a breaking change vertx should create that directory. This has been working and will work if only a single user starts vertx applications. There are many use cases where this is a reasonable approach. (Restricting directory access to owner as suggested below makes this a secure solution.)
I don't see any advantages not to create the directory; what benefit do we get from this breaking change?

tsegismont:
@vietj can you please take over the review of this PR, I will be afk for 2 weeks

pmlopes:
The current code already sets the right permissions on a single directory. This PR can be safely closed.

PR #3532, opened 2020-08-14
Add route support to http server metrics.

jekkel:
Motivation:

In preparation for https://github.com/vert-x3/vertx-web/issues/1431 and https://github.com/vert-x3/vertx-micrometer-metrics/issues/34

* adds `requestRouted(R requestMetric, String route)` to the http server metric SPI as proposed by @vietj [here](https://github.com/vert-x3/vertx-micrometer-metrics/issues/34#issuecomment-462356893)
* adds a `routed(String route)` to `HttpServerRequest` delegating to metrics to avoid exposing `requestMetric`

Companion PRs:
* https://github.com/vert-x3/vertx-web/pull/1682
* https://github.com/vert-x3/vertx-micrometer-metrics/pull/111
* https://github.com/vert-x3/vertx-dropwizard-metrics/pull/101

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpServerRequest.java
+++ b/src/main/java/io/vertx/core/http/HttpServerRequest.java
@@ -412,4 +412,15 @@ public interface HttpServerRequest extends ReadStream<Buffer> {
    */
   Map<String, Cookie> cookieMap();
 
+	/**
+	 * Marks this request as being routed to the given route. This is purely informational and is
+	 * being provided to metrics.
+	 *
+	 * @param route The route this request has been routed to.
+	 */
+	@Fluent
+	default HttpServerRequest routed(String route) {
+	  return this;
+  }
+
 }
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerStream.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerStream.java
@@ -130,7 +130,7 @@ abstract class Http2ServerStream extends VertxHttp2Stream<Http2ServerConnection>
   }
 
   HttpServerRequest routed(String route) {
-    if (METRICS_ENABLED) {
+    if (METRICS_ENABLED && !response.ended()) {
       HttpServerMetrics metrics = conn.metrics();
       if (metrics != null) {
         metrics.requestRouted(metric, route);
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerStream.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerStream.java
@@ -128,4 +128,14 @@ abstract class Http2ServerStream extends VertxHttp2Stream<Http2ServerConnection>
       }
     }
   }
+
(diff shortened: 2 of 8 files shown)


src/main/java/io/vertx/core/http/impl/Http2ServerStream.java
```diff
@@ -128,4 +128,13 @@ void handleClose() {
       }
     }
   }
+
+  void hasBeenRoutedTo(String routeName) {
```
vietj, line 132: rename it `routed` like the original method.

vietj:
I think we are missing a couple of tests to check that calling manually the `routed` will indeed call the metrics SPI

src/main/java/io/vertx/core/http/impl/Http2ServerStream.java
jekkel: I had it that way but it created an overload issue when not also returning the same type, which is not really possible here. Ideas?

jekkel: Done.

jekkel:
Thanks for reviewing! I refined the code a bit. Does this look better? Any ideas for additional tests?

vietj:
@jekkel I've made another comment

I've seen you implemented this for vert-micrometer-metrics, would you mind also doing the same for Dropwizard metrics ?

src/main/java/io/vertx/core/http/HttpServerRequest.java
```diff
@@ -412,4 +412,15 @@ default int cookieCount() {
...
   Map<String, Cookie> cookieMap();
 
+	/**
+	 * Marks this request as being routed to the given route. This is purely informational and is
+	 * being provided to metrics.
+	 *
+	 * @param routeName The name of the route this request has been routed to.
+	 */
+	@Fluent
+	default HttpServerRequest routed(String routeName) {
```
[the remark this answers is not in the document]
jekkel: gotcha. done.

jekkel:
> @jekkel I've made another comment
> 
> I've seen you implemented this for vert-micrometer-metrics, would you mind also doing the same for Dropwizard metrics ?

Here you go: https://github.com/vert-x3/vertx-dropwizard-metrics/pull/101

The build failed because of an infrastructure issue it seems.

vietj (changes requested):
look almost good, we need to ensure that the metrics SPI is not called when the stream has already ended, i.e when it is too late to report this, at this moment any call to routed should do nothing

jekkel:
Thanks for taking the time, I added a check so that routes only gets reported to metrics SPI if the response is not yet `ended()`. Is this what you had in mind?

vietj:
@jekkel that's what I meant, can you add a test for this as well ?

jekkel:
I already had added some assertions to the route metric test method but just pushed a refined version where this is being tested explicitly. I think this makes clear that it's intended behaviour instead of just a random implementation detail.

PR #3547, opened 2020-08-31
Added a method to CompositeFuture to retrieve all failure causes

slinkydeveloper:
Motivation:

Retrieving failures from a composite future requires some boilerplate, which can be avoided with this `causes` method.

### DIFF
--- a/src/main/java/io/vertx/core/CompositeFuture.java
+++ b/src/main/java/io/vertx/core/CompositeFuture.java
@@ -17,6 +17,8 @@ import io.vertx.core.impl.future.CompositeFutureImpl;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/CompositeFuture.java
```diff
@@ -246,4 +248,14 @@ default CompositeFuture onFailure(Handler<Throwable> handler) {
     }
     return list;
   }
+
+  /**
+   * @return a list of all the eventual failure causes. If no future failed, returns a list of null values.
+   */
+  @GenIgnore
+  default List<Throwable> causes() {
```
vietj, line 256: can you instead do like list() do that creates less objects

slinkydeveloper: done

vietj (approved):

PR #3571, opened 2020-09-10
Local lock improvements

vietj:
Various local improvements.

### DIFF
--- a/src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
+++ b/src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
@@ -30,16 +30,16 @@ public class LocalAsyncLocks {
 
   private class LockWaiter {
 
-    final Context context;
+    final ContextInternal context;
     final String lockName;
     final Promise<Lock> promise;
     final Long timerId;
 
-    LockWaiter(Context context, String lockName, long timeout, Promise<Lock> promise) {
-      this.context = context;
+    LockWaiter(ContextInternal context, String lockName, long timeout, Promise<Lock> promise) {
       this.lockName = lockName;
       this.promise = promise;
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
```diff
@@ -49,12 +47,30 @@
...
+        // Cleanup
+        waitersMap.compute(lockName, (s, list) -> {
+          if (list == null) {
+            // Already removed by nextWaiter
+            return null;
+          } else {
+            int idx = list.indexOf(LockWaiter.this);
+            if (idx == -1) {
+              // Already removed by nextWaiter
+              return null;
```
tsegismont: Shouldn't the list be returned instead?

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
```diff
@@ -49,12 +47,30 @@
...
+            }
+            int size = list.size();
+            ArrayList<LockWaiter> n = new ArrayList<>(size - 1);
```
tsegismont: Could we be more aggressive and check if we can return `null` when there's no other waiter or a `SingletonList` when there's just one?

tsegismont (commented):
LGTM, with a couple questions, thanks @vietj !

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncLocks.java
vietj: yes we need to consider this case as it could be an actual one and leave an empty list in the map, I'll try to address this with a test

tsegismont (approved):
LGTM

vietj:
pushed to fix failing CI that was failing for another reason (master was failing too)

tsegismont:
Still looks good to me! :wink:

PR #3572, opened 2020-09-12
Issues/cookie expires improvement

pmlopes:
Motivation:

Cookie expiration is not required to be executed immediately. Chrome for example will not remove the cookie after parsing the header, so it is recommended to clear the value as well as setting the max age to 0.


vietj (approved):

vietj:
does it need to be in 3.9.3 ?

pmlopes:
I've tagged it as enhancement because this is not fixing a bug but ensuring that any caches are invalidated asap at the user agent side.

vietj:
ok

PR #3574, opened 2020-09-14
Allow to define a timer with no delay (handler is invoked immediately)

Xophe:
Sometimes I want to wait for an arbitrary delay before to perform an action. This delay can be zero, but the method setTimer() does not accept zero, which leads to the following tiedous code :

```
if(delay > 0) {
  cb();
} else {
  vertx.setTimer(delay, l -> cb());
}
```
With this change, I'll be able to invoke setTimer() even if the delay is zero.

Another reason why I need this change is that sometimes I want to break a callback chain. Invoking setTimer(0, l -> cb()) is a good way to have a function invoked from the mainloop instead of from the current code, without introducing any delay. For example, into Javax/Swing you invoke SwingUtilities.invokeLater(runnable).


Xophe:
I'll retry in a better way.

PR #3581, opened 2020-09-23
Fixes Resolver delete race against file resolve

pmlopes:

Motivation:

Cleaning the FileSystem cache happens usually on a shutdown thread, outside the resolver monitor. It also changes the resolver state to set the cache dir object to null.

This has a couple of side effects:

1. A call to resolve while the delete is happening will use the "to be deleted directory" as root.
2. A resolve call during clear cache will see that the `cacheDir` object is `null` which will lead to resolved file to be written to the `${user.dir}`.
3. As `cacheDir` shared the same directory for all instances (by default) a clear for instance A will wipe the cache for instance B

The last item was also reported by https://github.com/eclipse-vertx/vert.x/issues/3510

This PR addresses the issue in the following way:

1. `cacheDir` is `final` and computed at `<ctor>` time.
2. `clearCache` will be synchronized on the `resolver` iif the is `cacheDir`
3. caches are suffixed with an UUID instead of sharing a common directory.

The last item only addresses the uniqueness and independence of caches not the full discussion on https://github.com/eclipse-vertx/vert.x/issues/3510 . As a note the mentioned issue warns about potential collisions, however wikipedia tells us that the change of a collision is 1 in a billion: https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random) . Given that most `unices` can handle a few million processes:

```
cat /proc/sys/kernel/pid_max
4194304
```

I'm not attempting to change the current use of UUIDs + on regular operations caches are deleted on VM shutdown, which means the probability will be even smaller.

### DIFF
--- a/src/main/java/io/vertx/core/file/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/file/impl/FileResolver.java
@@ -399,7 +399,7 @@ public class FileResolver {
       fileCacheDir = fileCacheDir.substring(0, fileCacheDir.length() - File.separator.length());
     }
 
-    // the cachedir will be prefixed a unique id to avoid eavesdroping from other processes
+    // the cacheDir will be suffixed a unique id to avoid eavesdropping from other processes/users
     // also this ensures that if process A deletes cacheDir, it won't affect process B
     String cacheDirName = fileCacheDir + "-" + UUID.randomUUID().toString();
     File cacheDir = new File(cacheDirName);
@@ -409,9 +409,11 @@ public class FileResolver {
     }
     // Add shutdown hook to delete on exit
     shutdownHook = new Thread(() -> {
-      // no-op if cache dir has been set to null
-      if (this.cacheDir == null) {
-        return;
+      synchronized (this) {
+        // no-op if cache dir has been set to null
+        if (this.cacheDir == null) {
+          return;
+        }
       }
 
       final Thread deleteCacheDirThread = new Thread(() -> {
@@ -433,17 +435,14 @@ public class FileResolver {
   }
 
   private void deleteCacheDir() throws IOException {
-    if (cacheDir == null || !cacheDir.exists()) {
-      return;
-    }
-    // save the state before we force a flip
     final File dir;
     synchronized (this) {
       if (cacheDir == null) {
         return;
       }
-      // disable the cache dir
+      // save the state before we force a flip
       dir = cacheDir;
+      // disable the cache dir
       cacheDir = null;
     }
     // threads will only enter here once, as the resolving flag is flipped above
--- a/src/main/java/io/vertx/core/file/impl/FileResolver.java
+++ b/src/main/java/io/vertx/core/file/impl/FileResolver.java
@@ -22,7 +22,6 @@ import java.io.InputStream;
 import java.net.URL;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
-import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.Enumeration;
 import java.util.UUID;
@@ -88,11 +87,10 @@ public class FileResolver {
   private static final Pattern JAR_URL_SEP_PATTERN = Pattern.compile(JAR_URL_SEP);
 
   private final File cwd;
-  private File cacheDir;
+  private final File cacheDir;
   private Thread shutdownHook;
   private final boolean enableCaching;
   private final boolean enableCpResolving;
-  private final String fileCacheDir;
 
   public FileResolver() {
     this(new FileSystemOptions());
@@ -101,7 +99,6 @@ public class FileResolver {
   public FileResolver(FileSystemOptions fileSystemOptions) {
     this.enableCaching = fileSystemOptions.isFileCachingEnabled();
     this.enableCpResolving = fileSystemOptions.isClassPathResolvingEnabled();
-    this.fileCacheDir = fileSystemOptions.getFileCacheDir();
 
     String cwdOverride = System.getProperty("vertx.cwd");
     if (cwdOverride != null) {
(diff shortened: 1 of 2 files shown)


tsegismont (approved):
LGTM

src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -375,9 +370,24 @@ private ClassLoader getClassLoader() {
     return cl;
   }
 
-  private synchronized void setupCacheDir(String id) {
-    String cacheDirName = fileCacheDir + "/file-cache-" + id;
-    cacheDir = new File(cacheDirName);
+  /**
+   * Will prepare the cache directory to be used in the application or return null if classpath resolving is disabled.
+   */
+  private synchronized File setupCacheDir(String fileCacheDir) {
```
vietj, line 376: this is called from the constructor so it does not need to be synchronized

src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -398,18 +408,20 @@ private synchronized void setupCacheDir(String id) {
       }
     });
     Runtime.getRuntime().addShutdownHook(shutdownHook);
+    return cacheDir;
   }
 
   private void deleteCacheDir() throws IOException {
-    Path path;
+    if (cacheDir == null || !cacheDir.exists()) {
```
vietj, line 415: instead here we should have something that will not block other threads when deleting the dir and just return, something like

```java
File dir;
synchronized(this) {
  if (cacheDir == null) {
    return;
  }
  dir = cacheDir;
  cacheDir = null;
}
if (dir.exists()) {
  FileSystemImpl.delete(dir.toPath(), true);
}
```

vietj (commented):
a couple of changes.

pmlopes:
I've applied the "fixes" from the comments and added a few extra checks:

* on close, if the shutdown hook is null, avoid enter a synchronized block.
* the shutdown hook thread, will not spawn a new thread with deadline to delete the cache if a delete is already in progress.
* the resolve file operation will throw ISE if the cache dir is being deleted

As validation tests pass on my side + the issue that was used to identify the problem (vertx-web tests) also pass on my side (no leaks of caches to the CWD or TMP.

src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -236,10 +242,15 @@ private synchronized File unpackFromFileURL(URL url, String fileName, ClassLoade
...
-      for (String file: listing) {
-        String subResource = fileName + "/" + file;
-        URL url2 = getValidClassLoaderResource(cl, subResource);
-        unpackFromFileURL(url2, subResource, cl);
+      if (listing != null) {
+        for (String file: listing) {
+          String subResource = fileName + "/" + file;
+          URL url2 = getValidClassLoaderResource(cl, subResource);
+          if (url2 == null) {
+            throw new VertxException("Invalid resource: " + subResource);
```
vietj, line 250: shouldn't we rather log this instead of throwing ?

src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -375,14 +386,34 @@ private ClassLoader getClassLoader() {
...
+    String cacheDirName = fileCacheDir + "-" + UUID.randomUUID().toString();
+    File cacheDir = new File(cacheDirName);
+
     if (!cacheDir.mkdirs()) {
       throw new IllegalStateException("Failed to create cache dir: " + cacheDirName);
     }
     // Add shutdown hook to delete on exit
     shutdownHook = new Thread(() -> {
+      // no-op if cache dir has been set to null
+      if (this.cacheDir == null) {
```
vietj, line 413: should read under synchronized

src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -398,18 +429,26 @@ private synchronized void setupCacheDir(String id) {
       }
     });
     Runtime.getRuntime().addShutdownHook(shutdownHook);
+    return cacheDir;
   }
 
   private void deleteCacheDir() throws IOException {
-    Path path;
+    if (cacheDir == null || !cacheDir.exists()) {
```
vietj, line 436: this should be done under synchronization

vietj (changes requested):
some changes

src/main/java/io/vertx/core/file/impl/FileResolver.java
pmlopes: Currently we throw npe, without a clear message on which resource caused it. I kept the behavior but you also see which resource triggered it.

pmlopes:
All changes added.

src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -101,29 +100,32 @@ public FileResolver() {
...
    */
   public void close() throws IOException {
-    synchronized (this) {
-      if (shutdownHook != null) {
-        // May throw IllegalStateException if called from other shutdown hook so ignore that
-        try {
-          Runtime.getRuntime().removeShutdownHook(shutdownHook);
-        } catch (IllegalStateException ignore) {
+    // avoid monitor, if already disabled
+    if (shutdownHook != null) {
```
vietj, line 118: you can remove this check as it is not synchronized and only check within the synchronized block.

src/main/java/io/vertx/core/file/impl/FileResolver.java
```diff
@@ -101,29 +100,32 @@ public FileResolver() {
...
+      synchronized (this) {
```
vietj, line 119: what you can do is

```java
Thread hook;
synchronized (this) {
  hook =  shutdownHook;
  shutdownHook = null;
}
if (hook != null) {
  // Remove hook
}
```

vietj: Like you did in `deleteCacheDir `

vietj (approved):

Issue #3406, opened 2020-05-12
Event-bus protocol versioning / scoping


vietj:
Vert.x 4 will ship with a new version of the event-bus wire protocol as well as cluster manager topology usage. We should have a way to isolate / detect invalid protocols.

An approach could be to have each event-bus node declare the protocol it handles when it publish its node in the cluster manager.

A solution leveraging an initial handshake at the wire level would be more difficult to implement because the initial version does not have such handshake. The cluster manager declarative solution would have the benefit to detect invalid nodes earlier too.

tsegismont:
## Current situtation

The format of messages sent from the event bus sender (client) to the recipient (server) is as follows:

```
+ length (int) + protocol version (byte) + payload (many fields) +
```

The server validates the protocol version when transforming a `Buffer` to a `ClusteredMessage` object:

- if the client version is greater than server version, the server throws an `IllegalStateException` (reported to the server context `exceptionHandler`)
- the client is not notified and fails with timeout

The server throws an ISE as well when it misses the message codec used by the client.

## Proposal

For `master` branch and then backport to `3.9` branch:

- create a new `io.vertx.core.eventbus.ReplyFailure`: `INTERNAL_ERROR`
- document that `INTERNAL_ERROR` is a fatal issue and there is no point in retrying to send the message
- server replies with `INTERNAL_ERROR` when protocol version is invalid or the message codec is missing

And then for `master` branch only:

- bump protocol version to `2` (4.0)

## Results

When 3.9.2 and below nodes connect to 3.9.3 and above nodes:

- if protocol version matches (eg 3.9.2 to 3.9.3), nothing happens
- if protocol version does not match (e.g. 3.9.2 to 4.0)
  - server node replies with `INTERNAL_ERROR`
  - client node fails with ISE (`INTERNAL_ERROR` is unknown) on the sender's context

When 3.9.3 and above nodes connect to 3.9.2 and below nodes:
- if protocol version matches (3.9.3 to 3.9.2), nothing happens
- if protocol version does not match (e.g. 4.0 to 3.9.2)
  - server node fails with ISE (reported to the server's context `exceptionHandler`)
  - the sender timeouts

When 3.9.3 and above nodes connect to 3.9.3 and above nodes:
- if protocol version matches (3.9.3 to 3.9.4), nothing happens
- if protocol version does not match (e.g. 3.9.3 to 4.0)
  - server node replies with `INTERNAL_ERROR`
  - sender receives reply failure

[...]

PR #3583, opened 2020-09-28
InternalError eventbus failure type

tsegismont:
This is the first part of the solution for #3406.

It allows to communicate an internal error from the receiver side (e.g. protocol mismatch, unknown codec) instead of failing silently.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/ReplyFailure.java
+++ b/src/main/java/io/vertx/core/eventbus/ReplyFailure.java
@@ -39,7 +39,7 @@ public enum ReplyFailure {
   /**
    * A fatal error occured while delivering the message. Do not retry to send.
    */
-  INTERNAL_ERROR;
+  ERROR;
 
   public static ReplyFailure fromInt(int i) {
     switch (i) {
@@ -50,7 +50,7 @@ public enum ReplyFailure {
       case 2:
         return RECIPIENT_FAILURE;
       case 3:
-        return INTERNAL_ERROR;
+        return ERROR;
       default:
         throw new IllegalStateException("Invalid index " + i);
     }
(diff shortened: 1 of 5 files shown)


tsegismont:
@vietj when this PR is approved, I will merge it and cherry-pick to `3.9` branch. Then I will update the `master` branch to bump the protocol version.

See the [_Proposal_](https://github.com/eclipse-vertx/vert.x/issues/3406#issuecomment-662918425) section of #3406.

vietj:
users might use a switch case on this enum and will be confused by this new addition.

so we need to keep this internal and not use the existing API (`ReplyFailure`) to carry this information.

e.g we can use an interna ReplyException subclass with the relevant information (e.g a specific internal code) and a related codec.

tsegismont:
If we use a `ReplyException`, which failure type value should it hold?

`NO_HANDLERS` and `RECIPIENT_FAILURE` do not sound right at all to me.
`null` is confusing and could break existing apps too.
`TIMEOUT` is confusing but at least it's consistent with what users would get without the change.

In any case, I don't believe a subclass/new codec are needed, using a specific failure code should be enough.

Ideally I'd like to have a way to convey the information to user (so that they can avoid to retry because it's useless with such errors). A specific failure code should do, even though it looks less elegant from my perspective.

vietj:
previously such errors were reported as `IllegalStateException` and as a RECIPIENT_FAILURE ?

tsegismont:
See the [Current situtation](https://github.com/eclipse-vertx/vert.x/issues/3406#issuecomment-662918425) section of #3406.

Previously receiver side would fail completely and the sender side would get a `TIMEOUT`

vietj:
ok that explains what I wanted to know

tsegismont:
Ok. What about the solution then?

vietj:
So I think we should go like you did but instead of `INTERNAL_ERROR` we should call it `ERROR` which is more broad and can convey errors such as networking error or other kind of errors.

tsegismont:
I've renamed the enum value as requested. PTAL

vietj (approved):

Issue #2747, opened 2018-12-04
http server metrics vertx.http.server.bytesSent is insufficient for bandwidth measurements


bckfnn:
The metrics collected by the SPI HttpServerMetrics.bytesWritten(..) is technical correct on average but for large downloads that spans multiple reporting intervals the result is misleading and hard to reason about.

Assuming a 10s reporting interval and a large file that take 60s to download from a vertx HttpServer, the metrics reported is 5 intervals of 0 bytes sent, followed by 1 interval with download activity far beyond physical capacity.

if bytesWritten() was called, not just at the end if the response, but more frequently (say for each 1Mb written) the metrics would be easier to make sense of.

PR #3586, opened 2020-09-30
Change the implementation of HTTP read/write metrics event to report …

### LINKED ISSUE #3585: HTTP streamed read/write event metrics reporting
The implementation of read/write metric events reports a single event with the total length when the stream lifecycle ends. While this is a correct way to proceed, it lacks of accuracy when streaming happens.

We can improve this behavior and instead report a read/write metric event when the accumulated value gets above a threshold (4K). e.g instead of reporting 17K at once, we might instead report (4K, 8K, 4K, 1K).

vietj:
…when the accumulated size gets above 4K. When the request/response ends the remaining bytes value is reported. Metrics implementation should not be affected by this change. - fixes #3585

Fixes also https://github.com/eclipse-vertx/vert.x/issues/2747

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -724,7 +724,6 @@ public class Http1xClientConnection extends Http1xConnectionBase<WebSocketImpl> 
 
   private void handleResponseChunk(Stream stream, ByteBuf chunk) {
     Buffer buff = Buffer.buffer(VertxHandler.safeBuffer(chunk, chctx.alloc()));
-    reportBytesRead(buff.length());
     stream.context.execute(buff, stream::handleChunk);
   }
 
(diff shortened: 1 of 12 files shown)


vietj:
can you have a look @jotak ?

jotak:
Thinking again about it .. I fear it doesn't fly well for everything. In micrometer impl, bytes read/write is a distribution summary, meaning that we're not only recording the throughput, but also the size distribution. Every time the `bytesRead` / `bytesWritten` functions are called, we're recording the given amount, and internally (in micrometer) it's keeping track not only of the total amount but also a counter of calls / distribution summary data.

The reason is that it gives volumetry/size information, in addition to throughput: average/percentiles on req/resp size in bytes. With this change, if I understand correctly, we'll see almost constant size of 4K.

So, it's up to you on which side you think the trade-off is better ; but if we choose to merge this, I think we should change our micrometers implementation as keeping distribution summary for constant 4K sizes is pointless.

jotak:
> With this change, if I understand correctly, we'll see almost constant size of 4K.

(or to be more accurate, that is going to screw up size metrics only when data >4K, all being capped at 4K ; sizes below 4K will still be ok)

vietj:
you mean the size of the entire distribution per request because there was a single write/read callbacks ? because currently for a given HTTP request there will be a single write/read reporting with the entire size.

vietj:
keep in mind that those are per socket reporting too

jotak (approved):
LGTM

PR #3587, opened 2020-10-01
Update HttpServerMetrics to use observability HttpRequest/HttpRespons…

vietj:
…e instead of HttpServerRequest and HttpServerResponse. In addition request/response ended event will now provide byte statistics for both client and server


vietj:
@jotak here is the PR for the changes we discussed

jotak (approved):
lgtm!

PR #3588, opened 2020-10-01
Fix NPE: call metrics callback in block finally

jotak:
Fix NPE, `request` being set to null in block `finally`. Call metrics hook just before.


jotak:
@vietj I think there was still an issue with metrics, here's a fix

vietj:
oh thanks I fixed it as well

PR #3592, opened 2020-10-05
New generic KeyStoreOptions

vietj:


src/test/java/io/vertx/core/http/Http1xTest.java
```diff
@@ -530,7 +530,7 @@ private void checkCopyHttpClientOptions(HttpClientOptions options, HttpClientOpt
     if (copy.getTrustOptions() instanceof PemTrustOptions) {
       assertEquals(((PemTrustOptions) options.getTrustOptions()).getCertValues(), ((PemTrustOptions) copy.getTrustOptions()).getCertValues());
     } else {
-      assertEquals(options.getTrustOptions().toJson(), copy.getTrustOptions().toJson());
+      // assertEquals(options.getTrustOptions().toJson(), copy.getTrustOptions().toJson());
```
pmlopes, line 533: was it just for test something? otherwise we can remove the else, i guess.

src/test/java/io/vertx/core/http/Http1xTest.java
```diff
@@ -828,7 +828,7 @@ private void checkCopyHttpServerOptions(HttpServerOptions options, HttpServerOpt
...
```
pmlopes, line 831: was it just for test something? otherwise we can remove the else, i guess.

src/main/java/io/vertx/core/net/KeyStoreOptions.java
```diff
@@ -0,0 +1,150 @@
...
+ * HttpServerOptions options = HttpServerOptions.httpServerOptions();
+ * options.setKeyStore(new JksOptions().setPath("/mykeystore.jks").setPassword("foo"));
+ * </pre>
+ *
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ * @author <a href="http://tfox.org">Tim Fox</a>
+ */
+public class KeyStoreOptions implements KeyCertOptions, TrustOptions {
+
+  private String type;
```
pmlopes, line 52: Perhaps we could have a default value for this. For example, on `auth` I was doing:

```java
 // Defaults
  private static final String TYPE = KeyStore.getDefaultType();
```

As this picks the runtime JDK default, for example on Java8 is `JKS` on java9 and up `PKCS12` and maybe on other jdk's or runtimes started with custom security config it will be something else, e.g.: `BKS`

pmlopes (commented):
I've a few comments, but nothing really blocking

src/test/java/io/vertx/core/http/Http1xTest.java
vietj: I will downcast to specific class and write new asserts

src/test/java/io/vertx/core/http/Http1xTest.java
vietj: I will downcast to specific class and write new asserts

src/main/java/io/vertx/core/net/KeyStoreOptions.java
vietj: that's dangerous I think

vietj: I mean it is fine to have type null by default and let user set an explicit value, if user does not want to set the value then using JksOptions or PfxOptions is just fine

vietj:
PTAL @pmlopes

pmlopes (approved):

Issue #3593, opened 2020-10-05
Set appropriate permissions for cache directory


vietj:

vietj:
See https://github.com/eclipse-vertx/vert.x/issues/3510

PR #3594, opened 2020-10-08
Issues/3593

pmlopes:
Motivation:

Fixes: #3593

Also addresses comments on: #3510

Cache directories are now guaranteed to be unique. While before the chance of collision was low due to the UUID suffix, this PR will throw ISE if a collision happens.

It also ensures that on posix systems (non windows) cache directories are created with 0700 permission just like the POSIX mkdtemp standard function does.


src/test/java/io/vertx/core/file/FileResolverTestBase.java
```diff
@@ -130,6 +132,19 @@ public void testResolveFileWithSpacesFromClasspath() throws Exception {
     }
   }
 
+  @Test
+  public void testCacheDirIsPosix0700() throws Exception {
+    if (!Utils.isWindows()) {
```
vietj, line 137: instead use Assume.assumeTrue(!Utils.isWindows());

vietj (changes requested):
Minor change for test

PR #3595, opened 2020-10-08
CompositeFuture.allOf, anyOf, joinOf with generics

### LINKED ISSUE #2058: CompositeFuture forces you to use raw types
There are currently three methods in `CompositeFuture` which force you to use raw types:
* `io.vertx.core.CompositeFuture.all(List<Future>)`
* `io.vertx.core.CompositeFuture.any(List<Future>)`
* `io.vertx.core.CompositeFuture.join(List<Future>)`

The causes compiler warnings to appear all over the place:

image

It should be possible to rewrite the method signature to e.g. `io.vertx.core.CompositeFuture.all(List<Future<?>>)`, which solves the raw types warning, but is not compatible with the previous declaration and this would break compatibility.

However there could be be a set of workarounds. One could be simply adding new methods like e.g. `allOf`. Or putting everything in a new interface `TypedCompositeFuture`, which would avoid the collision.

julianladisch:
CompositeFuture.all(List<Future>), CompositeFuture.any(List<Future>) and
CompositeFuture.join(List<Future>) use the raw type of Future because
the code generator for other languages doesn't support wildcard extensions.
IDEs and other code checkers warn about the raw type because the compiler
cannot do type checks.

Jens Reimann (ctron) suggests to create allOf, anyOf and joinOf methods
in #2058. This avoids a breaking change with the existing methods and
allows not to generate the generic methods for other languages.

Resolves #1994, resolves #2058.

### DIFF
--- a/src/main/java/io/vertx/core/CompositeFuture.java
+++ b/src/main/java/io/vertx/core/CompositeFuture.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2020 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -17,8 +17,6 @@ import io.vertx.core.impl.future.CompositeFutureImpl;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 
 /**
  * The composite future wraps a list of {@link Future futures}, it is useful when several futures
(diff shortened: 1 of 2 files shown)


vietj:
for the moment we cannot accept this as we would like this to avoid using `@GenIgnore`

julianladisch:
Are there any other suggestion how the issue can be resolved?

vietj:
do your own wrapper in the meantime I think

vietj:
I spent time trying to have wildcard working in codegen but it did not work out easily, I'll try revisiting this again soon

julianladisch:
Rejected.

Workaround: do your own wrapper.

vietj:
I think it could now accepted in vertx 4 because this method is not codegen
anymore.

PR #3605, opened 2020-10-14
Filesystem classloader resolution

vietj:


vietj:
@pmlopes this is a draft PR for allowing the creation of an (internal) FileSystem that resolve resources using a classloader.

- the `FileResolver` was refactored to encapsulate cache operations in a `FileCache` class so 2 instances of a `FileResolver` will use the same instance to access the Vert.x resolver cache.
- the `FileSystem` was updated to use `FileResolver` instead of `VertxInternal#resolveFile`
- `VertxInternal` now can create a `FileSystem` using a class loader, this shall be used by `StaticHandler` when passed a class loader

PR #3607, opened 2020-10-19
Verticle deployment class loader - fixes #3606

### LINKED ISSUE #3606: Verticle deployment class loader
Verticle deployment has been using the implicit thread context classloader for the `VerticleFactory` and the `io.vertx.core.Context` class loader. This classloader is used indirectly by code that needs to resolve file in some places (e.g `HttpServerResponse#sendFile(String)`). Until now the only possibility to use a specific classloader was the set a temporary thread context classloader in some places (such as deploying a verticle or sending a file).

The `DeploymentOptions` will now hold an optional classloader that will be passed to the `VerticleFactory` and associated with the `Context`. When no classloader is provided, the current thread context classloader is used.

vietj:


vietj:
@pmlopes @cescoffier this is the deployment context classloader PR

vietj:
I will then provide a PR for vertx-web that test the behavior `StaticHandler` using this option.

pk-work (approved):

PR #3608, opened 2020-10-20
List optimisation

stuartwdouglas:
These provide some minor performance improvements to reduce object allocation. In particular:

- Don't allocate a list on a getAll call if there are no headers
- Don't create a StringBuilder if not required, as most paths don't contain a '.' character


src/main/java/io/vertx/core/http/impl/HttpUtils.java
```diff
@@ -126,6 +126,20 @@ public static String normalizePath(String pathname) {
       return "/";
     }
 
+    boolean needsChange = false;
```
vietj, line 129: `boolean needsChange = pathname.charAt(0) != '/'`

stuartwdouglas: I think this way is clearer when you come back and read it as to exactly what is going on. I also don't think it makes any real difference either way.

vietj:
there are actually failing tests in the CI 

```
Failed tests: 
12482  HttpUtilsTest.testMultipleSlashPath1:119 expected:</[]blah> but was:</[/]blah>
12483  HttpUtilsTest.testMultipleSlashPath2:124 expected:</[]blah> but was:</[//]blah>
12484  HttpUtilsTest.testMultipleSlashPath3:129 expected:</foo/[]blah> but was:</foo/[/]blah>
12485  HttpUtilsTest.testMultipleSlashPath4:134 expected:</foo/[blah]/> but was:</foo/[/blah//]/>
12486  HttpUtilsTest.testMultipleTrailingSlashes1:164 expected:</blah/[]> but was:</blah/[/]>
12487  HttpUtilsTest.testMultipleTrailingSlashes2:169 expected:</blah/[]> but was:</blah/[//]>
```

stuartwdouglas:
Fixed

Sanne:
up :)

PR #3611, opened 2020-10-23
Add a tracing policy to enable fine grained tracing per component con…

vietj:
…figuration.

- https://github.com/eclipse-vertx/vertx-sql-client/pull/793
- https://github.com/eclipse-vertx/vertx-tracing/pull/10

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
@@ -22,6 +22,7 @@ import io.vertx.core.impl.logging.Logger;
 import io.vertx.core.impl.logging.LoggerFactory;
 import io.vertx.core.spi.tracing.TagExtractor;
 import io.vertx.core.spi.tracing.VertxTracer;
+import io.vertx.core.tracing.TracingPolicy;
 
 import java.util.Iterator;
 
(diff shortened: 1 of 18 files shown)


vietj:
ping @jotak

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
```diff
@@ -155,10 +156,11 @@ public void next() {
           bus.metrics.messageDelivered(m, message.isLocal());
         }
         if (tracer != null && !src) {
-          message.trace = tracer.receiveRequest(context, message, message.isSend() ? "send" : "publish", message.headers(), MessageTagExtractor.INSTANCE);
+          message.trace = tracer.receiveRequest(context, TracingPolicy.PROPAGATE, message, message.isSend() ? "send" : "publish", message.headers(), MessageTagExtractor.INSTANCE);
```
jotak, line 159: no configurable policies for the event bus? is that on purpose?

src/main/java/io/vertx/core/eventbus/impl/ReplyHandler.java
```diff
@@ -45,7 +45,8 @@
 
   private void trace(Object reply, Throwable failure) {
     VertxTracer tracer = context.tracer();
-    if (tracer != null && src) {
+    Object trace = this.trace;
```
jotak, line 48: why this new Object? thread-safety concerns?

src/main/java/io/vertx/core/tracing/TracingPolicy.java
```diff
@@ -0,0 +1,35 @@
...
+
+  /**
+   * Propagate an existing trace.
+   */
+  PROPAGATE,
+
+  /**
+   * Reuse an existing trace or create a new trace when no one exist.
+   */
+  REQUIRED
```
jotak, line 33: as said in the other PR I think I would prefer "Always" 
or "CREATE_OR_PROPAGATE" would be very explicit but very verbose as well...

src/main/java/io/vertx/core/eventbus/impl/ReplyHandler.java
vietj: yes, reading a ref and then using it

src/main/java/io/vertx/core/tracing/TracingPolicy.java
vietj: I'll use always as you recommend

jotak (approved):

Issue #3616, opened 2020-10-29
Don't disable resource leak detector if netty property is set


jvican:
#### Describe the feature

Don't disable resource leak detector in netty if the netty system property for the resource leak detector is set. Vertx disables it in a static initializer and it would be best that before disabling it checks if [`io.netty.leakDetection.level`](https://github.com/netty/netty/blob/e9b28e76a366dadaf6da4375c503f670ec3e3ae4/common/src/main/java/io/netty/util/ResourceLeakDetector.java#L46) is set. If it is set, then vertx won't disable the resource leak detector.

#### Use cases

This is useful for those folks that want to check that no resource leaks are ever found in their code using the current vertx release as well as future releases.

vietj:
can you contribute a patch for this @jvican ?

jvican:
Yeah, will take care of this one

PR #3620, opened 2020-10-30
Don't disable resource leak detector if set by user

jvican:
Fixes https://github.com/eclipse-vertx/vert.x/issues/3616

### DIFF
--- a/src/main/java/io/vertx/core/impl/VertxImpl.java
+++ b/src/main/java/io/vertx/core/impl/VertxImpl.java
@@ -86,8 +86,12 @@ public class VertxImpl implements VertxInternal, MetricsProvider {
   private static final int NETTY_IO_RATIO = Integer.getInteger(NETTY_IO_RATIO_PROPERTY_NAME, 50);
 
   static {
-    // Netty resource leak detection has a performance overhead and we do not need it in Vert.x
-    ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.DISABLED);
+    // Disable Netty's resource leak detection to reduce the performance overhead if not set by user
+    if (System.getProperty("io.netty.leakDetection.level") != null ||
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -86,8 +86,12 @@
   private static final int NETTY_IO_RATIO = Integer.getInteger(NETTY_IO_RATIO_PROPERTY_NAME, 50);
 
   static {
-    // Netty resource leak detection has a performance overhead and we do not need it in Vert.x
-    ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.DISABLED);
+    // Disable Netty's resource leak detection to reduce the performance overhead if not set by user
+    if (System.getProperty("io.netty.leakDetection.level") != null ||
```
vietj, line 90: why 2 distinct property names ?

jvican: The second property name is a legacy one still used by some applications. I can remove it if you want but I preferred to add it for the sake of completeness, as migrating away from old system properties is not always trivial.

vietj: can you add a comment about it ?

jvican: Sure thing 👍

jvican: Done @vietj

vietj:
can you back-port to 3.9 branch ?

vietj:
thanks for the contribution

jvican:
@vietj sure thing, will send a PR to backport this to 3.9 shortly. Thanks for the help seeing this through.

Issue #3619, opened 2020-10-30
Allow users to change the vert.x context propagation through threads


jvican:
#### Describe the feature

The `VertxImpl` implementation uses a `ThreadLocal` called `stickyContext` to propagate the context when `getOrCreateContext` or `getContext` are called outside of a pool of vert.x threads.

However, there are other ways that the state can be propagated in other executor pools. For example, effect libraries such as [Monix](https://github.com/monix/monix) have their own way of propagating state local to an execution with `TaskLocal` -- this is an equivalent mechanism to the context locals that vert.x uses internally to work around the limitation that handlers can be run in different handlers and yet be part of the same semantic execution trace.

I'd like to request that vert.x allows users to bring their own mechanism to propagate the vert.x context.

That can be achieved in two ways:

- Making `getContext` part of the `Vertx` API just like `getOrCreateContext` is -- this way, I can create a `WrappedVertx` extending `Vertx`, wrapping the default `vertx: VertxImpl` and overriding `getContext` and `getOrCreateContext`.
- Adding a new abstraction `LocalManager`, similar in name and spirit to `ClusterManager`, that I can implement as a service and that vertx loads through SPI. The default instance of `LocalManager` will use the thread local sticky context implementation with the weak reference.

I prefer the second solution over the first one. Do you have any thoughts?

#### Contribution

I can open a pull request implementing this feature.

vietj:
what are the actual use cases for this feature ?

vietj:
stickyContext is not a way to propagate a context, it is a way to guarantee that the same context will always be used when called from a non vertx thread in order to avoid races.

jvican:
> stickyContext is not a way to propagate a context, it is a way to guarantee that the same context will always be used when called from a non vertx thread in order to avoid races.

Yeah, that's propagation for me. To be more explicit, Vert.x wants to make sure it propagates the context for all runnables that run in a thread not managed by the framework.

> what are the actual use cases for this feature ?

I'm building a framework using Monix, the mentioned effect system library, that propagates local state with a mechanism different than thread locals.

PR #3622, opened 2020-10-31
Add `ContextManager` service abstraction

jvican:
This PR implements a prototype for #3619.  This is my first non-trivial
contribution to the project, please let me know if I got something
wrong. I hope this prototype informs a bit better what I intend to do,
I'm not sure I explained my intentions correctly in the ticket.

----

The context manager is a new abstraction that allow users to implement
their own service providers to change the default, thread-local-based
mechanism to guarantee that the same context is used from threads that
are not managed by vert.x (for example, when users use the `WebClient`
in an external thread pool).

This new abstraction intends to imitate how the `ContextInternal` API
works.  As the previous code already had a cast from
`ContextInternal.currentContext` to `AbstractContext`, which is an
internal class, the new context manager returns a generic public
instance of `Context` and relies on `VertxImpl` to cast it to whichever
context it wants to use.

The new interface is designed so that implementors don't need to use
internal classes (e.g. `ContextInternal` or `AbstractContext`) to
implement the new interface.

Context manager implementors should not change the stored context
reference and the java doc clearly specifies that. That is to guarantee
context manager implementors don't change the context stored by vert.x
and possibly change its type, causing the internal cast to fail.

### DIFF
--- a/src/main/java/io/vertx/core/impl/VertxFactory.java
+++ b/src/main/java/io/vertx/core/impl/VertxFactory.java
@@ -13,6 +13,7 @@ package io.vertx.core.impl;
 
 import io.vertx.core.*;
 import io.vertx.core.file.impl.FileResolver;
+import io.vertx.core.impl.context.ContextManagerImpl;
 import io.vertx.core.net.impl.transport.Transport;
 import io.vertx.core.spi.VertxMetricsFactory;
 import io.vertx.core.spi.VertxTracerFactory;
@@ -21,6 +22,7 @@ import io.vertx.core.spi.cluster.NodeSelector;
 import io.vertx.core.spi.cluster.impl.DefaultNodeSelector;
 import io.vertx.core.spi.metrics.VertxMetrics;
 import io.vertx.core.spi.tracing.VertxTracer;
+import io.vertx.core.spi.context.ContextManager;
 
 import java.util.Collection;
 import java.util.Objects;
@@ -75,13 +77,13 @@ public class VertxFactory {
   }
 
   public Vertx vertx() {
-    VertxImpl vertx = new VertxImpl(options, null, null, createMetrics(), createTracer(), createTransport(), createFileResolver());
+    VertxImpl vertx = new VertxImpl(options, null, null, createMetrics(), createTracer(), createTransport(), createFileResolver(), createContextManager());
     vertx.init();
     return vertx;
   }
 
   public void clusteredVertx(Handler<AsyncResult<Vertx>> handler) {
-    VertxImpl vertx = new VertxImpl(options, createClusterManager(), createNodeSelector(), createMetrics(), createTracer(), createTransport(), createFileResolver());
+    VertxImpl vertx = new VertxImpl(options, createClusterManager(), createNodeSelector(), createMetrics(), createTracer(), createTransport(), createFileResolver(), createContextManager());
     vertx.initClustered(options, handler);
   }
 
@@ -157,6 +159,14 @@ public class VertxFactory {
       }
     }
     return tracer;
+  }
+
+  private ContextManager createContextManager() {
+    ContextManager contextManager = ServiceHelper.loadFactoryOrNull(ContextManager.class);
+    if (contextManager == null) {
+      contextManager = new ContextManagerImpl();
+    }
+    return contextManager;
   }
 
   private FileResolver createFileResolver() {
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/impl/VertxFactory.java
```diff
@@ -159,6 +161,14 @@ private VertxTracer createTracer() {
     return tracer;
   }
 
+  private ContextManager createContextManager() {
```
jvican, line 164: I don't store anything in the factory so that the context manager doesn't share the same context manager across many vert.x instances, which would leak its stateful nature. One vert.x instance, one context manager. I explored adding here a context manager factory but looked more verbose and I didn't think it was useful to add another layer of indirection.

vietj:
I think instead we want to have a single method e.g `getContext` and the implementation will create an event loop context and associate it instead.

There are two reasons for this:

- this abstraction might prevent some critical JIT inlining to happen (because of two extra method calls) so a single call is preferrable
- it allows the creation of non event-loop contexts by the manager and it can be interesting. E.g a worker context could be returned instead

jvican:
Thanks for the review @vietj!

> I think instead we want to have a single method e.g getContext and the implementation will create an event loop context and associate it instead.

In the `resolveHooks` and `createSharedWorkerExecutor` call sites inside `VertxImpl`, `getContext` doesn't create a new context if there is none available, I think we will still need `getContext` to cover those use cases.

> this abstraction might prevent some critical JIT inlining to happen (because of two extra method calls) so a single call is preferrable
  
Good point! I agree this is a performance sensitive change.

I think it's OK if `setContext` is potentially not inlined because it's called only on misses, which should be bounded and happen infrequently, but two questions about inlining `getContext`:

- If we add the context creation logic to the implementation of `getContext`, will the JIT not be less likely to inline the method because the method size will be larger? I'm concerned that will decrease the inlining budget even more than keeping the implementation of `getContext` as is now.
- Do you want me to check with jitwatch that `getContext` is inlined after this change? If so, which JVM do you want me to use? This way, we can guarantee the new abstraction doesn't imply a performance hit.
  
Also, I believe we can help the JIT inline `getContext` even more by marking the implementation class
`ContextManagerImpl` final, as that gives the JIT extra points to avoid dynamic dispatch and inline the implementations.

I personally prefer to keep the interface unchanged and keep `getContext` and `setContext` separate because it makes the semantics of the new abstraction clearer and doesn't conflate accessing and creating a context in one operation.
  
> it allows the creation of non event-loop contexts by the manager and it can be interesting. E.g a worker context could be returned instead

Yeah! This is a nice idea. I didn't add a `createContext` interface because I'm concerned that would force implementations of the service to depend on internals of the project to call `createEventLoopContext` and `createWorkerContext`, which are only defined in `VertxInternal` and that depend on other internals such as `CloseHooks`, `Deployment` or `WorkerPool`. Do you have any ideas to avoid making those public and still allow the context manager to create the context?

jvican:
@vietj Let me know how you'd like me to proceed here, it'd be lovely if we can add this in 4.0.0 as I'd like to use it as soon as possible.

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -515,13 +516,7 @@ public long scheduleTimeout(ContextInternal context, Handler<Long> handler, long
...
 
   public AbstractContext getContext() {
-    AbstractContext context = (AbstractContext) ContextInternal.current();
-    if (context != null && context.owner() == this) {
-      return context;
-    } else {
-      WeakReference<AbstractContext> ref = stickyContext.get();
-      return ref != null ? ref.get() : null;
-    }
+    return (AbstractContext) contextManager.getContext(this);
```
vietj, line 519: So actually the ContextInternal.current() should remain here (as it does not cost an extra call) and the else statement should be in the context manager, sorry to not have been clear enough

vietj: In other words, the call should only be called when we are not on the current thread.

src/main/java/io/vertx/core/spi/context/ContextManager.java
```diff
@@ -0,0 +1,49 @@
...
+ * The context manager guarantees that vert.x operations running in the
+ * same "execution trace" share the same context to avoid race conditions
+ * when executing vert.x handlers.
+ *
+ * The default implementation of this context manager propagates the context
+ * using a thread local, so that code running subsequent to the first context
+ * creation can access it. Users can bring their own manager implementation
+ * to use a different mechanism different to thread locals.
+ */
+public interface ContextManager {
```
vietj, line 30: The name `ContextManager` is too general and does not describe precisely what we want here which is associating a non vertx thread with a context.

vietj:
we can try progress toward a solution as best effort, however this is not currently a 4.0.0 priority.

PR #3639, opened 2020-11-10
Fixes #3629: runs the upload handler on the caller context

### LINKED ISSUE #3629: [Regression] Vertx4 HttpServerRequest.uploadHandler() erases context.
I have a code snippet (link to the github repo below) that works on vertx4 milestone4, but not with vertx4 >= milestone5.
At first I thought it could be related to https://github.com/eclipse-vertx/vert.x/issues/3601, but this does not seem to be the case.

# Problem
Vertx context (`Vertx.currentContext()`) is lost inside `HttpServerRequest.uploadHandler()`

# Expected behavior
Vertx context (`Vertx.currentContext()`)  is kept inside `HttpServerRequest.uploadHandler()`

# Code example
```
public class MyHandler implements Handler<RoutingContext> {
    @Override
    public void handle(RoutingContext routingContext) {
        final HttpServerRequest event = routingContext.request().getDelegate();
        event.setExpectMultipart(true);

        Context context = Vertx.currentContext(); // Context is NOT null here
        event.uploadHandler(event1 -> {
            Single.create(new io.vertx.rx.java.SingleOnSubscribeAdapter<>(fut -> {
                Context context2 = Vertx.currentContext(); // Context2 is null here
            })).subscribe(o -> routingContext.end("check"));
        });

    }
}
```

As I said, with vertx4 milestone 4, both `context` and `context2` are *not* null, while from milestone 5 onward, `context2` is `null`.

# Github project with reproducible test
https://github.com/oscarfh/vertx-upload-handler-context
The test `ATest#start_server` will fail, but if you downgrade to vertx milestone 4, it passes.

pmlopes:

Motivation:

Fixes #3629

The upload handler (not being a future) was being called without any guarantees on the caller context. This PR ensures that calls always run on the right context.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/NettyFileUploadDataFactory.java
+++ b/src/main/java/io/vertx/core/http/impl/NettyFileUploadDataFactory.java
@@ -46,7 +46,8 @@ class NettyFileUploadDataFactory extends DefaultHttpDataFactory {
       size);
     Handler<HttpServerFileUpload> uploadHandler = lazyUploadHandler.get();
     if (uploadHandler != null) {
-      uploadHandler.handle(upload);
+      // run the handler on the caller context
+      context.runOnContext(v -> uploadHandler.handle(upload));
     }
     return nettyUpload;
   }
--- a/src/main/java/io/vertx/core/http/impl/NettyFileUploadDataFactory.java
+++ b/src/main/java/io/vertx/core/http/impl/NettyFileUploadDataFactory.java
@@ -46,7 +46,7 @@ class NettyFileUploadDataFactory extends DefaultHttpDataFactory {
       size);
     Handler<HttpServerFileUpload> uploadHandler = lazyUploadHandler.get();
     if (uploadHandler != null) {
-      context.execute(upload, uploadHandler);
+      context.dispatch(upload, uploadHandler);
     }
     return nettyUpload;
   }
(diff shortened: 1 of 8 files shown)


src/main/java/io/vertx/core/http/impl/NettyFileUploadDataFactory.java
```diff
@@ -46,7 +46,8 @@ public FileUpload createFileUpload(HttpRequest httpRequest, String name, String
       size);
     Handler<HttpServerFileUpload> uploadHandler = lazyUploadHandler.get();
     if (uploadHandler != null) {
-      uploadHandler.handle(upload);
+      // run the handler on the caller context
+      context.runOnContext(v -> uploadHandler.handle(upload));
```
vietj, line 50: instead try `context.execute(upload, uploadHandler)`

jvican: What's exactly the difference between the two @vietj? I've always used `context.runOnContext` because when I tried to use `context.execute` it would run the handlers outside of the context.

vietj: Actually it should have be `context.dispatch()`, look at the latest version.

context.execute will execute the task on the context, if the current event loop is already the context then the task is executed directly.

vietj:
failed to merge this for 4.0.0 and somebody provided a similar PR we merged

PR #3640, opened 2020-11-12
Event bus improvements

vietj:
The three first commits simplify a bit the event-bus serializer implementation.
The fourth commit rewrites the event-bus serializer dequeuing algorithm to use a tail recursion style to avoid stack overflows.


vietj:
@tsegismont you can review each commit separately to avoid mixing concerns

tsegismont (approved):
@vietj thanks, LGTM

Just one comment: I think we could move back the logic that's now in the `add` method to the `Serialized` task class and call check pending directly from the `queue` method.
Perhaps that would improve readability.

But that's not blocking, I can propose a later PR to verify this and you decide.

vietj:
ok @tsegismont

PR #3641, opened 2020-11-16
Minor update of Serializer for readability

tsegismont:
When a task is complete, we need to:

- complete the original promise (provided by the event bus class)
- make sure task processing continues by polling the queue and checking pending tasks

Instead of implementing this behavior in a single method (SerializerQueue#add), we can separate the concerns in other methods for clarity.


tsegismont:
@vietj here's the minor update I talked about when reviewing #3640

src/main/java/io/vertx/core/eventbus/impl/clustered/Serializer.java
```diff
@@ -127,30 +124,26 @@ void close() {
         tasks.remove().internalPromise.tryFail("Context is closing");
       }
     }
+  }
 
-    private class SerializedTask<U> implements Handler<AsyncResult<U>> {
+  private class SerializedTask<U> {
```
vietj, line 129: why are you removing this ? it allows to save a instance creation

tsegismont: Because with the changes in this PR, it's not needed anymore.

tsegismont: I mean, it's not needed to save an instance creation anymore.

vietj: `task.future().onComplete(v -> processed())` does instantiate a lambda unless we cache it.

If `SerializerTask` is an handler then it could call in handle the `processed()` method

tsegismont: I've updated the PR

src/main/java/io/vertx/core/eventbus/impl/clustered/Serializer.java
```diff
@@ -73,7 +75,8 @@ private void close(Promise<Void> promise) {
     promise.complete();
   }
 
-  private class SerializerQueue {
+  @SuppressWarnings({"rawtypes", "unchecked"})
+  private class SerializerQueue implements Handler<AsyncResult> {
```
vietj, line 79: what is the type arg for AsyncResult ?

tsegismont: It's the same as of `SerializedTask`. But we can't put a wildcard here, the compiler does not accept it.

tsegismont:
@vietj any other comments/concerns?

PR #3642, opened 2020-11-16
JavaDoc that JsonObject#copy and JsonArray#copy make a deep copy

julianladisch:
The JavaDoc should state whether the copy is shallow or deep.


tsegismont (approved):
LGTM, thank you

PR #3643, opened 2020-11-16
Add Future.toCompletableFuture()

julianladisch:
Many use cases require a CompletableFuture, for example
`future.toCompletionStage().toCompletableFuture().get()`
`future.toCompletionStage().toCompletableFuture().isDone()`
`future.toCompletionStage().toCompletableFuture().completeExceptionally(failure)`

Use the shorthand Future.toCompletableFuture method for concise code:
`future.toCompletableFuture().get()`
`future.toCompletableFuture().isDone()`
`future.toCompletableFuture().completeExceptionally(failure)`

Reference:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html#toCompletableFuture--

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -317,14 +317,14 @@ public interface Future<T> extends AsyncResult<T> {
   }
 
   /**
-   * Bridges this Vert.x future to a {@link CompletionStage} instance.
+   * Bridges this Vert.x future to a {@link CompletableFuture} instance.
    * <p>
-   * The {@link CompletionStage} handling methods will be called from the thread that resolves this future.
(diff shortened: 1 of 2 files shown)


tsegismont:
@julianladisch the choice of exposing `CompletionStage` only was debated on vertx-dev and we made a decision.

Be careful with:

`future.toCompletionStage().toCompletableFuture().get()`

Most often the future will be completed on an event loop, so this will block it.

julianladisch:
Closing to nudge people not to run a blocking get() on the event thread.

jponge:
Something else to consider is that `CompletableFuture` is both a future and a promise, so it would have been weird to complete a Vert.x future by completing a CF :-)

PR #3653, opened 2020-11-18
Allow space in Windows file path

### LINKED ISSUE #3630: Filesystem.readDir fail on file with space in filename
### Version

3.9.4

### Context

Filesystem methods for read directory failing when directory contains file with space character in filename.

### Root cause

The code is failing in FileResolver.getValidClassLoaderResource because isValidCachePath returns false
```
    private static URL getValidClassLoaderResource(ClassLoader cl, String fileName) {
        URL resource = cl.getResource(fileName);
        return resource != null && !isValidCachePath(fileName) ? null : resource;
    }
```
In isValidCachePath 
```
if (!CACHE_PATH_CHECKER.test(c)) { //where c = 32
     return false;
}
```

### Steps to reproduce

1. Put "some file.txt" into resources.
2. Use Filesystem.readDir(path, handler).

### Extra

Windows 7, Azul Java 11

julianladisch:
A space is a valid character in Windows file path, but
not at the end of a directory or file name.

The path validation is adjusted to reject a space or dot
at the end of a file or directory name.

The \ character is always replaced by a / character so that
it doesn't get URL encoded.

Fixes #3630


vietj:
@julianladisch I'm working on a PR that run test on windows, we should review and merge this PR after the window test PR.

vietj:
@julianladisch see https://github.com/eclipse-vertx/vert.x/pull/3657

vietj:
thanks @julianladisch I took inspiration from your PR and fixed it, I mostly tried to have correct tests for windows.

PR #3655, opened 2020-11-18
Replace SystemUtils.IS_OS_LINUX

julianladisch:
SystemUtils cannot instantiate when running under
Windows and Java 11:

```
java.lang.ExceptionInInitializerError: null
        at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
        at java.base/java.lang.String.substring(String.java:1874)
        at org.apache.commons.lang.SystemUtils.getJavaVersionAsFloat(SystemUtils.java:1133)
        at org.apache.commons.lang.SystemUtils.<clinit>(SystemUtils.java:818)
```

The replacement uses the same os.name string comparison:
https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/src-html/org/apache/commons/lang/SystemUtils.html#line.1153


tsegismont:
Can you explain why it does not work with Java 11 on Windows?

julianladisch:
This is the relevant code:
https://github.com/apache/commons-lang/blob/LANG_2_4/src/java/org/apache/commons/lang/SystemUtils.java#L1115-L1142
```
public static final float JAVA_VERSION_FLOAT = getJavaVersionAsFloat();
private static float getJavaVersionAsFloat() {
    if (JAVA_VERSION_TRIMMED == null) {
        return 0f;
    }
    String str = JAVA_VERSION_TRIMMED.substring(0, 3);    // "11".substring(0, 3) fails
    if (JAVA_VERSION_TRIMMED.length() >= 5) {
        str = str + JAVA_VERSION_TRIMMED.substring(4, 5);
    }
    try {
        return Float.parseFloat(str);
    } catch (Exception ex) {
        return 0;
    }
}
```

`OpenJDK Runtime Environment 18.9 (build 11+28)` has this `java.version` string: `11`

This is two characters long.

The code expects three characters like `1.8` or five characters like `1.3.1`.

Two characters make `substring(0, 3)` fail:

```
testClusterPublicHostHintFromClusterManager(io.vertx.core.eventbus.ClusterHostTest)  Time elapsed: 6.419 sec  <<< ERROR!
java.lang.ExceptionInInitializerError: null
        at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3319)
        at java.base/java.lang.String.substring(String.java:1874)
        at org.apache.commons.lang.SystemUtils.getJavaVersionAsFloat(SystemUtils.java:1133)
        at org.apache.commons.lang.SystemUtils.<clinit>(SystemUtils.java:818)
        at io.vertx.core.eventbus.ClusterHostTest.setUp(ClusterHostTest.java:26)
```

src/test/java/io/vertx/core/eventbus/ClusterHostTest.java
```diff
@@ -14,16 +14,24 @@
...
 public class ClusterHostTest extends VertxTestBase {
+  private static final String OS_NAME = getOsName();
 
   @Override
   public void setUp() throws Exception {
     super.setUp();
-    Assume.assumeTrue(SystemUtils.IS_OS_LINUX);
+    Assume.assumeTrue(OS_NAME.startsWith("Linux") || OS_NAME.startsWith("LINUX"));
+  }
+
```
vietj: can you instead use Netty's `PlatformDependant` util class ?

julianladisch: Yes, done.

vietj:
can you rebase this on latest master ?

PR #3659, opened 2020-11-20
Fix completion handling when removing sub

### LINKED ISSUE #3638: ClusteredEventBus.removeRegistration does not call completionHandler on failure
### Version

3.9.3, present in last 3.9, it seems to be fixed in master

### Context

Closing clustered Vert.x with **custom-provided Infinispan cluster manager**. The CM is closed in `VertxInternal.addCloseHook(...)` handler.

In method `ClusteredEventBus.removeSub` the `completionHandler` is invoked if and only if the `subs.remove(...)` succeeds; if that fails only an error is logged but the handler is not invoked. Consequently the handler for `vertx.close(...)` is never called.

In my case the call fails because the cluster manager is already closed when the eventbus is being unregistered.

### Steps to reproduce

```java
DefaultCacheManager cacheManager = ...
options.setClusterManager(new InfinispanClusterManager(cacheManager));
Vertx.clusteredVertx(options, result -> {
   if (result.failed()) {
      /* ... */
   }
   Vertx vertx = result.result();
   /* ... */
   ((VertxInternal) vertx).addCloseHook(handler -> {
      cacheManager.stop();
      handler.handle(Future.succeededFuture());
   });
   /* ... */
});
```

### Extra

If there is a way to properly close the CM upon Vert.x close please let me know.

tsegismont:
Fixes #3638

Completion handler, if present, should be called in all cases.
When the cluster manager could not find the subscription, it's not necessary to complete with a failure.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
@@ -409,19 +409,12 @@ public class ClusteredEventBus extends EventBusImpl {
 
   private void removeSub(String subName, ClusterNodeInfo node, Handler<AsyncResult<Void>> completionHandler) {
     subs.remove(subName, node, ar -> {
-      if (!ar.succeeded()) {
-        log.error("Failed to remove sub", ar.cause());
-      } else {
-        if (ar.result()) {
-          if (completionHandler != null) {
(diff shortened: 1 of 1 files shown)


tsegismont:
@vietj this is the behavior of 4.0 already, no need for a `master` branch PR.

vietj (approved):

Issue #3617, opened 2020-10-29
Support for setting local source port on WebClientOptions


pantinor:
#### Read me

As a user of WebClient, I would like the ability to configure the local source port in addition to the local address, in cases where a ephemeral source port is not acceptable in deployed production environment.

#### Describe the feature

Addition of setLocalPort on WebClientOptions

#### Use cases

Some servers may have firewall rules in place that limit certain ports or ranges by the source port of the request.  

#### Contribution

It would be best if the vertx team could implement the feature.  If you need assist for us to contribute please let me know, and provide a pointer or brief description of where this addition is most appropriate for implementing.

hradecek:
I've create a PR [1], which is adding a new options `localPort` for `HttpClientOptions` and `NetClientOptions`.
--
[1] https://github.com/eclipse-vertx/vert.x/pull/3660

pantinor:
@hradecek Are we delivering it to WebClientOptions with this PR as well, or is that required to be separate PR ?

vietj:
This should not be in `WebClientOptions` because this applies to all connections created by the client. Instead it should be a `localPort` setting on `HttpRequest` that would set the local port on the `RequestOptions` that are created for `HttpClient`.

So we should proceed to first have a contribution in master to add local port support to `RequestOptions` and then a contribution here.

vietj:
can you reopen in vertx-web project this issue, it is the wrong project

PR #3660, opened 2020-11-21
Customized local port on Net/Http client

hradecek:
**Support for setting local source port**
*Enhancement link*: https://github.com/eclipse-vertx/vert.x/issues/3617

Added new options to `HttpClientOptions` and `NetClientOptions` named `localPort` which is then passed to underlying `io.netty.bootstrap.Bootstrap` at `Transport#configure`.

If  `localPort` is not specified then default value - `0` is used, which means that system will picks the port by itself (ephemeral port), no change in functionality in comparing to current behavior.

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -149,6 +149,10 @@ Set the idle timeout unit. If not specified, default is seconds.
 Set the local interface to bind for network connections. When the local address is null,
  it will pick any local address, the default local address is null.
 +++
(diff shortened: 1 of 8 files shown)


vietj:
For HTTP I think we should use instead `RequestOptions` so its specified per request and not globally

vietj:
For NetClient, I think we could introduce a `ConnectOptions` that would carry it.

vietj:
as I explained this PR cannot work as is, we should proceed as I explained and provide a local port on `RequestOptions` instead to specify the local port when the HTTP connection is asked.

PR #3663, opened 2020-11-23
Clustered consumer keeps local registration when unregistered

### LINKED ISSUE #3662: Clustered consumer keeps local registration when unregistered
A method name was changed in `EventBusImpl` but not updated in `ClusteredEventBus` while re-working the clustering SPI.
This might have happened while rebasing #3132 on `master`.

The bug was not caught because we didn't have a test for unregistration.

tsegismont:
Fixes #3662


vietj (approved):

vietj:
tests seem to pass but it looks like the failure is coming from the junit publish results.

PR #3667, opened 2020-11-24
Rework generic KeyStoreOptions

vietj:


src/main/java/io/vertx/core/net/KeyStoreOptionsBase.java
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ * @author <a href="http://tfox.org">Tim Fox</a>
+ */
+class KeyStoreOptionsBase implements KeyCertOptions, TrustOptions {
+
+  private KeyStoreHelper helper;
+  private String type;
+  private String password;
+  private String path;
```
sberyozkin, line 37: @vietj Hi Julien - it looks good, can this base class include `provider` now ?

vietj: yes, but how can we test it ?

pmlopes: I think you can use "jce" as provider to load a jceks file. Although this provider is always present if I'm not mistaken.

vietj: ok, I'll have a look, perhaps we can write an integration test with a mock provider that delegates to JCE ?

vietj: it is actually used `KeyStore.getInstance("JKS").getProvider().getName()` that returns the _SUN_ value

vietj:
I updated the PR with the provider, I think now this is ready for review.

src/main/java/io/vertx/core/net/KeyStoreOptions.java
```diff
@@ -26,33 +27,29 @@
...
- * options.setKeyStore(new KeyStoreOptions().setType("JKS").setPath("/mykeystore.jks").setPassword("foo"));
+ * options.setKeyCertOptions(new KeyStoreOptions().setType("JKS").setPath("/mykeystore.jks").setPassword("foo"));
  * </pre>
  *
  * Or directly provided as a buffer:
  *
  * <pre>
  * Buffer store = vertx.fileSystem().readFileBlocking("/mykeystore.jks");
- * options.setKeyStore(new JKSOptions().setType("JKS").setValue(store).setPassword("foo"));
+ * options.setKeyCertOptions(new JKSOptions().setType("JKS").setValue(store).setPassword("foo"));
```
pmlopes, line 37: `new KeyStoreOptions()`

pmlopes (approved):
Just a small typpo on the javadoc

PR #3669, opened 2020-11-25
Intermittent failure in testClusteredUnregistration

tsegismont:
The test failed intermittently with Hazelcast on CI.
It failed consistently with Vert.x nodes configured as lite members of the Hazelcast cluster.

When a subscription is removed, it can take some time before the cluster manager invokes our listeners.
It is then possible that we still have in cache a registration that has been effectively removed.

The fix consists in intercepting the registration update event to make sure it has been processed.
Then we can try sending again and we should get the NO_HANDLERS error as expected.


tsegismont:
This fixes the issue with Hazelcast CM build.

tsegismont:
@vietj the build failure looks unrelated

PR #3679, opened 2020-11-29
using empty array for Collection.toArray()

dfa1:
Motivation:

There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
    
In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.

### DIFF
--- a/src/main/java/io/vertx/core/CompositeFuture.java
+++ b/src/main/java/io/vertx/core/CompositeFuture.java
@@ -73,7 +73,7 @@ public interface CompositeFuture extends Future<CompositeFuture> {
    * When the list is empty, the returned future will be already completed.
    */
   static CompositeFuture all(List<Future> futures) {
-    return CompositeFutureImpl.all(futures.toArray(new Future[futures.size()]));
+    return CompositeFutureImpl.all(futures.toArray(new Future[0]));
   }
 
   /**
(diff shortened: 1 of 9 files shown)


julianladisch:
The posted motivation text is the same as https://github.com/JetBrains/intellij-community/blob/master/plugins/InspectionGadgets/src/inspectionDescriptions/ToArrayCallWithZeroLengthArrayArgument.html and should have proper attribution.
For background information please visit https://shipilev.net/blog/2016/arrays-wisdom-ancients/

dfa1:
thank you @julianladisch for your time, please have another look at the commit message

tsegismont (approved):
Thank you @dfa1

dfa1:
you're welcome! I'm happy to contribute something back to the community :)

PR #3680, opened 2020-12-02
Hostname validation by default, fixes SSL/TLS certificate host mismatch

julianladisch:
Vert.x (and Netty) disable hostname validation of SSL/TLS
certificates by default. This opens a back door for man-in-the-middle
(MITM) attacks because attackers only need to present a valid SSL/TLS
certificate for a different hostname to successfully intercept the
connection.

Details: https://cwe.mitre.org/data/definitions/297.html

Developers using Vert.x can enable hostname validation if they call
`.setHostnameVerificationAlgorithm("HTTPS")`.
However, 6 out of 7 examples presented on
https://vertx.io/docs/vertx-core/java/#_client_trust_configuration
don't enable it. It is likely that some developers forget to
enable hostname verification.

Security by default is state of the art and required by GDPR articles
32 and 25. Libraries should ship with secure defaults.

Therefore Vert.x should enable SSL/TLS hostname validation by default.

Netty will enable hostname validation by default for Netty 5:
https://github.com/netty/netty/issues/8537
Netty got security vulnerability notices because of this issue:
https://snyk.io/vuln/SNYK-JAVA-IONETTY-1042268
https://github.com/netty/netty/issues/10362

### DIFF
--- a/src/main/asciidoc/net.adoc
+++ b/src/main/asciidoc/net.adoc
@@ -448,9 +448,16 @@ To enable SSL on a NetClient the function setSSL(true) is called.
 
 ==== Client trust configuration
 
-If the {@link io.vertx.core.net.ClientOptionsBase#setTrustAll trustALl} is set to true on the client, then the client will
-trust all server certificates. The connection will still be encrypted but this mode is vulnerable to 'man in the middle' attacks. I.e. you can't
-be sure who you are connecting to. Use this with caution. Default value is false.
+By default, the client checks that the certificate presented by the server is trusted, and that the host name the
+client wants to connect to is listed in the server certificate.
+
+Disabling one or both of these checks makes the connection vulnerable to 'man in the middle' attacks. I.e. you can't
+be sure who you are connecting to. Use this with caution.
+
+If the {@link io.vertx.core.net.ClientOptionsBase#setTrustAll trustAll} is set to true on the client, then the client will
+trust all server certificates, i.e. certificates issued by an attacker.
+
+The default value for trustAll is false.
 
 [source,$lang]
 ----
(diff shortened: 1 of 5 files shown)


vietj:
I don't think it is correct to set the default value to HTTPS in NetClientOptions, setting this value should be done in HttpClientOptions subclass only

vietj:
@julianladisch can you change to configure the algorithm in the `HttpClientOptions` subclass ?

vietj:
Actually Vert.x is currently fine, your PR does care about `NetClient` which is a generic TCP client and should not be bothered by the hostname verification algorithm by default. This algorithm depends on the underlying exchanged protocol , e.g it's HTTP for HTTP but also for some others) like vertx-mail-client that does use this value https://github.com/vert-x3/vertx-mail-client/blob/master/src/main/java/io/vertx/ext/mail/impl/SMTPConnectionPool.java#L66

The current HTTP client instead use an option named `verifyHost` that is enabled by default, you can see in `SSLHelper`

```java
  public SSLHelper(HttpClientOptions options, KeyCertOptions keyCertOptions, TrustOptions trustOptions) {
    this((ClientOptionsBase) options, keyCertOptions, trustOptions);
    if (options.isVerifyHost()) {
      this.endpointIdentificationAlgorithm = "HTTPS";
    }
  }
```

I think we can keep the documentation part to warn the user that when using a TCP client with SSL then out of the box nothing is done and there might be a MIDM attack and that the verification can be enabled.

vietj:
can you simply update the doc ?

vietj:
please resubmit another pr with doc change

src/main/asciidoc/net.adoc
```diff
@@ -448,9 +448,16 @@ To enable SSL on a NetClient the function setSSL(true) is called.
 
 ==== Client trust configuration
 
-If the {@link io.vertx.core.net.ClientOptionsBase#setTrustAll trustALl} is set to true on the client, then the client will
-trust all server certificates. The connection will still be encrypted but this mode is vulnerable to 'man in the middle' attacks. I.e. you can't
-be sure who you are connecting to. Use this with caution. Default value is false.
+By default, the client checks that the certificate presented by the server is trusted, and that the host name the
```
hutchig, line 451: link

Issue #703, opened 2013-07-25
Custom SSL trust managers


SlavikBaranov:
Support for custom trust managers would allow to implement custom trust stores. For example, store trusted certificates in a database so that they can be added/removed at runtime.

See this discussion in the Google group: https://groups.google.com/forum/#!topic/vertx/-grdlyiP2Ig

PR #3719, opened 2020-12-20
Support for generic Key and Trust material

Hakky54:
Describe the feature
While using the webclient of vertx from the following project: https://github.com/vert-x3/vertx-web I discovered that vertx core has couple of options to enable ssl. It should be of the type KeyCertOptions and TrustOptions. The core library already provides utilities to help the end-user by providing JksOptions, KeyStoreOptions, PemTrustOptions and PfxOptions.

It is providing a-lot ways for the end-user, but I had the feeling that it was missing one option. What if someone wanted to supply a KeyManagerFactory or KeyManager as KeyCertOptions? Or if someone wanted to supply a TrustManagerFactory or TrustManager? That would not be possible out of the box. The end-user needs to implement KeyCertOptions and TrustOptions and provide a custom logic to enable it.

Use cases
This will enable the end-user to supply their own custom constructed key and trust material, for example a trust manager which is able to use multiple different keystore files, for example: Stackoverflow - Registering multiple keystores in JVM
In an existing environment where someone wants to migrate to vertx and has already a keymanager/trustmanager, by just supplying it it doesn't need to use JksOptions, KeyStoreOptions, KeyStoreOptons, PemTrustOptions or PfxOptions to configure vertx. They would be able to supply just only KeyManager, KeyManagerFactory, TrustManager or TrustManagerFactory and adapt much easier.

With this PR the following will be possible:

```java
TrustManager trustManager = ...; //initialised trustManager
KeyManager keyManager = ...; //initialised keyManager

WebClientOptions webClientOptions = new WebClientOptions();
webClientOptions.setTrustOptions(new TrustManagerOptions(trustManager);
webClientOptions.setKeyCertOptions(new KeyManagerOptions(keyManager);
webClientOptions.setSsl(true);

WebClient client = WebClient.create(Vertx.vertx(), webClientOptions);
```

or with the following snippet:

```java
TrustManagerFactory tmf = ...; //initialised trustManagerFactory
KeyManagerFactory kmf = ...; //initialised keyManagerFactory

WebClientOptions webClientOptions = new WebClientOptions();
webClientOptions.setTrustOptions(new TrustManagerOptions(tmf);
webClientOptions.setKeyCertOptions(new KeyManagerOptions(kmf);
webClientOptions.setSsl(true);

WebClient client = WebClient.create(Vertx.vertx(), webClientOptions);
```

I have discovered that back in 2013 someone requested this feature, see here: https://github.com/eclipse-vertx/vert.x/issues/703

Do you guys think this option has additional value to the library? And if so do you think something needs to be adjusted?
I was not very creative about the class names, so love to get feedback.


Hakky54:
This PR is an implementation for the following feature request: https://github.com/eclipse-vertx/vert.x/issues/3720

vietj:
this makes sense however, I think we should make the API more concise and not introduce a new options object.

Instead we should simply have a static methods on the related interfaces, i.e:

```java
webClientOptions.setTrustOptions(TrustOptions.wrap(trustManager));
webClientOptions.setKeyCertOptions(KeyCertOptions.wrap(keyManager));
```

Hakky54:
Hi @vietj,

Thank you for the feedback. It make sense, in that way it would be less complex for the end-user and very convenient to use. So I have refactored the code, what I did was:
- Changed the access modifier of the classes and constructors from public to default, the following classes are fully encapsulated and not accessible to anyone. The following classes are not accessible:
  - KeyManagerFactoryWrapper
  - KeyManagerFactoryOptions
  - KeyManagerOptions
  - TrustManagerFactoryWrapper
  - TrustManagerFactoryOptions
  - TrustManagerOptions
- KeyCertOptions contains a default method named wrap which takes a X509KeyManager and returns an instance of KeyCertOptions
- TrustOptions contains a default method named wrap which takes a TrustManager and returns an instance of TrustOptions

What do you think of the current PR? Any other things what needs to be adjusted

Hakky54:
@vietj Would you like me to adjust anything else?

src/main/java/io/vertx/core/net/KeyManagerOptions.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package io.vertx.core.net;
+
+import javax.net.ssl.KeyManager;
+
+/**
+ * @author Hakan Altindag
+ */
+class KeyManagerOptions extends KeyManagerFactoryOptions {
```
vietj, line 19: why is this class necessary ?

src/main/java/io/vertx/core/net/TrustManagerOptions.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package io.vertx.core.net;
+
+import javax.net.ssl.TrustManager;
+
+/**
+ * @author Hakan Altindag
+ */
+class TrustManagerOptions extends TrustManagerFactoryOptions {
```
vietj, line 19: why is this class necessary ?

src/main/java/io/vertx/core/net/KeyManagerOptions.java
Hakky54: It is not necessary, I wanted to decouple the new KeyOptions by having two implementations, 1. KeyManagerOptions and 2. KeyManagerFactoryOptions as one can be build from the keymanagerfactory and the other one from just only a keymanager. But if you prefer to have a single implementation I can move the constructor to KeyManagerFactoryOptions

Hakky54:
I have refactored it, the TrustManagerOptions and KeyManagerOptions are removed, slightly adjusted the KeyManagerFactoryOptions and TrustManagerFactoryOptions. Could you have another look at it, do you. want something else to be changed too?

Hakky54:
By the way do you think `TrustManagerFactoryWrapper` `TrustManagerFactoryOptions` `KeyManagerFactoryOptions` `KeyManagerFactoryWrapper` should be moved to the impl package or stay at the current package?

vietj:
I'm quite greedy and usually try to avoid un-necessary language declarations because they can have an impact on the runtime (JIT, heap)

src/main/java/io/vertx/core/net/KeyManagerFactoryOptions.java
```diff
@@ -0,0 +1,59 @@
...
+package io.vertx.core.net;
+
+import io.vertx.core.Vertx;
+
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.X509KeyManager;
+import java.util.function.Function;
+
+/**
+ * @author Hakan Altindag
```
vietj, line 21: Missing doc

src/main/java/io/vertx/core/net/KeyManagerFactoryWrapper.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.KeyManagerFactorySpi;
+import javax.net.ssl.ManagerFactoryParameters;
+import java.security.KeyStore;
+import java.security.Provider;
+import java.util.Objects;
+
+/**
+ * @author Hakan Altindag
```
vietj, line 26: Missing decription.

Hakky54:
I added some javadoc, let me know if that needs to be adjusted!

vietj:
LGTM, will wait for CI to finish

Hakky54:
Awesome, thank you very much Julien! It was quite fun to contribute 😄

vietj:
you're welcome

PR #3727, opened 2020-12-30
Add a lock-free implementation of a Future

essobedo:
## Motivation:

Avoid the overhead of synchronized blocks by using a `volatile` reads and writes instead 

## Modifications:

* Use an `AtomicInteger` to store the state of the `Future` in order to make compare-and-swap and `volatile` reads and writes
* Rely on piggybacking to read/modify the member fields `value` and `listener`

**NB:** Please note that I did not use `Unsafe` / `VarHandle` instead of  an `AtomicInteger` for a better compatibility with the different Java versions

### DIFF
--- a/src/main/java/io/vertx/core/impl/future/FutureImpl.java
+++ b/src/main/java/io/vertx/core/impl/future/FutureImpl.java
@@ -19,19 +19,64 @@ import io.vertx.core.impl.NoStackTraceThrowable;
 
 import java.util.ArrayList;
 import java.util.Objects;
-import java.util.function.Function;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
- * Future implementation.
+ * A thread-safe / lock-free implementation of a {@link Future} relying on compare-and-swap, busy-waiting
+ * and piggybacking.
  *
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ * @param <T> The result type of the future.
  */
 class FutureImpl<T> extends FutureBase<T> {
 
-  private static final Object NULL_VALUE = new Object();
-
+  /**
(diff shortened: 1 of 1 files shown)


essobedo:
@vietj Hi and happy new year! Could you please tell me what you think of this PR?

vietj:
@essobedo for now we are not considering it immediately because we want to carry a larger effort on this subject. We also haven't been worried so far but we might be because of some changes in Vert.x 4

essobedo:
@vietj when you say "a larger effort on this subject", what subject do you refer to exactly?

vietj:
I'm thinking of studying synchronized usage in vertx to see if that's necessary, harmfull, should be replaced, etc... and also benchmarks to see the impact of this in real life projects.

essobedo:
@vietj do you believe that this kind of PRs could be interesting for the project or should I close it?

vietj:
I think we can keep it

essobedo:
@vietj ok cool. One last question, may I propose other PRs like this one and #3728 or it is useless as long as you did not do your study?

vietj:
Anything regarding locking and concurrency should rather wait.

vietj:
BTW if you are interested to contribute there are other areas where you can
help

essobedo:
@vietj yes why not, I'm always interested in contributing to open source projects but do you mean tickets labeled with `help wanted`?

vietj:
yes those are good start for small features.

we do have coarser grained features that can be contributed and we can
discuss that on mailing list or on Discord

knotenpunkt:
@essobedo why you did close this issue?

essobedo:
@knotenpunkt the real question is more: why keeping it open? :-) In short "Spring Cleaning" of my PRs. This PR is open for too long now to be treated one day moreover it is outdated, so it is better to close it.

PR #3728, opened 2020-12-31
Add a lock-free implementation of a CloseFuture

essobedo:
## Motivation:

Avoid the overhead of synchronized blocks by using a volatile reads and writes instead.

## Modifications:

* Use an `AtomicInteger` to store the state of the `CloseFuture` in order to make compare-and-swap and `volatile` reads and writes
* Rely on piggybacking to read/modify the member field `resource`

**NB 1:** The code is meant to be backward compatible so it supports the fact that we can call `init(null)` after having initialized the `CloseFuture` with a non null value
**NB 2:** Please note that I did not use `Unsafe` / `VarHandle` instead of  an `AtomicInteger` for a better compatibility with the different Java versions

### DIFF
--- a/src/main/java/io/vertx/core/impl/CloseFuture.java
+++ b/src/main/java/io/vertx/core/impl/CloseFuture.java
@@ -16,56 +16,106 @@ import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Promise;
 
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 
 /**
  * An helper class for managing close operations.
+ * <p>
+ * A thread-safe / lock-free implementation of a {@link CloseFuture} relying on compare-and-swap, busy-waiting
+ * and piggybacking.
  */
 public class CloseFuture implements Future<Void>, Closeable {
 
+  /**
+   * All possible states of the {@code CloseFuture}, the initial state is <i>NEW</i> then the possible state transitions are:
+   * <p>
+   * <ul>
+   *   <li><i>NEW</i> -> <i>INITIALIZING</i> -> <i>INITIALIZED</i> -> <i>CLOSED</i>
+   * </ul>
+   *
+   * <p>
+   * The description of all possible states:
+   * <table>
+   *   <col width="25%"/>
+   *   <col width="75%"/>
+   *   <thead>
+   *     <tr><th>State</th><th>Description</th></tr>
+   *   <thead>
+   *   <tbody>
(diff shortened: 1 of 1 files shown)


jvican (approved):
LGTM :+1: I'm curious, do we have any performance numbers after this change? It makes sense to use this approach to do less work than before, but some insights into how this makes vert.x fare speed wise would be welcome.

essobedo:
@jvican thx for the revue and the feedback. 

About you question, well, as usual it depends on the use case. If you mostly write, CAS operations are more costly than synchronized blocks especially with a lot of concurrent threads but if you mostly read (what I assumed), `volatile` reads are much more cheaper than synchronized blocks.

For a better understanding, let's say that we want to compare

```
public boolean isClosed1() {
    return state.get() == CLOSED;
}
```
with
```
public synchronized boolean isClosed2() {
    return closed;
}
```

With one thread on my local machine (MBP 2,8 GHz Intel Core i7), I get:

```
Benchmark                     Mode   Samples         Mean   Mean error    Units
o.s.MyBenchmark.isClosed1     avgt        10        1,303        0,007    ns/op
o.s.MyBenchmark.isClosed2     avgt        10        2,902        0,185    ns/op
```

So more than **twice faster**

With max threads (`Runtime.getRuntime().availableProcessors()`), I get:

```
Benchmark                     Mode   Samples         Mean   Mean error    Units
o.s.MyBenchmark.isClosed1     avgt        10        3,723        0,255    ns/op
o.s.MyBenchmark.isClosed2     avgt        10      666,265       57,456    ns/op
```

So more than **200 times faster**

On the other hand, If I want to compare the current `init` method with this new version under the same conditions, I get:

One Thread:

```
Benchmark                     Mode   Samples         Mean   Mean error    Units
o.s.MyBenchmark.init1         avgt        10       16,258        0,300    ns/op
o.s.MyBenchmark.init2         avgt        10        8,220        0,199    ns/op
```

So **twice slower**

Max Threads:

```
Benchmark                     Mode   Samples         Mean   Mean error    Units
o.s.MyBenchmark.init1         avgt        10     1686,056      232,578    ns/op
o.s.MyBenchmark.init2         avgt        10      456,712       21,903    ns/op
```

So more than **3 times slower**.

To summarize, it worths it if concurrent calls to `isClosed` is more likely than concurrent calls to `init` what I assumed.

I hope it answered your question. Assuming that you still believe that this PR worths it, it would be great If you could also review the other PR of the same type https://github.com/eclipse-vertx/vert.x/pull/3727  :-)

Issue #3434, opened 2020-06-03
Enable service loadable external ExecutorService Factory via SPI


hutchig:
#### Describe the feature

Provide an SPI Interface for an executor service that can be used to fulfill executeBlocking requests and Java service load and use an implementation of the SPI that is found on the classpath at runtime.

#### Use cases

I want to be able to run a Vert.x application in a managed environment that makes use of Vert.x in its implementation to run 'blocking' work asynchronously but I need to be able to control various aspects of the threads and threadpool that is used in order to be able to maintain my requirements on system management, tuning and so on.

As I do not have direct control of the application source code I want to be able to get
Vert.x to use my executor service without making changes to the current Vert.x APIs
or requiring the application to call any new APIs.

I do not want to have a large surface area of new API's for tuning and thread control
in Vert.x that I have to get approved, implement and then call and I also want to ability to handle
more than one Vert.x application that issues executeBlocking calls in the JVM at
one time. I want to be able to externalize my configuration in a way that is idiomatic
to my environment and this can be acheived by encapsulating this in my own implementation
of the SPI.

Similarly, I do not think I have a requirement to interject at the level of the Context class
(or the class that implements executeBlocking). I want to continue to have Vert.x handle
such issues as 'ordered', metrics plug-in points, handler callbacks and so on. 

However, SPI insertion 'after' task queueing is too low level, as to maintain recognisable interfaces in the SPI we would probably 'wrap' any Callable/Runnable that gets submitted inorder to retrieve what contexts we needed when it was run.

It may make sense to have the SPI as a vanilla ExecutorService and leave in place
the Vert.x WorkerPool in order for that to handle metrics or other related requirements
but have WorkerPool::executor return the service loaded ExecutorService that has
been loaded off the classpath.

##### OpenLiberty embedding Vert.x

The first use case I have for this is to use it to enable the `@Blocking` annotation in
SmallRye Reactive Messaging 2.0+ in the context of the Liberty application server.

Note that as the service loaded ExecutorService may have a thread pool that grows
**and shrinks**, although Vert.x may schedule `ordered=true` blocking tasks one at
a time it may not be possible to run all tasks from a single Vert.x context on a the
same thread from the pool.

The SPI could be the standard j.u.c.ExecutorService but it is probably better to have 
a Vert.x specific interface that can be used to obtain the ExecutorService as well as to provide better support for 'named' worker pools being backed by the external ExecutorService.

#### Contribution

[...]

PR #3729, opened 2021-01-03
Service loadable executor prototype

hutchig:
Motivation:

My goal is to open a code based discussion for an approach that might address #3434. 
This have a Vert.x SPI that provides the ability to use a foreign ExecutorService
for worker tasks.

This is to enable easier consumption of SmallRye technologies that make use of Vert.x in
environments with custom ExecutorServices (in the first instance OpenLiberty).

This is a draft PR for now and I am very happy to make any further improvements or changes that might make
this change more likely to be adopted either while it is a draft or when it is turned into a 'standard' PR.


vietj:
can we have this PR onto master instead of 3.9 ?

hutchig:
Hi @vietj yes certainly I will do a PR onto master directly now.

I did it to the 3.9 line initially because of https://github.com/smallrye/smallrye-reactive-messaging/blob/0225686abdc52dbec1253d8ae8c29531bcf70d3a/pom.xml#L60
which is the dependency for the specific use case I was tasked to solve.

Thank you very much for commenting on this.

vietj:
can we review this PR @hutchig ?

hutchig:
Yes, that would be great @vietj I went through the CI tests this morning and none of the failures seem to be related to this PR as far as I can see. Thank you. As always, happy to change anything at all or add tests, doc etc.

vietj:
I reran CI one the two build failed

vietj:
I think we could simplify a bit the implementation

1. currently the implementation uses `new ThreadPoolExecutor(..)` and `Executors.newFixedThreadPool` which seem to be similar

2. always use `ExecutorServiceFactory` to create a thread pool, so when no `ExecutorServiceFactory` is provided by the SPI then we use an internal implementation that returns a fixed thread pool.

hutchig:
The two different forms were there in the code already and I did not want to alter the existing behaviour - I will make it as you suggest above, thanks for the feedback @vietj .

hutchig:
@vietj For "always use ExecutorServiceFactory to create a thread pool," I can use my head and not apply that to transport.eventLoopGroup but will take what you say and apply it to the internalBlockingExec.

vietj:
@hutchig yes indeed that's the point

src/test/java/io/vertx/test/core/LogContainsRule.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+/**
+ * A generic log scanner used by testcases to check that messages have appeared
+ * in the JUnit test logs.
+ *
+ */
+public class LogContainsRule implements TestRule {
```
vietj, line 35: we have a similar rule already existing : `io.vertx.test.core.BlockedThreadWarning`, can you reuse this class and perhaps make it more generic to it works in this case too ?

src/test/java/io/vertx/core/spi/executor/ExecutorFactoryMetricsTest.java
```diff
@@ -0,0 +1,54 @@
...
+ * 
+ * It performs the same tests as MetricsContextTest but using an external
+ * {@link java.util.concurrent.ExecutorService} for worker tasks
+ * 
+ * To enable different tests to have particular ExecutorServiceFactory's (or
+ * none), we only put the particular ExecutorServiceFactory .class file on the
+ * classpath and make it service loadable as part of each test and are careful
+ * to remove it afterwards.
+ */
+public class ExecutorFactoryMetricsTest extends MetricsContextTest {
```
vietj, line 38: I don't think we need this as it tests metrics behavior.

src/test/java/io/vertx/core/spi/executor/Utils.java
```diff
@@ -0,0 +1,187 @@
...
+
+import org.junit.rules.ExternalResource;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+import io.vertx.test.core.LogContainsRule;
+
+public class Utils {
```
vietj, line 39: ServiceHelperTest is using also Java compiler to do something similar, can you update the ServiceHelperTest to use this class instead?

src/test/java/io/vertx/test/core/LogContainsRule.java
hutchig: Yes, will do.

src/test/java/io/vertx/core/spi/executor/ExecutorFactoryMetricsTest.java
hutchig: Agreed.

src/test/java/io/vertx/core/spi/executor/Utils.java
hutchig: Yes OK will do, probably tomorrow now.

vietj:
in master I removed this compilation code and replaced with maven compiler
execution:
https://github.com/eclipse-vertx/vert.x/commit/82e0c4d690146c59c62439697ddba575b281c4e0

src/test/java/io/vertx/test/core/LogContainsRule.java
hutchig: Done. (Hope @tsegismont is OK with the author tag being left as I learned how to do the superclass from this one.)

src/test/java/io/vertx/core/spi/executor/ExecutorFactoryMetricsTest.java
hutchig: Done.

src/test/java/io/vertx/core/spi/executor/Utils.java
hutchig: Having looked at 82e0c4d more closely, does the above still stand or are you going to backport
what master does?

hutchig: I chose that it was part of the test to move SPI service classes on and off the classpath in the tests to try to make them more realistic - that is why I did not have them compiled and on the classpath all at once. However, its true that the ServiceLoader provider configuration files alone could be used and it is not our job to test Java ServiceLoading. I copied the style and code from one that Clement wrote.

@vietj What would you like me to do:

1. backport 82e0c4d and eliminate the compilation from the tests in my PR
2. as mentioned above, make 3.9 ServiceHelper and my PR tests use a common Utils method (maybe relocated out of executor/Utils to a more generic location)
3. nothing, as you are going to backport 82e0c4d at some point
?

hutchig:
I was asked to check with @cescoffier (with an '@' here) if he knew of any reason why this would be incompatible with the `@Blocking` in SmallRye Reactive messaging.

cescoffier:
@hutchig it should be fine.

src/test/java/io/vertx/core/spi/executor/Utils.java
vietj: option 1. is my favorite, it's a relevant cleanup

hutchig: That's fine by me. I will learn more as it was doing mnv stuff I have not done that way before
and I agree it is more efficient not to compile 'in' the test.
I will push another commit on Monday.

hutchig:
CI Test failed on:
```
Failed tests: 
  EventBusRegistrationRaceTest.theTest:82->AsyncTestBase.assertEquals:235 expected:<0> but was:<1>
Tests run: 4287, Failures: 1, Errors: 0, Skipped: 14
```

pom.xml
```diff
@@ -571,6 +571,62 @@
...
+              <goal>testCompile</goal>
+            </goals>
+            <configuration>
+              <compileSourceRoots>
+                <compileSourceRoot>src/test/classpath/servicehelper</compileSourceRoot>
+              </compileSourceRoots>
+              <outputDirectory>${project.build.directory}/classpath/servicehelper</outputDirectory>
+            </configuration>
+          </execution>
+          <execution>
```
hutchig, line 591: The reason that there is a second clause here is that there is a pattern in the test code of creating a new classloader and adding a directory such as the  _${project.build.directory}/classpath/servicehelper_ above to it's classpath. If I had a single folder, all the test classloaders would have combined classpaths and in order to seperate them, the smallest chunk of XML that I can duplicate is a an <execution>.

src/test/java/io/vertx/core/spi/executor/Utils.java
hutchig: Done.

vietj:
is it ready for another review @hutchig

hutchig:
Yes please, that would be great @vietj thank you.

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -1082,6 +1087,13 @@ void release() {
     }
   }
 
+  public class DefaultExecutorServiceFactory implements ExecutorServiceFactory {
```
vietj, line 1090: can we instead use `FixedThreadPoolExecutorServiceFactory` that seems to be the same class ?

src/test/java/io/vertx/core/spi/executor/Utils.java
```diff
@@ -0,0 +1,126 @@
...
+   * methods
+   */
+  static Method EXECUTE = null;
+  static Method SHUTDOWN_NOW = null;
+  static {
+    try {
+      EXECUTE = ExecutorService.class.getMethod("execute", Runnable.class);
+      SHUTDOWN_NOW = ExecutorService.class.getMethod("shutdownNow");
+    } catch (NoSuchMethodException | SecurityException e) {
+      throw new IllegalStateException(e);
```
vietj, line 54: I think that will prevent the class loading instead we should set null fields and fail in the other methods with an assertion that check the fields are not null before using them or throwing an IllegalStateException at this place

src/main/java/io/vertx/core/impl/VertxImpl.java
vietj: my bad, FixedThreadPoolExecutorServiceFactory is in test package, forget it.

vietj:
beside the minor static initializer throwing IllegalStateException it seems fine to me.

src/test/java/io/vertx/core/spi/executor/Utils.java
hutchig: Ok I will make it so. Originally the methods parameters were not typed, only string names but what you suggest is better now they are the actual method types.

hutchig: Done, is ready for another review @vietj I think. Thank you again.

hutchig:
fixes #3434

hutchig:
I could hug you, thank you.

vietj:
I will port this to master with the new service provider

PR #3731, opened 2021-01-04
Connection pool alternative implementation

vietj:
A new connection pool implementation that is able to handle multiple event loops and maintain some of the existing pool features.

- [x] decouple HTTP connection from current pool implementation
- [x] move the responsibility of recycling connection to the borrower and not the connection listener for the current pool
- [x] pool API with simple lock based implementation
- [x] simple lock implementation
- [x] lock free implementation
- [x] HTTP client integration
- [x] [SQL client integration](https://github.com/eclipse-vertx/vertx-sql-client/pull/874)
- [x] [Mail client integration](https://github.com/gaol/vertx-mail-client/tree/new_connpool_from_4.0)
- [x] [Redis integration](https://github.com/vert-x3/vertx-redis-client/pull/287)
- [x] Waiter cancellation
- [x] Connection selection strategy
- [x] Connection capacity increase
- [x] Cleanup
- [x] Documentation
- [x] HTTP client LIFO policy


vietj:
@gaol you can look at integrating the new pool for mail-client

vietj:
and provide feedback on this WIP too

gaol:
@vietj added to my list :)

vietj:
rebased on master

src/main/java/io/vertx/core/net/impl/clientconnection/Lease.java
```diff
@@ -0,0 +1,19 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.net.impl.clientconnection;
+
+public interface Lease<C> {
```
gaol, line 13: Missing Javadoc

src/main/java/io/vertx/core/net/impl/pool/Connector.java
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.net.impl.pool;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Handler;
+import io.vertx.core.impl.EventLoopContext;
+import io.vertx.core.net.impl.clientconnection.ConnectResult;
+
+public interface Connector<C> {
```
gaol, line 18: Missing Javadoc

src/main/java/io/vertx/core/net/impl/pool/ConnectionPool.java
```diff
@@ -0,0 +1,78 @@
...
+public interface ConnectionPool<C> {
+
+  /**
+   * Acquire a connection from the pool.
+   *
+   * @param context the context
+   * @param weight the weight
+   * @param handler the callback
+   */
+  void acquire(EventLoopContext context, int weight, Handler<AsyncResult<Lease<C>>> handler);
```
gaol, line 42: `EventLoopContext` prevents the cases that maybe `DuplicateContext`, please refer to [pr comment to mail-client](https://github.com/vert-x3/vertx-mail-client/pull/160#issuecomment-768231304) for more detail.

gaol (commented):
It will be better to have docs about the difference between `io.vertx.core.net.impl.pool.ConnectionPool` and `io.vertx.core.net.impl.clientconnection.Pool`, and how to choose the pool to use for a specific case.

src/main/java/io/vertx/core/net/impl/pool/NonBlockingSynchronization1.java
```diff
@@ -0,0 +1,36 @@
...
+  private final S state;
+
+  public NonBlockingSynchronization1(S state) {
+    this.state = state;
+  }
+
+  @Override
+  public void execute(Action<S> action) {
+    q.add(action);
+    if (s.incrementAndGet() == 1) {
```
franz1981, line 19: I would do instead:
```java
      if (s.get() > 0) {
         return;
      }
      while (s.compareAndSet(0, 1)) {
         try {
            Action<S> a;
            while ((a = q.poll()) != null) {
               Runnable post = a.execute(state);
               if (post != null) {
                  post.run();
               }
            }
         } finally {
            s.set(0);
            // full barrier here:
            // the q emptiness check cannot happen before 
            if (q.isEmpty()) {
               return;
            }
         }
      }
```
This should never let any actions to be dropped, awaiting to be executed forever, *but* it has another issue: fairness.

If other threads keep on submitting actions the risk is that a single thread will take the whole execution hit without completing, leading to a huge time occupying its own `Thread`.
If the executor is running on the event loop it means that no other tasks will be executed on that thread..

vietj: thanks.

the fairness si fine because it is for a pool with a bound number of requests to be performed. worth case I would set a threshold and when too many actions are performed then a task is executed on the event loop to continue processing.

vietj: Note that I think we will not keep this implementation but instead use the `NonBlockingSynchronization2` that we discussed, perhaps you can have a look as well ?

src/main/java/io/vertx/core/net/impl/pool/ConnectionPool.java
vietj: can you for now extract the event loop context out of the DuplicateContext with instanceof ?

vietj:
@gaol after this pool is merged we will simply drop the previous pool as this one should achieve the same but in a better way

src/main/java/io/vertx/core/net/impl/pool/ConnectionPool.java
gaol: `DuplicateContext` is a package level class.

src/main/java/io/vertx/core/net/impl/pool/NonBlockingSynchronization2.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.netty.util.internal.PlatformDependent;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class NonBlockingSynchronization2<S> implements Synchronization<S> {
+
+  private final Queue<Action<S>> q = PlatformDependent.newMpscQueue(Integer.MAX_VALUE);
```
franz1981, line 11: Change this with `PlatformDependent::newMpscQueue` with no args to use an unbounded q

src/main/java/io/vertx/core/net/impl/pool/NonBlockingSynchronization2.java
```diff
@@ -0,0 +1,43 @@
...
+    this.state = state;
+  }
+
+  @Override
+  public void execute(Action<S> action) {
+    q.add(action);
+    while (true) {
+      int v = s.get();
+      if (v == 0) {
+        if (s.compareAndSet(0, 1)) {
```
franz1981, line 25: I see an unecessary op here: it's ok just for the very first attempt of `s.compareAndSet(0, 1)` that we fail fast and just move on, terminating `execute`, while is mandatory for the other subsequent loop iteration that we check for `q::isEmpty`. 
I would group 
```java
          Action<S> a;
          while ((a = q.poll()) != null) {
            Runnable post = a.execute(state);
            if (post != null) {
              post.run();
            }
          }
```
into its own method eg `pollAndExecute` and will separate the 2 cases:
```java
q.add
if (s.get() != 0 || !s.compareAndSet(0, 1)) {     
    return;
}
do {
    try {
        pollAndExecute(action);
    } finally {
        s.set(0);
    }
} while (!q.isEmpty() && s.compareAndSet(0,1));
```

franz1981 (changes requested):

src/main/java/io/vertx/core/net/impl/pool/NonBlockingSynchronization2.java
franz1981: Just thinking loud and not related with this PR, but if we pack the loop logic:
```java
do {
    try {
        pollAndExecute(q, action);
    } finally {
        s.set(0);
    }
} while (!q.isEmpty() && s.compareAndSet(0, 1));
```
into a separate method ie `lockedPollAndExecute` and we get an event loop group instance, we can just send back a continuation (with s.get() == 1, to save other threads to try the same) instead of keeping a single event loop thread busy for "too long" (and I would add a check with a configurable busy time eg 100 us).

src/test/benchmarks/io/vertx/benchmarks/SynchronizationBenchmark.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @TearDown
+  public void tearDown() {
+    latch.countDown();
+  }
+
+  @Benchmark
+  public void lock1() {
+    synchronization1.execute(action);
```
franz1981, line 92: We're interested when a specific action is going to be completed too, because, if not, we just risk to measure the cost of submission and just using an unbunded q to accumulate tasks would perform best then any possible implementation (until OOM).
In addition, this bench need repeatable condition and using RTT can achieve steady state.
I would change it like this:
```java
@Param
int burstSize;

@Param
int producerDelay;

// consumerDelay is used on action execution to simulare real CPU work
@Param 
int consumerDelay;

// create an enum that use this to inject a specific synchronization impl instead of using N different methods
@Param
private String lockType;

// PerThreadState is a Thread scoped state that hold resettable actions
@Benchmark
public void lock(PerThreadState state) {
   if (producerDelay > 0) {
      BlackHole.consumeCPU(producerDelay);
   }
   // this is going to execute all actions and spin wait completion -> can use a single AtomicLong somewhere
   state.executeActionsAndWaitCompletion(sync);
}
```

franz1981 (changes requested):

src/test/benchmarks/io/vertx/benchmarks/SynchronizationBenchmark.java
vietj: it is actually the intention to measure the submission cost to compare the two implementations. I think what you describe can be another benchmark.

franz1981: Yep but the submission cost depends a lot if the queue is streaming or is near empty...and that's not reproducible because depends by the backlog of tasks...
Benchs like this work better while having a steady state and the RTT burst ones can help to reach it, that's why I have proposed it :)

src/main/java/io/vertx/core/net/impl/pool/ConnectionPool.java
vietj: I gave some thoughts to this and in this case I think we need to address this in the mail client.

I believe the mail client should distinguish the context of the TCP connection and the context of the client interacting with the mail asynchronous API. 

Both are separate, specially with pooled connections, e.g a client uses a pooled connection, the connection is created with the context of this client, the connection goes back in the pool, another client comes and reuse this connection and the two contexts might be different. This is the shared pool scenario and this is something that we don't handle well in Vert.x 3 with Vert.x/Netty based connections and we are addressing in Vert.x 4.

The current example of this is the SQL/HTTP client that behave differently in V3/V4, assuming two verticles are using a shared client:
- in V3, a verticle can acquire a connection created by another verticle and get a callback on the other verticle context
- in V4, the callback will happen on the correct context

The solution to this is that the connection and the client API have each their own context and both use message passing to communicate. Of course if the context is the same, then this is a straight execution.

You can check Http1xServerRequest/Http1xConnection to see how it operates. We need I think first to refactor the mail client to behave likewise and then the usage of this pool will be more natural.

gaol: Thanks for the info, I think it can be addressed in mail client.

vietj:
@gaol I refactored a bunch of stuff in the pool, so you need to update your WIP

gaol:
> @gaol I refactored a bunch of stuff in the pool, so you need to update your WIP

done :)

vietj:
thanks Lin!

gaol:
@vietj will `pool.close()` reset the `pool.size()` to `0` ? please refer to the test case [here](https://github.com/gaol/vert.x/blob/2f0bec87acc702b74bb3d9eda7e899359c02df87/src/test/java/io/vertx/core/net/impl/pool/ConnectionPoolTest.java#L446-L451).

vietj:
thanks for the feedback, indeed it was not, now it does with a simple fix.

Btw you should not rely on the size() operation of the pool to take
decisions, this operations is mostly for testing purpose. Using size() can
lead to data races.

src/main/java/io/vertx/core/http/impl/SharedClientHttpStreamEndpoint.java
```diff
@@ -0,0 +1,157 @@
...
+  @Override
+  public boolean isValid(HttpClientConnection connection) {
+    return connection.isValid();
+  }
+
+  void checkExpired() {
+    pool.evict(conn -> !conn.isValid(), ar -> {
+      if (ar.succeeded()) {
+        List<HttpClientConnection> lst = ar.result();
+        lst.forEach(HttpConnection::close);
```
gaol, line 95: Since `pool.evict()` removes the connections from pool, and  `HttpConnection.close()` tries to remove it from pool again via the `evictHandler, (listener.onRemove())`, does it make sense to mark the evicted connections as `evicted = true` here, and call `decRefCount()` to keep the ref counts consistent in the Endpoint ?

vietj: I don't think so because Http1xClientConnection/Http2ClientConnection have an `evicted` flag to do this.

vietj: I think we could add a test for HTTP/1.x that uses a short keep alive timeout to evict connection from the pool and when connection is closed, a new request is made to create a new connection. This way we can ensure that the endpoint is never depleted ?

gaol: > I don't think so because Http1xClientConnection/Http2ClientConnection have an evicted flag to do this.

When the connection is evicted using `pool.evict()`, the `evicted` flag of the connection is still `false`, because the `evictHandler` is not invoked yet, it is called on `HttpConnection::close`

gaol: > I think we could add a test for HTTP/1.x that uses a short keep alive timeout to evict connection from the pool and when connection is closed, a new request is made to create a new connection. This way we can ensure that the endpoint is never depleted ?

Sorry that I may not make it clear, I think there is a [test case](https://github.com/eclipse-vertx/vert.x/pull/3731/commits/eb2f9427a2c2e4e316dc6d7e6331fa959d44af3a) to cover what I concerned.

It does not cause issues in mail client, I think it is because the [listernr.onRemove()](https://github.com/eclipse-vertx/vert.x/pull/3731/files#diff-7695048cda6da12d4c94190f135d7f25e864955644b466403e5179849f03d00cR65) doesn't do anything since[ pool.slots[removed.index] != removed](https://github.com/eclipse-vertx/vert.x/pull/3731/files#diff-47e2e3f2721793f39461349a5aa942a0f36118b3bcd405044cc2e62e9afbdc56R290) is true.

So we are good :)

vietj: Yes this case can happen but it is expected. Thanks for the feedback.

BTW if you can review the internal pool actions to find bugs, it would help a lot :-)

vietj:
@gaol another set of small API changes

vietj:
@gaol please pay attention to the weight that has been changed to a max size + kind

src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
```diff
@@ -0,0 +1,781 @@
...
+    private ConnectSuccess(Slot<C> slot, ConnectResult<C> result, PoolWaiter<C> waiter) {
+      this.slot = slot;
+      this.result = result;
+      this.waiter = waiter;
+    }
+
+    @Override
+    public Runnable execute(SimpleConnectionPool<C> pool) {
+
+      int capacity = pool.capacityFactors[(int)result.weight()];
```
gaol, line 266: Shouldn't `result.weight()`be the `kind`?

src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
```diff
@@ -0,0 +1,781 @@
...
+
+    int[] capacities = new int[maxSizes.length];
+    int maxCapacity = 1;
+    int numSlots = 0;
+    for (int i = 0;i < maxSizes.length;i++) {
+      int maxSize = maxSizes[i];
+      if (maxSize < 1) {
+        throw new IllegalArgumentException();
+      }
+      maxCapacity *= maxSize;
```
gaol, line 205: Not sure if I understand here, if 
`Http1xConnection: {max-pool: 5, capacity: x}`, `Http2Connection: {max-pool: 2, capacity: y}`, should the `maxCapacity = 5x + 2y` ?

According to your equation, we get `maxCapacity = 5 * 2 = 10, http1x.capacity = 2, http2.capacity = 5`. added them up again, we get `maxCapacity = 20` if above sentence is correct. :)

src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
vietj: it actually is, it is just that result.weight() needs to be renamed after we remove the current connection pool the `ConnectResult` class belongs to.

src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
vietj: what we can get at most is:

- 2 HTTP/2 connections (5+5)
- 5 HTTP/1 connections (2+2+2+2+2)
- 1 HTTP/2 connection + 3 HTTP/1 connections (5+2+2+2)

vietj:
this is now ready for review

tsegismont (commented):
Great design @vietj !

src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
```diff
@@ -169,6 +169,11 @@ public boolean isValid() {
     return expirationTimestamp == 0 || System.currentTimeMillis() <= expirationTimestamp;
   }
 
+  @Override
+  public long lastResponseReceivedTimestamp() {
+    return 0L;
```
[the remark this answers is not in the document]
vietj: yes, but LIFO does not make much sense for HTTP/2 because of multiplexing

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor1.java
```diff
@@ -0,0 +1,51 @@
...
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.net.impl.pool;
+
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class CombinerExecutor1<S> implements Executor<S> {
```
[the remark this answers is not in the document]
vietj: actually I will rather remove this class as it is not used

src/main/java/io/vertx/core/net/impl/pool/SemaphoreExecutor.java
```diff
@@ -0,0 +1,38 @@
...
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.net.impl.pool;
+
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class SemaphoreExecutor<S> implements Executor<S> {
```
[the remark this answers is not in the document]
vietj: I think indeed we should remove it

src/main/java/io/vertx/core/net/impl/pool/ConnectionPool.java
```diff
@@ -0,0 +1,125 @@
...
+   * @param handler the callback handler with the result
+   */
+  void acquire(EventLoopContext context, int kind, Handler<AsyncResult<Lease<C>>> handler);
+
+  /**
+   * Acquire a connection from the pool.
+   *
+   * @param context the context
+   * @param listener the waiter event listener
+   * @param kind the connection kind wanted
```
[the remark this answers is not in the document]
vietj: I will

src/main/java/io/vertx/core/http/impl/SharedClientHttpStreamEndpoint.java
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+class SharedClientHttpStreamEndpoint extends ClientHttpEndpointBase<Lease<HttpClientConnection>> implements PoolConnector<HttpClientConnection> {
+
+  /**
+   * LIFO pool selector.
+   */
+  private static final BiFunction<PoolWaiter<HttpClientConnection>, List<PoolConnection<HttpClientConnection>>, PoolConnection<HttpClientConnection>> LIFO_SELECTOR = (a, b) -> {
```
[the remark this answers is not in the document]
vietj: right :-)

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -633,11 +607,8 @@ private void request(
             if (followRedirects != null) {
               req.setFollowRedirects(followRedirects);
             }
-            if (timerID >= 0L) {
-              if (!vertx.cancelTimer(timerID)) {
-                req.reset(0);
-                return;
-              }
+            if (timeout > 0L) {
+              // Maybe later ?
```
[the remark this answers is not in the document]
vietj: it is called because it will act then when no data is received when the connection is used.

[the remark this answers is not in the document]
vietj: let me elaborate, previously the pool did not have timeout so this call was done before getting a connection from the pool and when the timeout was firing the connection was delivered to a waiter anyway.

now that the pool supports timeout, the set timeout is used when acquiring the connection. when the connection is acquired we need then to set a timeout to signal when no data has been received.

src/main/java/io/vertx/core/net/impl/pool/SemaphoreExecutor.java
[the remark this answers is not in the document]
vietj: we will keep it for a while, actually I would like to write a benchmark for it.

vietj:
I will merge this pR and then the SQL client one.

zekronium:
Does the pool allow different instances of httpclient(via WebClient) to share connections amongst each other?

If so does that happen automatically?

vietj:
not sure what you mean here.

Vert.x 4.0 allows WebClient/HttpClient to be shared among various
event-loop / worker safely but will use a single event loop that is
assigned when the pool is created.

The new 4.1 pool does the same however, connections can have various
event-loop provided by the caller.

zekronium:
Got it, so now the httpclient will execute the request on the callers eventloop and not on the "creators" eventloop.

My question was that some other httpclients (for example apache or OkHttp) have a decoupled ConnectionPool that you can share amongst different instances of httpclient. That comes in handy when different instances(with different configurations for example) might also hit the same endpoints hence reusing the connection some other httpclient created but currently released. Also, when the client gets closed it leaves its connections free so other clients can use them if needed.

knotenpunkt:
> The new 4.1 pool does the same however, connections can have various event-loop provided by the caller.

@vietj can have, but not must have? how you can configure this behavior?

and one more questions: which way is faster, creating for each context/verticle an own pool or use this shared pool?
i think the messaging-overhead is in the one-pool for different contexts/verticle a little bit higher?

maybe you can add some notes to the documentation to the new pool-behavior, where to create the pools and how to use them 
and what are the benefits or none-benefits if you use one pool vs a lot of pools

laurentvaills:
We would also be very interested in the answer to the questions raised by @knotenpunkt , to help us in our migration to Vert.x 4.1 .

vietj:
Hi,

here is the pool behavior when a connection is needed

1/ use an available connection with the same event-loop
2/ otherwise try to create a connection with the same event-loop if there
is room for creating a connection to the database
3/ otherwise reuse any available connection
4/ otherwise get queued until a connection is available

In addition if there is pipelining at the pool level (i.e using pool
operations) then the step 1/ will chose the connection with the least
number of inflight queries.

So if you use the new pool like the previous pool (i.e confined in a
verticle) then it will behave like previously.

When you are sharing a pool between verticles, then the pool will use the
caller event-loop when it needs one to create a connection, so if 3
verticles are using a pool then it will contain connections having a mix of
event loop of these three verticles.

Performance wise, I believe it will not make a difference because the
latency to the database is always the dominant factor. It might be
different on a benchmark (like TechEmpower) which are very specific
workloads. So the main benefit is to distribute the database load on
several event loops.

We might add in the future createShared() methods that will behave like
above but instead of using the caller event-loop it will use all the
available event-loops of the vertx instance (it will likely be
configurable).

I will add this to the javadoc.

Julien

Issue #3688, opened 2020-12-08
No message codec for type should probably return a Future.failedFuture not throw


doctorpangloss:
### Version

4.0.0.CR2

### Context

Sending a message on the event bus with a missing codec throws an exception instead of returning a failed future.

```
java.lang.IllegalArgumentException: No message codec for type: class com.hiddenswitch.framework.impl.ConfigurationRequest
	at io.vertx.core.eventbus.impl.CodecManager.lookupCodec(CodecManager.java:101)
	at io.vertx.core.eventbus.impl.EventBusImpl.createMessage(EventBusImpl.java:232)
	at io.vertx.core.eventbus.impl.EventBusImpl.request(EventBusImpl.java:109)
	at io.vertx.core.eventbus.EventBus.request(EventBus.java:81)
```

### Do you have a reproducer?

```
class SomeClass {
}
Vertx.vertx().eventBus().publish("1", new SomeClass());
```

vietj:
that's a fair point

tsegismont:
I gave some thought to this and I don't believe we should change anything.

First, this is not a regression in Vert.x 4 but the usual behavior of `EventBus`.

More importantly, a missing codec is a fatal issue which can't be recovered: something wrong happened during packaging or the app initialization code is broken. Such problems are usually caught during testing.

PR #3733, opened 2021-01-05
Allow for SPI provided ExecutorServiceFactory

hutchig:
Signed-off-by: Gordon Hutchison <Gordon.Hutchison@gmail.com>

Motivation:

My goal is to open a code based discussion for an approach that might address #3434. 
This have a Vert.x SPI that provides the ability to use a foreign ExecutorService
for worker tasks.

This is to enable easier consumption of SmallRye technologies that make use of Vert.x in
environments with custom ExecutorServices (in the first instance OpenLiberty).

This is a draft PR for now and I am very happy to make any further improvements or changes that might make
this change more likely to be adopted either while it is a draft or when it is turned into a 'standard' PR.

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -780,6 +780,16 @@ Worker executors can be configured when created:
 
 NOTE: the configuration is set when the worker pool is created
 
+You can provide your own `ExecutorService` that backs the worker executors by implementing the `ExecutorServiceFactory`
+interface and making your implementation loadable by the Java `ServiceLoader` mechanism. Your
+`ExecutorService` should use the Vert.x `ThreadFactory` provided on the `createExecutor` call to create
+threads in the worker pool and respect the maximum concurrency limits Vert.x passes. 
+
+[source,$lang]
+----
+{@link examples.spi.executor.FixedThreadPoolExecutorServiceFactory#createExecutor}
+----
+
 == Metrics SPI
 
 By default Vert.x does not record any metrics. Instead it provides an SPI for others to implement which can be added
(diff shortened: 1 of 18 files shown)


hutchig:
ToDos
- [ ] Create a commit to reduce any line changes in imports reordering from Eclipse (VertxImpl.java)
- [ ] Resolve any documentation changes required into this or a separate PR

hutchig:
Dear @vietj, I think the best place for my doc would be in https://vertx.io/docs/vertx-core/java by adding some text at the end of https://vertx.io/docs/vertx-core/java/#blocking_code , via a 'src/main/asciidoc/index.adoc' change. Would that make sense to you, or somewhere better? If so I will add a commit to this PR or at your preference.

(The other alternative I thought about might be having a sentence in the https://vertx.io/docs/#integration section with the same style as the others and a short bit of text in a folder like https://vertx.io/docs/vertx-embedding/ but the sections in '#integration' look like they all refer to add-ons that are separately addressable by maven and the 'integration' is not really the same sort as embedding vert.x in something else. I was considering this sort of approach as it would be a 'place' where multiple issues that are relevant to embedding Vert.x might be found and using a foreign executor might be just one. Of course I could do a https://vertx.io/docs/vertx-embedding when a second interesting issue is identified but I think  the docs for things like vertx.logger-delegate-factory-class-name are already good as they are. I am also aware of http://www.julienviet.com/advanced-vertx-guide/#_embedding-vert-x . That document has a lot of good material in it and it would be good for it to have a safe/permanent home (which it may have already, I don't know :-) ).

hutchig:
See also #3688 3749

hutchig:
This PR should not be merged as it is. It is written to the 3.x approach.
In 4.x
 _the proposal is that when Vertx.vertx() is called by reactive messenging, 
then a service loadable class is invoked and can customize the returned vertx instance
and can declare to use a specific worker pool
instead of the default one_
I will close this for now to be tidy and prevent merging consideration.
Work will progress on #3729

PR #3744, opened 2021-01-12
Add replace and replaceIfPresent with ttl to AsyncMap api

zyclonite:
Motivation:

* Add `replace` and `replaceIfPresent` with `ttl` to AsyncMap api
* Fix a small bug in `putIfAbsent` where the timer for the new object is only removed if the old had a ttl set
* Add tests for new methods


src/main/java/io/vertx/core/shareddata/AsyncMap.java
```diff
@@ -167,6 +167,23 @@ default void replace(K k, V v, Handler<AsyncResult<@Nullable V>> resultHandler)
...
+   * @param resultHandler  the result handler will be passed the previous value
+   */
+  default void replace(K k, V v, long ttl, Handler<AsyncResult<@Nullable V>> resultHandler) {
+    replace(k, v, ttl).onComplete(resultHandler);
+  }
+
+  /**
+   * Same as {@link #replace(K, V, long, Handler)} but returns a {@code Future} of the asynchronous result
+   */
+  Future<@Nullable V> replace(K k, V v, long ttl);
```
tsegismont, line 185: This is a new API method, we should provide a default impl that returns a `Future` failed with `UnsupportedOperationException`.

src/main/java/io/vertx/core/shareddata/AsyncMap.java
```diff
@@ -184,6 +201,24 @@ default void replaceIfPresent(K k, V oldValue, V newValue, Handler<AsyncResult<B
...
+   * @param resultHandler the result handler
+   */
+  default void replaceIfPresent(K k, V oldValue, V newValue, long ttl, Handler<AsyncResult<Boolean>> resultHandler) {
+    replaceIfPresent(k, oldValue, newValue, ttl).onComplete(resultHandler);
+  }
+
+  /**
+   * Same as {@link #replaceIfPresent(K, V, V, long, Handler)} but returns a {@code Future} of the asynchronous result
+   */
+  Future<Boolean> replaceIfPresent(K k, V oldValue, V newValue, long ttl);
```
tsegismont, line 220: This is a new API method, we should provide a default impl that returns a `Future` failed with `UnsupportedOperationException`.

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncMapImpl.java
```diff
@@ -93,9 +93,7 @@ private void removeIfExpired(K k) {
     long timerId = vertx.setTimer(ttl, l -> removeIfExpired(k));
     Holder<V> existing = map.putIfAbsent(k, new Holder<>(v, timerId, ttl, timestamp));
     if (existing != null) {
-      if (existing.expires()) {
-        vertx.cancelTimer(timerId);
-      }
+      vertx.cancelTimer(timerId);
```
tsegismont, line 96: Nice catch! Can you please:

- add a new test case for this
- file a separate issue to keep track of the bug (that we will close by merging this PR)

We use issues to generate the release notes so a dedicated issue will make it clear that a bug was fixed.

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncMapImpl.java
```diff
@@ -148,6 +163,25 @@ private void removeIfExpired(K k) {
...
+    long timerId = vertx.setTimer(ttl, l -> removeIfExpired(k));
+    Holder<V> h = new Holder<>(newValue, timerId, ttl, timestamp);
+    Holder<V> result = map.computeIfPresent(k, (key, holder) -> {
+      if (holder.value.equals(oldValue)) {
+        if (holder.expires()) {
+          vertx.cancelTimer(holder.timerId);
+        }
+        return h;
+      }
+      vertx.cancelTimer(timerId);
```
tsegismont, line 179: You cannot do this here because the handler may be called several times if concurrent modifications are executed.

src/test/java/io/vertx/core/shareddata/AsyncMapTest.java
```diff
@@ -457,6 +457,19 @@ public void testMapReplaceClusterSerializableObject() {
     testMapReplace(new SomeClusterSerializableObject("foo"), new SomeClusterSerializableObject("bar"), new SomeClusterSerializableObject("quux"));
   }
 
+  @Test
+  public void testMapReplaceTtl() {
+    getVertx().sharedData().<String, String>getAsyncMap("foo", onSuccess(map -> {
```
tsegismont, line 462: Can you also test the other case (`replace` returns a `Future` completed with previous value)?

src/test/java/io/vertx/core/shareddata/AsyncMapTest.java
```diff
@@ -529,6 +542,22 @@ public void testMapReplaceIfPresentClusterSerializableObject() {
     testMapReplaceIfPresent(new SomeClusterSerializableObject("foo"), new SomeClusterSerializableObject("bar"), new SomeClusterSerializableObject("quux"));
   }
 
+  @Test
+  public void testMapReplaceIfPresentTtl() {
+    getVertx().sharedData().<String, String>getAsyncMap("foo", onSuccess(map -> {
```
tsegismont, line 547: Can you also test the other case (`replaceIfPresent` returns a `Future` completed with `false`)?

tsegismont (changes requested):
Thank you @zyclonite !

src/main/java/io/vertx/core/shareddata/AsyncMap.java
ruslansennov: Hi @tsegismont 
What about https://github.com/vert-x3/vertx-hazelcast/pull/132 ?
Can we do something similar with default method?

tsegismont: Hi, the purpose of providing such a default implementation is to avoid a breaking change when releasing. In the other PR, we're talking about a feature that, as far as we know, is only supported in Hazelcast.

So we can't do something similar, it's a different matter.

src/main/java/io/vertx/core/shareddata/impl/LocalAsyncMapImpl.java
zyclonite: i tried to test it now... old and new and it actually never happens because even if the new timer tries to delete the old record it cannot be deleted if not put with ttl (not expiring old)

so it's more a cosmetic cleanup

zyclonite: @tsegismont added the test case but should i still create an issue for it?

tsegismont: @zyclonite no, it's fine

tsegismont (approved):
LGTM, thank you @zyclonite

tsegismont:
@zyclonite after it's merged, would you mind taking care of follow-up PRs in `vertx-infinispan`, `vertx-hazelcast`, `vertx-ignite`? And file an issue for `vertx-zookeeper`?

zyclonite:
@tsegismont sure can do

tsegismont:
@zyclonite thank you!

zyclonite:
@tsegismont Sorry, I completely forgot about that PR

For hazelcast there is an open feature request
https://github.com/hazelcast/hazelcast/issues/10015

Zookeeper does only support setting a ttl on create

and for Infinispan and Ignite the PRs are referenced

tsegismont:
@zyclonite I'll wait for `master` branch to be the 4.1 baseline before merging. In the meantime, I'll review the PR for ISPN and Ignite. Thanks!

zyclonite:
@tsegismont i guess we can continue here or are there any open topics?

tsegismont:
Thanks for the heads-up @zyclonite !

PR #3742, opened 2021-01-10
New parser for length field frame protocols

EmadAlblueshi:
Signed-off-by: Emad Alblueshi <emad.albloushi@gmail.com>

Motivation:

The parser handles length fields byte(1), short(2), medium(3), int(4) and long(8) for field based protocols

addresses https://github.com/eclipse-vertx/vert.x/issues/3613 with simple API

Conformance:

Your commits should be signed and you should have signed the Eclipse Contributor Agreement as explained in https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md
Please also make sure you adhere to the code style guidelines: https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines

PR #3746, opened 2021-01-12
New parser for length field based protocols

EmadAlblueshi:

Motivation:

A new `LengthFieldParser` that handles length field based protocols it uses the `RecordParser` internally with simple API it supports the following :

1. Length fields `byte(1)`, `short(2)`, `medium(3)`, `int(4)` and `long(8)`.
2. Maximum frame length the default is `Integer.MAX_VALUE` and can be changed explicitly.
3. Length field offset to skip explicit length while parsing.

It's rewritten instead of https://github.com/eclipse-vertx/vert.x/pull/3742 as @vietj and @tsegismont suggested in [vertx-dev google group](https://groups.google.com/g/vertx-dev/c/iLX9aYgcRwU).

Thanks!

### DIFF
--- a/src/main/java/io/vertx/core/parsetools/LengthFieldParser.java
+++ b/src/main/java/io/vertx/core/parsetools/LengthFieldParser.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.parsetools;
+
+import io.vertx.codegen.annotations.Fluent;
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.Handler;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.parsetools.impl.LengthFieldParserImpl;
+import io.vertx.core.streams.ReadStream;
+
+/**
+ *  Parser for length field frame protocols
+ * * <p/>
+ *  The parser handles length fields byte(1), short(2), medium(3), int(4) and long(8) in the {@link Buffer}
+ *  with length field offset to skip explicit length while parsing.
+ * * <p/>
+ *  the default maximum frame length is {@link Integer#MAX_VALUE} and can be changed explicitly.
+ * * <p/>
+ *  The {@link #exceptionHandler(Handler)} is called when the frame exceeds the maximum length
+ *  or the length is invalid "less or equal to zero".
+ * * <p/>
+ * @author <a href="mailto:emad.albloushi@gmail.com">Emad Alblueshi</a>
+ */
+
+@VertxGen
+public interface LengthFieldParser extends ReadStream<Buffer>, Handler<Buffer> {
+
+
+  /**
+   * Create a new {@code LengthFieldParser} instance.
+   *
+   * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
+   */
+  static LengthFieldParser newParser(int length) {
+    return new LengthFieldParserImpl(length, 0, Integer.MAX_VALUE, null);
+  }
+
+  /**
+   * Create a new {@code LengthFieldParser} instance.
+   *
+   * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
+   * @param stream the wrapped of read stream
+   */
+  static LengthFieldParser newParser(int length, ReadStream<Buffer> stream) {
+    return new LengthFieldParserImpl(length, 0, Integer.MAX_VALUE, stream);
+  }
+
+  /**
+   * Create a new {@code LengthFieldParser} instance.
+   *
+   * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
+   * @param offset the offset of the field
+   * @param stream the wrapped of read stream
+   */
+  static LengthFieldParser newParser(int length, int offset, ReadStream<Buffer> stream) {
+    return new LengthFieldParserImpl(length, offset, Integer.MAX_VALUE, stream);
+  }
+
+  /**
+   * Create a new {@code LengthFieldParser} instance.
+   *
+   * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
+   * @param offset the offset of the field
+   * @param max the maximum length of the frame
+   * @param stream the wrapped of read stream
+   */
+  static LengthFieldParser newParser(int length, int offset, int max, ReadStream<Buffer> stream) {
+    return new LengthFieldParserImpl(length, offset, max, stream);
+  }
+
+  /**
+   * Create a new {@code LengthFieldParser} instance.
+   *
+   * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
+   * @param offset the offset of the field
+   * @param max the maximum length of the frame
+   */
+  static LengthFieldParser newParser(int length, int offset, int max) {
+    return new LengthFieldParserImpl(length, offset, max, null);
+  }
+
+  /**
+   * Create a new {@code LengthFieldParser} instance.
+   *
+   * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
+   * @param offset the offset of the field
+   */
+  static LengthFieldParser newParser(int length, int offset) {
+    return new LengthFieldParserImpl(length, offset, Integer.MAX_VALUE, null);
+  }
+
+  @Override
+  LengthFieldParser pause();
+
+  @Override
+  LengthFieldParser resume();
+
+  @Override
+  LengthFieldParser fetch(long amount);
+
+  @Fluent
+  LengthFieldParser endHandler(Handler<Void> endHandler);
+
+  @Fluent
+  LengthFieldParser handler(Handler<Buffer> handler);
+
+  @Fluent
+  LengthFieldParser exceptionHandler(Handler<Throwable> handler);
+
+}
--- a/src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
+++ b/src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
+ *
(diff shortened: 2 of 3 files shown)


src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
```diff
@@ -0,0 +1,141 @@
...
+              String err = frameLength <= 0 ? "Frame length : " + frameLength + " <= 0" :
+                "Frame length is too large current: " + frameLength + " max: " + max;
+              IllegalStateException ex = new IllegalStateException(err);
+              if (exceptionHandler != null) {
+                exceptionHandler.handle(ex);
+              } else {
+                throw ex;
+              }
+            } finally {
+              frameLength = 0;
```
tsegismont, line 77: It would be better to set the `RecordParser` handlers to `null`. This will ensure the underlying stream is closed eagerly.

src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+
+public class LengthFieldParserImpl implements LengthFieldParser {
+
+  private Handler<Throwable> exceptionHandler;
+  private final RecordParser parser;
+
+  private long frameLength = -1;
+  private final int length;
+  private final int offset;
```
tsegismont, line 32: The purpose of the offset is to skip any field in the frame that precedes the length field. But after the length has been read, the user might want to know what was the value of the skipped field.

Looking at the implementation below, it seems the user will only get fields that come after the length fields.

The user should be able to tell if they want to receive the whole frame (including the length field) or only the fields that come after the length-field. They could do it with a boolean param provided when creating this parser.

src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
```diff
@@ -0,0 +1,141 @@
...
+                exceptionHandler.handle(ex);
+              } else {
+                throw ex;
+              }
+            } finally {
+              frameLength = 0;
+            }
+          } else {
+            parser.fixedSizeMode((int) frameLength);
+            fetch(1);
```
tsegismont, line 81: Are you sure this is needed?

tsegismont (changes requested):
Thank you @EmadAlblueshi and sorry for the delay.

If we add this to Vert.x core, we also need the doc/examples to be updated.

src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
EmadAlblueshi: This is the weird part which confuses everyone. If I did not add it the `LengthFieldParser#fetch()` in general won't behave as expected!

src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
EmadAlblueshi: The parser won't work correctly in case of null stream is provided in the constructor :-) we should keep in mind that the current parser and `RecordParser` handle both cases.

1) If stream is null the exception is thrown and **keep reading**. 
2) If stream is not null an exception is thrown and **stop reading** then the user can handle the exception and close the stream explicitly.

I discovered this after heavy testing for the above two cases using `RecordParser` alone with fixed and delimited modes!

So the `frameLength = 0`  covers both cases accordingly.

src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
EmadAlblueshi: `skip = true`
``` 
 BEFORE PARSING (14 bytes)                     AFTER PARSING (12 bytes)
 +----------+----------+----------------+      +----------+-----+
 | offset   |  Length  | Actual Content |----->| Actual Content |
 | byte(1)  | byte(1)  | "HELLO, WORLD" |      | "HELLO, WORLD" |
 +----------+----------+----------------+      +----------+-----+
```
`skip = false`

```
 BEFORE PARSING (14 bytes)                      AFTER PARSING (14 bytes)
 +----------+----------+----------------+      +----------+----------+----------------+
 |  offset  |  Length  | Actual Content |----->|  offset  |  Length  | Actual Content |
 |  byte(1) | byte(1)  | "HELLO, WORLD" |      |  byte(1) | byte(1)  | "HELLO, WORLD" |
 +----------+----------+----------------+      +----------+----------+----------------+
```

@tsegismont right?

EmadAlblueshi: The first one is already covered if we simply provide a boolean ’skip = true’ 

but if ‘skip = false’ I need to know to add it asap.

tsegismont: I think we need to support the first two cases.

EmadAlblueshi: @tsegismont Done

EmadAlblueshi:
@tsegismont ping :-)

src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
EmadAlblueshi: Another thing that I should clarify is that when the user invokes `fetch(1)` the use won't get the first item in the queue even if there are more than one item so it's important to have it in the parser to keep the chronological order of the items that needs to be fetched once it's there.

src/main/java/io/vertx/core/parsetools/LengthFieldParser.java
```diff
@@ -19,84 +19,70 @@
...
-    return new LengthFieldParserImpl(length, 0, Integer.MAX_VALUE, stream);
-  }
-
   /**
    * Create a new {@code LengthFieldParser} instance.
    *
    * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
    * @param offset the offset of the field
-   * @param stream the wrapped of read stream
+   * @param skip skips prior field(s) of the frame body
```
tsegismont, line 48:
```suggestion
   * @param skip if set to {@code true}, only the frame body is emitted by the parse, if {@code false}, the whole frame is emitted
```

src/main/java/io/vertx/core/parsetools/LengthFieldParser.java
```diff
@@ -19,84 +19,70 @@
...
   }
-
   /**
    * Create a new {@code LengthFieldParser} instance.
    *
    * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
    * @param offset the offset of the field
-   * @param max the maximum length of the frame
-   * @param stream the wrapped of read stream
+   * @param skip skips prior field(s) of the frame body
```
tsegismont, line 58:
```suggestion
   * @param skip if set to {@code true}, only the frame body is emitted by the parse, if {@code false}, the whole frame is emitted
```

src/main/java/io/vertx/core/parsetools/LengthFieldParser.java
```diff
@@ -19,84 +19,70 @@
...
+    return new LengthFieldParserImpl(length, offset, skip, max, null);
   }
-
   /**
    * Create a new {@code LengthFieldParser} instance.
    *
    * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
    * @param offset the offset of the field
-   * @param max the maximum length of the frame
+   * @param skip skips prior field(s) of the frame body
```
tsegismont, line 69:
```suggestion
   * @param skip if set to {@code true}, only the frame body is emitted by the parse, if {@code false}, the whole frame is emitted
```

src/main/java/io/vertx/core/parsetools/LengthFieldParser.java
```diff
@@ -19,84 +19,70 @@
...
+  static LengthFieldParser newParser(int length, int offset, boolean skip, ReadStream<Buffer> stream) {
+    return new LengthFieldParserImpl(length, offset, skip, Integer.MAX_VALUE, stream);
   }
-
   /**
    * Create a new {@code LengthFieldParser} instance.
    *
    * @param length the length of the field : byte(1), short(2), medium(3), int(4) or long(8)
    * @param offset the offset of the field
+   * @param skip skips prior field(s) of the frame body
```
tsegismont, line 80:
```suggestion
   * @param skip if set to {@code true}, only the frame body is emitted by the parse, if {@code false}, the whole frame is emitted
```

src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
tsegismont: When you create a `Flowable` from a `ReadStream` with `vertx-rx`, we need a way to close the stream if the subscription is canceled. This is done by setting handlers to `null`.

So this method must be prepared to receive a `null` handler.

tsegismont (changes requested):
Hi @EmadAlblueshi , thanks for the updates. I was able to take another look today.

Can you update the very long test? I must admit I'm not prepared to review it. It seems many tests share the same code. Perhaps it can be factored out?

EmadAlblueshi:
> Hi @EmadAlblueshi , thanks for the updates. I was able to take another look today.
> 
> Can you update the very long test? I must admit I'm not prepared to review it. It seems many tests share the same code. Perhaps it can be factored out?

I will try my best to refactor it.

src/main/java/io/vertx/core/parsetools/impl/LengthFieldParserImpl.java
EmadAlblueshi: I think I should add `this.handler(null)`

tsegismont:
Thank you Emad

>

July-X:
@EmadAlblueshi bro,release it,come on!!! @vietj @tsegismont

moxi-born:
It seems all test passed, so  when will it be released?

EmadAlblueshi:
Hi @moxi-born @July-X Sorry for long time I was busy in my new job!

Hi @tsegismont as far as I remember you told me that the test code was so long and needs to be refactored. do you have any recommendations?

tsegismont:
@EmadAlblueshi I'll come back to this after 4.4.2 is released

tsegismont:
Hi @EmadAlblueshi , I'm going to close this PR. Indeed, we decided not to maintain new features of `RecordParser` for a purpose that can be achieved leveraging Netty codecs.

We have the [Using Netty TCP codecs](https://vertx.io/docs/guides/advanced-vertx-guide/#_using_netty_tcp_codecs) section in the advanced Vert.x guide that explains how to proceed.

It's not an ideal outcome but, thanks to your efforts, we've been able to explore the idea in details.

PR #3749, opened 2021-01-18
Add support for specifying thread subclasses

### LINKED ISSUE #3767: Vertx thread customization
In some environment in which Vert.x is embedded it can be desirable to customize or subclass the base thread used by Vert.x (`VertxThread`).

FroMage:
As discussed previously, this is required by Quarkus in order to create custom threads with faster thread locals as fields.

Let me know if there's anything else you need in there. Note that this is for the 3.x branch, let me know if this is the wrong target branch.

Should I also make a similar PR for 4.0?

Fixes #3767


src/main/java/io/vertx/core/VertxThreadFactoryCreator.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.concurrent.TimeUnit;
+
+import io.vertx.core.impl.BlockedThreadChecker;
+import io.vertx.core.impl.VertxThreadFactory;
+
+/**
+ * Extension point to allow creating custom {@code VertxThreadFactory} instances.
+ */
+public interface VertxThreadFactoryCreator {
```
vietj, line 11: can you rename to `VertxThreadFactoryProvider`, we already have similar plugins that are named this way (like `ResolverProvider`), I think also it's a better name.

also it should be in the package `io.vertx.core.spi` instead

vietj:
Actually I think we don't need to have a provider and instead we can use in 3 the Java service loader mechanism.

As we are going to have a single thread factory most of the time and this is a very low level setting we should use java ServiceLoader for this at least in 3.x . When no factory is found then it uses the default thread factory.

Can you rework this PR to use service loader to load the factory and fallback to the regular thread factory when no one is present ? You can make `VertxThreadFactory` an interface in `io.vertx.core.spi` package and rename the current `VertxThreadFactory` to an impl.

geoand:
@FroMage if you no longer have time to do follow up on this, I can implement @vietj's suggestions

hutchig:
This would be a very welcome feature for embedding Vert.x into other environments too.

FroMage:
> @FroMage if you no longer have time to do follow up on this, I can implement @vietj's suggestions

Yes please. The changes look fine, BTW. We should have to make sure the Quarkus service file is visible to Vert.x.

geoand:
Cool, I'll update it most likely tomorrow

geoand:
@FroMage can you grant me access to your fork?

FroMage:
Done?

geoand:
Yup, thanks 👍

geoand:
@vietj I made a small update to the PR to simply make `VertxThreadFactoryCreator` an SPI class and load it from the ServiceLoader.

I don't see how to apply your suggestion of making `VertxThreadFactory` and SPI class as we need to invoke it with constructor arguments (unless you had in mind to add an extra method like `init` to it).

vietj:
@geoand you are right I missed the arguments, can you rename `VertxThreadFactoryCreator` to `VertxThreadFactoryProvider` ?

geoand:
@vietj done

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -129,6 +130,7 @@ static void clusteredVertx(VertxOptions options, Transport transport, Handler<As
   private final TimeUnit maxEventLoopExecTimeUnit;
   private final CloseHooks closeHooks;
   private final Transport transport;
+  private final VertxThreadFactoryProvider threadFactoryCreator;
```
vietj, line 133: rename field to `threadFactoryProvide`

geoand: Ah shoot :(

geoand: Fixed

src/main/java/io/vertx/core/spi/VertxThreadFactoryProvider.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.concurrent.TimeUnit;
+
+import io.vertx.core.impl.BlockedThreadChecker;
+import io.vertx.core.impl.VertxThreadFactory;
+
+/**
+ * Extension point to allow creating custom {@code VertxThreadFactory} instances.
+ */
+public interface VertxThreadFactoryProvider {
```
vietj, line 11: missing licence header

vietj:
One missing licence header and we are good to go.

hutchig:
I will adapt #3733 to look like this one where relevant.

src/main/java/io/vertx/core/spi/VertxThreadFactoryProvider.java
geoand: Fixed

hutchig:
If there was a PR for 4 I could 'rebase' #3733 to be on top of when that is merged - assuming of course that this one is going to progress as quickly as it has. 
#3729 would be the ExecutorServiceFactory SPI PR that would go on top of this 3749+eclipse-vertx:3.9

geoand:
Once this gets merged, I can do a follow up against Vert.x 4, based on @vietj's feedback

vietj:
@geoand Vert.x 4 will likely use the VertxFactory instead of using service loader

geoand:
Understood. Let me know if you want me to do anything WRT to Vert.x 4

geoand:
So just to be clear, you want me to add the ability to configure `VertxThreadFactory` in `VertxFactory`, correct?

Also, are we good to go with this PR?

src/main/java/io/vertx/core/spi/VertxThreadFactoryProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Creates a new {@code VertxThreadFactory} with the following parameters.
+     * @param prefix
```
jponge, line 26: No Javadoc? 😉

geoand: I'll add it :)

geoand: I'll add it in a few minutes

geoand: Done

vietj:
@geoand not in 3.9, this would be in master. In master `VertxFactory` is a builder like API that is used to configure services required by a vertx instance. It is an alternative way to create a vertx instance that provides more configuration and allows to keep VertxOptions without these specific customizations. We might introduce an service loader mechanism to provide the factory for use cases that @hutchig requires. We are fine with this PR.

geoand:
> @geoand not in 3.9, this would be in master. In master `VertxFactory` is a builder like API that is used to configure services required by a vertx instance.

Right, so in `master` you want me to add to add the ability to configure `VertxThreadFactory` in `VertxFactory`, right?

hutchig:
I am confused with regard to https://github.com/eclipse-vertx/vert.x/pull/3729#issuecomment-753842962 both that PR and this one were for SmallRye Reactive Messaging's use of Vert.x.

hutchig:
Can I look to progress #3729 in a similar pattern to this PR now then?

vietj:
I will provide in master a way for configuring the vertx instance with a
builder and then we can adapt this PR to use it.

vietj:
you can

Issue #3760, opened 2021-01-25
Create server form upload attribute when completed rather than at the end of the request


vietj:
The current server form upload processing will create the form multi-map when the HTTP server request ends. Some attributes might use `ByteBuf` that can be recycled by the request decoder overwriting the attribute value. We should create the request form attributes as soon as they are completed to avoid this case.

This issue was found when upgrading to Netty 4.1.52.Final, this version changes the implementation of the post request decoder and this situation happens with very large file upload (32M) and was found in vertx-web-client tests.

PR #3761, opened 2021-01-25
Avoid lazy server form upload attribute creation

vietj:
Fix for #3760


vietj:
Reverted as create undesirable side effects (regressions).

PR #3771, opened 2021-01-26
Vert.x Service Provider bootstrap

vietj:
Provide a single entry point for configuring Vert.x with service providers, see https://github.com/eclipse-vertx/vert.x/issues/3772


vietj:
scoped for 4.1.0 but it could be I think in 4.0.1 as there are no external breaking changes

PR #3781, opened 2021-01-29
duplicate message in `ClusteredEventBusTestBase.testPublish`

jo5ef:
I think the intention of `ClusteredEventBusTestBase.testPublish` is to

1. start 3 nodes, register handlers on 2 of these nodes
1. wait for both handlers to be registered
1. publish *one* message on the remaining node
1. wait for both handlers to receive the message

However, `testPublish` currently publishes *two* messages, one without waiting. Depending on timing, this can cause `testComplete` to be called prematurely. As a consequence, tests [sometimes fail for `vertx-ignite`](https://github.com/vert-x3/vertx-ignite/issues/105).

This PR fixes this and I've also verified that tests for `vertx-hazelcast`, `vertx-zookeeper` and `vertx-infinispan` still pass with this change.


tsegismont (approved):
Nice catch! Thank you @jo5ef and @zyclonite !

PR #3782, opened 2021-01-29
Allow enabling Netty resource leak detection

pepijnve:
The current code sets the Netty ResourceLeakDetector to disabled if either of the Netty system properties is set to a non-null value. I believe this is the exact opposite of what you want, you want to disable it if both are not set instead.

Motivation:

I'm trying to enable Netty resource leak detection in my own code, but because I'm setting the system property and I'm loading vert.x as well, the level gets set to disabled instead of the value I specified.


BillyYccc (approved):

EmadAlblueshi (approved):

vietj:
right, it was introduced by https://github.com/eclipse-vertx/vert.x/commit/25bbc9fe10c46b086a2967c437de23e7003e6fcf#diff-38e15c129e2f81ad425c1bbe0005b26df5a661c05b6c2938d774d95879959cef

PR #3788, opened 2021-02-04
Missing ByteBuf releases fix

vietj:


src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -69,6 +70,9 @@
   private static final Logger log = LoggerFactory.getLogger(Http1xClientConnection.class);
 
   private static final Handler<Object> INVALID_MSG_HANDLER = msg -> {
+    if (msg instanceof ReferenceCounted) {
```
franz1981, line 73: I believe there should be an Util class from Netty to do the same

vietj: ok, if you know which one :-)

franz1981: `ReferenceCountUtil::release`!

src/main/java/io/vertx/core/http/impl/Http1xConnectionBase.java
```diff
@@ -40,13 +50,36 @@
...
     synchronized (this) {
       w = webSocket;
     }
     if (w != null) {
-      w.context.execute(msg, w::handleFrame);
+      w.context.execute(frame, w::handleFrame);
+    }
+  }
+
+  private WebSocketFrameInternal decodeFrame(io.netty.handler.codec.http.websocketx.WebSocketFrame msg) {
```
franz1981, line 63: If each case is equally possible, it's better to use a
`static final IdentityHashMap<Class<?>,FrameType` initialized with the handled types, given that `System::identityHashCode` is an intrinsic, msg::getClass too and `Class` should be singleton in the same class loader AFAIK!

src/main/java/io/vertx/core/http/impl/Http2UpgradedClientConnection.java
```diff
@@ -228,25 +233,51 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
+          } else {
+            super.userEventTriggered(ctx, evt);
+          }
+        }
+
+        @Override
+        public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
+          if (buffered != null) {
+            Deque<Object> messages = buffered;
+            buffered = null;
```
franz1981, line 270: why not reusing `buffered` insteadd of nulling it out?

src/main/java/io/vertx/core/http/impl/Http1xConnectionBase.java
vietj: thanks, feel free to do a PR because it is another concern than this PR and the code was just moved around :-)

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
vietj: I'll use it asap then to cleanup this

vietj: https://github.com/eclipse-vertx/vert.x

src/main/java/io/vertx/core/http/impl/Http2UpgradedClientConnection.java
vietj: because it is used as a marker by `channelRead`

vietj: and in practice if that's not null then it means that something bad happen

PR #3793, opened 2021-02-08
Add SubstitutedContext and make DuplicatedContext use delegate TaskQueue

### LINKED ISSUE #3790: DuplicatedContext can cause unexpected behavior
The following code will run a bunch of blocking tasks all with `ordered=true`
```java
public static void main(String[] args){
Context context0 = Vertx.vertx().getOrCreateContext();
context0.runOnContext(ignored0 -> {
	Vertx vertx = Vertx.currentContext().owner();
	vertx.eventBus().consumer("eb_address", m -> runTask(m.body().toString()));
	runTask("Task 1");
	runTask("Task 3");
	vertx.eventBus().publish("eb_address", "Task 2");
	vertx.setTimer(1, t -> {
		runTask("Task 4");
		vertx.eventBus().publish("eb_address", "Task 5");
	});
});
}

public static void runTask(String name){
        Vertx.currentContext().executeBlocking(promise -> {
	        System.out.println(name + " start");
	        try {
		        Thread.sleep(1000);
	        }catch(Exception e){
		        promise.fail(e);
		        return;
	        }
	        System.out.println(name + " done");
	        promise.complete();
        }, true);
}
```
This is the output:
```
Task 1 start
Task 2 start
Task 5 start
Task 1 done
Task 3 start
Task 2 done
Task 5 done
Task 3 done
Task 4 start
Task 4 done
```
Given that `ordered=true`, the output is odd, as conventional wisdom (and the documentation) would say that for any `Task n start` the message that should always follow it is `Task n done`. Of course ordering is only the case when the calls take place within the same context, and they all are called from the same context. Right?
I know that's not the case, but for any person that hasn't read the implementation, they should reasonably expect that everything that I wrote above is all run from the same context. That's the core of this issue, certain handlers are called with a `DuplicatedContext` and it causes unusual behavior because it creates a view of the parent context with the blocking `TaskQueue` being independant of the parent. Also, these side effects are completely undocumented.

I understand why `DuplicatedContext` exists, and I also know why we made the `DuplicatedContext` use its own `TaskQueue`, but I don't think the different `TaskQueue` was a good idea. It will cause issues for developers that will not be discovered immediately and in its current form make implementing the scheduler for RxJava imposible (For reasons why see: https://github.com/vert-x3/vertx-rx/issues/245#issuecomment-773415774) Personally I think it should be removed, and jdbc-client should revert back to the old solution (it appears to be the only thing that should use `TaskQueue`, but it doesn't look like it was properly implemented, see:  https://github.com/vert-x3/vertx-jdbc-client/issues/203#issuecomment-773719617). Even though that is what I think, I don't think that is something that is going to happen, so I have a simpler solution. Do not create a seperate `TaskQueue` unless explicitly specified. This way the `DuplicatedContext` behaves exactly like its parent for all actions (though of course the `localData` map is not preserved between the two)

I'm already busy on a PR to make a couple of changes around `DuplicatedContext`

Gattag:
- `DuplicatedContext` uses delegate `TaskQueue` instead of its own
- Adds a `SubstitutedContext`, a context based off a delegate with certain elements that can be changed
- Also fixed a synchronization issue in `WorkerExecutorImpl` (Noticed it and through it in there)

Still needs more tests, working on it, but waiting till I get some feedback on this.

This will required a change or two in JDBC, but otherwise shouldn't break anything.

Fixes #3790

Signed-off-by: Gregory Tracy greg@thetracys.net

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerImpl.java
@@ -16,6 +16,7 @@ import io.vertx.core.*;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.*;
 import io.vertx.core.impl.ContextInternal;
+import io.vertx.core.impl.ContextSubstitution;
 import io.vertx.core.impl.EventLoopContext;
 import io.vertx.core.impl.future.PromiseInternal;
 import io.vertx.core.impl.VertxInternal;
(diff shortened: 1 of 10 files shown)


vietj:
this seems like a lot of changes and additions, I'll have a look myself

Gattag:
I think that second commit I made was not the right solution, but I panicked

Gattag:
Too much complexity

PR #3800, opened 2021-02-10
VertxThreadFactory service provider

vietj:


vietj:
@geoand @FroMage this is the port of the thread customization to 4, can you have a look?

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -1124,6 +1128,20 @@ public synchronized SharedWorkerPool createSharedWorkerPool(String name, int poo
...
 
+  private ThreadFactory createThreadFactory(long maxExecuteTime, TimeUnit maxExecuteTimeUnit, String prefix, boolean worker) {
+    AtomicInteger threadCount = new AtomicInteger(0);
+    return runnable -> {
+      VertxThread thread = threadFactory.newVertxThread(runnable, prefix + threadCount.getAndIncrement(), worker, maxExecuteTime, maxExecuteTimeUnit);
+      checker.registerThread(thread, thread);
+      // Vert.x threads are NOT daemons - we want them to prevent JVM exit so embededd user doesn't
+      // have to explicitly prevent JVM from exiting.
+      // I know the default is false anyway, but just to be explicit-  Vert.x threads are NOT daemons
+      // we want to prevent the JVM from exiting until Vert.x instances are closed
```
geoand, line 1139: Seems like the comment is duplicated :)

FroMage (approved):
LGTM, but this is a different API to the one we have in 3.x, right? So worth keeping in mind for https://github.com/quarkusio/quarkus/issues/14779

src/main/java/io/vertx/core/impl/VertxImpl.java
geoand: Also, should this really a lambda?
I don't know about Vert.x, but in Quarkus we try to stay away from lambdas as they have a slightly higher metaspace cost than regular classes.

geoand (approved):
I added a couple minor comments, but otherwise LGTM

vietj:
yes it's different but it achieves the same right ?

vietj:
if I would have done it myself for 3.x, that's the way I would have done it

FroMage:
> yes it's different but it achieves the same right ?

Sure, but it's not binary compat, so one more thing for us to fix when we switch to 4.0.

> if I would have done it myself for 3.x, that's the way I would have done it

is it too late? was it released?

vietj:
it has not been released, but it seems trivial

src/main/java/io/vertx/core/impl/VertxImpl.java
vietj: it's a lambda for readability, we have many lambda in other places and it seems it never raised an issue.

vietj:
@FroMage I added a comment in the Quarkus 2 / Vert.x4 issue for max to record this change in the issue.

FroMage:
Thanks.

PR #3802, opened 2021-02-11
Worker verticle execution should always be serialized

vietj:


src/test/java/io/vertx/core/ContextTest.java
```diff
@@ -514,29 +516,38 @@ private void checkDuplicate(ContextInternal ctx, ContextInternal duplicated) thr
...
+
+  private void testDuplicateWorkerConcurrency(BiConsumer<ContextInternal, Runnable> task) throws Exception {
+    ContextInternal worker = ((VertxInternal)vertx).createWorkerContext();
+    ContextInternal[] contexts = new ContextInternal[] { worker.duplicate(), worker.duplicate()};
+    waitFor(contexts.length);
+    AtomicBoolean owner = new AtomicBoolean();
+    CountDownLatch latch = new CountDownLatch(contexts.length);
+    for (ContextInternal context : contexts) {
+      task.accept(context, () -> {
+        assertFalse(owner.get());
```
guidomedina, line 535: There is a chance 2 threads executing this and none will be the owner, you need to make an atomic comparison:
```
assertTrue(owner.compareAndSet(false, true));
```
then you don't need to set owner to true on after the `try`

vietj: good catch, if there is no bug then it is ok, but if that's buggy (which is what we are testing here) indeed it is racy

src/test/java/io/vertx/core/ContextTest.java
```diff
@@ -514,29 +516,37 @@ private void checkDuplicate(ContextInternal ctx, ContextInternal duplicated) thr
...
+      task.accept(context, () -> {
+        try {
+          assertTrue(owner.compareAndSet(false, true));
+          Thread.sleep(200);
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
+        } finally {
+          owner.set(false);
+        }
+        latch.countDown();
```
guidomedina, line 543: This is a minor but in if something goes wrong the latch is never decremented, I usually decrement required `latch`es in the finally block just in case, that way the main coordinating thread will be able to progress in any case.

guidomedina: Actually, now I remember why I have 2 try blocks, here is how I would had that code:
```
try {
  assertTrue(owner.compareAndSet(false, true));
  try {
    Thread.sleep(200);
  } catch (InterruptedException e) {
    Thread.currentThread().interrupt();
  } finally {
   // if we reach this finally the owner was successfully set to true by this thread
    owner.set(false);
  }
} finally {
  // Regardless of success or any exception always decrement the latch
  latch.countDown();
}
```

vietj:
if there is an exception then the test fails.

guidomedina:
> if there is an exception then the test fails.

I don't know what the test engine does when an assertion fails, does it halt the whole test method skipping the latch await call or does it proceed to wait and then fails the test? if is the latter then always decrementing the latch is advised.

vietj:
I'll do as you advertise

vietj:
@guidomedina I updated the PR do you want to take another look ?

guidomedina:
> @guidomedina I updated the PR do you want to take another look ?

Looks good, I'm guessing that you tested without the patch with 4.0.x and both of these test cases failed

vietj:
yes they need to be changed to the correct behavior

PR #3808, opened 2021-02-12
Remove the default Jackson codec descriptor in vertx-core jar

vietj:


vietj:
@cescoffier ping

pom.xml
```diff
@@ -357,7 +357,7 @@
         <artifactId>maven-failsafe-plugin</artifactId>
         <version>2.19.1</version>
         <executions>
-          <execution>
+<!--          <execution>
```
cescoffier, line 360: any reason to keep it commented?

src/test/java/io/vertx/it/CustomJsonCodec.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @Override
+  public <T> T fromValue(Object json, Class<T> toValueType) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public String toString(Object object, boolean pretty) throws EncodeException {
+    return new StringBuilder().append(object).reverse().toString();
```
cescoffier, line 27: Almost cryptographic.

src/main/java/io/vertx/core/spi/json/JsonCodec.java
```diff
@@ -21,7 +21,15 @@
  */
 public interface JsonCodec {
 
-  JsonCodec INSTANCE = ServiceHelper.loadFactory(JsonCodec.class);
+  static JsonCodec loadCodec() {
+    JsonCodec instance = ServiceHelper.loadFactoryOrNull(JsonCodec.class);
```
cescoffier, line 25: JsonCodec should have an `ordinal` method, and the returned instance should be the one with the highest ordinal. We may end up with multiple implementations on the classpath, and we need to have a deterministic way to retrieve always the same one.

cescoffier (changes requested):

src/main/java/io/vertx/core/spi/json/JsonCodec.java
cescoffier: BTW, the same should be applied to the Vert.x 4 JsonFactory.

cescoffier: One way would be to introduce a `Sorted` interface with a single `ordinal()` method, and in `io.vertx.core.ServiceHelper#loadFactories(java.lang.Class<T>, java.lang.ClassLoader)`, if the given interface implements `Sorted`, just sort the collections (reverse as we want highest ordinals first).

pom.xml
vietj: no, it is my mistake, I commented this for testing quicker

src/main/java/io/vertx/core/spi/json/JsonCodec.java
vietj: we don't expect to have another implementation than the quarkus one, so I think it is not necessary for the moment.

cescoffier:
Quarkus will have multiple implementations (Jackson, JsonB, Gson...)

src/main/java/io/vertx/core/spi/json/JsonCodec.java
vietj: the point is that anyone would anyway use the same value to get higher priority at the end. I've seen similar things in the past and it never really worked as far as I could observe.

What we could do instead is something that applies all codecs and each codec will set itself as a codec unless there is already one codec here. The quarkus one would always set itself whatever the value is.

vietj: Something like

```java
  default JsonCodec init(JsonCodec previous) {
    if (previous == null) {
      return this;
    } else {
      return previous;
    }
  }
```

that the load method will call. Then quarkus can have an implementation that always return this.

vietj:
and those can be on the classpath simultaneously ?

how will ordering be controlled then ?

vietj:
in that case, isn't it better if Quarkus provides a single JsonCodec which
will resolve itself ordering with extensions, so the priority logic is in
Quarkus and can be changed easily ?

cescoffier:
On Fri 12 Feb 2021 at 12:36, Julien Viet <notifications@github.com> wrote:

> and those can be on the classpath simultaneously ?
>


Potentially yes.


> how will ordering be controlled then ?


Each implementation would have a default ordinal, overridable by
configuration.

cescoffier:
On Fri 12 Feb 2021 at 12:39, Julien Viet <notifications@github.com> wrote:

> in that case, isn't it better if Quarkus provides a single JsonCodec which
> will resolve itself ordering with extensions, so the priority logic is in
> Quarkus and can be changed easily ?


No, because it would mean having an implementation that would depend on all
the different possibilities.

vietj:
I just meant this implementation would have its own ordering internally. I'm just affraid that this is a quarkus issue and the logic at some point will be limited and require change / update in the logic that would be in vertx. So it's more like a delegation of responsibility.

cescoffier:
On Fri 12 Feb 2021 at 12:43, Julien Viet <notifications@github.com> wrote:

> I just meant this implementation would have its own ordering internally.
> I'm just affraid that this is a quarkus issue and the logic at some point
> will be limited and require change / update in the logic that would be in
> vertx. So it's more like a delegation of responsibility.
>
And that’s what we can’t do. One way would be to add a second SPI, but
that’s unnecessary.  In that case, it would be preferable to not have the
vertx SPI at all and just a setter.

BTW, the ordinal approach is widely used (CDI priority, config source
ordinals...)

vietj:
I fear that after adding the ordinal, people will ask to have system properties to override ordinal and bring more complexity and might lead to breaking change when we need to update it.

vietj:
in practice is there a  need for ordinal in 3.9 ? I would prefer if we can avoid un-necessary work

cescoffier:
Yes, same as for 4.0.

vietj:
alright sir

vietj:
@cescoffier order has been ported from master to this PR

cescoffier:
Thanks.

PR #3810, opened 2021-02-12
Make sure metrics and cluster manager can be customized in Launcher

### LINKED ISSUE #3804: Cannot modify metrics options via sysprop or custom Launcher
A custom Launcher is often used when you want to setup things like metrics options programmatically before Vert.x is started.

Since #3771 , it is no longer possible to do this because the builder init method is invoked before the Launcher callback.

Then it's too late to customize the metrics implementation.

This applies to other special options too, I think (e.g. tracing, cluster manager).

EDIT: actually, it's no longer possible to activate some metrics options with sysprops. From Dropwizard docs:

```
-Dvertx.metrics.options.enabled=true -Dvertx.metrics.options.jmxEnabled=true
```

tsegismont:
Fixes #3804

### DIFF
--- a/src/main/java/io/vertx/core/impl/launcher/commands/BareCommand.java
+++ b/src/main/java/io/vertx/core/impl/launcher/commands/BareCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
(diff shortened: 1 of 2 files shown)


vietj:
the test fails is that expected ?

tsegismont:
@vietj looking

tsegismont:
@vietj PR updated, checks are all green (except the Windows one which is not finished yet)

tsegismont:
@vietj rebased on master

vietj:
before merging this, can you check it still works with
vertx-dropwizard-metrics ?

tsegismont:
Yes it works.

PR #3811, opened 2021-02-15
JsonFactory ordering mechanism

vietj:


vietj:
@cescoffier ping, here is the order mechanism in master

vietj:
@cescoffier can you have a look ?

src/main/java/io/vertx/core/spi/JsonFactory.java
```diff
@@ -21,7 +22,32 @@
...
+
+  /**
+   * The order of the factory. If there is more than one matching factory they will be tried in ascending order.
+   *
+   * @implSpec returns {@link Integer#MAX_VALUE}
+   *
+   * @return  the order
+   */
+  default int order() {
+    return Integer.MAX_VALUE;
```
cescoffier, line 49: Shouldn't you use a lower value?

If an implement omits that method (because it does not care) it will be always used.

cescoffier: Typically in your test, if one of the implementations does not provide the method, it will be first.

cescoffier (changes requested):

src/main/java/io/vertx/core/spi/JsonFactory.java
vietj: actually it sorts by ascending order and takes the first one, so factory that don't implement this will always be chosen last.

cescoffier: Oh sorry! Misread it! 

So all good for me.

cescoffier (approved):

PR #3813, opened 2021-02-17
feat: completion function to MessageConsumer returns a future

wem:
feat: completion function to MessageConsumer returns a future

Alternative to completionHandler.

Motivation:

https://github.com/vert-x3/vertx-lang-kotlin/issues/190

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/MessageConsumer.java
+++ b/src/main/java/io/vertx/core/eventbus/MessageConsumer.java
@@ -87,8 +87,20 @@ public interface MessageConsumer<T> extends ReadStream<Message<T>> {
    * Optional method which can be called to indicate when the registration has been propagated across the cluster.
    *
    * @param completionHandler the completion handler
+   * @deprecated Please use {@link #registrationCompletion(Handler)} or {@link #registrationCompletion()}
    */
+  @Deprecated
   void completionHandler(Handler<AsyncResult<Void>> completionHandler);
+
+  /**
+   * Optional method which can be called to indicate when the registration has been propagated across the cluster.
+   */
+  void registrationCompletion(Handler<AsyncResult<Void>> completionHandler);
+
+  /**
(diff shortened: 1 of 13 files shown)


src/main/java/io/vertx/core/eventbus/MessageConsumer.java
```diff
@@ -90,6 +90,11 @@
    */
   void completionHandler(Handler<AsyncResult<Void>> completionHandler);
 
+  /**
+   * Like {@link #completionHandler(Handler)} but returns a {@code Future} of the asynchronous result.
+   */
+  Future<Void> completion();
```
vietj, line 96: I think instead we should name it `registrationCompletion`

vietj: to clarify it is related to the consumer registration.

src/main/java/io/vertx/core/eventbus/MessageConsumer.java
```diff
@@ -90,6 +90,11 @@
    */
   void completionHandler(Handler<AsyncResult<Void>> completionHandler);
 
+  /**
+   * Like {@link #completionHandler(Handler)} but returns a {@code Future} of the asynchronous result.
+   */
+  Future<Void> registrationCompletion();
+
```
vietj, line 97: we also need a `registrationCompletion(Handler<AsyncResult<Void>>)` method to go along with this

wem: I think after this round i understand your intention :)

vietj:
can you rebase on latest master ?

vietj:
I think the main issue here is that a consumer can be registered or unregistered multiple times and it is not clear what a registration future would mean, we need to clarify this before taking any action

tsegismont:
Superseded by #4645

PR #3836, opened 2021-03-08
Fix HttpClientRequest setURI(uri), query(), path()

julianladisch:
* Invalidate query cache in setURI(uri). This fixes a bug
  where query() returns an old and invalid value, the
  new unit test run against the old code demonstrates this.
* Use boolean newQuery to invalidate the query cache as
  HttpUtils.parseQuery(uri) may return null.
* Remove redundant code from path() as
  HttpUtils.parsePath(uri) never returns null.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientRequestBase.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientRequestBase.java
@@ -42,7 +42,6 @@ public abstract class HttpClientRequestBase implements HttpClientRequest {
   private String uri;
   private String path;
   private String query;
-  private boolean newQuery = true;
   private final PromiseInternal<HttpClientResponse> responsePromise;
   private Handler<HttpClientRequest> pushHandler;
   private long currentTimeoutTimerId = -1;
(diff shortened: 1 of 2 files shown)


vietj:
I think we can achieve the same without the boolean and simply setting the cached query to null

julianladisch:
@julianladisch wrote in the description:
> Use boolean newQuery to invalidate the query cache as HttpUtils.parseQuery(uri) may return null.

@vietj wrote:
> I think we can achieve the same without the boolean and simply setting the cached query to null

What should happen if HttpUtils.parseQuery(uri) returns null as demonstrated in the `setURI("")` test case? This happens for all URIs that don't contain a `?`.

vietj:
I think it will still work

julianladisch:
Yes, it will return the correct result but it will not cache the result but call HttpUtils.parseQuery(uri) again and again. This happens for the frequent case where the URI doesn't have a `?`.

If we don't care about the query caching then we should completely remove the caching and simply call parseQuery every time.

vietj:
I think then we should simply avoid lazy loading and recompute the cached
value everytime URI is set instead.

vietj:
thanks for the contribution.

PR #3838, opened 2021-03-08
Always throw IndexOutOfBoundsException in CompositeFuture

julianladisch:
CompositeFutureImpl#future(index) throws
* ArrayIndexOutOfBoundsException if index == results.length
* IndexOutOfBoundsException if index < 0 or index > results.length

This fix changes the index == results.length case to also throw IndexOutOfBoundsException.


vietj:
thanks for the contribution

PR #3841, opened 2021-03-09
Support for passing URL in RequestOptions

### LINKED ISSUE #3803: Support for passing URL in RequestOptions
#### Describe the feature

The `io.vertx.core.http.RequestOptions` in Vert.x 4.0.0 allows to set up instance with absolute Uri as a `String` via `RequestOptions().setAbsoluteUri(...)`.  Could be this extended also for providing Uri as a `java.net.URL` ? 

#### Use cases

In our calls we have a request uri directly in `java.net.URL` . For creating `RequestOptions` instance, we have to convert `URL` to `String`, enabling us to call `setAbsoluteUri`. But inside of this `String` is immediately parsed back to URL via `RequestOptions.parseUrl()` which is ineffective (as `parseUrl` correctly mentions it in `Note`).

#### Contribution

I have zero experience with contributing to any library but this could be easy done via splitting `setAbsoluteUri` into parsing and processing method:
```java
  public RequestOptions setAbsoluteURI(String absoluteURI) {
    Objects.requireNonNull(absoluteURI, "Cannot set a null absolute URI");
    return this.setAbsoluteURI(parseUrl(absoluteURI));
  }

  public RequestOptions setAbsoluteURI(URL url) {
    Objects.requireNonNull(absoluteURI, "Cannot set a null absolute URI");
    Boolean ssl = false;
    int port = url.getPort();
    String relativeUri = url.getPath().isEmpty() ? "/" + url.getFile() : url.getFile();
    String protocol = url.getProtocol();
    switch (protocol) {
      case "http":
        if (port == -1) {
          port = 80;
        }
        break;
      case "https": {
        ssl = true;
        if (port == -1) {
          port = 443;
        }
        break;
      }
      default:
        throw new IllegalArgumentException();
    }
    this.uri = relativeUri;
    this.port = port;
    this.ssl = ssl;
    this.host = url.getHost();
    return this;
  }
```

tsegismont:
Convenience method that avoids converting to String then re-parsing when users already have a URL object

Closes #3803

### DIFF
--- a/src/main/java/io/vertx/core/http/RequestOptions.java
+++ b/src/main/java/io/vertx/core/http/RequestOptions.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+import static io.vertx.codegen.annotations.GenIgnore.PERMITTED_TYPE;
+
 /**
  * Options describing how an {@link HttpClient} will connect to make a request.
  *
@@ -372,7 +374,7 @@ public class RequestOptions {
    * @param url the uri to use
    * @return a reference to this, so the API can be used fluently
    */
-  @GenIgnore
+  @GenIgnore(PERMITTED_TYPE)
   public RequestOptions setAbsoluteURI(URL url) {
     Objects.requireNonNull(url, "Cannot set a null absolute URI");
     Boolean ssl = Boolean.FALSE;


src/main/java/io/vertx/core/http/RequestOptions.java
```diff
@@ -333,7 +346,12 @@ private URL parseUrl(String surl) {
   public RequestOptions setAbsoluteURI(String absoluteURI) {
     Objects.requireNonNull(absoluteURI, "Cannot set a null absolute URI");
     URL url = parseUrl(absoluteURI);
-    Boolean ssl = false;
+    apply(url);
+    return this;
+  }
+
+  private void apply(URL url) {
```
vietj, line 353: I think that this could simply be in the body `setAbsoluteURI(URL absoluteURI)` instead of being an independent method.

vietj (changes requested):
minor modif

src/main/java/io/vertx/core/http/RequestOptions.java
tsegismont: Done

src/main/java/io/vertx/core/http/RequestOptions.java
```diff
@@ -333,7 +333,19 @@ private URL parseUrl(String surl) {
...
+    return setAbsoluteURI(url);
+  }
+
+  /**
+   * Like {@link #setAbsoluteURI(String)} but using an {@link URL} parameter.
+   *
+   * @param url the uri to use
+   * @return a reference to this, so the API can be used fluently
+   */
+  @GenIgnore
```
pmlopes, line 345: Make it a permitted type, so it can be used by rx, kotlin, and so...

src/main/java/io/vertx/core/http/RequestOptions.java
```diff
@@ -333,7 +333,19 @@ private URL parseUrl(String surl) {
...
+  public RequestOptions setAbsoluteURI(URL url) {
```
pmlopes, line 346: `URL` used to give trouble with native images compiled without ssl support. if this is switched to `URI` it is always known to work and it matches better with the method name.

src/main/java/io/vertx/core/http/RequestOptions.java
tsegismont: Done

src/main/java/io/vertx/core/http/RequestOptions.java
tsegismont: Internally, we use `URL`. The original issue was filed by a user who works with `URL`. If users have a `URI` they can still make the conversion.

Do you know if there is an issue tracking the problem you mentioned? In practice, would `native-image` with just any program that has a method with `URL` parameter? Or only programs that actually use that method?

In the latter case, I believe we can tell users to avoid using that method as a workaround.

PR #3845, opened 2021-03-10
EventBus node selector should always initialize with initial state request

tsegismont:
Otherwise some tests can fail because selector is initialized with intermediate data.


tsegismont:
After this one is merged and a new snapshot deployed, we can rerun the actions in https://github.com/vert-x3/vertx-hazelcast/pull/139

PR #3846, opened 2021-03-10
Give ClusterManager a chance to know if NodeSelector wants updates for an address

tsegismont:
This can be useful to reduce load/traffic.
Indeed, cluster managers such as HZ/ISPN/Ignite must query the backend to get a fresh view of the data.
Consequently, if the NodeSelector is not interested, this query can be avoided

### DIFF
--- a/src/main/java/io/vertx/core/spi/cluster/NodeSelector.java
+++ b/src/main/java/io/vertx/core/spi/cluster/NodeSelector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
  *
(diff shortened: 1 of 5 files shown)


tsegismont:
@zyclonite can you please take a look as well?

This can be useful in Vert.x Ignite too.

zyclonite:
What would be the case where the NodeSelector does not want updates?

tsegismont:
@zyclonite for example when no verticle on this node is trying to send messages to the address that gets notifications. This may happen when you deploy verticles on the cluster nodes which do different things.

tsegismont:
@zyclonite I've added a test for the case when selector does not want/need updates.

zyclonite:
yes, makes sense

it will improve HA deployments as well

looks good

jponge (approved):

PR #3867, opened 2021-03-20
Fix timeout reply exception message in ReplyHandler

zyclonite:
Cosmetic fix for the message logged in the ReplyHandler when a ReplyException is thrown on a timeout

In the current implementation the below `3` stands for the timer ID not the milliseconds of the timeout
`Timed out after waiting 3(ms) for a reply. address:`

This PR puts the timeout in milliseconds back into the message


zyclonite:
@vietj this should be a quick one ;)

vietj:
right thanks

Issue #3882, opened 2021-04-04
ClusteredEventBus publishes local subscription topics on the cluster


elventear:
### Version

4.0.3

### Context

Migrating socket application from stand alone to cluster mode using Zookeper. I have noticed that all socket write IDs are being published to the cluster, which seemed suspicious and possible scalability problem for very busy network applications.

Further reviewing to the code showed that `NetSocket` registration should only expose it's write ID locally, which makes sense to me. https://github.com/eclipse-vertx/vert.x/blob/d7605ababc5d86334184c32ce2a123cfac6c3fdb/src/main/java/io/vertx/core/net/impl/NetSocketImpl.java#L105

### Do you have a reproducer?

I have not created a reproducer, but I have been reviewing the code and comparing Vert.X 3.9.6 and 4.0.3. Comparing the two branches the behavior is not consistent and it seems to me there is a bug. 

When adding a new registration I see that the event bus, when clustered resolves to the following interactions in order.

On Vert.X 3.9.6:
1. Call `addRegistration` on `EventBusImpl.java`. https://github.com/eclipse-vertx/vert.x/blob/d2b24fd5bbb4c2672dab4af0f7b35275751fadcc/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java#L235
2. Call  overridden `addRegistration` in `ClusteredEventBus.java`. This call explicitly looks at the `localOnly` flag and does not register the address in the cluster if its local only. https://github.com/eclipse-vertx/vert.x/blob/d2b24fd5bbb4c2672dab4af0f7b35275751fadcc/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java#L186

On Vert.X 4.0.3:

1. Call `addRegistration`on `EventBusImpl.java`. https://github.com/eclipse-vertx/vert.x/blob/d7605ababc5d86334184c32ce2a123cfac6c3fdb/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java#L238
2. Call  overridden `onLocalRegistration` in `ClusteredEventBus.java`. This call does not take a `localOnly` flag and ends up registering the address in the cluster *always*. https://github.com/eclipse-vertx/vert.x/blob/d7605ababc5d86334184c32ce2a123cfac6c3fdb/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java#L128

### Steps to reproduce

1. Start application in clustered mode with Zookeper.
[...]

PR #3883, opened 2021-04-05
Do not publish local consumer addressses to cluster

elventear:
Motivation:

Provides a solution for bug reported in #3882.  TLDR; local only addresses should not be published to cluster, this returns behavior closer to what is observer in Vert.X 3.9.x where a `localOnly` consumer will not generate a registration in the cluster.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -237,11 +237,11 @@ public class EventBusImpl implements EventBusInternal, MetricsProvider {
 
   protected <T> HandlerHolder<T> addRegistration(String address, HandlerRegistration<T> registration, boolean replyHandler, boolean localOnly, Promise<Void> promise) {
     HandlerHolder<T> holder = addLocalRegistration(address, registration, replyHandler, localOnly);
-    onLocalRegistration(holder, promise);
+    onLocalRegistration(holder, promise, localOnly);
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -400,6 +400,10 @@ protected String generateReplyAddress() {
     sendOrPubInternal(newSendContext(message, options, handler, writePromise));
   }
 
+  public boolean isAddressLocal(String address) {
```
elventear, line 403: This seemed to be necessary to support the hybrid use case when addresses are published consumed locally only in some verticals but global in other others.

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -174,6 +174,10 @@ public MessageImpl createMessage(boolean send, String address, MultiMap headers,
     } else if (sendContext.options.isLocalOnly()) {
       super.sendOrPub(sendContext);
     } else {
+      if (isAddressLocal(sendContext.message.address())) {
```
elventear, line 177: If address is also registered locally, then also deliver it using the local EventBus implementation.

vietj:
assigned to you @tsegismont

tsegismont:
Closed related issue (#3882) as duplicate.

Issue #3886, opened 2021-04-08
JsonParserImpl fails after resume


MikhailFokanov:
### Version

4.0.0

It is regression bug, it works well with 3.9.4

### Context

We `pipeTo(WriteStream<T> dst)` method in JsonParser to implement batching of items, which are returned from web service. When the drainHandler is called, it calls `resume()` in JsonParserImpl. These methods are called at random time.

### Reproducer

* https://gist.github.com/MikhailFokanov/4137dcbaabbd033df0f081fca7101379

If the ROWS_NUMBER < 1000 it works well, otherwise fails.

### Steps to reproduce

1. Run the reproducer

tsegismont:
This might be a duplicate of https://github.com/eclipse-vertx/vert.x/issues/3822

@MikhailFokanov can you try with 4.0.3 please?

MikhailFokanov:
@tsegismont With 4.0.3 it still doen't work.

adamdickmeiss:
The code calls pause and resume from different threads than the Verticle running the parser. Is that safe?

tsegismont:
@adamdickmeiss you are right, JsonParser is not thread-safe and pause/resume should be invoked on the same thread that feeds the parser.

@MikhailFokanov I changed the test to this:

```java
        parser.handler(event -> {
            if (recordsCount.incrementAndGet() % 50 == 0) {
                System.out.println("Records returned so far: " + recordsCount.get());
            }
            vertx.setTimer(10, l -> {
              if (isPaused.get()) {
                isPaused.set(false);
                parser.resume();
              } else {
                isPaused.set(true);
                parser.pause();
              }
            });
        });
```

Then the test fails with a more meaningful message:

```
com.fasterxml.jackson.core.JsonParseException: Still have 95 undecoded bytes, should not call 'feedInput'
 at [Source: UNKNOWN; line: 1, column: 1931428]

	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1851)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:712)
	at com.fasterxml.jackson.core.json.async.NonBlockingJsonParser.feedInput(NonBlockingJsonParser.java:99)
	at io.vertx.core.parsetools.impl.JsonParserImpl.handle(JsonParserImpl.java:199)
	at io.vertx.core.parsetools.impl.JsonParserImpl.handle(JsonParserImpl.java:41)
[...]

PR #3893, opened 2021-04-14
Possible fix for #3886

adamdickmeiss:
pendingEvents quene introduced.

Test case missing for this PR.


vietj:
we need a test for this

adamdickmeiss:
The test case reveals a more fundamental problem with the stream
that feeds JsonParser in #3886. It's as it does not respect
pause and will continue to feed. So it seems this is not
a solution but a work-around may use a lot of memory.

vietj:
can you provide a reproducer for the non respected pause ?

adamdickmeiss:
> can you provide a reproducer for the non respected pause ?

I'll try.. I added a flag in JsonParserImpl to see if my assertion was correct. It seems not.. Ie it does seem to be able to occur without that claim. I guess that means the problem might still be in JsonParserImpl and not in the stream being used.

adamdickmeiss:
Got it now.. stream==null in original case and that causes trouble.

src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java
```diff
@@ -221,6 +232,16 @@ public void end() {
...
+      while (demand > 0L && !pendingEvents.isEmpty()) {
+        if (demand != Long.MAX_VALUE) {
+          demand--;
+        }
+        JsonEvent jsonEvent = pendingEvents.removeFirst();
+        Handler<JsonEvent> handler = this.eventHandler;
+        if (handler != null) {
+          handler.handle(jsonEvent);
+        }
+      }
```
julianladisch, line 244: The while loop should be inside the try clause. If the eventHandler throws an exception we want `finally` `emitting = false;`

adamdickmeiss: Right

vietj:
were do we stand now @adamdickmeiss ?

src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java
```diff
@@ -72,15 +75,18 @@ public JsonParser pause() {
 
   @Override
   public JsonParser resume() {
-    return fetch(Long.MAX_VALUE);
+    demand = Long.MAX_VALUE;
+    checkPending();
```
vietj, line 79: why do we need this change ?

src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java
```diff
@@ -114,6 +120,10 @@ public JsonParser handler(Handler<JsonEvent> handler) {
   }
 
   private void handleEvent(JsonEvent event) {
+    if (demand == 0L) {
+      pendingEvents.addLast(event);
+      return;
```
vietj, line 125: can you avoid return and instead use else-if ?

vietj:
@adamdickmeiss I believe the internal of the current `JsonParser` to rely on the internal parser to handle unconsumed tokens is flawed. I refactored the `JsonParser` to consume all tokens and buffer events instead (like your PR do) and went for more drastic changes in https://github.com/eclipse-vertx/vert.x/pull/3912

adamdickmeiss:
If PR #3912 solves the problem of #3886 then we're good and this can be closed (duplicate work)

vietj:
can you review code in the PR ?

adamdickmeiss:
> can you review code in the PR ?
> […](#)

OK

vietj:
thanks for the reproducer and analysis @adamdickmeiss it definitely helped a lot

PR #3894, opened 2021-04-14
Expose AsyncFile size

### LINKED ISSUE #3892: AsyncFile: allow access to file size
In https://github.com/quarkusio/quarkus/pull/16286 we realised that `AsyncFile` does not provide access to the file count, so if we're using it to pump bytes into a response we can't set the `Content-Length` header. We can't trust the `getReadLength()` method as it defaults to `Long.MAX_SIZE` if not set, and does not reflect the file size. We also can't access the file size by asking the file system, since we have no access to the file path.

Could we get access to the file size, or if not, at least to the path?

tsegismont:
Closes #3892

AsyncFileImpl can get the file size from the underlying AsynchronousFileChannel.
But it was not exposed.

### DIFF
--- a/src/main/java/io/vertx/core/file/AsyncFile.java
+++ b/src/main/java/io/vertx/core/file/AsyncFile.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/file/AsyncFile.java
```diff
@@ -181,4 +181,30 @@
    */
   @Fluent
   AsyncFile setReadBufferSize(int readBufferSize);
+
+  /**
+   * Like {@link #size()} but blocking.
+   *
+   * @throws FileSystemException if an error occurs
+   */
+  default long sizeBlocking() {
```
pmlopes, line 190: Why a default here? It seems to be implemented on the impl class.

tsegismont:
In case someone has an alternative implementation.

>

src/main/java/io/vertx/core/file/AsyncFile.java
```diff
@@ -181,4 +181,30 @@
...
+    if (handler != null) {
+      future.onComplete(handler);
+    }
+  }
+
+  /**
+   * @return the size of the file, or {@code -1} if not supported
+   */
+  default Future<Long> size() {
+    return Future.succeededFuture(-1L);
```
vietj, line 208: does it happen to have unsupported operations for this ?

src/main/java/io/vertx/core/file/AsyncFile.java
jponge: Same question as Paulo :)

src/main/java/io/vertx/core/file/AsyncFile.java
tsegismont: Same answer :smiley:

Seriously, if you both believe I'm over-cautious, I'm fine with removing this.

cc @pmlopes

src/main/java/io/vertx/core/file/AsyncFile.java
tsegismont: I removed the default implementations. PTAL

pmlopes (approved):

PR #3901, opened 2021-04-23
Ensure that Vert.x respects the top level CL

vietj:


src/main/java/io/vertx/core/impl/VertxThread.java
```diff
@@ -28,6 +28,7 @@
   private final TimeUnit maxExecTimeUnit;
   private long execStart;
   private ContextInternal context;
+  private ClassLoader topLevelTCCK;
```
vietj, line 31: TCCL, not TCCK

vietj:
ping @cescoffier @stuartwdouglas

cescoffier (approved):
LGTM

src/main/java/io/vertx/core/impl/VertxThread.java
```diff
@@ -28,6 +28,7 @@
   private final TimeUnit maxExecTimeUnit;
   private long execStart;
   private ContextInternal context;
+  private ClassLoader topLevelTCCL;
```
stuartwdouglas, line 31: Can we also get a setter for this? 

For Quarkus dev mode we have an issue where the vert.x instance is long lived, so it hangs onto the first deployments CL, and subsequent restarts result in Vert.x having the wrong CL.

stuartwdouglas (changes requested):

src/main/java/io/vertx/core/impl/VertxThread.java
vietj: I think the top level CL is set to null when it is not executing a Vert.x stack and it is not cached

stuartwdouglas: Ah, mis-read this this

stuartwdouglas (approved):

PR #3903, opened 2021-04-26
Local codecs

### LINKED ISSUE #3378: Clarify that HTTP keep alive value 0 means no eviction
Currently it is unclear about what setting HTTP keep alive value to 0 means. This specific value should imply that the connection is not evicted from the pool until it is closed.

slinkydeveloper:
Fix #3378

This PR adds the concept of local codecs, which is a way to simplify sending copyable and immutable data through the local event bus.

### DIFF
--- a/src/main/java/docoverride/eventbus/Examples.java
+++ b/src/main/java/docoverride/eventbus/Examples.java
@@ -14,7 +14,10 @@ package docoverride.eventbus;
 import io.vertx.core.eventbus.DeliveryOptions;
 import io.vertx.core.eventbus.EventBus;
 import io.vertx.core.eventbus.MessageCodec;
+import io.vertx.core.shareddata.Shareable;
 import io.vertx.docgen.Source;
+
+import java.util.function.Function;
 
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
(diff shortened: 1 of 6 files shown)


src/main/java/docoverride/eventbus/Examples.java
```diff
@@ -38,14 +41,29 @@ public void example11(EventBus eventBus, MessageCodec myCodec) {
     eventBus.send("orders", new MyPOJO());
   }
 
+  public void example12(EventBus eventBus) {
+    // Class does not implement Shareable
+    eventBus.registerDefaultCodec(MyPOJO.class, MessageCodec.localCodec(MyPOJO.class, MyPOJO::copy));
+
+    // Class implements Shareable
+    eventBus.registerDefaultCodec(MyPOJO.class, MessageCodec.localCodec(MyPOJO.class));
```
pmlopes, line 49: If it's `Shareable` wouldn't it be enough just to say:

```java
eventBus.registerDefaultCodec(MyPOJO.class);
```

It feels a bit redundant to have to declare a codec.

slinkydeveloper:
@pmlopes TBH my initial plan was to change the `EventBus` interface directly, something like:

```
eventBus.registerShareableDefaultCodec(MyPOJO.class);
eventBus.registerImmutableClassDefaultCodec(MyPOJO.class);
```

But that would not give the indication to the user of the "name" of the codec (we can write it in the docs). How do they sound like?

PrateekSahni (approved):

tsegismont:
Closing as won't fix. A solution could be to use a default selector https://github.com/eclipse-vertx/vert.x/issues/4365

PR #3912, opened 2021-05-06
Refactor JsonParser to consume all tokens and buffer events instead

vietj:


src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java
```diff
@@ -393,50 +388,42 @@ protected void _closeInput() {
     }
   }
 
-  private class BufferingHandler implements Handler<JsonToken> {
+  private static final EnumMap<JsonEventType, JsonToken> mm = new EnumMap<JsonEventType, JsonToken>(JsonEventType.class);
```
tsegismont, line 391: Not sure where this is used. Did I miss it?

tsegismont (approved):
@vietj great work!

tsegismont:
@vietj don't forget to close #3886

src/main/java/io/vertx/core/parsetools/impl/JsonParserImpl.java
vietj: you didn't miss anything, at some point I thought I would need it but finally I did not, so I will clean this up

adamdickmeiss:
I didn't get a chance to review.. Much cleaner solution than my hack.

fetch.. I still can't get my head around the fetch and demand handling.. In my book there should just not be arithmetic on demand when the value is MAX. So why not just use:
```
if (demand != Long.MAX_VALUE) {
  demand += amount;
}
```
Is it because you anticipate overflow for fetch in general?

IDEA says ..

Field `tokens` may be final:321

Private field `mm` is never used : 391

Inner class `BufferingHandler` may be `static`

PR #3931, opened 2021-05-18
Add alias property in JksOptions for net/http servers

### LINKED ISSUE #3926: Please enhance JksOptions to support selection of SSL key in keystore using an alias property
For context, I asked this question originally for Quarkus (see:  https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Any.20way.20to.20specify.20SSL.20key.20alias.20when.20using.20a.20keystore.20.3F ) and was asked to raise an issue for a Vert.x change.

There is currently no way to choose the SSL key alias for cases where a java key store may contain more than one key, thus no way to ensure the correct key is used when presenting the server certificate for SSL.

tsegismont:
Fixes #3926

This can be useful if users don't control the JKS file and it contains
multiple entries.

In this case, the SSL engine chooses the first one by default but it
might not what the user expects.


vietj:
this brings some coupling between SSLHelper and JksOptions which is something I would like to avoid. When an alias is used, I think we could create a new KeyStore that contains the alias we extract from the original KeyStore instead.

tsegismont:
@vietj ok, I'm not sure in which part of the code we'd do this though. Can you help?

vietj:
I think we could try hook in KeyStoreHelper static method `loadKeyStoreOptions` with an extra alias parameter that will load the specific alias and create new keystore with this alias.

tsegismont:
Thanks for the pointer.

>

tsegismont:
Superseded by #3933 3933

PR #3933, opened 2021-05-19
Add alias property in KeyStore options for net/http servers

### LINKED ISSUE #3926: Please enhance JksOptions to support selection of SSL key in keystore using an alias property

tsegismont:
Fixes #3926

This can be useful if users don't control the JKS file and it contains
multiple entries.

In this case, the SSL engine chooses the first one by default but it
might not be what the user expects.


src/main/java/io/vertx/core/net/KeyCertOptions.java
```diff
@@ -39,6 +39,17 @@
...
+  /**
+   * Like {@link #getKeyManagerFactory}, except the underlying keystore can be filtered.
+   * <p>
+   * This should be called only when creating a {@link KeyManagerFactory} for a TCP or HTTP server.
+   *
+   * @param vertx the vertx instance
+   * @return the key manager factory
+   * @see JksOptions#setAlias(String)
+   */
+  KeyManagerFactory getKeyManagerFactory(Vertx vertx, boolean filter) throws Exception;
```
vietj, line 51: is this used because the operation should only happen on the server ?

tsegismont: Yes

vietj:
I think we can avoid it since, the user simply should not set an alias on a
client JKS options

tsegismont:
Done

>

vietj:
shouldn't this option also apply to keystore in general ? e.g it should work too for `PfxOptions` or the generic `KeyStoreOptions` new in 4.0

vietj:
@tsegismont can you make this option generic for any keystore ?

tsegismont:
@vietj PTAL

vietj:
I think we are missing in `ssl.txt` the steps you did to create the new keystore, so when it gets invalid, anyone knows the procedure to generate it.

tsegismont:
@vietj I think I made them valid for the next 100 years ;-) But point taken.

tsegismont:
@vietj done

vietj:
waiting for CI green light and then we can merge this

vietj (approved):

PR #3953, opened 2021-06-03
Feature. VerticleGroup is a list of verticle that share the same context.

wang007:
when you haved a list of verticle that hope to run on the same context to avoid data race, you can used VerticleGroup.


wang007:
#3956

PR #3964, opened 2021-06-07
Do not deliver remote message to localOnly consumers

### LINKED ISSUE #3959: Local consumer gets distributed event if a non-local consumer with same address is present
### Version

4.1.0

### Context

See https://github.com/blindpirate/local-consumer-bug for reproduction

In short, if a local consumer and non-local consumer with same address are registered, the local consumer also receives message from other Vertx cluster node.

### Steps to reproduce

This is reproducible with both ZK and Hazelcast cluster manager, so I think it's more likely to be a Vertx core issue.

- ZooKeeper:
  - Run `./gradlew installDist` to package the application.
  - Start ZooKeeper at `2181`, e.g. `docker run -d -p 2181:2181 zookeeper:3.6.2`
  - Run `PORT=8080 CONSUMER=1 app/build/install/app/bin/app` start the event consumer.
  - In another terminal, run `PORT=8081 app/build/install/app/bin/app` to start the event producer.
  - No message displayed in event consumer. This is expected because now only local consumer is registered.
  - Now kill the consumer and producer.
  - Run `PORT=8080 CONSUMER=1 ENABLE_DISTRIBUTED_CONSUMER=1 app/build/install/app/bin/app` start the event consumer.
  - In another terminal, run `PORT=8081 ENABLE_DISTRIBUTED_CONSUMER=1 app/build/install/app/bin/app` to start the event producer.
  - Now you can see the local consumer gets a message from distributed event bus.
- Hazelcast:
  - Run `./gradlew installDist` to package the application. 
  - Run `PORT=8080 CONSUMER=1 MANAGER=hazelcast app/build/install/app/bin/app` start the event consumer.
  - In another terminal, run `PORT=8081 MANAGER=hazelcast app/build/install/app/bin/app` to start the event producer.
  - No message displayed in event consumer. This is expected because now only local consumer is registered.
  - Now kill the consumer and producer.
  - Run `PORT=8080 CONSUMER=1 MANAGER=hazelcast ENABLE_DISTRIBUTED_CONSUMER=1 app/build/install/app/bin/app` start the event consumer.
  - In another terminal, run `PORT=8081 MANAGER=hazelcast ENABLE_DISTRIBUTED_CONSUMER=1 app/build/install/app/bin/app` to start the event producer.
  - Now you can see the local consumer gets a message from distributed event bus.

tsegismont:
Fixes #3959

Consider a node that has both local and clustered consumers registered for an address.
When a remote message is received (sent from another node), calling deliverMessageLocally would lead to message being possibly delivered to one of the local consumers.
This is not what users expect.

With a variation of ConcurrentCyclicSequence, we can make sure we deliver:

- to all consumers when a message was sent from this node
- to clustered consumers only when a message was sent from another node

### DIFF
--- a/src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
+++ b/src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
@@ -45,13 +45,19 @@ public class ClusteredEventBusTest extends ClusteredEventBusTestBase {
   @Test
   public void testLocalHandlerNotVisibleRemotely() throws Exception {
     startNodes(2);
+    AtomicInteger received = new AtomicInteger();
+    vertices[1].eventBus().consumer(ADDRESS1).handler(msg -> {
+      if (received.incrementAndGet() == 2) {
+        vertices[1].setTimer(1000, id -> testComplete());
+      }
+    });
     vertices[1].eventBus().localConsumer(ADDRESS1).handler(msg -> {
-      fail("Should not receive message");
+      fail("Local consumer should not receive message");
     });
     vertices[0].eventBus().send(ADDRESS1, "foo");
     vertices[0].eventBus().publish(ADDRESS1, "foo");
-    vertices[0].setTimer(1000, id -> testComplete());
     await();
+    assertEquals(2, received.get());
   }
 
   @Test
--- a/src/main/java/io/vertx/core/eventbus/impl/ClusteredEventBusHandlers.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/ClusteredEventBusHandlers.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.eventbus.impl;
+
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@SuppressWarnings("rawtypes")
+public class ClusteredEventBusHandlers implements Handlers {
+
+  private static final HandlerHolder[] EMPTY_ARRAY = new HandlerHolder[0];
+
+  private final HandlerHolder[] elements;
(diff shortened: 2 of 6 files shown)


tsegismont:
@vietj sharing a draft PR to get some feedback first and see what CI says.

src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
```diff
@@ -45,13 +45,19 @@
...
+    });
     vertices[1].eventBus().localConsumer(ADDRESS1).handler(msg -> {
-      fail("Should not receive message");
+      fail("Local consumer should not receive message");
     });
     vertices[0].eventBus().send(ADDRESS1, "foo");
     vertices[0].eventBus().publish(ADDRESS1, "foo");
-    vertices[0].setTimer(1000, id -> testComplete());
     await();
+    assertEquals(2, received.get());
```
vietj, line 60: It is incorrect from the testing unit perspective: when this statement is reached, `testComplete()` has been called and the test is considered over.

When the test passes it is fine, when the test fails (e.g a regression) then a failure is reported by `AsyncTestBase` that will fail with an `IllegalStateException` instead of the correct error to report.

It can be fixed either by calling junit's `assertEquals` method, or by using a latch without a `testComplete()` to be called

vietj:
I can see duplicated code (`ConcurrentCyclicSequence`) and I think we should try to do better than that

src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
tsegismont: Ok I will update it

tsegismont:
@vietj right. Actually I was thinking about removing `ConcurrentCyclicSequence` and having a single class instead of 4.

I believe there was an attempt to make it a reusable collection-like container but actually it's only used by the EventBus. Perhaps it's used in other modules? I will check.

tsegismont:
It seems it's not used elsewhere: https://github.com/search?q=org%3Aeclipse-vertx+org%3Avert-x3+ConcurrentCyclicSequence&type=code

vietj:
I think that we could provide a Predicate<T> in the
ConcurrentCyclicSequence that would skip instances that do not satisfy the
predicate and implement both cases.

tsegismont:
This was my first idea but I didn't pursue because:

- implementing `size()` would require iterating over the elements
- in the worst case, invoking `next()` may lead to iterator over (n-1) elements before finding the only element that matches the predicate again

It seemed to me that the effort/price to maintain a generic container (which is actually not used elsewhere) was not justified.

vietj:
I think important concerns here are

- how many elements are we talking about ?
- how often does this case happen ?

it seems to affect 3.9 and that's the first time it's ever reported.

PR #3975, opened 2021-06-19
Add file paths to exception details

imanushin:
*Motivation:*
Add exception details for the most of file IO access errors.

*Explain here the context, and why you're making that change, what is the problem you're trying to solve.*
Explain errors for developers. It is hard to determine the missing file if you do loop-based operations and you can't debug.

Conformance:


vietj:
can we avoid adding a new class only for the static methods ? thanks

imanushin:
@vietj , I don't know how it can be done unfortunately. May I ask you about some advice?

For current moment this class is used from 4 other (please check screenshot from the IntelliJ Idea below). Moreover, callers use different methods.

The my initial idea was to put message formats into the code, however that produced a lot of copy-pasted texts.

I had also another idea to put error text generation into the one of existing classes, however this breaks Single Responsibility principle.

May I ask you for any advise, how can I avoid special class for error message generation?

image

vietj:
I think having such static methods as package private in FileSystemImpl is
fine

imanushin:
@vietj, done, thank you.

src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
```diff
@@ -602,6 +602,26 @@ public String createTempFileBlocking(String dir, String prefix, String suffix, S
...
+
+  static String getFileCopyErrorMessage(String from, String to) {
+    return getFileDualOperationErrorMessage("copy", from, to);
+  }
+
+  static String getFileMoveErrorMessage(String from, String to) {
+    return getFileDualOperationErrorMessage("move", from, to);
+  }
+
+  static String getFileDualOperationErrorMessage(String verb, String from, String to) {
```
vietj, line 621: what means dual here ?

imanushin: It means that two paths are involved here. Operation ```delete file``` requires single file path, however operation ```copy file``` requires two paths (from and to).

I decided to extract function to cover this case. So logs will have both paths.

imanushin:
@vietj , may I ask you to check if there are corrections needed/suggested in PR?

src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
```diff
@@ -602,6 +602,26 @@ public String createTempFileBlocking(String dir, String prefix, String suffix, S
     return createTempFileInternal(dir, prefix, suffix, perms).perform();
   }
 
+  static String getFileAccessErrorMessage(String verb, String path) {
```
vietj, line 605: rather _action_ than _verb_

vietj:
Added a minor comment and then we are good to go

imanushin:
@vietj , thank you, done

vietj:
can you rebase on latest master :-) ?

imanushin:
@vietj , rebase had been done, thank you

vietj:
thank you

Issue #3977, opened 2021-06-21
Let VertxInternal#sharedTCPServers  allow other server to be used


wang007:
#### purpose
Write netty code directly in vert.x  Reuse both vert.x threading model and vert.x components

because 
```java
public <S extends TCPServerBase> Map<ServerID, S> sharedTCPServers(Class<S> type) {
    if (NetServerImpl.class.isAssignableFrom(type)) {
      return (Map<ServerID, S>) sharedNetServers;
    } else if (HttpServerImpl.class.isAssignableFrom(type)) {
      return (Map<ServerID, S>) sharedHttpServers;
    } else {
      throw new IllegalStateException();
    }
  }
```
So you can't do it
```java
public class DemoServer extends TCPServerBase {
  
  public DemoServer(VertxInternal vertx, NetServerOptions options) {
    super(vertx, options);
  }

  @Override
  public Future<Void> close() {
    return null;
  }
}
```

#### Describe the feature

VertxInternal#sharedTCPServers(Class type) implement allow other server class to be used

#### Use cases

```java
public class DemoServer extends TCPServerBase {
  
  public DemoServer(VertxInternal vertx, NetServerOptions options) {
    super(vertx, options);
  }

  @Override
  public Future<Void> close() {
    return null;
  }
}


public class SomeVerticle extends AbstractVerticle {
  @Override
  public void start() throws Exception {
    DemoServer server = DemoServer.create(vertx, options);
    // netty codec,decode, handle
   server.listen(port);
  }

```

#### Contribution
I can contribute this feature

vietj:
sure thing but that would remain an internal class though

vietj:
well actually why don't you simply wrap a NetServer ?

wang007:
> well actually why don't you simply wrap a NetServer ?

because write netty code directly in vert.x.  Netty has a rich encoder/decoder library and reuse the Netty Codec directly.

Either way, is it convenient to allow TCPServerBase to be used by the Other Server Class

vietj:
you can reuse Netty codecs with NetSocket :
http://www.julienviet.com/advanced-vertx-guide/

wang007:
> ChannelPipeline pipeline = so.channelHandlerContext().pipeline();
> pipeline.addFirst(codec);

That seems to work.

Under the discussion of the China user group, Some of you are familiar with Netty and want to start with Raw Netty. But they want to reuse the ver's threading model and components. 

[...]

PR #3978, opened 2021-06-21
VertxInternal#sharedTCPServers allow other server to be used

wang007:
Let VertxInternal#sharedTCPServers allow other server to be used #3977


vietj:
I'm wondering if somehow we should not try instead to make an HTTP server to be a Net server

wang007:
> I'm wondering if somehow we should not try instead to make an HTTP server to be a Net server

Sorry,  I didn't get your point.  = =!

PR #3982, opened 2021-06-22
Keystore enhancements

### LINKED ISSUE #3969: Non-exitent KeyStore alias is not reported as an error
### Version

4.1.0

### Context

While integrating the alias feature (#3926/#3933, thanks for that!) in Quarkus (https://github.com/quarkusio/quarkus/pull/17884), I noticed that I can specify a bogus alias via `io.vertx.core.net.KeyStoreOptions.setAlias(String)` and vert.x just doesn't mind.
This makes it unecessary user-unfriendly to find configuration mistakes.

### Do you have a reproducer?

I don't have a small and simple reproducer but the problem can be observed in Quarkus.

* https://github.com/quarkusio/quarkus.git

### Steps to reproduce

1. `mvn -Dquickly -T1.5C`
2. In`integration-tests/vertx-http/src/main/resources/application.properties` change `quarkus.http.ssl.certificate.key-store-key-alias=server` to `=foobar`
3. `mvn clean verify -f integration-tests/vertx-http/ -Dtest=Http2TestCase`
4. Test fails with a `handshake_failure` when the test request is made instead of erroring out early on startup

### Extra

* Discussion in Quarkus PR: https://github.com/quarkusio/quarkus/pull/17884#discussion_r650659134

### LINKED ISSUE #3970: Introduce support for key password (not store password!) in KeyStoreOptions
#### Describe the feature

There is `io.vertx.core.net.KeyStoreOptions.setPassword(String)` for the store and `.setAlias(String)` to select a specific key from the store, but there is nothing to set the _password for a specific key_.

#3452 seems closely related.

#### Use cases

In "enterprisy setups"™, more often than not you have multipe key pairs in a single key store and each of it is protected by an individual password. In such cases, just having a keystore password is not sufficient.
FWIW: Spring and WildFly both have something like `key-password` in config.

Btw, something like `.setKeyPassword(String)` would make sense to me.
In that context, maybe `.setAlias(String)` should be renamed to `.setKeyAlias(String)` to make things clearer.
See also this discussion over at Quarkus: https://github.com/quarkusio/quarkus/pull/17884#issuecomment-860579222

tsegismont:
Fixes #3969, fixes #3970

Changes in this PR:

- fail if `alias` option is set but it does not exist in the keystore
- add `aliasPassword` option to be able to retrieve the key when it has its own password (instead of using the store's password)


tsegismont:
@famod can you please take a look? Thank you

src/main/java/io/vertx/core/net/KeyStoreOptionsBase.java
```diff
@@ -59,6 +60,7 @@ protected KeyStoreOptionsBase(KeyStoreOptionsBase other) {
     this.path = other.path;
     this.value = other.value;
     this.alias = other.alias;
+    this.alias = other.aliasPassword;
```
vietj, line 63: that seems incorrect

src/main/java/io/vertx/core/net/KeyStoreOptionsBase.java
```diff
@@ -154,6 +156,23 @@ public KeyStoreOptionsBase setAlias(String alias) {
     return this;
   }
 
+  /**
+   * @return the alias password for a server certificate when the keystore has more than one, or {@code null}
```
vietj, line 160: more than one password ? it reads ambiguous (could be alias)

vietj (changes requested):
Some changes

famod:
@tsegismont I'll try to have a look later this week. Sorry, really busy.
But thanks a lot for taking care of it!

src/main/java/io/vertx/core/net/KeyStoreOptionsBase.java
vietj: ping @tsegismont

tsegismont:
```suggestion
    this.aliasPassword = other.aliasPassword;
```

src/main/java/io/vertx/core/net/KeyStoreOptionsBase.java
tsegismont:
```suggestion
   * @return the password for the server certificate designated by {@link #getAlias()}, or {@code null}
```

src/main/java/io/vertx/core/net/KeyStoreOptionsBase.java
```diff
@@ -154,6 +156,23 @@ public KeyStoreOptionsBase setAlias(String alias) {
...
+   */
+  public String getAliasPassword() {
+    return aliasPassword;
+  }
+
+  /**
+   * Set the alias password for a server certificate when the keystore has more than one.
```
tsegismont, line 167:
```suggestion
   * Set the password for the server certificate designated by {@link #getAlias()}.
```

tsegismont:
@vietj the linked issues already have the `milestone` field set. Do we need a `milestone` on this PR too?

vietj:
yes because it makes the issue appear in GitHub filters

tsegismont:
OK. Any other comment to this PR?

famod:
Wasn't able to test it myself yet (in Quarkus), but LGTM!
I'll try to integrate a vert.x snapshot in Quarkus later this week, to see if it's working there as well as expected. But I cannot promise anything.

vietj (approved):
👍

PR #4005, opened 2021-07-07
Do not create a VertxTracer if TracingOptions are null

Ultranium:
A fix for the Issue https://github.com/eclipse-vertx/vertx-tracing/issues/24

Starting with the Vert.x 4.0.3, if there is a `vertx-opentracing` artifact in the classpath and no TracingOptions set for a Vertx instance, the `IllegalArgumentException: Service name must not be null or empty` exception is raised on a Vertx instance initialization.

This patch avoids instantiation of a new VertxTracer if the TracingOptions are null.

### DIFF
--- a/src/main/java/io/vertx/core/spi/VertxTracerFactory.java
+++ b/src/main/java/io/vertx/core/spi/VertxTracerFactory.java
@@ -31,7 +31,7 @@ public interface VertxTracerFactory extends VertxServiceProvider {
 
   @Override
   default void init(VertxBuilder builder) {
-    if (builder.tracer() == null) {
+    if (builder.options().getTracingOptions() != null && builder.tracer() == null) {
       builder.tracer(tracer(builder.options().getTracingOptions()));
     }
   }


vietj:
thanks, can you add a test for this case ?

Ultranium:
Sorry, I don't yet feel competent enough to write a test for something like this.

I checked the [VertxFactoryTest.java](https://github.com/eclipse-vertx/vert.x/blob/11ab144725e551be591118831377101015e1392c/src/test/java/io/vertx/core/impl/VertxFactoryTest.java) and found that it will be kind of tricky to check if a Tracer was created or not, because to do so you have to provide a TracingOptions() with a fake VertxTracerFactory, and it ruins the whole test case with null TracingOptions.

vietj:
@Ultranium I was able to create an integration test for this

vietj:
I commited your fix along with a test, thanks for reporting this and investigating @Ultranium

PR #4026, opened 2021-07-21
Make methods on io.vertx.core.impl.WorkerPool public

Sanne:
This would be very useful to improve integration in Quarkus.

Thanks!

### DIFF
--- a/src/main/java/io/vertx/core/impl/WorkerPool.java
+++ b/src/main/java/io/vertx/core/impl/WorkerPool.java
@@ -28,11 +28,11 @@ public class WorkerPool {
     this.metrics = metrics;
   }
 
-  ExecutorService executor() {
+  public ExecutorService executor() {
     return pool;
   }
 
-  PoolMetrics metrics() {
+  public PoolMetrics metrics() {
     return metrics;
   }
 


tsegismont (approved):
LGTM

vietj:
Do you intent to use the worker pool directly ? using it directly will not report metrics as this is the raw pool.

Sanne:
@vietj we have a very weird requirement.. it relates to what I sent you an email about (subject "VertxThread.DISABLE_TCCL") and this patch I made yesterday to reset the right classloader context on all running Threads in the executor:

 - https://github.com/quarkusio/quarkus/pull/18912/commits/0e412f1cd8671480dcc6ccb58b95a5ab9d52df00

What I need to do is grab a reference to the Executor, which in our case will be of type `EnhancedQueueExecutor` from JBoss Threads, so to access all running threads and reset the classloader:

https://github.com/quarkusio/quarkus/pull/18912/commits/0e412f1cd8671480dcc6ccb58b95a5ab9d52df00#diff-a38e0d86cf6a637c19b6e0a0e23959f644886bdcc0f0e5615ce7cfa0e6bc9909R138-R141

There's other alternatives, but making these getters public seemed quite harmless as the package is marked "internal" anyway so hopefully there can be a little flexibility :)

vietj:
yes this is fine from the visibility perpective in Vertx internal API, I
was only worried about the usage on your side

vietj:
can you cherry pick on 4.1 branch @Sanne ?

Sanne:
> can you cherry pick on 4.1 branch @Sanne ?

done! https://github.com/eclipse-vertx/vert.x/pull/4029

Thanks

PR #4028, opened 2021-07-22
Reuse emit func in FutureBase

drriguz:
A refactor of the emitSuccess/emitFailure func, removed duplicated code blocks


vietj:
this is done on purpose, because the issue of this PR is that it will instante an object because of `listener::onSuccess` and we want to avoid this creation

drriguz:
Thanks for your clarification :p

PR #4041, opened 2021-07-27
Use KeyStore.getDefaultType() for internal KeyStore type, but use 'PKCS12' over 'JKS' if available

### LINKED ISSUE #4031: PKCS12 Keystore format when running in a FIPS-restricted JRE
### Version

3.9.8, but later versions appear to have the same issue.

### Context

We are developing a number of applications that need to be able to run in a JRE that has all non-FIPS compliant encryption formats disabled/removed. We have a `BCFKS` KeyStore, and we are able to read it in, however, we get an error because Vert.x then tries to initialize an internal `PKCS12` KeyStore, which is not present in our environment:
```
io.vertx.core.VertxException: java.security.KeyStoreException: PKCS12 not found
```
We traced the issue to the `createEmptyKeyStore` function in `/src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java`, which hard-codes the new empty keystore to the `PKCS12` format. We verified that we could create a custom build of Vert.x with the hard-coded value changed to `BCFKS`, and Vert.x then successfully runs on the affected environment, however this is not a practical long-term solution. 

### Do you have a reproducer?

To reproduce, you will need to remove the `PKCS12` KeyStore format from the JRE's `lib/security/java.security` file. To do this, find the list of providers and comment out the line containing `com.sun.net.ssl.internal.ssl.Provider`. Save the file, and then read in a (non-PKCS12) KeyStore using Vert.x.

kjrokos:
Use KeyStore.getDefaultType() for internal KeyStore type, but bias against using 'JKS' in favor of 'PKCS12'

Resolves #4031 

This solution:
* Allows a user to specify the keystore type by editing the `keystore.type` in the `security.java` file
* Allows a user to override the global keystore type by passing in `-Dkeystore.type=...`
* No regression from previous implementation for Java 8 users 
  * Keeps Java 8 users on 'PKCS12', if they are using the default 'JKS' keystore type)
  * Will gracefully fallback to 'JKS' if it is the default and 'PKCS12' is not available 

This seems like the best balance of configurability while maintaining the previous behavior. That said, I'm more than happy to switch approaches here if necessary.


src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2021 VMware, Inc. and others
```
kjrokos, line 2: I believe this conforms to both my employer's guidelines and yours, but I'm a bit new to this - please let me know if anything is amiss here from your POV. Thanks.

vietj: you need to sign the Eclipse Contributor Agreement to validate this PR and thus you become a contributor of the foundation.

vietj: there is a check in the PR that says you are not covered by the ECA yet.

kjrokos: I've signed the ECA and looks like the check is passing now.

jponge: You need to keep `Contributors to the Eclipse Foundation`

jponge: Technically this is what you are doing, you are a contributor to the Eclipse Foundation because you signed the ECA to contribute. This is per-foundation guidelines.

Otherwise we'd have every company adding their own name here, and it would be a mess.

jponge (changes requested):

jponge (commented):
Thanks for reverting the copyright line

jponge (approved):
Looks good to me, but I'll let @vietj / @pmlopes on security stuff

pmlopes (approved):

vietj:
do we need to document this as a breaking change @pmlopes ?

vietj:
I think also we should add to this PR title that when PKCS12 is available then it takes precedence over JKS

vietj:
can you rebase on latest master branch ? thanks

kjrokos:
Yup, done

vietj:
thanks, that allowed CI to pass

PR #4068, opened 2021-08-17
Context: allow storage and retrieval of instances of any type

### LINKED ISSUE #3952: Allow other types beyond String being used as key to store state in Context
#### Describe the feature

In class `io.vertx.core.Context` : methods `getLocal`, `putLocal`, `removeLocal` all require a key of type String.

Would be great to relax this requirement and accept any object types (Object?).

#### Use cases

We have to store a custom object with sub-scope identifiers as well; the current API is forcing us to encode additional scope as post-fixes on a shared root String, but this implies we have to allocate new String instances on each access.

Being able to use a composite object would still require allocation, but encoding of parts of scoping would be really cheap. Not leat, String comparison isn't very efficient when we have to use a shared prefix.

#### Contribution

I'm happy to contribute, if I can be guided towards which solutions are acceptable. It seems it might just be a matter of changing the signature on the API, but I'm not sure of the implications or additional requirements. For example, is this state supposed to be serializable?

tsegismont:
Closes #3952

The Context interface used to restrict storage and retrieval of objects in context to String instances.

But the underlying storage permits to use instances of any type.

With this change, users can now store any type of objects in the context, without having to cast to ContextInternal and work with data maps directly.

### DIFF
--- a/src/main/java/io/vertx/core/Context.java
+++ b/src/main/java/io/vertx/core/Context.java
@@ -196,7 +196,7 @@ public interface Context {
    * @param <T>  the type of the data
    * @return the data
    */
-  <T> T get(String key);
+  <T> T get(Object key);
 
   /**
    * Put some data in the context.
@@ -206,7 +206,7 @@ public interface Context {
    * @param key  the key of the data
    * @param value  the data
    */
-  void put(String key, Object value);
+  void put(Object key, Object value);
 
   /**
    * Remove some data from the context.
@@ -214,7 +214,7 @@ public interface Context {
    * @param key  the key to remove
    * @return true if removed successfully, false otherwise
    */
-  boolean remove(String key);
+  boolean remove(Object key);
 
   /**
    * Get some local data from the context.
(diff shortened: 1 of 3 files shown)


tsegismont:
cc @Sanne 

@vietj I checked the documentation, there is no need for a change. If this is merged, I will create an entry in the breaking changes page.

Sanne:
@tsegismont thanks for taking care of this!

Sanne:
BTW, not sure if I'd consider it a "breaking change" , it should be backwards compatible no?

tsegismont:
@Sanne You're welcome.

Although users source code wouldn't have to change, they will have to recompile their application. In that sense, it is a breaking change.

vietj:
note : this should be listed as a binary breaking change in the wiki page

tsegismont:
@vietj right. Is the PR ok to merge?

Sanne:
you could also keep the original methods and deprecate them?

vietj:
I would rather avoid that (un-necessary noise) if the application does not
need to updated to continue working

vietj:
backport PR : https://github.com/eclipse-vertx/vert.x/pull/4073

vietj:
@Sanne if we can avoid this in 4.1 branch I would prefer, how does that sound to you ?

Sanne:
Sure no rush at all, we simply use the internal API in HR and that's working fine currently.

PR #4070, opened 2021-08-18
Don't create a Promise from the finalizer and un-necessary remove HTTP client finalizer

stuartwdouglas:
Motivation:

When this finalizer is run it attempts to notify the empty promise on the IO thread. 

This can result in spawning a new IO thread, which will then inherit the
TCCL and ThreadGroup of the finalizer thread.
    
This can still be an issue if the client is not closed properly, but
currently there is no way to avoid it even if you do close it properly.

In addition the HTTP client finaliser is not necessary as client cleanup is achieved by the closeFuture listener that is either close directly or by the NetClient finaliser.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
@@ -646,13 +646,4 @@ public class HttpClientImpl implements HttpClient, MetricsProvider, Closeable {
       throw new IllegalStateException("Client is closed");
     }
   }
-
-  @Override
-  protected void finalize() throws Throwable {
-    // Make sure this gets cleaned up if there are no more references to it
-    // so as not to leave connections and resources dangling until the system is shutdown
-    // which could make the JVM run out of file handles.
-    close((Handler<AsyncResult<Void>>) Promise.<Void>promise());
-    super.finalize();
-  }
 }
--- a/src/main/java/io/vertx/core/net/impl/NetClientImpl.java
+++ b/src/main/java/io/vertx/core/net/impl/NetClientImpl.java
@@ -280,7 +280,7 @@ public class NetClientImpl implements MetricsProvider, NetClient, Closeable {
     // Make sure this gets cleaned up if there are no more references to it
     // so as not to leave connections and resources dangling until the system is shutdown
     // which could make the JVM run out of file handles.
-    close((Handler<AsyncResult<Void>>) Promise.<Void>promise());
+    closeFuture.close();
     super.finalize();
   }
 }


src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -646,13 +646,4 @@ private void checkClosed() {
       throw new IllegalStateException("Client is closed");
     }
   }
-
-  @Override
-  protected void finalize() throws Throwable {
```
Sanne, line 651: I agree :)

src/main/java/io/vertx/core/net/impl/NetClientImpl.java
```diff
@@ -280,7 +280,7 @@ protected void finalize() throws Throwable {
     // Make sure this gets cleaned up if there are no more references to it
     // so as not to leave connections and resources dangling until the system is shutdown
     // which could make the JVM run out of file handles.
-    close((Handler<AsyncResult<Void>>) Promise.<Void>promise());
+    closeFuture.close();
```
Sanne, line 283: I'd remove the method?

stuartwdouglas: That would be my preference, but it is up to @vietj . If he agrees I will remove both of them.

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
stuartwdouglas: This one just calls NetClientImpl.close(), and it has its own finalizer.

vietj:
I disagree with this change, removing the finalize method will solve your problem but raise new issue for other users as the finalizer is used to dispose resources.

if you can describe precisely the actual issue we can find a better solution to solve it.

vietj:
the HTTP client also needs to cleanup resources like connection pools

vietj:
If the finalize method close the client whereas it was previously explicitly closed by the Quarkus (or any application) then that is a bug and we must solve it. When the client is closed before it is eligible to GC, then the finalizer should be a no-op and no IO thread would be spawn.

Note that indeed using here the acceptor thread does not seem adequate and instead we could use Netty's GlobalEventExecutor instead.

stuartwdouglas:
@vietj NetClient already has a finalize() method, the HttpClient finalize() just calls close() again, on an object that already has its own finalize method. 

The NetClient changes stops the finalize() method dispatching what is basically an empty task to a Netty event loop. This is a problem as if it causes a new thread to be created you end up with an IO thread that belongs to the 'system' thread group and with the wrong TCCL.

Basically this happens because when you pass the promise into the close() method, but cast it to Handler, it then gets wrapped in another promise. The close future then attempts to dispatch to the IO thread do execute what is basically an empty promise. Just calling closeFuture.close() directly in the finalize method avoids all this.

vietj:
I see more clearly now, the HttpClient piggy back on the NetClient finalize
method.

since the HttpClient add itself as a listener of the closeFuture then it
gets cleaned too.

vietj:
thanks for the explanation it was not obvious that the HttpClient gets cleaned too.

I will add a couple of comments

vietj:
basically we should add somewhere in HttpClient that its resources gets cleaned by the delegated NetClient finalizer

src/main/java/io/vertx/core/net/impl/NetClientImpl.java
vietj: @Sanne we should at some point get rid of finalizer but we need to keep this cleanup. I know that Java 9 has a cleaner that can be used by Vert.x still relies on Java 8.

vietj:
back-ported https://github.com/eclipse-vertx/vert.x/commit/b3d46bcb1740c730dc8d4c0a87124fc4ad99411d

src/main/java/io/vertx/core/net/impl/NetClientImpl.java
Sanne: I understand and no rush :)

But I suppose you know well that a finalize() method is not guaranteed to be invoked, and could be invoked multiple times? In case you prefer to keep it, the methods it calls should at least be idempotent, so to allow a double-close event.

I haven't tried it but I think you should be able to replace any finalizer with a ReferenceQueue  + PhantomReference combination, which would give you:

1. better quarantees of it being invoked at all
2. still compatible with Java 8
3. will also work when finalize() no longer works

Might be worth a shot? No pressure, just writing down the idea for future.

vietj:
I think reference queue requires to have a dedicated thread taking care of
it, or at least a periodic check.

For now I think I want to have all possible options

PR #4071, opened 2021-08-19
Introduce a flag to disable ContextClassLoader control by VertxThread

Sanne:
Motivation:

During the investigation of a classloader issue in Quarkus, we had reasons to believe that the ContextClassLoader reset mechanism in `VertxThread` had a good chance to be the cause of the problem.

After further analysis we figured that this wasn't the case, so there isn't a concrete known issue solved by this, but also it's not hard to think of situations in which this would cause real issues, and more in general in Quarkus it's not at all useful for Vert.x to control the TCCL.

So I'd suggest introducing this flag: should be harmless and it has the potential to prevent further related issues. Maybe even save some CPU cycles :)

cc/ @stuartwdouglas @gsmet

### DIFF
--- a/src/main/java/io/vertx/core/impl/ContextImpl.java
+++ b/src/main/java/io/vertx/core/impl/ContextImpl.java
@@ -55,6 +55,9 @@ abstract class ContextImpl extends AbstractContext {
   private static final String DISABLE_TIMINGS_PROP_NAME = "vertx.disableContextTimings";
   static final boolean DISABLE_TIMINGS = Boolean.getBoolean(DISABLE_TIMINGS_PROP_NAME);
 
+  private static final String DISABLE_TCCLSWAPPING_PROP_NAME = "vertx.disableTCCLSwapping";
+  static final boolean DISABLE_TCCLSWAPPING = Boolean.getBoolean(DISABLE_TCCLSWAPPING_PROP_NAME);
+
   protected final VertxInternal owner;
   protected final JsonObject config;
   private final Deployment deployment;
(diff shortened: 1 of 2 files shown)


stuartwdouglas:
I don't know if we want this. I would prefer we go down the `Supplier<ClassLoader>` route that gives us explicit control over the threads.

The issue is that if something somewhere breaks the TCCL of a vert.x thread (e.g. the finalizer issue we saw) then everything falls apart, and it is incredibly difficult to debug why the TCCL is now wrong (as you don't get the error at the code that has set the wrong TCCL, but generally on the next task the thread performs).

I think that safety is more important than saving a few cycles here. If we want this for benchmarks then I am ok with it, but I don't think we want this in the general sense.

Sanne:
I think we need both?

Wasn't thinking of this PR as an alternative to your supplier idea, but
complementary.

Sanne:
p.s. I wasn't at all suggesting this for performance reasons, just to make
sure nothing would swap any ClassLoader behind our back. The thought that
it might also save some CPU cycles is just a side effect.

stuartwdouglas:
My concern with this PR is that if something does swap a CL behind our back, vert.x won't restore it, so the bad CL hangs around forever.

Sanne:
> My concern with this PR is that if something does swap a CL behind our back, vert.x won't restore it, so the bad CL hangs around forever.

Ok good point - but this will also make it harder to diagnose such issues.  Besides, I'm not really sure vert.x would restore the "right" CL so you'd end up with the wrong one forever?

I'll close this based on the fact you're not convinced, but I still think we need to be more defensive about such things happening. I thought preventing the CLs to be swapped at all would be a good start.

PR #4079, opened 2021-08-23
Implement shared HttpClient support

### LINKED ISSUE #3361: Shared HttpClient support
A few Vert.x clients (e.g. JDBC, Cassandra, Mongo) have `createShared` methods. The purpose of these creation methods is that, when you deploy multiple verticle instances, a single resource (JDBC datasource, Mongo client, Cassandra session) is created and reused.

While this requirement was obvious for Vert.x clients wrapping existing libraries, most built-in clients don't have them. So if a user has to open at most `C` connections to an HTTP backend, and deploys `V` verticle instances, it is advised to create an `HttpClient` with `C/V` as max pool size.

On the upside, the user will get maximum performance because the HttpClient will be operated on the verticle context. On the downside, the configuration process is not straightforward.

For Vert.x 4, HttpClient (and some other built-in clients too) have been made safe to use on different contexts. So it is possible to add `createShared` methods now.

The benefit is simplified configuration. The drawback is that this will require an extra hop between the context that actually created the `HttpClient` and the context using it. This should be acceptable by most apps though (the user pays at least this price when sending messages over the eventbus).

tsegismont:
Closes #3361

The shared client operates on its own context instead of the creation context.
This is to avoid keeping around a context that could be closed (like the context of an undeployed verticle).

However, we make sure replies to requests are emitted on the sender context.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
@@ -169,7 +169,8 @@ public class HttpClientImpl implements HttpClient, MetricsProvider, Closeable {
     httpCM = httpConnectionManager();
     if (options.getPoolCleanerPeriod() > 0 && (options.getKeepAliveTimeout() > 0L || options.getHttp2KeepAliveTimeout() > 0L)) {
       PoolChecker checker = new PoolChecker(this);
-      timerID = vertx.setTimer(options.getPoolCleanerPeriod(), checker);
+      ContextInternal timerContext = vertx.createEventLoopContext();
+      timerID = timerContext.setTimer(options.getPoolCleanerPeriod(), checker);
     }
 
     closeFuture.add(netClient);
--- a/src/main/java/io/vertx/core/http/impl/SharedHttpClient.java
+++ b/src/main/java/io/vertx/core/http/impl/SharedHttpClient.java
@@ -1,0 +1,207 @@
+/*
+ * Copyright (c) 2011-2021 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.http.impl;
+
+import io.vertx.core.*;
+import io.vertx.core.http.*;
+import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.shareddata.LocalMap;
+
+import java.util.List;
+import java.util.function.Function;
+
+public class SharedHttpClient implements HttpClient, Closeable {
+
+  private static final String MAP_NAME = "__vertx.shared.httpClients";
+  public static final String DEFAULT_CLIENT_NAME = "SharedHttpClient.DEFAULT";
+
+  private final VertxInternal vertx;
+  private final String name;
+  private final HttpClient delegate;
+
+  private SharedHttpClient(VertxInternal vertx, String name, HttpClient delegate) {
+    this.vertx = vertx;
+    this.name = name;
+    this.delegate = delegate;
+  }
+
+  public static SharedHttpClient create(VertxInternal vertx, String name, HttpClientOptions options) {
+    LocalMap<String, HttpClientHolder> localMap = vertx.sharedData().getLocalMap(MAP_NAME);
+    HttpClient client;
+    HttpClientHolder current, candidate;
+    for (; ; ) {
+      current = localMap.get(name);
+      if (current != null) {
+        candidate = current.increment();
+        if (candidate != null && localMap.replaceIfPresent(name, current, candidate)) {
+          client = candidate.get();
+          break;
+        }
+      } else {
+        candidate = new HttpClientHolder();
+        if (localMap.putIfAbsent(name, candidate) == null) {
+          candidate = candidate.init(vertx, options);
+          client = candidate.get();
+          localMap.put(name, candidate);
+          break;
+        }
+      }
+    }
+    return new SharedHttpClient(vertx, name, client);
+  }
+
+  @Override
+  public void close(Handler<AsyncResult<Void>> handler) {
+    Future<Void> future = close();
+    if (handler != null) {
+      future.onComplete(handler);
+    }
+  }
+
+  @Override
+  public Future<Void> close() {
+    Promise<Void> promise = vertx.promise();
+    LocalMap<String, HttpClientHolder> localMap = vertx.sharedData().getLocalMap(MAP_NAME);
+    HttpClientHolder current, candidate;
+    for (; ; ) {
+      current = localMap.get(name);
+      candidate = current.decrement();
+      if (candidate == null) {
+        if (localMap.removeIfPresent(name, current)) {
+          current.close().onComplete(promise);
+          break;
+        }
+      } else if (localMap.replace(name, current, candidate)) {
+        promise.complete();
+        break;
+      }
+    }
+    return promise.future();
+  }
+
+  @Override
+  public void close(Promise<Void> completion) {
+    close().onComplete(completion);
+  }
+
+  @Override
+  protected void finalize() throws Throwable {
+    // Make sure the shared client count gets decreased if there are no more references to this instance
+    close();
+    super.finalize();
+  }
+
+  @Override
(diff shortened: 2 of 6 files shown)


src/main/java/io/vertx/core/http/impl/WrappedHttpClient.java
```diff
@@ -0,0 +1,150 @@
...
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.MultiMap;
+import io.vertx.core.http.*;
+
+import java.util.List;
+import java.util.function.Function;
+
+public class WrappedHttpClient implements HttpClient {
```
vietj, line 23: this class is not necessary

src/main/java/io/vertx/core/http/impl/SharedHttpClient.java
```diff
@@ -0,0 +1,92 @@
...
+
+package io.vertx.core.http.impl;
+
+import io.vertx.core.*;
+import io.vertx.core.http.HttpClient;
+import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.shareddata.LocalMap;
+
+public class SharedHttpClient extends WrappedHttpClient implements Closeable {
```
vietj, line 20: the client needs to override finalize so it get released when it is gc-ed like the current HTTP client

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -120,6 +112,7 @@
 
 
   private final VertxInternal vertx;
+  private final ContextInternal context;
```
vietj, line 115: the client should not depend on a specific context

src/main/java/io/vertx/core/http/impl/SharedHttpClient.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Override
+  public Future<Void> close() {
+    Promise<Void> promise = vertx.promise();
+    LocalMap<String, HttpClientHolder> localMap = vertx.sharedData().getLocalMap(MAP_NAME);
+    HttpClientHolder current, candidate;
+    for (; ; ) {
+      current = localMap.get(name);
+      candidate = current.decrement();
```
vietj, line 74: we should be using map compute instead of removeIfPresent / replace

src/main/java/io/vertx/core/http/impl/WrappedHttpClient.java
tsegismont: ok

src/main/java/io/vertx/core/http/impl/SharedHttpClient.java
tsegismont: will do

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
tsegismont: what's the issue? it would depend on a context only when shared

src/main/java/io/vertx/core/http/impl/SharedHttpClient.java
tsegismont: :+1:

tsegismont: Looking again, I did not use compute because we need a way to determine when to actually close the underlying client.

Also, I think using remove/replace is not too critical in this case, considering the low concurrency and frequency of invocations.

src/main/java/io/vertx/core/http/impl/SharedHttpClient.java
tsegismont: Actually HttpClientImpl does not override finalize. Only NetClientImpl does. But it looks ok to me given this is where the resources cleanup needs to happen. Or am I missing something?

vietj:
a client is not attached to a given context.

the event-loop it will use depends on the usage and not when the client was
created.

vietj:
HttpClientImpl piggies back from the usage of NetClient.

if you look at the CloseFuture of NetClient, you will spot that the
HttpClientImpl add a listener to cleanup pools when the NetClient is closed
after it has been finalized.

In this case it is different: we want that a GC-ed client wrapped is closed
so the share count is decreased.

tsegismont:
>a client is not attached to a given context.

This is not completely true because a timer is scheduled on the context that creates the client.
Is it fine to use a context after the corresponding verticle has been undeployed?

>HttpClientImpl piggies back from the usage of NetClient.
> ...

Thanks for the details, I see what you mean now.

vietj:
On Tue, Aug 24, 2021 at 4:14 PM Thomas Segismont ***@***.***>
wrote:

> a client is not attached to a given context.
>
> This is not completely true because a timer is scheduled on the context
> that creates the client.
>
it can be any context, the context / event-loop does not matter. I want to
avoid relating a client to a context.

> Is it fine to use a context after the corresponding verticle has been
> undeployed?
>

what do you mean?

tsegismont:
>what do you mean?

Suppose you have an HttpClient instance. Then you deploy a verticle, providing the client in the constructor.
You make a request with the client in the verticle, so a connection is created and bound to the context of the verticle (via the pool slot).
After the response is received, you undeploy the verticle. If keep alive is on and the connection is used, the requests made by other parts of the code are handled on the undeployed verticle context.

Perhaps this is not a problem, I'm just wondering. Although the context keeps a reference to the deployment (and hence the verticle instances)

tsegismont:
@vietj there's an interesting [failure](https://github.com/eclipse-vertx/vert.x/pull/4079/checks?check_run_id=3486944490) here. It did not show up before I reverted the changes to `HttpClientImpl`.

After the verticle is undeployed, the keepalive timeout is no longer enforced. I think this happens because timer tasks created on a deployment context are canceled when deployment is closed.

The issue is fixed by the last commit.

zero88:
@vietj @tsegismont
what happens if I create 1 `HTTPClient` after starting `Vert.x` (might be in root verticle), then deploy many verticles later and other verticles reuse this client without create new one (this is basic case)?
That is my current implementation(v4). Also, my holder is indexed by `HTTPClientOptions hash` instead of `name`. And I don't see different (or not sure) about fork context per verticle will affect to `HTTPClient`

In previous version (v3.6), I did it quite similar your solution but get some troubles to increase/decrease count when inject close handler per HTTPClient.

One thing I have not yet cleared about `HTTPClient` is: (based on my knowledge `HTTPClient` is a holder that keeps a list of `NetClient` connection per host, a.k.a `map`)
- net pool will shared among hosts or allocate per host?
- http client pool will closed after all Net connections is inactive, right?

Thanks

tsegismont:
>what happens if I create 1 HTTPClient after starting Vert.x (might be in root verticle), then deploy many verticles later and other verticles reuse this client without create new one (this is basic case)?

It should work.

>net pool will be shared among hosts or be allocated per host?

Per-host, nothing new here.

>http client pool will closed after all Net connections is inactive (keep alive timeout), right?

The HTTP client will not be closed when all connections have been closed due to keep-alive timeout.
It is only closed automatically when all verticles which have a reference to it are undeployed.

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -128,6 +128,7 @@
   private final boolean keepAlive;
   private final boolean pipelining;
   private final CloseFuture closeFuture;
+  private final ContextInternal timerContext;
```
vietj, line 131: we don't need to keep a reference on it as the timer task can simply call vertx.setTimer that will schedule the call on the same context

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -138,6 +139,8 @@ public HttpClientImpl(VertxInternal vertx, HttpClientOptions options, CloseFutur
     this.metrics = vertx.metricsSPI() != null ? vertx.metricsSPI().createHttpClientMetrics(options) : null;
     this.options = new HttpClientOptions(options);
     this.closeFuture = closeFuture;
+    ContextInternal ctx = vertx.getOrCreateContext();
```
vietj, line 142: no need to get or create context, simply call `vertx.createEventLoopContext()`

tsegismont: done

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
tsegismont: done

vietj:
Merged with https://github.com/eclipse-vertx/vert.x/pull/4172

PR #4083, opened 2021-08-31
Adding netty's HexDump option to Vertx ClientOptions https://github.com/netty/netty/pull/11420

Unev:
Motivation:

Able/Disable the hex dump of messages in Netty's LoggingHandler.

By default, in Vertx, when we enable the logging enable, netty dumps the data in Hex format along with other log level messages.

Dumping of the data in Hex format may not always be needed, so added a new constructor for this here
https://github.com/netty/netty/pull/11420

In this commit, just added the enable HexDump in all the config locations with default value = true, will try to use the new constructor in the next commit.

Conformance:


src/main/java/io/vertx/core/net/NetworkOptions.java
```diff
@@ -68,6 +74,7 @@ public NetworkOptions() {
     reuseAddress = DEFAULT_REUSE_ADDRESS;
     trafficClass = DEFAULT_TRAFFIC_CLASS;
     logActivity = DEFAULT_LOG_ENABLED;
+    hexDumpEnabled = DEFAULT_HEXDUMP_ENABLED
```
Unev, line 77: `hexDumpEnabled = DEFAULT_HEXDUMP_ENABLED;`

Unev:
```suggestion
    hexDumpEnabled = DEFAULT_HEXDUMP_ENABLED;
```

vietj:
it is not clear how this relates to the current logging option

Unev:
Current logging Option, if enabled, prints all the log messages along with data, which is being transformed,  in hexadecimal format.

And let's say If I want to have all the logging messages but not the hex dumping of data, then it is not possible currently.

So introducing one more option to have the hexadump enabled or not. Defaults to enable.

vietj:
ok I see more clearly now thanks

Unev:
Hi @vietj , could you please havelook at this again?

This is needed where we want to log the network activity without dumping the message that is being transferred.
thanks,

vietj:
This PR is not complete, it only add an option but this option is never used.

Unev:
Hi @vietj, 
Added use cases for the above optional field.

Thanks

Unev:
Hi @vietj , can you pls have a look at this.

Unev:
Hi @vietj , can you please have look at this.

Thanks,

vietj:
so looking at this, I believe it is better to replace the `hexDumpEnabled` boolean by the actual enum provided in Netty and name it _activityLogFormat_ with the default being `SIMPLE`

vietj:
ping @Unev

Unev:
@vietj, Ack. Will do the required changes.

Thanks.

vietj:
thanks for the contribution

Unev:
Thank you.

PR #4088, opened 2021-09-11
Update Http1xServerConnection.java

TFyre:
Ensure `request.metric` is set for all requests regardless if response is in progress and using small buffers (HTTP Keep-Alive)

Motivation:

Fixes a NPE encountered when updating metrics due to the `request.metric` object not being set in `handleMessage` when a response is pending.
Conditions to trigger this NPE :
* return a response that is larger than the default quarkus.resteasy.vertx.response-buffer-size
* concurrent incomming requests using KeepAlive

Using Quarkus 2.2.2.FINAL and Vert.x 4.1.3

```
2021-09-11 13:07:17,071 ERROR [io.ver.ext.web.RoutingContext] (vert.x-eventloop-thread-3) Unhandled exception in router: java.lang.NullPointerException
        at io.quarkus.micrometer.runtime.binder.vertx.VertxHttpServerMetrics.requestRouted(VertxHttpServerMetrics.java:80)
        at io.quarkus.micrometer.runtime.binder.vertx.VertxHttpServerMetrics.requestRouted(VertxHttpServerMetrics.java:28)
        at io.vertx.core.http.impl.Http1xServerRequest.routed(Http1xServerRequest.java:658)
        at io.quarkus.vertx.http.runtime.AbstractRequestWrapper.routed(AbstractRequestWrapper.java:272)
        at io.vertx.ext.web.impl.HttpServerRequestWrapper.routed(HttpServerRequestWrapper.java:395)
        at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:142)
        at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
        at io.vertx.ext.web.impl.RoutingContextImpl.doFail(RoutingContextImpl.java:591)
        at io.vertx.ext.web.impl.RoutingContextImpl.fail(RoutingContextImpl.java:184)
        at io.vertx.ext.web.impl.RoutingContextImpl.fail(RoutingContextImpl.java:173)
        at io.vertx.ext.web.impl.RoutingContextImplBase.handleInHandlerRuntimeFailure(RoutingContextImplBase.java:189)
        at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:156)
        at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
        at io.vertx.ext.web.impl.RouterImpl.handle(RouterImpl.java:55)
[...]

PR #4091, opened 2021-09-13
HTTP server request events wrong metrics events for pipelined requests

vietj:


src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
```diff
@@ -558,6 +566,25 @@ private void onEnd() {
     }
   }
 
+  private void reportRequestComplete(Http1xServerRequest request) {
```
TFyre, line 569: No need for parameter anymore

src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
```diff
@@ -558,6 +566,25 @@ private void onEnd() {
...
+    HttpServerMetrics metrics = conn.metrics;
+    if (metrics != null) {
+      metrics.requestEnd(request.metric(), request, request.bytesRead());
+      conn.flushBytesRead();
+    }
+  }
+
+  private void reportRequestBegin(Http1xServerRequest request) {
```
TFyre, line 577: No need for parameter anymore

vietj:
right thanks!

src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
vietj: thanks I updated

TFyre:
The only other question:
https://github.com/eclipse-vertx/vert.x/pull/4088#issuecomment-917979726
Should it not calculate the total time for the request (including time spent waiting in the queue)?

vietj:
well in practice pipelining should not be used, so most of the time it
will be correct

vietj:
that being said we could indeed initialize it maybe all the time when
the request has been received ?

in that case the metrics implementation will see interleaved requests
begin/end like:

begin request 1
begin request 2
end request 1
begin request 3
end request 2
....

vietj:
also that would make code inside the application that calls the `routed` like the metrics incorrect, because the request end event would be emitted before the application calls the `routed` method.

TFyre:
This should be fine, since the metric is tied to the request... and a new request is being created for every message in `handleMessage`

Issue #3654, opened 2020-11-18
File resolver does partial cached class path directory listing


vietj:
The implementation takes care of unpacking the full content of an file's parent directory to ensure that listing this directory will return all the entries in the class path.

This works only for a single level and therefore the issue happens with 2 directories, e.g resolving the file `/foo/bar/juu` leads to listing `/foo/bar` to be correct but `/foo` will only contain `/foo/bar`.

There are several possible strategies to solve this:

- simply unpack everything recursively: this is simple but can lead to performance issues, i.e it might introduce some spikes when cache is populated
- add some state to record whether the listing is full or partial, this could be saved on disk (a dot file)
- we only cache file and not directories, so directory listing would trigger a class path resolution and content listing, note this impacts the `FileResolver` API that always returns a `File` for a directory or a file.

Reproducer: modify `FileResolverTestBase`

```java
  @Test
  public void testReadFileInDirThenReadDir() {
    Buffer buff = vertx.fileSystem().readFileBlocking("webroot/subdir/subfile.html");
    assertEquals(buff.toString(), "<html><body>subfile</body></html>");
    Function<List<String>, Set<String>> fx = l -> l.stream().map(path -> {
      int idx = path.lastIndexOf(File.separator);
      return idx == -1 ? path : path.substring(idx + 1);
    }).collect(Collectors.toSet());
    Set<String> names = fx.apply(vertx.fileSystem().readDirBlocking("webroot/subdir"));
    assertEquals(names, new HashSet<>(Arrays.asList("subdir2", "subfile.html")));
    names = fx.apply(vertx.fileSystem().readDirBlocking("webroot"));
    assertEquals(names, new HashSet<>(Arrays.asList("subdir", "somefile.html", "someotherfile.html", "this+that")));
  }
```

pmlopes:
The Linked PR will fix this issue

PR #4099, opened 2021-09-20
Cache related improvements to resolver

pmlopes:

Motivation:

This PR attempts to address some common pattern that isn't addressed by the resolver and is requiring workarounds in several modules, for example:

* vertx-json-schema
* vertx-web-openapi
* reactiverse/es4x
* ...

Consider the following jar file contents:

```
my-classpath-jar-file.jar
├── a/
│   ├── a.txt
├── b/
│   ├── b.txt
```

When a user requests a file resolution for `a/a.txt`:

```java
vertxInternal.resolveFile("a/a.txt")
```

The resolver will extract all files under `a/` and return the resolved file `/tmp/vertx-cache-UUID/a/a.txt`.

What usually happens is that in modules like the ones mentioned above, files may have relationships, so we need to resolve other files relatively to their location. For example `a.txt` may refer to `../b/b.txt`. which means we will call the resolve method again with:

```java
vertxInternal.resolveFile("/tmp/vertx-cache-UUID/a/../b/b.txt")
```

Since the resolver didn't extract the full contents of the jar, in the current implementation it will return the file `/tmp/vertx-cache-UUID/a/../b/b.txt` which doesn't exist.

What this PR is doing is adding an extra step.

If a file to be resolved, doesn't exist in the `cacheDir` it can be for 2 reasons:

1. the file really doesn't exist
2. the file is a resource that hasn't been resolved yet

This PR will then attempt to check if the requested path is insude the cache dir, and if it is, rebase it to a relative path and continue, so for the `b.txt` file, this happens:

```
vertxInternal.resolveFile("/tmp/vertx-cache-UUID/a/../b/b.txt")
// the file doesn't exists
// it falls under the cache dir
// it get rewritten to: `/a/../b/b.txt`
// is exists as a resource
// the resource gets extracted to the cache
// we return: "/tmp/vertx-cache-UUID/b/b.txt"
```

This fix allows us to avoid doing double resolutions on other modules and rely on undocumented behavior:

```
vertxInternal.resolveFile("") // <-- returns the cache dir (but we never defined "" as a special value) if there is no `vertx.cwd` system property
```

And properly resolve the cache entries with the right value.


pmlopes:
This PR is now extended with a few extra features:

* Fixes #3654 
* Speed up on known missing classpath resources.

For every classpath resource that is missing, we keep track of the resource name in a bloom filter, if further resolutions use that name again, we avoid going over the class path search and extract

pmlopes:
@vietj I've build this branch locally and tested that it will ensure `vertx-json-schema`, `vertx-web-openapi` and `es4x` will be able to resolve files without having to guess cache locations (which was a process that didn't work all the time as it relied on some undocumented behavior of the file resolver).

pmlopes:
Removed the use of the inverted bloom filter. Although the filter can help speeding up lookups for known missing resources, it will not prevent bad actors from looking up known missing resources by adding a permutation of the resource name, voiding the whole speedup expectation.

Issue #3916, opened 2021-05-10
Wrong implementation of how to identify cookies


fbuetler:
### Version

Version: 4.0.3

### Context

[RFC 6265 HTTP State Management Mechanism, 5.3 Storage Model, Step 11](https://datatracker.ietf.org/doc/html/rfc6265#section-5.3) defines the tuple (name, domain, path) as the cookie identifier. 
However, the implementation of [Http1xServerResponse](https://github.com/eclipse-vertx/vert.x/blob/4.0.3/src/main/java/io/vertx/core/http/impl/Http1xServerResponse.java#L787) and [Http2ServerResponse](https://github.com/eclipse-vertx/vert.x/blob/4.0.3/src/main/java/io/vertx/core/http/impl/Http2ServerResponse.java#L716) use only the cookie name as identifier.

### Steps to reproduce

1. Set two cookies with the same name but different path in a response header with
```java
public void handle(RoutingContext ctx) {
  Cookie cookie = Cookie.cookie("SOME_NAME", "foo").setPath("/");
  Cookie sameCookieWithDifferentPath = Cookie.cookie("SOME_NAME", "bar").setPath("/somePath");

  ctx.response().addCookie(cookie);
  ctx.response().addCookie(sameCookieWithDifferentPath);
  ctx.response().end();
}
```
2. Expect 2 cookies on the client, but only get 1.

vietj:
assigned to you @pmlopes

pmlopes:
This one is tricky to solve. Given that the uniqueness should be the tuple (name + domain + path) and the public API exposes just a simple `Map`, we either need to break and change it to be a `MultiMap` or define a well-known format for looking up cookies. 

Say that we define that the key in the map is the string `NAME/PATH` which is trivial to split as the path begins on the first `/`, it makes the lookup and removal more expensive as we need to consider using a `TreeMap` and perform `subMap(key, key + Character.MAX_VALUE)` to create a view filtering the prefix, and then we still need to decide how to handle more than 1 entry.

fbuetler:
I would rather not do some string magic to create a key. Simply, because these values may already contains the meta character we would use.

I would rather:
* Create a `Tuple` object, wrapping name, domain and path, that we would use as a key i.e. `Map<Tuple, ServerCookie>`.  `Tuple` would be a new class implementing the `Comparable` interface, overwriting the `equals` method and maybe the `hashCode` method.
[...]

PR #4105, opened 2021-09-22
Implementing the changes for #3916

pmlopes:

Motivation:

Address the issue mentioned in the title.


pmlopes:
Updated the original proposal from `List` to `Set`.

The cookie jar should contain unique cookies, using a list would not solve the problem. Given that a cookie unique identifier according to the spec is the tuple `<name, domain, path>` having the interface implement the `Comparable` interface too allows us to use a `TreeSet` to hold the cookies, which will ensure that addition and uniqueness according to the rules are enforced.

This also works for the case where users work with the set directly instead of the helper methods.

pmlopes:
@fbuetler, @vietj could you have a review to this PR? It follows the full discussion from the original issue and only deprecates 1 method as recommended by @vietj.

pmlopes:
There is still one small usability issue with this PR that I am not fully happy with. Users can now properly work with cookies and locate the right cookie in the cookie set.

The problem is that if a user doesn't use the helpers to add a cookie, the final result may feel strange. For example:

```java
// my cookies set has [ foo=bar ]
cookies().add(Cookie.cookie("foo", "barista"));
```

Will not replace the cookie because given that both cookies have the same key the comparator returns 0 according to the RFC, the set implementation will not replace the element.

I'm now thinking if it would be better going back with the original idea of having a `List` but it's not a real list but a wrapper that controls the insertions to ensure that there are no duplicates, and on duplicate, replace the underlying entry.

WDYT?

pmlopes:
I've addressed the issue mentioned above by creating a custom container. This way we isolate the cookie jar logic into a single class (instead of having it spread across several classes) and the addition of cookies are consistent either by using the helpers or the container itself.

src/main/java/io/vertx/core/http/impl/CookieJar.java
```diff
@@ -0,0 +1,294 @@
...
+/**
+ * Vert.x cookie jar implementation. A Cookie Jar is a simple wrapped list that behaves like a Set with one difference.
+ * On adding an existing item, the underlying list is updated.
+ *
+ * This call also provides extra semantics for handling invalidation of cookies, which aren't plain removals as well as
+ * quick search utilities to find all matching cookies by name or by unique key.
+ *
+ * @author <a href="http://pmlopes@gmail.com">Paulo Lopes</a>
+ */
+public class CookieJar implements Set<ServerCookie> {
```
vietj, line 26: why not extend AbstractSet ?

src/main/java/io/vertx/core/http/HttpServerRequest.java
```diff
@@ -478,24 +477,62 @@ default StreamPriority streamPriority() {
...
+  default Map<String, Cookie> cookieMap() {
+    return cookies()
+      .stream()
+      .collect(Collectors.toMap(Cookie::getName, cookie -> cookie));
+  }
+
+  /**
+   * Returns a set of parsed cookies that match the given name, or an empty set. Several cookies may share the same name
+   * but have different keys. A cookie is unique by its {@code <name, domain, path>} tuple.
+   *
```
vietj, line 526: we should document whether the returned Set can be modified or not and if modifications will be reflected against the HTTP resposne

src/main/java/io/vertx/core/http/HttpServerRequest.java
```diff
@@ -478,24 +477,62 @@ default StreamPriority streamPriority() {
...
+   * @param name the name to be matches
+   * @return the matching cookies or {@code null}
+   */
+  @Nullable Set<Cookie> cookies(String name);
+
+  /**
+   * @return a list with all cookies in the cookie jar.
```
vietj, line 533: it says returns a List but it's actuallyh returning a Set

vietj:
should we have a method that removes cookies with a filter ? i.e removeCookies(Function<Cookie, Boolean>)

src/main/java/io/vertx/core/http/impl/CookieJar.java
pmlopes: Good point

src/main/java/io/vertx/core/http/HttpServerRequest.java
pmlopes: Documentation has been updated to define the contract and tests were added to ensure it's correct.

pmlopes:
> should we have a method that removes cookies with a filter ? i.e removeCookies(Function<Cookie, Boolean>)

I'm not sure about that one. We already have:

```java
Cookie removeCookie(String name)
Cookie removeCookie(String name, String domain, String path)
Set<Cookie> removeCookies(String name)

Cookie removeCookie(String name, boolean invalidate)
Cookie removeCookie(String name, String domain, String path, boolean invalidate)
Set<Cookie> removeCookies(String name, boolean invalidate)
```

I think adding that one it would not bring anything extra, just 2 ways of achieving the same. If a user really wants to go on a remove with very specific requirements, then it is still possible with the set API:

```java
Iterator<Cookie> it = request.response().cookies().iterator()

while(it.next()) {
  Cookie cookie = it.next();
  boolean outcome = // Here the logic for define what should happen
  if (outcome) {
    it.remove();
  }
}
```

But If you see it as a added value, I can add it as it is trivial!

PR #4111, opened 2021-09-28
Small optimization for local `BufferMessageCodec`

kristian:
If the underlying Netty `ByteBuf` is read-only, there is no need to `.copy()` the buffer when it is being locally sent via the event bus.

Motivation:

A very simple check for whether the underlying Netty `ByteBuf` is read-only can significantly improve the performance of `Buffer`s being exchanged via the (local) event bus. Especially for large buffers the `.copy()` operation is expensive and if the underlying buffer is anyways read-only, there is no reason for copying the buffer in the first place.


kristian:
I cannot seem to get the ECA voter green... I have a pretty old Eclipse account, associated to the same e-mail address that I use on GitHub (kris@***.lc), I *did* sign the ECA, as well as I am working for a Eclipse Member, for which my companies member committer agreement covers the ECA. Thus I really cannot understand why the ECA check fails. See screenshot attached.

SNAG--20210928-00

vietj:
can you elaborate when this happens and add a test for this ?

kristian:
@vietj sure, happy to do so! I also did restructure the commit a little, to make it more useful. Let me first elaborate on the reason: We are currently optimizing our event bus transfers and as we are transferring large byte buffers, that returning them is the last activity in the sending verticle, copying them during the local event bus transfer is a overhead that we are trying to overcome. Instead of creating an own `ImmutableBuffer` class (like we attempted [here](https://github.com/SAP/neonbee/pull/69/commits/06b3f9d8880f9ef20a01038d3f2484bb6141b733) and found out that `BufferImpl` doesn't respect the `Buffer` interface [see my new first commit]) I found that making the underlying `ByteBuf` read-only instead, would be a very convenient method of making the whole `Buffer` read-only. This would allow to call `Buffer.buffer(b.getByteBuf().asReadOnly())` to optimize local event bus transfers. However in the amended commit I have added the tests you have asked for and actually extended the `Buffer` interface with a `as/isReadOnly` similar to `ByteBuf`. This makes it more useful, even to others wanting to optimize, because it will be clear calling `eb.send(b.asReadOnly())` is more efficient, others might also want to use this method instead, depending on their use-case. Hope that is fine, please let me know if any further changes are required.

vietj:
thanks it makes sense if you are wrapping Netty buffers that are read-only

vietj:
I do agree we should optimise for read-only buffers however there is an issue with the `asReadOnly` method that returns a read-only version of this buffer:

1/ the original buffer can still be modified so it is an hint for the codec that the buffer should not be copied and can be shared and it can be dangerous (a user might modify the original buffer).

2/ Netty 5 new buffer API changed the semantic of read-only, `asReadOnly` is replaced by `makeReadOnly` so that would make migration to Vert.x 5 harder and force to deprecate or remove the `asReadOnly` method

vietj:
we actually implemented the read-only optimization in the copy() method of Buffer which solves the same issue but in the buffer : https://github.com/eclipse-vertx/vert.x/issues/4136

PR #4113, opened 2021-10-01
Allow to register default codecs that can handle sub-types

kristian:
Implements #4087. Imagine having multiple "Ducks" that you want to send via the wire:

- `GreatDuck`
- `SmallDuck`
- `DuckyMcDuck`

All ducks implement the `Duck` interface and a `toJson` method.

This change allows to implement one `DuckMessageCodec` which can handle any duck by calling the `toJson` method on the object.

PR #4117, opened 2021-10-04
Message codec inheritance

vietj:


vietj:
Can you have a look @pmlopes ?

vietj:
Please have a look @kristian

pmlopes (commented):
I think we should also have a minimal example, so users could see how to implement a generic codec. For example, just delegate to `Object{In|Out}putStream` yet stressing that this is not the safest or most performance recommended way to implement.

src/main/asciidoc/override/eventbus.adoc
```diff
@@ -18,6 +18,14 @@ you don't have to specify the codec on each send in the delivery options:
...
 ----
 
+Default message codec performs a lookup based on the class of the body.
+
+This lookup happens after the system codecs are examined, i.e a default codec
+for `io.vertx.core.buffer.Buffer` will not be used.
+
+Since Vert.x 4.2, the lookup resolves on the super class of the body class when the body class does not have a codec, recursively yet
+excluding `java.lang.Object`.
+
```
pmlopes, line 28: Should also mention that generics are not taken into consideration when looking for the codec.

src/main/java/io/vertx/core/eventbus/impl/CodecManager.java
```diff
@@ -46,103 +46,111 @@
...
-      codec = defaultCodecMap.get(body.getClass());
-      if (codec == null) {
-        throw new IllegalArgumentException("No message codec for type: " + body.getClass());
+      if (body == null) {
+        return NULL_MESSAGE_CODEC;
+      } else {
+        Class<?> clazz = body.getClass();
+        if (clazz == String.class) {
+          return STRING_MESSAGE_CODEC;
+        } else if (Buffer.class.isAssignableFrom(clazz)) {
```
kristian, line 73: Would it be better to check for BufferImpl here? This is an important difference to understand, because checking `isAssignableFrom` means that every implementation of `Buffer` is forced to used the build-in `BufferMessageCodec`. Different implementations of `Buffer` will not work. See also related [1].

[1] https://github.com/eclipse-vertx/vert.x/pull/4111/commits/acb42fafba589c42968bf2136395358195c2dfa8

src/main/java/io/vertx/core/eventbus/impl/CodecManager.java
```diff
@@ -46,103 +46,111 @@
...
+          if (codec == null) {
+            throw new IllegalArgumentException("No message codec for type: " + clazz);
+          }
+          return codec;
+        }
+      }
+    }
+  }
+
+  private MessageCodec resolveCodec(Class<?> clazz) {
```
kristian, line 110: `resolveCodec` will resolve any available default codec, even if it doesn't match the source objects class. As suggested previously, what do you think of adding a:

```java
default boolean supportsGenericTypes {
  return false
}
```

To the `MessageCodec` interface and then changing the implementation of this to:

private MessageCodec resolveCodec(Class<?> clazz) {
 
 
`resolveCodec` will resolve any available default codec, even if it doesn't match the source objects class. As suggested previously, what do you think of adding a:

```java
default boolean supportsGenericTypes {
  return false
}
```

To the `MessageCodec` interface and then changing the implementation of this:

```java
private MessageCodec resolveCodec(Class<?> resolve, Class<?> clazz) {
    MessageCodec codec;
    codec = codecsByClass.get(resolve);
    if (codec != null && !resolve.equals(clazz) && !codec.supportsGenericTypes()) {
       codec = null; // this codec does not support converting generic types
    }
    if (codec == null) {
      Class<?> parent = clazz.getSuperclass();
      if (parent != Object.class && parent != null) {
        codec = resolveCodec(parent, clazz);
      }
    }
    return codec;
  }
```

That'd make it a non-breaking change for one and also make it more safe, because codecs will have to specifically announce if they want to be supporting generic types.

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -193,13 +193,14 @@ public EventBus unregisterCodec(String name) {
 
   @Override
   public <T> EventBus registerDefaultCodec(Class<T> clazz, MessageCodec<T, ?> codec) {
```
kristian, line 195: Do you want to allow <? super T, ?> here in this PR? See #4113

src/main/asciidoc/override/eventbus.adoc
```diff
@@ -24,7 +24,8 @@ This lookup happens after the system codecs are examined, i.e a default codec
 for `io.vertx.core.buffer.Buffer` will not be used.
```
kristian, line 24: I think this is maybe a bad example, because `Buffer` is a interface, not a class. Changing the example and maybe even adding a own comment about that interfaces are not considered when trying to find a codec would be advisable. What do you think?

pmlopes: Maybe:

> This lookup happens after the system codecs (these include the java primitive types, `Buffer`, `JsonObject` and `JsonArray`) are examined, i.e a default codec for `io.vertx.core.json.JsonObject` will not be used.

src/main/java/io/vertx/core/eventbus/impl/CodecManager.java
vietj: if we would not check Buffer, then any buffer implementation that is not a vertx heap buffer would be rejected

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
kristian: Hey @vietj, it is the other way around isn‘t it? Registering a class to be handled by a:

```java
MessageCodec<? super T, ?>
```

codec means: „Class XYZ is handled by a codec that can handle everything UP-TO that class“, so it is an upper-bound, no lowee-bound. The `<? super T, ?>` don‘t make any assumption about what it is converted into, it just makes it clear that a codec maybe can handle to convert *from* more than just the specific class (it can convert the specific class and all it‘s decendents).

You still only check for the specific class T, in finding the right codec, meaning there is still a 1:1 mapping between class and codec only that you don‘t need to have individual codecs build for types that have a common way of getting serialized. If the codec then spits out the type, or sub type in desirilization is part of the second generic argument which is ? even at the moment. So you cannot be sure right now even for default codecs.

src/main/java/io/vertx/core/eventbus/impl/CodecManager.java
kristian: Well yes, that would again speak for the `accept` method you suggested in your comment below, because due to the checks of the built-in codecs taking precedence over the custom codecs, your statements on the other hand means: every Buffer implementstion that is not a Vert.x heap buffer is forced to be converted into one, if it is sent via the event bus. I am fine with that, however that make custom implementations of `Buffer` very limited in their usefullness.

vietj:
we discussed whether this is the right way to implement this behaviour and we believe that we can come with something more flexible in Vert.x 4.3 based on the fact that user code could set the codec name to use based on the object type.

We are closing Vert.x 4.2 development and focusing on bug fixes before the GA expected within a couple of weeks. We'll this PR open nevertheless. We will work on an RFC for this and share it with the community in the coming weeks.

tsegismont:
As commented above, a more flexible solution is described in #4365

PR #4129, opened 2021-10-11
Make sure exception stack trace is logged

tsegismont:
If the Throwable is passed to log() method, its exception message only is added to the logs.


tsegismont:
@vietj I've just removed the milestone on this PR given the related issue has it already.

PR #4148, opened 2021-10-27
Project Loom: Update thread factory code to accept virtual threads

Jotschi:
This PR contains a collection of patches which I used to experiment with [Project Loom](https://openjdk.java.net/projects/loom/) support in Vert.x.

The patches prepare Vert.x so that it can accept a custom thread factory which can provide virtual threads.
I did use the SPI mechanism for `VertxThreadFactory` to provide a custom JDK 18 based factory.
The code for this factory and corresponding tests can be found in my [vertx-loom](https://github.com/Jotschi/vertx-loom) project.

# Changes

* I had to create a wrapper since Virtual Threads can't be extended any more. Thus I needed to create the `VertxThread` interface.
* In some places it was needed to identify virtual threads. In order to keep the codebase JRE 8 friendly I created the `JavaCompatUtil` utility which helps to identify virtual threads.

# Caveat

* `VirtualVertxThreadImpl` objects can no longer be fetched via `Thread.currentThread()` when virtual threads are used. This poses a problem when handling dispatch calls (see `AbstractContext`). This issue most likely also renders the `BlockedThreadChecker` inoperable.
* Using Virtual Threads may also affect classloader handling / hierarchy. I have not investigated this at the moment.
* Large thread pool sizes for the eventloop (400+) can cause Netty to stall during request processing due to pinned carrier threads. The usage of `Select#select()` requires synchronization and this makes use of a monitor lock. [Details on loom-dev mailinglist](https://mail.openjdk.java.net/pipermail/loom-dev/2021-October/003000.html) - A warning on this has been added to the VertxVirtualThreadFactory.
* `ContextImpl#executeIsolated` is still lacking a virtual thread friendly implementation. Maybe the given task could just be executed in a dedicated virtual thread.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -367,6 +367,7 @@
         </configuration>
         <executions>
           <execution>
+            <?m2e ignore ?>
             <id>generate-sources</id>
             <goals>
               <goal>process</goal>
--- a/src/main/java/io/vertx/core/Context.java
+++ b/src/main/java/io/vertx/core/Context.java
@@ -15,7 +15,7 @@ import io.vertx.codegen.annotations.Fluent;
 import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.Nullable;
 import io.vertx.codegen.annotations.VertxGen;
-import io.vertx.core.impl.VertxThread;
+import io.vertx.core.VertxThread;
 import io.vertx.core.json.JsonObject;
 
 import java.util.List;
--- a/src/main/java/io/vertx/core/VertxThread.java
+++ b/src/main/java/io/vertx/core/VertxThread.java
@@ -1,0 +1,32 @@
+package io.vertx.core;
+
+import io.vertx.core.impl.BlockedThreadChecker;
+import io.vertx.core.impl.ContextInternal;
+
+public interface VertxThread extends BlockedThreadChecker.Task {
+
+	/**
+	 * @return the current context of this thread, this method must be called from
+	 *         the current thread
+	 */
+	ContextInternal context();
+
+	boolean isWorker();
+
+	StackTraceElement[] getStackTrace();
+
+	void setDaemon(boolean b);
+
+	/**
+	 * Check whether the thread execution time can be tracked.
+	 * @return
+	 */
+	boolean isTrackable();
+
+	/**
+	 * Return the actual {@link Thread} that corresponds to the {@link VertxThread}.
+	 * @return
+	 */
+	Thread getThread();
+
+}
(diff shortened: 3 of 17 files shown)


Jotschi:
> the goal of this draft is to discuss and see what we can get so feel free to change JDK to make it work like you need to

Okay. I have reverted the multi-release changes. The build is now again passing. But I think it is best to focus on the wrapper idea for now.

Jotschi:
Just to document this:

I was able to reduce the needed changes to the current codebase by switching to a different approach.
The key idea is to keep a reference to the VertxThread object acessible via a threadlocal.

This can be done by injecting a small call to the started thread which stores the reference:
```java
    this.instance = Thread.ofVirtual().name(name).factory().newThread(() -> {
      VirtualVertxThreadImpl.setVertexThread(this);
      runnable.run(); 
    });
```

Neeed changes to core:

* InboundBuffer#checkThread
* ContextInternal#current
* AbstractContext#beginDispatch / AbstractContext#endDispatch

The Virtual Thread extends Thread but redirects methods to the internal virtual thread:

```java
public class VirtualVertxThreadImpl extends VertxThread {

  private Thread instance;
  
  private static final ThreadLocal<VirtualVertxThreadImpl> THREAD_REF = new ThreadLocal<>();

  public VirtualVertxThreadImpl(Runnable runnable, String name, boolean worker, long maxExecTime,
    TimeUnit maxExecTimeUnit) {
    super(runnable, name, worker, maxExecTime, maxExecTimeUnit);
    this.instance = Thread.ofVirtual().name(name).factory().newThread(() -> {
      VirtualVertxThreadImpl.setVertexThread(this);
      runnable.run(); 
    });
  }

  @Override
  public StackTraceElement[] getStackTrace() {
    return instance.getStackTrace();
  }

  @Override
  public void start() {
    this.instance.start();
  }

  @Override
  public void interrupt() {
    this.instance.interrupt();
  }

  @Override
  public boolean isInterrupted() {
    return this.instance.isInterrupted();
  }

  public Thread unwrap() {
    return instance;
  }
  
  public static VertxThread getVertexThread() {
    return THREAD_REF.get();
  }

  private static void setVertexThread(VirtualVertxThreadImpl virtualVertxThread) {
    THREAD_REF.set(virtualVertxThread);
  }
}
```

This is a bit smelly but needed since VirtualThreads can't be extended. I don't see any other way if large scale refactoring can't be done. (e.g. because of third party library thread support).

I tested this in the [vertx-loom/vertx-custom-thread-factory-alt](https://github.com/Jotschi/vertx-loom/compare/vertx-custom-thread-factory-alt) branch

PR #4162, opened 2021-11-09
Adding full details on what are the requirements for the SPI

pmlopes:

Motivation:

Users trying to implement a custom JsonCodec face application failures because the expectations of the SPI are not well documented. For example, nothing is stated over how to encode `JsonObject/JsonArray` or the RFCs vert.x implements.

This PR is mostly a documentation update explaining the expectations so users can write working codecs.

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -533,7 +533,7 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
    * @return the string encoding
    */
   public String encode() {
-    return Json.CODEC.toString(this, false);
+    return Json.CODEC.toString(this.list, false);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -533,7 +533,7 @@ public JsonArray clear() {
    * @return the string encoding
    */
   public String encode() {
-    return Json.CODEC.toString(this, false);
+    return Json.CODEC.toString(this.list, false);
```
vietj, line 536: this seems to be more than a detail, is it a bug ?

pmlopes: As I was debugging to figure out the behavior I noticed that we can reduce jackson work, as it internally doesn't need to do a 1st lookup on the registered classes for see that `JsonArray` needs some extra processing. Which is unwrapping the list.

So the behavior didn't change, the stacktraces are just shorter because we help jackson.

pmlopes: Same applies to `JsonObject`

PR #4167, opened 2021-11-13
fix "setTimer throws when delay < 1ms"

mtomfjc:
Motivation:

Conformance:


vietj:
that would change the current behavior

PR #4172, opened 2021-11-18
Tsegismont shared http client

### LINKED ISSUE #3361: Shared HttpClient support

### LINKED ISSUE #4175: HTTP client fixed event loop pool
HTTP client uses event loops provided by actual requests to assign connection event-loops. In some case it is desirable to have the client use its own event loop as a pool, like when it is shared between verticles or statically.

vietj:
fixes #3361
fixes #4175

### DIFF
--- a/src/main/asciidoc/http.adoc
+++ b/src/main/asciidoc/http.adoc
@@ -1397,6 +1397,36 @@ or close the client instance.
 Alternatively you can set idle timeout using {@link io.vertx.core.http.HttpClientOptions#setIdleTimeout(int)} - any
 connections not used within this timeout will be closed. Please note the idle timeout value is in seconds not milliseconds.
 
+=== HTTP client sharing
+
+You can share an HTTP client between multiple verticles or instances of the same verticle. Such client should be created outside
+of a verticle otherwise it will be closed when the verticle that created it is undeployed
+
+[source,$lang]
+----
+{@link examples.HTTPExamples#httpClientSharing1}
+----
+
+You can also create a shared HTTP client in each verticle:
+
+[source,$lang]
+----
+{@link examples.HTTPExamples#httpClientSharing2}
+----
+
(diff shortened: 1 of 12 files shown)


src/main/asciidoc/http.adoc
```diff
@@ -1397,6 +1397,36 @@ or close the client instance.
...
+
+You can also create a shared HTTP client in each verticle:
+
+[source,$lang]
+----
+{@link examples.HTTPExamples#httpClientSharing2}
+----
+
+The first time {@link io.vertx.core.Vertx#createSharedHttpClient(java.lang.String)} is called it will create the
+client. Subsequent calls will reuse this client and create a lease to this client. The client exists until all leases are closed.
```
tsegismont, line 1418:
```suggestion
client. Subsequent calls will reuse this client and create a lease to this client. The client is closed after all leases have been closed.
```

tsegismont (approved):
It looks good to me @vietj

PR #4179, opened 2021-11-25
Allow loading PKCS#11 keystores (HSM)

pmlopes:

Motivation:

PKCS#11 (Hardware Smart Cards) do not provide a keystore file path or buffer, instead they are expected to be loaded with a `null` inputstream.

Vert.x core on the other hand, assumed that `null` wasn't a valid value.

This PR attempts to fix this by allowing `null` to be used.


vietj:
can we have a test for this case ?

PR #4196, opened 2021-12-10
Upgrade log4j2 to 2.15.0 for CVE-2021-44228

zyclonite:
just to be on the safe side

@vietj shall we do that for all the other maintained branches as well?


pmlopes:
We already have this on master. We needs to backport to 4.0 and 3.9 still and same across remaining repos

zyclonite:
@pmlopes thank you. wasn't there when i filed this pr

PR #4199, opened 2021-12-13
Fix the NPE problem of FileSystemImpl

mashirofang:
file.listFiles may return null while using  windows system , this method may occurs  NPE 

test code :
```
        String file = System.getProperty("user.home") + "/Application Data";
        FileSystem fileSystem = Vertx.vertx().fileSystem();
        List<String> list = fileSystem.readDirBlocking(file);
```

System: windows10
JDK: 1.8.0_291

### DIFF
--- a/src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
+++ b/src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
@@ -1045,11 +1045,16 @@ public class FileSystemImpl implements FileSystem {
             } else {
               files = file.listFiles(fnFilter);
             }
-            List<String> ret = new ArrayList<>(files.length);
-            for (File f : files) {
-              ret.add(f.getCanonicalPath());
+            if (null == files || files.length == 0) {
+              return new ArrayList<>();
+            } else {
+              List<String> ret = new ArrayList<>(files.length);
+              for (File f : files) {
+                ret.add(f.getCanonicalPath());
+              }
+              return ret;
             }
-            return ret;
+
           }
         } catch (IOException e) {
           throw new FileSystemException(getFolderAccessErrorMessage("read", p), e);
--- a/src/test/java/io/vertx/core/file/FileSystemTest.java
+++ b/src/test/java/io/vertx/core/file/FileSystemTest.java
@@ -2336,4 +2336,14 @@ public class FileSystemTest extends VertxTestBase {
     file.close(onSuccess(v -> testComplete()));
     await();
   }
+  
+  @Test
+  public void testWindowsApplicationDataDirs() {
+    Assume.assumeTrue(Utils.isWindows());
+    String file = System.getProperty("user.home") + "/Application Data";
+    Assume.assumeTrue(new File(file).exists());
+    List<String> dirs = vertx.fileSystem().readDirBlocking(file);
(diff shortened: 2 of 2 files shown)


src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
```diff
@@ -1045,11 +1045,16 @@ public boolean accept(File dir, String name) {
             } else {
               files = file.listFiles(fnFilter);
             }
-            List<String> ret = new ArrayList<>(files.length);
-            for (File f : files) {
-              ret.add(f.getCanonicalPath());
+            if (null == files || files.length == 0) {
+              return new ArrayList<>();
```
pmlopes, line 1049: Why not `Collections.emptyList()` ?

src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
```diff
@@ -1045,11 +1045,16 @@ public boolean accept(File dir, String name) {
...
+            } else {
+              List<String> ret = new ArrayList<>(files.length);
```
pmlopes, line 1051: If we make the above immutable, this list should be too, for consistency

pmlopes (changes requested):
Small comments, open for discussion.

pmlopes:
It would be nice to have a test case that reproduces the NPE for regression testing.

src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
mashirofang: Because `Collections.emptyList()` is an immutable list . Creating a new instance would be compatible with previous version

mashirofang:
> It would be nice to have a test case that reproduces the NPE for regression testing.

I add an test case for reproducing the NPE of (windows) file system  just now

src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
pmlopes: I get the reasons, I am just thinking, shouldn't this list actually be immutable? For example, the underlying data is an array. We wouldn't be able to add entries to the array because its size is fixed. (We can mutate elements, though). So, my question was more, should we align the behavior of the underlying API with the `FileSystem` API?

WDYT @vietj ?

vietj: I agree we need to return a new array list here, perhaps we can optimize and instead return a list with a minimal initial capacity as it should not be updated

vietj: e.g `new ArrayList(0)` will not allocate an array

src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
vietj: we cannot really make this immutable otherwise it will be a breaking change

vietj:
can you rebase on lastest master soon so we can merge this for 4.2.2 ? @mashirofang

mashirofang:
I mis operated git while rebase ,  so I retry to fork & commit again;
 @vietj  #4202

PR #4244, opened 2022-01-13
Return failed future when executeBlocking task is rejected

### LINKED ISSUE #4209: `executeBlocking` throw exception instead of return a failed future after the `workerPool` closed, thus caused `RejectedExecutionException`
### Questions
When using `HazelcastClusterManager`, it would thrown `RejectedExecutionException` sometimes after closed the `vertx` instance:

``` log
2021-12-21 11:05:30,529 (:) [ERROR][io.vertx.core.impl.ContextImpl] Unhandled exception
java.util.concurrent.RejectedExecutionException: Task io.vertx.core.impl.ContextImpl$$Lambda$210/0x0000000800ece910@554cbcc6 rejected from java.util.concurrent.ThreadPoolExecutor@5b8b481f[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 8]
	at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2065) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:833) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1365) ~[?:?]
	at io.vertx.core.impl.ContextImpl.executeBlocking(ContextImpl.java:172) ~[vertx-core-4.2.2.jar:4.2.2]
	at io.vertx.core.impl.ContextImpl.executeBlocking(ContextImpl.java:137) [vertx-core-4.2.2.jar:4.2.2]
	at io.vertx.core.impl.EventLoopContext.executeBlocking(EventLoopContext.java:22) [vertx-core-4.2.2.jar:4.2.2]
	at io.vertx.core.Vertx.executeBlocking(Vertx.java:581) ~[vertx-core-4.2.2.jar:4.2.2]
	at io.vertx.spi.cluster.hazelcast.impl.Throttling.lambda$run$5(Throttling.java:100) ~[vertx-hazelcast-4.2.2.jar:4.2.2]
```

The `executeBlocking` method throw the exception at [Line#179](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/impl/ContextImpl.java#L179) when failed to execute the command ([because the `workerPool` closed earlier then the netty event loop](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/impl/VertxImpl.java#L809-L813)), is it works as intended? Maybe we should wrapped the exception with a failed future instead, thus the caller could handle the exception or just ignore it.

### Version
4.2.2

tsegismont:
Fixes #4209

When Vert.x is closed, there's a small timeframe which allows tasks to be submitted on worker pool but they will never get executed because the pool is closed.

Instead of throwing the runtime exception, we can return a failed future and give the user a chance to deal with it.

### DIFF
--- a/src/main/java/io/vertx/core/impl/ContextImpl.java
+++ b/src/main/java/io/vertx/core/impl/ContextImpl.java
@@ -176,7 +176,7 @@ abstract class ContextImpl extends AbstractContext {
       if (metrics != null) {
         metrics.rejected(queueMetric);
       }
-      throw e;
+      promise.fail(e);
     }
     return fut;
   }


vietj (changes requested):
It is missing a test (e.g closing a vertx instance before calling execute blocking should work)

I am not sure it works, because the promise future will not be able to execute the listeners because it will use the context itself.

So it should use a context less promise I think.

vietj:
also this should be recorded in the changelog I think because it changes the behavior, although the impact is not important, still worth to record this.

tsegismont:
Closing as my analysis was wrong.

Looking again, I believe the issue is Vert.x Hazelcast invokes `executeBlocking` after Vert.x has been closed.

PR #4247, opened 2022-01-18
[Issue-4246] Support vfs protocol in the default FileResolver implementation

### LINKED ISSUE #4246: Support vfs protocol in the default FileResolver implementation
#### Read me

When an embedded Vertx is used for a war deployment which reads resources in the bundle, and it is deployed to a WildFly server, it fails to read with error message: `Invalid url protocol: vfs`.

It is more like a feature than a bug from Vert.x point of view.

For more information, please refer to https://github.com/gaol/tests/tree/master/vertx/vfs-resolver for the reproducer.

#### Describe the feature

The default FileResolver implementation [supports several URL protocols](https://github.com/eclipse-vertx/vert.x/blob/4.2.3/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java#L235-L239) to resolve the file path, I would like to add the support of `vfs`.

`vfs` comes from [jboss-vfs](https://github.com/jbossas/jboss-vfs) project, which is used inside of WildFly server for managing the files of the deployed applications, any resource loaded by the class loader(ModuleClassLoader) is a vfs URL in the server.

Using a customized FileResolver implementation looks too much than just one line modification in the default implementation, so it is great to have it supported in Vert.x.

#### Use cases

When embedded Vertx is used in web application, and it is deployed to WildFly server, it faces this problem.
Like when [PropertyFileAuthentication](https://vertx.io/docs/vertx-auth-properties/java/) is used to read authentication properties from the bundle.

#### Contribution

I can send PR for this.

gaol:
Motivation:

Fixes #4246 

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
+++ b/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
@@ -237,6 +237,7 @@ public class FileResolverImpl implements FileResolver {
       case "bundleresource": // Equinox
       case "jrt": // java run-time (JEP 220)
       case "resource":  // substratevm (graal native image)
+      case "vfs":  // jboss-vfs
         return unpackFromBundleURL(url, isDir);
       default:
         throw new IllegalStateException("Invalid url protocol: " + prot);


src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
```diff
@@ -237,6 +237,7 @@ private File unpackUrlResource(URL url, String fileName, ClassLoader cl, boolean
       case "bundleresource": // Equinox
       case "jrt": // java run-time (JEP 220)
       case "resource":  // substratevm (graal native image)
+      case "vfs":  // jboss-vfs
```
vietj, line 240: this seems very specific and not something we need to support

vietj: well we can support this as best effort to some extend, i.e without extra specific code

vietj: I'm just worried if we do support this, then we'll have to add specific custom code to handle this

gaol: Agree to support this as best effort to an extend without extra code changes :)

If there are other vfs related issues after this, it should be done by a customized FileResolver.

PR #4256, opened 2022-01-24
Fix issue 4255 - HTTP header values verifier is laxer than what's advised in the RFCs

NilsRenaud:
According to :
- https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
- https://datatracker.ietf.org/doc/html/rfc5234#appendix-B.1
- https://datatracker.ietf.org/doc/html/rfc7540#section-10.3

For both HTTP/1.1 and HTTP/2.0 versions, bytes allowed for header value are : CR (0x0D), LF (0x0A), SP (0x20), HTAB (0x09) and ranges 0x21 to 0x7E and 0x80 to 0xFF (although the last range is considered as obsolete)

For more information, see the related issue : https://github.com/eclipse-vertx/vert.x/issues/4255

Signed-off-by: Nils Renaud <renaud.nils@gmail.com>

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpUtils.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpUtils.java
@@ -847,29 +847,38 @@ public final class HttpUtils {
     for (int i = 0;i < value.length();i++) {
       char c = value.charAt(i);
       switch (c) {
-        case 0x1c:
-        case 0x1d:
-        case 0x1e:
-        case 0x1f:
-        case 0x00:
-        case '\t':
-        case '\n':
-        case 0x0b:
-        case '\f':
-        case '\r':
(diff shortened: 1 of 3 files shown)


NilsRenaud:
Not sure why these tests have failed, it does not seem related to my changes 🤔

vietj:
Can you make the `Http1xTest#testHeaderNameStartsOrEndsWithControlChars` test generic and test the forbidden chars validated by value header validator ?

NilsRenaud:
> Can you make the `Http1xTest#testHeaderNameStartsOrEndsWithControlChars` test generic and test the forbidden chars validated by value header validator ?

This was not possible without enforcing the header name validation too.
So I did it, putting all the header name related changes in an other commit. Let me know what you think.

NilsRenaud:
I'm not sure why windows tests have failed. But that does not seem to be related to my changes :/

vietj:
we do have some flacky tests on windows

PR #4257, opened 2022-01-25
Introduce VertxBindException

geoand:
Motivation:

This is useful for Quarkus since it introspects
the BindException in order to figure out which port
caused the conflict.
Without this enhancement, Quarkus resorts to guessing (see [this](https://github.com/quarkusio/quarkus/blob/2.7.0.CR1/core/runtime/src/main/java/io/quarkus/runtime/ApplicationLifecycleManager.java#L158))
and this guess is [not correct](https://github.com/quarkusio/quarkus/issues/22967) when there are various
servers used (like an HTTP server and a GRPC server)

### DIFF
--- a/src/main/java/io/vertx/core/net/VertxBindException.java
+++ b/src/main/java/io/vertx/core/net/VertxBindException.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.net;
+
+import java.net.BindException;
+import java.net.SocketAddress;
+
+/**
+ * An exception that is meant to stand in for {@link BindException} and provide information
+ * about the target which caused the bind exception.
+ */
+public class VertxBindException extends BindException {
+
+  private final SocketAddress bindTarget;
+
+  public VertxBindException(BindException cause, SocketAddress bindTarget) {
+    this.bindTarget = bindTarget;
+    initCause(cause);
+  }
+
+  public SocketAddress getBindTarget() {
+    return bindTarget;
+  }
+}
--- a/src/main/java/io/vertx/core/net/impl/AsyncResolveConnectHelper.java
+++ b/src/main/java/io/vertx/core/net/impl/AsyncResolveConnectHelper.java
@@ -19,8 +19,10 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.net.VertxBindException;
 import io.vertx.core.net.SocketAddress;
 
+import java.net.BindException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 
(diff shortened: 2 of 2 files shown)


geoand:
@cescoffier @vietj if you agree with this change, I'd be glad to add a test (provided you let me know where I should add it) and take the PR out of draft

cescoffier:
Makes sense to me.

vietj:
it's not clear what it solves, can you elaborate ?

src/main/java/io/vertx/core/net/ContextualBindException.java
```diff
@@ -0,0 +1,33 @@
...
+package io.vertx.core.net;
+
+import java.net.BindException;
+import java.net.SocketAddress;
+
+/**
+ * An exception that is meant to stand in for {@link BindException} and provide information
+ * about the target which caused the bind exception.
+ */
+public class ContextualBindException extends RuntimeException {
```
vietj, line 21: which not extend BindException ?

vietj, line 21: `VertxBindException` than `ContextualBindException`

vietj:
isn't quarkus able to remember the socket address when it reaches a bind exception ?

src/main/java/io/vertx/core/net/ContextualBindException.java
cescoffier: BinException is not a RuntimeException, but in this case, we could extend BindException.

geoand: Because `BindException` does not provide a constructor that takes a cause

vietj:
I can't see code that tries to guess the port in https://github.com/quarkusio/quarkus/blob/2.7.0.CR1/core/runtime/src/main/java/io/quarkus/runtime/ApplicationLifecycleManager.java#L158

geoand:
> isn't quarkus able to remember the socket address when it reaches a bind exception ?

Not really, because at the point where the exception is caught, the bind exception could be thrown the HTTP server, the gRPC server or anything else starting using Vert.x. 

> https://github.com/quarkusio/quarkus/blob/2.7.0.CR1/core/runtime/src/main/java/io/quarkus/runtime/ApplicationLifecycleManager.java#L158

It's guessing in the sense that it assumes that only the HTTP or HTTPS ports can be the source of the BindException

src/main/java/io/vertx/core/net/ContextualBindException.java
vietj: you can still all `initCause(Throwable)` @geoand

geoand: Done

vietj: not using an exception extending BindException would be a breaking change

geoand: Huh, I never knew about `initCause`...

Thanks! Changed

vietj:
can't the address set on the application before listen is called and when a bind exception is met, then the socket address can be obtained from the application instead of looking at the configuration

geoand:
> can't the address be set on the application before listen is called and when a bind exception is met, then the socket address can be obtained from the application instead of looking at the configuration

Sure, but if the application is binding to multiple ports (HTTP and gRPC for example), then when it's time to catch the exception, you don't know which one caused the failure.

vietj:
oh right.

could the binding process could keep a reference to the server it is starting ?

that being said having this exception carrying extra info could be
interesting in the vanilla case too

geoand:
> could the binding process could keep a reference to the server it is starting ?

Yeah, we could likely do that as well in various places

geoand:
Actually, it looks like this change would not cover the gRPC case... 

So I think I'll close this (for now at least) and deal with the issue in Quarkus

vietj:
vertx gRPC uses Vert.x HTTP Server, so it should work

on what is based Quarkus gRPC ?

geoand:
> on what is based Quarkus gRPC

It is based on Vert.x gRPC, but I didn't see the same stacktrace involved.

In any case, I opened https://github.com/quarkusio/quarkus/pull/23186 on Quarkus

PR #4260, opened 2022-01-28
Introduce support for pcap traffic capturing of network traffic

geoand:
**Motivation**:

The idea here is to allow users to analyze HTTP traffic consumed and produced via Vert.x using Wireshark.
This capabity relies on Netty's `io.netty.handler.pcap.PcapWriteHandler` and has also been mentioned in https://github.com/vert-x3/issues/issues/566.

P.S. I tested this by also patching Quarkus to use this change and tried an application that uses RESTEasy Reactive and the Reactive REST Client.
Here is what the Wireshark screenshot looks like:
Screenshot from 2022-01-28 16-51-38

### DIFF
--- a/src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
+++ b/src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
@@ -28,6 +28,7 @@ import io.vertx.core.Handler;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.datagram.DatagramSocket;
 import io.vertx.core.datagram.DatagramSocketOptions;
+import io.vertx.core.http.impl.VertxPcapWriteHandler;
 import io.vertx.core.impl.AddressResolver;
 import io.vertx.core.impl.Arguments;
 import io.vertx.core.impl.ContextInternal;
@@ -77,6 +78,9 @@ public class DatagramSocketImpl implements DatagramSocket, MetricsProvider {
     context.nettyEventLoop().register(channel);
     if (options.getLogActivity()) {
       channel.pipeline().addLast("logging", new LoggingHandler(options.getActivityLogDataFormat()));
+    }
+    if ((options.getPcapCaptureFile() != null) && !options.getPcapCaptureFile().isEmpty()) {
+      channel.pipeline().addLast("pcapCapturing", new VertxPcapWriteHandler(options.getPcapCaptureFile()));
     }
     VertxMetrics metrics = vertx.metricsSPI();
     this.metrics = metrics != null ? metrics.createDatagramSocketMetrics(options) : null;
--- a/src/main/java/io/vertx/core/http/HttpClientOptions.java
+++ b/src/main/java/io/vertx/core/http/HttpClientOptions.java
@@ -414,6 +414,12 @@ public class HttpClientOptions extends ClientOptionsBase {
   }
 
   @Override
+  public HttpClientOptions setPcapCaptureFile(String pcapCaptureFile) {
+    super.setPcapCaptureFile(pcapCaptureFile);
+    return this;
+  }
+
+  @Override
   public HttpClientOptions setTrafficClass(int trafficClass) {
     super.setTrafficClass(trafficClass);
     return this;
--- a/src/main/java/io/vertx/core/http/HttpServerOptions.java
+++ b/src/main/java/io/vertx/core/http/HttpServerOptions.java
@@ -285,6 +285,12 @@ public class HttpServerOptions extends NetServerOptions {
   }
 
   @Override
+  public HttpServerOptions setPcapCaptureFile(String pcapCaptureFile) {
+    super.setPcapCaptureFile(pcapCaptureFile);
+    return this;
+  }
+
+  @Override
   public HttpServerOptions setTrafficClass(int trafficClass) {
     super.setTrafficClass(trafficClass);
     return this;
--- a/src/main/generated/io/vertx/core/dns/DnsClientOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/dns/DnsClientOptionsConverter.java
@@ -20,6 +20,11 @@ public class DnsClientOptionsConverter {
   public static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json, DnsClientOptions obj) {
     for (java.util.Map.Entry<String, Object> member : json) {
       switch (member.getKey()) {
+        case "activityLogFormat":
+          if (member.getValue() instanceof String) {
+            obj.setActivityLogFormat(io.netty.handler.logging.ByteBufFormat.valueOf((String)member.getValue()));
+          }
+          break;
         case "host":
           if (member.getValue() instanceof String) {
             obj.setHost((String)member.getValue());
@@ -54,6 +59,9 @@ public class DnsClientOptionsConverter {
   }
 
   public static void toJson(DnsClientOptions obj, java.util.Map<String, Object> json) {
+    if (obj.getActivityLogFormat() != null) {
+      json.put("activityLogFormat", obj.getActivityLogFormat().name());
+    }
     if (obj.getHost() != null) {
       json.put("host", obj.getHost());
     }
--- a/src/main/generated/io/vertx/core/eventbus/EventBusOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/eventbus/EventBusOptionsConverter.java
@@ -25,6 +25,11 @@ public class EventBusOptionsConverter {
             obj.setAcceptBacklog(((Number)member.getValue()).intValue());
           }
           break;
+        case "activityLogDataFormat":
+          if (member.getValue() instanceof String) {
+            obj.setActivityLogDataFormat(io.netty.handler.logging.ByteBufFormat.valueOf((String)member.getValue()));
+          }
+          break;
         case "clientAuth":
           if (member.getValue() instanceof String) {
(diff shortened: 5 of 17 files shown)


vietj:
we should have it also on NetSocket as it remains useful for other parts of vertx stack like MQTT or SQL client that use NetSocket

geoand:
Good point. I'll have a look on Monday

geoand:
Capturing also add to the `NetClientImpl` pipeline

vietj:
I think that the option declaration could simply be in `TCPSSLOptions` or perhaps `NetworkOptions` since it could work for UDP as well ?

geoand:
Indeed, PR updated

FroMage:
Very cool!

geoand:
@vietj is there anything else that this PR needs, or should I mark it ready for review?

src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -267,6 +272,7 @@
 
   private boolean shared;
   private String name;
+  private String pcapCaptureFile;
```
vietj, line 275: this should be removed in favor of `NetworkOptions` settings

src/main/java/io/vertx/core/http/HttpServerOptions.java
```diff
@@ -169,6 +174,7 @@
   private boolean webSocketPreferredClientNoContext;
   private int webSocketClosingTimeout;
   private TracingPolicy tracingPolicy;
+  private String pcapCaptureFile;
```
vietj, line 177: this should be removed in favor of `NetworkOptions `

src/main/java/io/vertx/core/http/HttpClientOptions.java
geoand: Good point, fixed

src/main/java/io/vertx/core/http/HttpServerOptions.java
geoand: Fixed

src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
```diff
@@ -80,6 +81,9 @@ private DatagramSocketImpl(VertxInternal vertx, DatagramSocketOptions options) {
     if (options.getLogActivity()) {
       channel.pipeline().addLast("logging", new LoggingHandler());
     }
+    if ((options.getPcapCaptureFile() != null) && !options.getPcapCaptureFile().isEmpty()) {
+      channel.pipeline().addLast("logging", new VertxPcapWriteHandler(options.getPcapCaptureFile()));
```
Unev, line 85: I guess, handler name can be changed here to "pcapCapturing" ?

geoand: Good catch, fixed!

vietj: thanks @Unev

vietj:
I did a quick review of the `VertxPcapWriteHandler` and I can see a few issues with the static metadata map and the pcap write handler.

1. it seems to me that when a channel is closed it will close the underlying output stream and all other concurrent channel use will not be able to write anymore.

2. the shared static map access seems to be racy, e.g two `FileOutputStream` could be created for the same file.

Overall this seems to come from the fact that we haven't properly defined the lifecycle of the pcap files and therefore we should first define how this should be used.

e.g the pcap file could be created when the HTTP server is starting and closed when the HTTP server is closed, etc...

geoand:
> it seems to me that when a channel is closed it will close the underlying output stream and all other concurrent channel use will not be able to write anymore.

Although theoretically true, `close` is called by `handlerRemoved` which properly guards handles the case of multiple writers.

> the shared static map access seems to be racy, e.g two FileOutputStream could be created for the same file.

I don't see how could this happen, as `Metadata` (which contains the reference to the OutputStream) is only created via `fileToMetadata.computeIfAbsent(pcapCaptureFile, Metadata::new);`

geoand:
Fair points, I'll come back tomorrow with answers

Unev:
Hi,  what happens incase of HTTPS(SSL/TLS), can we have ability dump encrypted message?

geoand:
> what happens incase of HTTPS(SSL/TLS), can we have ability dump encrypted message?

Yes, the messages are shown decrypted (in my original screenshot, one of the HTTP calls is actually an HTTPS call)

geoand:
It seems like this does have issues with concurrent writes. I'll have to investigate further

geoand:
The only reliable way I have found so far to get this to work for multiple event loop threads under load is have each instance write to its own file and then have the user invoke `mergecap` in order to create a capture file that has all the information.

It's not a great user experience, but it's not terrible either...

vietj:
for now I'm really interested by the client and server use case, i.e how /
when does the user start and end capture.

I think that we should have an API to control start/stop recording to give
user the control of the lifecycle of the dump and then we could have
predefined ways to control this, e.g when the HTTP server closes then the
pcap file is flushed and closed, etc...

geoand:
> for now I'm really interested by the client and server use case, i.e how /
when does the user start and end capture.

The way things are now, the bytes are written straight to a FileOutputStream, so the capture files are pretty much immediately available on the file system.

vietj:
yes but the question is when does this stop ?

geoand:
It does not stop

geoand:
> I think that we should have an API to control start/stop recording

How do you envision this would work? How would the user enable this?

geoand:
How is this for an alternative:

We introduce an SPI in Vert.x that allows for customizing the Netty Handler chain. That way in Quarkus we can introduce the the PcapHandler in whatever way we like.
WDYT?

geoand:
Bump :)

geoand:
Ah, I wasn't aware of `NetSocketInternal#channelHandlerContext`. That is super useful, so yeah, it would be great to have it for the HTTP API as well in order to cater for advanced use cases :)

jvican:
Just wanted to say that I think this is actually very useful for me as a developer, especially if we can make it work with WebSockets. It's been frustrating to debug TCP communication (especially when it's https) and to know exactly what is the JVM receiving or not (to detect bugs or find out what's going on), so this would improve my life considerably! :+1:

vietj:
perhaps this should be unified along with logging, as I see this more as a way to log network activity rather than dumping TCP packets.

geoand:
> If possible, In Quarkus, I'd like to avoid having to go down to that low-level just to be able to add a handler :)

To expand on this, what I would like is to be able to access the pipeline of the HTTP server and be able to add a handler.

vietj:
cast `HttpConnection` to `ConnectionBase`

geoand:
> cast HttpConnection to ConnectionBase

The problem with that however is that by the time I have a connection, I can no longer capture the first request, or am I missing something?

vietj:
I think we can make this work in Vert.x with the proper configuration, i.e I would like to reuse the current log activity configuration to have the activity logged as a pcap format.

Currently activity logging is in `NetworkOptions` with
- `logActivity` boolean
- `activityLogDataFormat` enum (Netty) added recently in 4.2.x

I think we could keep the `logActivity` and replace the Netty format by our own format class that would be either the current activity logging or a file in pcap format on the disk.

geoand:
> @geoand you can already do this with NetSocketInternal (add handlers) can you check with this API ? if that works we could expose the same API for HTTP.

Is this going to be considered after all?

vietj:
the current NetSocketInternal API works and you can access the channel handler context

geoand:
But that is not used for the server part, right?

vietj:
that is used for TCP server and client.

geoand:
But not HTTP, which is what we really care about (for starters :))

vietj:
would the proposal about improving vertx activity logging by a pcap file an option that works for quarkus ?

geoand:
Yeah, I think that would work just fine

vietj:
I'll have a look and reuse your work and provide a PR for review

geoand:
Coolio

geoand:
> I'll have a look and reuse your work and provide a PR for review

Did you have time to come up with something @vietj? If so, I'd love to take a look and see how we could use it in Quarkus

vietj:
not yet, I will do that after 4.3 is released.

geoand:
Thanks for the update

pmlopes:
Note that netty5 will move pcap to a external module. Shouldn't this feature too be an external module?

https://github.com/netty/netty/wiki/Netty-5-Migration-Guide

vietj:
I think we can do that in vertx 5 that will use Netty 5

for the moment we can keep like this

we will have the same issue for HA proxy support at least

yeikel:
I am not familiar with this but it looks very interesting. I would love to see a guide about this feature when it is released :)

geoand:
> > I'll have a look and reuse your work and provide a PR for review
> 
> Did you have time to come up with something @vietj? If so, I'd love to take a look and see how we could use it in Quarkus

@vietj did anything come out of this?

M-o-e-si (commented):
The arts is perfecting he's art ... love this broo but I don't know where to start I install Wireshark  but "capture"

vietj:
Actually I do not understand the use case for this over using tcpdump with wireshark directly instead, what is the benefit ?

geoand:
One very large benefit is that you can use this to capture TLS traffic which is really hard to do with a normal capture

yeikel:
I can confirm that analysis of tls traffic is very difficult in production environments

Issue #4084, opened 2021-09-02
setTimer should not throw when delay < 1


alamothe:
### Version

4.1.2

### Context

When `setTimer` is called with delay <1, it will throw `IllegalArgumentException`.

This behavior is unacceptable as it can easily take down the whole service. 

Instead, it simply needs to do `max(delay, 1)`.

Note that `node.js` doesn't throw in this case. There is no sane argument why it should throw. `setTimer(0)` simply means execute as soon as possible.

tsegismont:
If you need to schedule a task for execution asap, you can use `io.vertx.core.Vertx#runOnContext`.

If you're willing to provide a PR we may consider it.

PR #4261, opened 2022-01-30
Fix issue 4084 - setTimer should not throw exception for delay < 1ms

muhammad-abubakar:
Motivation:
Fixes https://github.com/eclipse-vertx/vert.x/issues/4084

If time set less than 1 ms in `io.vertx.core.Vertx#setTimer` then it throws an exception. Behaviour is changed to use max(delay, 1) to enforce 1ms restriction rather than throwing an exception

Conformance:


vietj:
actually, using a timer value of 0 should instead of creating a timer call `runOnContext` and not schedule a task

muhammad-abubakar:
> 

Thanks for the review. You mean it should schedule the task right away on the `EventLoop`  inside `InternalTimerHandler`? If yes then what about periodic tasks?

PR #4262, opened 2022-02-01
HTTP compression

vietj:


src/main/asciidoc/http.adoc
```diff
@@ -729,6 +729,14 @@ the more the level increases.
 By default - if compression is enabled via {@link io.vertx.core.http.HttpServerOptions#setCompressionSupported} - Vert.x will use '6' as compression level,
 but the parameter can be configured to address any case with {@link io.vertx.core.http.HttpServerOptions#setCompressionLevel}.
 
+=== HTTP compression algorithms
+
+Vert.x supports out of the box deflate and gzip provided out of the box by the JDK
+
+Vert.x supports extra algorithms
+
```
cescoffier, line 737: something missing here.

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -181,7 +182,13 @@ private HttpRequest createRequest(
     }
     if (options.isTryUseCompression() && request.headers().get(ACCEPT_ENCODING) == null) {
       // if compression should be used but nothing is specified by the user support deflate and gzip.
-      request.headers().set(ACCEPT_ENCODING, DEFLATE_GZIP);
+      CharSequence acceptEncoding = DEFLATE_GZIP;
+      if (Brotli.isAvailable()) {
```
cescoffier, line 186: Could you isolate this call in its own method? The idea is that it would ease writing a substitution is needed (the substitute would return `false`)

src/main/asciidoc/http.adoc
vietj: yes, it's on hold for now until API is final

src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
vietj: yes that's fine to do this

vietj:
ping @cescoffier

cescoffier (approved):
Thanks!

LGTM!

PR #4274, opened 2022-02-10
Add forEach method on HeadersMultiMap that avoids object allocation

geoand:
Motivation:

When iterating over the HeadersMultiMap (as vertx-http-proxy) does,
it makes sense to not have to allocate a useless Map.Entry container.
This new method provides that functionality

### DIFF
--- a/src/main/java/io/vertx/core/MultiMap.java
+++ b/src/main/java/io/vertx/core/MultiMap.java
@@ -21,6 +21,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 
 /**
@@ -72,6 +74,19 @@ public interface MultiMap extends Iterable<Map.Entry<String, String>> {
   List<String> getAll(CharSequence name);
 
   /**
+   * Allows iterating over the entries in the map
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  default void forEach(BiConsumer<String, String> action) {
+    forEach(new Consumer<Map.Entry<String, String>>() {
+      @Override
+      public void accept(Map.Entry<String, String> entry) {
+        action.accept(entry.getKey(), entry.getValue());
+      }
+    });
+  }
+
+  /**
    * Returns all entries in the multi-map.
    *
    * @return A immutable {@link java.util.List} of the name-value entries, which will be
@@ -80,7 +95,7 @@ public interface MultiMap extends Iterable<Map.Entry<String, String>> {
   @GenIgnore(GenIgnore.PERMITTED_TYPE)
   default List<Map.Entry<String, String>> entries() {
     List<Map.Entry<String, String>> entries = new ArrayList<>();
-    forEach(entries::add);
+    forEach((Consumer<Map.Entry<String, String>>)entries::add);
     return entries;
   }
 
--- a/src/main/java/io/vertx/core/http/impl/headers/HeadersMultiMap.java
+++ b/src/main/java/io/vertx/core/http/impl/headers/HeadersMultiMap.java
@@ -332,7 +332,16 @@ public final class HeadersMultiMap extends HttpHeaders implements MultiMap {
   public void forEach(Consumer<? super Map.Entry<String, String>> action) {
     HeadersMultiMap.MapEntry e = head.after;
     while (e != head) {
-      action.accept(new AbstractMap.SimpleEntry<>(e.key.toString(), e.value.toString()));
+      action.accept(e.stringEntry());
+      e = e.after;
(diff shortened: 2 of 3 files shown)


src/main/java/io/vertx/core/http/impl/headers/HeadersMultiMap.java
```diff
@@ -337,6 +337,14 @@ public void forEach(Consumer<? super Map.Entry<String, String>> action) {
     }
   }
 
+  public void forEach(BiConsumer<String, String> action) {
```
vietj, line 340: maybe we can have this in the `Multimap` API and instead pas `Handler<Map.Entry<CharSequence, CharSequence>>` ? (annotated with `@GenIgnore(GenIgnore.PERMITTED_TYPE)`)

src/main/java/io/vertx/core/MultiMap.java
```diff
@@ -71,6 +73,19 @@ static MultiMap caseInsensitiveMultiMap() {
   @GenIgnore(GenIgnore.PERMITTED_TYPE)
   List<String> getAll(CharSequence name);
 
+  /**
+   * Allows iterating over the entries in the map
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
```
vietj, line 79: why is this creating less instances than the `forEach<Consumer<>>` ?

src/main/java/io/vertx/core/http/impl/headers/HeadersMultiMap.java
vietj: so this might lead to a class cast exception because sometimes entries can be not strings, e.g `io.netty.handler.codec.http.AsciiString`, instead we should have I think:

```java
  @Override
  public void forEach(Consumer<? super Map.Entry<String, String>> action) {
    HeadersMultiMap.MapEntry e = head.after;
    while (e != head) {
      Map.Entry<String, String> entry;
      if (e.key instanceof String && e.value instanceof String) {
        entry = (Map.Entry) e;
      } else {
        entry = new AbstractMap.SimpleEntry<>(e.key.toString(), e.value.toString());
      }
      action.accept(entry);
      e = e.after;
    }
  }
```

we should test this case which is the HTTP server case as headers are provided by Netty has `AsciiString` implementation of `CharSequence`

geoand: Would that not have to be done in the existing `forEach` as well? 
I pretty much copied that and removed the allocation of the Map.Entry.

vietj: perhaps we should have on MapEntry a method that transform it to a Map.Entry<String, String> so it can be called when we need it

geoand: Added

vietj (commented):
can you add a test with a CharSequence entry iterated by forEach ?

geoand:
Done :)

src/main/java/io/vertx/core/http/impl/headers/HeadersMultiMap.java
```diff
@@ -332,7 +332,16 @@ public String get(String name) {
...
-      action.accept(new AbstractMap.SimpleEntry<>(e.key.toString(), e.value.toString()));
+      action.accept(e.stringEntry());
+      e = e.after;
+    }
+  }
+
+  public void forEach(BiConsumer<String, String> action) {
+    HeadersMultiMap.MapEntry e = head.after;
+    while (e != head) {
+      Map.Entry<String, String> consumedEntry = e.stringEntry();
```
vietj, line 343: in this case actually we don't need to call stringEntry, we can directly call `getkey()` / `getValue()` and then `toString()`, which avoids to create an entry when it contains a non string char sequence

geoand: Okay, updated

src/main/java/io/vertx/core/http/impl/headers/HeadersMultiMap.java
```diff
@@ -332,7 +332,15 @@ public String get(String name) {
   public void forEach(Consumer<? super Map.Entry<String, String>> action) {
     HeadersMultiMap.MapEntry e = head.after;
     while (e != head) {
-      action.accept(new AbstractMap.SimpleEntry<>(e.key.toString(), e.value.toString()));
+      action.accept(e.stringEntry());
+      e = e.after;
+    }
+  }
+
+  public void forEach(BiConsumer<String, String> action) {
```
vietj, line 340: missing `@Override`

geoand: Fixed

vietj (approved):

PR #4283, opened 2022-02-21
Support customizing BlockedThreadChecker

dano:
Signed-off-by: Dan O'Reilly <oreilldf@gmail.com>

Adds support for customizing the implementation of BlockedThreadChecker, exposing an API that allows users to pass a callback that will be executed whenever a thread is blocked for longer than the allowed duration. To use the API, users have to call the `blockedThreadChecker()` getter on `VertxInternal`.

**Motivation:**
In my application, I want to be able to restart the JVM if thread stay blocked for an extended period of time, so that my application can potentially self-heal if something goes very wrong and a thread ends up stuck forever. It's currently not possible to do this (at least as far as I can tell). Adding this new API makes it possible.

### DIFF
--- a/src/main/java/io/vertx/core/impl/btc/BlockedThreadEvent.java
+++ b/src/main/java/io/vertx/core/impl/btc/BlockedThreadEvent.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.impl.btc;
+
+/**
+ * A class containing status details about how long a particular thread has been blocked,
+ * and how long it is allowed to be blocked before warnings start getting logged. All times
+ * and durations are in nanoseconds.
+ */
+public class BlockedThreadEvent {
+  private final Thread thread;
+  private final long maxExecTime;
+  private final long duration;
+  private final long warningExceptionTime;
+
+  /**
+   * Create an instance of BlockedThreadEvent
+   *
+   * @param thread The thread being checked
+   * @param duration The duration the thread has been blocked, in nanoseconds
+   * @param maxExecTime The max execution time the thread is allowed, in nanoseconds
+   * @param warningExceptionTime The max time a thread can be blocked before stack traces get logged, in nanoseconds
+   */
+  public BlockedThreadEvent(Thread thread, long duration, long maxExecTime, long warningExceptionTime) {
+    this.thread = thread;
+    this.duration = duration;
+    this.maxExecTime = maxExecTime;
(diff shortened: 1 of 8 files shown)


src/test/java/io/vertx/core/BlockedThreadCheckerTest.java
```diff
@@ -119,4 +122,36 @@ public void start() throws InterruptedException {
...
+    ((VertxInternal) newVertx).blockedThreadChecker().setThreadBlockedHandler(bte -> {
+      Assert.assertEquals(NANOSECONDS.convert(maxWorkerExecuteTime, maxWorkerExecuteTimeUnit), bte.maxExecTime);
+      Assert.assertEquals(NANOSECONDS.convert(warningExceptionTime, warningExceptionTimeUnit), bte.warningExceptionTime);
+      cl.countDown();
+    });
+    DeploymentOptions deploymentOptions = new DeploymentOptions();
+    deploymentOptions.setWorker(true);
+    newVertx.deployVerticle(verticle, deploymentOptions);
+    await();
+    awaitLatch(cl);
```
vietj, line 155: avoid using a latch here, instead use `waitFor(2)` at the beginning of the test and `complete()` twice (one of `testComplete()` and one to replace the latch count down.

src/main/java/io/vertx/core/impl/BlockedThreadChecker.java
```diff
@@ -52,24 +83,28 @@ public void run() {
...
           }
         }
       }
     }, intervalUnit.toMillis(interval), intervalUnit.toMillis(interval));
   }
 
-  synchronized void registerThread(Thread thread, Task checked) {
+  /**
+   * Specify the handler to run when it is determined a thread has been blocked for longer than allowed.
+   *
```
vietj, line 100: we should specify that the handler will be call on the checker thread and not an event-loop

vietj:
the `BlockedThreadChecker` class should be moved to a `impl.btc` package and the nested class event should become a regular class instead.

src/test/java/io/vertx/core/BlockedThreadCheckerTest.java
dano: @vietj I tried to do it this way initially, but the `testComplete()` and `assert*` methods provided by the base class have checks to make sure they are called from an event loop thread, which means the test fails when they get called from the BlockedThreadChecker thread.

I could use `runOnContext` to get back on an event loop thread inside of the `threadBlockedHandler`, though. Do you prefer that to using the `CountdownLatch`?

vietj:
you can call `disableThreadChecks()` to disable thread checks

src/main/java/io/vertx/core/impl/btc/BlockedThreadEvent.java
```diff
@@ -0,0 +1,39 @@
...
+
+package io.vertx.core.impl.btc;
+
+/**
+ * A class containing status details about how long a particular thread has been blocked,
+ * and how long it is allowed to be blocked before warnings start getting logged. All times
+ * and durations are in nanoseconds.
+ */
+public class BlockedThreadEvent {
+  public final Thread thread;
```
vietj, line 20: use accessor instead, e.g `thread()`, `maxExecTime()`, etc...

vietj (changes requested):
minor change on `BlockedThreadEvent` and we are good to go

vietj:
thanks for the contribution @dano

PR #4300, opened 2022-03-09
Use the same logger name in BlockedThreadChecker to avoid regression when logger has been customized for this purpose, plus a couple of minor improvements to the blocked thread checker

vietj:


src/main/java/io/vertx/core/impl/btc/BlockedThreadChecker.java
```diff
@@ -36,7 +38,7 @@
     TimeUnit maxExecTimeUnit();
   }
 
-  private static final Logger log = LoggerFactory.getLogger(BlockedThreadChecker.class);
+  private static final Logger log = LoggerFactory.getLogger("io.vertx.core.impl.BlockedThreadChecker");
```
vietj, line 41: @cescoffier ping

cescoffier: Thanks!

PR #4305, opened 2022-03-13
[ISSUE-4304] clusterPublicPort should be ignored in toJson() result if     it is the default value

### LINKED ISSUE #4304: clusterPublicPort in EventBusOptions should be ignored in toJson() result if it is the default value
### Questions

`-1` is the default value of `clusterPublicPort` in `EventBusOptions`, but setting this default value will fail:

```java

  @Test
  public void testEventBusOptions() {
    VertxOptions vertxOptions1 = new VertxOptions();
    JsonObject json = vertxOptions1.toJson();
    VertxOptions vertxOptions2 = new VertxOptions(json);  //  <== this line leads to exception
    assertEquals(json, vertxOptions2.toJson());
  }
```

The exception is:

```java
java.lang.IllegalArgumentException: clusterPublicPort p must be in range 0 <= p <= 65535
	at io.vertx.core.eventbus.EventBusOptions.setClusterPublicPort(EventBusOptions.java:651)
	at io.vertx.core.eventbus.EventBusOptionsConverter.fromJson(EventBusOptionsConverter.java:55)
	at io.vertx.core.eventbus.EventBusOptions.<init>(EventBusOptions.java:165)
	at io.vertx.core.VertxOptionsConverter.fromJson(VertxOptionsConverter.java:45)
	at io.vertx.core.VertxOptions.<init>(VertxOptions.java:181)
	at io.vertx.core.VertxOptionsTest.testEventBusOptions(VertxOptionsTest.java:321)
```

This does not make sense as we cannot re-construct the VertxOptions using the JsonObject exported from a default VertxOptions instance.

### Version

latest `master` branch

### Do you have a reproducer?

Please refer to https://gist.github.com/gaol/7593cebcb2d1328d4ed50ece5ee1ac70 for more detail.

### Contribution

I will provide PR for the fix

gaol:
Motivation:

Fixes #4304 

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/EventBusOptions.java
+++ b/src/main/java/io/vertx/core/eventbus/EventBusOptions.java
@@ -173,6 +173,10 @@ public class EventBusOptions extends TCPSSLOptions {
   public JsonObject toJson() {
     JsonObject json = new JsonObject();
     EventBusOptionsConverter.toJson(this, json);
+    final String clusterPublicPortName = "clusterPublicPort";
+    if (json.containsKey(clusterPublicPortName) && json.getInteger(clusterPublicPortName) == DEFAULT_CLUSTER_PUBLIC_PORT) {
+      json.remove(clusterPublicPortName);
+    }
     return json;
   }
 
--- a/src/test/java/io/vertx/core/VertxOptionsTest.java
+++ b/src/test/java/io/vertx/core/VertxOptionsTest.java
@@ -315,6 +315,14 @@ public class VertxOptionsTest extends VertxTestBase {
   }
 
   @Test
+  public void testDefaultJsonVertxOptions() {
+    VertxOptions vertxOptions1 = new VertxOptions();
+    JsonObject json = vertxOptions1.toJson();
+    VertxOptions vertxOptions2 = new VertxOptions(json);
+    assertEquals(json, vertxOptions2.toJson());
+  }
+
+  @Test
   public void testJsonOptions() {
     VertxOptions options = new VertxOptions(new JsonObject());
 


src/main/java/io/vertx/core/eventbus/EventBusOptions.java
```diff
@@ -647,7 +647,7 @@ public int getClusterPublicPort() {
    * @return a reference to this, so the API can be used fluently
    */
   public EventBusOptions setClusterPublicPort(int clusterPublicPort) {
-    if (clusterPublicPort < 0 || clusterPublicPort > 65535) {
+    if (clusterPublicPort != DEFAULT_CLUSTER_PUBLIC_PORT && (clusterPublicPort < 0 || clusterPublicPort > 65535)) {
```
vietj, line 650: I think actually a better fix would be to update the `toJson()` method and remove the `clusterPublicPort` when it is `-1` so we don't end up in this case

gaol: OK,that makes sense and looks better :)

gaol:
Updated to remove `clusterPublicPort` from the result of `toJson()` method if the value is `-1`.

I am not sure if it is possible to have `codegen` supports `ignore default value`, that will be even better solution if it can support. :)

vietj:
@gaol there is no such possibility in codegen but it has been designed to only generate static methods so `toJson()` method have the ability to perform additional tweaks like this, so it's the most appropriate way to do that

vietj:
can you provide a PR for the 4.2 branch @gaol ?

gaol:
@vietj sure, done at: https://github.com/eclipse-vertx/vert.x/pull/4307 :)

PR #4317, opened 2022-03-22
Add isPaused() method to HttpServerRequest

lsutic:
Motivation:

Sometimes it is necessary to pause the request before handling it - for example if you intend to make some async calls and then call `toWebSocket()` on it. During that processing, the modules you call may in turn themselves want to pause the request.

In order to write such modules and make them composable, the module must be able to return the request to the state it was previously in (flowing or fetch). That is, you want to be able to do something like:

    boolean wasPaused = request.isPaused();
    request.pause(); // unconditional pause
    ... do stuff with a paused request...
    // restore the state of the request to whatever it was before
    if (!wasPaused) {
        request.resume()
    }

In order to do that you need to know if the request is paused or not. This PR adds an `isPaused()` method to the `HttpServerRequest` for that purpose.

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerRequest.java
@@ -300,6 +300,13 @@ public class Http2ServerRequest extends Http2ServerStream implements HttpServerR
   }
 
   @Override
+  public boolean isPaused() {
+    synchronized(conn) {
+      return super.isPaused();
+    }
+  }
+
+  @Override
   public HttpServerRequest resume() {
     return fetch(Long.MAX_VALUE);
   }
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerRequest.java
@@ -301,7 +301,7 @@ public class Http2ServerRequest extends Http2ServerStream implements HttpServerR
 
   @Override
   public boolean isPaused() {
-    synchronized(conn) {
+    synchronized (conn) {
       return super.isPaused();
     }
   }
--- a/src/main/java/io/vertx/core/http/HttpServerRequest.java
+++ b/src/main/java/io/vertx/core/http/HttpServerRequest.java
@@ -432,6 +432,13 @@ public interface HttpServerRequest extends ReadStream<Buffer> {
   boolean isEnded();
 
   /**
+   * Is the request paused?
+   *
+   * @return true if paused
+   */
+  boolean isPaused();
+
+  /**
    * Set a custom frame handler. The handler will get notified when the http stream receives an custom HTTP/2
    * frame. HTTP/2 permits extension of the protocol.
    *
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
@@ -493,6 +493,13 @@ public class Http1xServerRequest implements HttpServerRequestInternal, io.vertx.
   }
 
   @Override
+  public boolean isPaused() {
+    synchronized(conn) {
+      return pendingQueue().isPaused();
+    }
+  }
+
+  @Override
   public HttpServerRequest customFrameHandler(Handler<HttpFrame> handler) {
     return this;
   }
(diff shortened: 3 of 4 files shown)


src/main/java/io/vertx/core/http/impl/Http2ServerRequest.java
```diff
@@ -299,6 +299,13 @@ public HttpServerRequest pause() {
     return this;
   }
 
+  @Override
+  public boolean isPaused() {
+    synchronized(conn) {
+      return super.isPaused();
```
lsutic, line 305: This only takes the `VertxHttp2Stream.isPaused` and wraps it in a `synchronized(conn)` block. Should perhaps have different names for this and the superclass method, but couldn't come up with a good parallel to `doPause` - `doIsPaused`?

vietj:
@lsutic you can pause the request if thats not paused, there won't be any side effect

vietj:
I'm not a big fan of this change because it means that maybe this should be on the ReadStream interface but obviously we can't add this to ReadStream without side effects (and if we do add it to HttpServerRequest then ppl will ask for this to be in ReadStream). So I'm fine if that's go in `HttpServerRequestInternal` instead.

vietj:
actually I'm not sure this is needed because `toWebSocket()` will resume the request if that's paused

vietj:
also it seems that does not work if the stream in in fetch mode, because you would change the fetch value to Long.MAX_VALUE

lsutic:
@vietj the problem related to websockets, and the request processing done before we call `request.toWebSocket()`:

1. When we receive an HTTP request, we need to pause it if we're going to do any async calls. We are going to do async calls. So we pause it.
2. We then have logging, auth, some other modules, that are all added as handlers for the request:

    interface Handler {
        void handle(RoutingContext ctx)
    }

The context + request now passes through these in order.
3. Finally, we get to the "real" handler, which does a bit of processing and then calls `ctx.request().toWebSocket()`. It is important that the request remains paused until we get here, otherwise the `toWebSocket()` call will fail.

The question now is how to implement a typical `Handler` as outlined above. The `Handler` implementation may have to pause the request for whatever reason. The question is what the handler should do with a request that it has paused: should it leave the request paused, or should it resume the request?

When I wrote this PR I assumed that a "good" handler should restore the request to the state it had when it received it before passing it on to the next handler: that is, a good, modular handler should, if it receives a paused handler, leave it paused, and if it receives a flowing handler, return it to the flowing state before passing it on to the next handler.

But based on what you say here: "it seems that does not work if the stream in in fetch mode, because you would change the fetch value to Long.MAX_VALUE", I'm starting to think that the correct way to write a generic, modular `Handler` is:

1. If you explicitly need to pause the request, pause it.
2. If you need to read the body of the request, resume it.
3. Otherwise don't touch the pause/resume state.

vietj:
are we talking of Vert.x Web here ?

lsutic:
Yes.

lsutic:
I ended up rewriting our code according to the rules I listed:

1. If you explicitly need to pause the request, pause it.
2. If you need to read the body of the request, resume it.
3. Otherwise don't touch the pause/resume state.

And it worked. So I don't need the `isPaused()` method any more.

vietj:
I think we should make this clear in vertx-web and take this in account to improve the API in vertx 5

vietj:
@pmlopes can you look at the rules that @lsutic gave and give your opinion ? do all vertx-web handler respect this ?

pmlopes:
In web I recently changed its internals to properly handle upgrades and parsing. Before we were pausing resume the request all the time and doing upgrade headers check to avoid resuming.

A sane alternative is to start the web processing in pause mode, so when upgrades happen, core automatically resumes. For body handler I did the same and it reduces lots of checks in web.

So like @lsutic said, we can live without the extra API.

lsutic:
FYI, the rules that we ended up with were:

1. Pause the request before any async calls.
2. If you need to read the body of the request, resume it.
3. Otherwise don't touch the pause/resume state.

pmlopes:
These are also the rules web does. This will be visible from 4.3.0

vietj:
I think we should have these rules in the docs of vertx-web handler

pmlopes:
@vietj and we do. The PR where this change was introduced, include docs update to mention this!

vietj:
nice!

Issue #4323, opened 2022-03-31
Websocket Handshake timeout


vietj:
The WebSocket client does not have a handshake timeout, we can reuse the `WebSocketOptions#timeout` to configure this.

PR #4324, opened 2022-03-31
WebSocket handshake completion timeout

vietj:
The HttpClient does not have a timeout for handshake completion. After the connection has been established with the server, we can set a timeout to fire if the server has not responded with the handshake response timely.

See #4323

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -855,6 +855,7 @@ public class Http1xClientConnection extends Http1xConnectionBase<WebSocketImpl> 
     boolean allowOriginHeader,
     WebsocketVersion vers,
     List<String> subProtocols,
+    long handshakeTimeout,
     int maxWebSocketFrameSize,
     Promise<WebSocket> promise) {
     try {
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -898,6 +908,9 @@ synchronized void toWebSocket(
         !options.isSendUnmaskedFrames());
 
       WebSocketHandshakeInboundHandler handshakeInboundHandler = new WebSocketHandshakeInboundHandler(handshaker, ar -> {
+        if (timer > 0L) {
```
AlexanderAA, line 911: @vietj should this possibly be changed to `if (timer >= 0L) {`?

When I set .setTimeout(timeout) in my websocket client, the very first  call to `vertx.setTimer(handshakeTimeout, id -> closeInternal())` returns `0`. 
Because of that, the timer is not cancelled after and closes connection - after a successful handshake.

I am able to avoid this by initializing the timer counter before websocket connection setup: `vertx.cancelTimer(vertx.setTimer(1, ignored -> {}));`.

Is there some other configuration that I might be missing, that helps to avoid this behaviour?

PR #4327, opened 2022-04-01
bugfix: internal NetServer of ClusterEventBus closed too early. fixes #4326

### LINKED ISSUE #4326: internal NetServer of ClusterEventBus closes too early causing ConnectionHolder failures
### Version

VertX 4.2.6

### Context

[ClusterEventBus uses NetServer#listen](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java#L90) internally but this adds a 
[closehook on context](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/net/impl/TCPServerBase.java#L185) while ClusterEventBus expects it is in control of the LifeCycle and [wants to close the server later](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java#L110).

This leads to other nodes sending / publishing clustered messages to instantly loose the connection and triggers their [close()](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java#L141)

But because the [subscriptions have not yet been updated in NodeSelector](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/spi/cluster/NodeSelector.java#L68) the producers connect again to the closed server.

causing loads of:
`i.v.c.e.i.c.ConnectionHolder - Connecting to server 41650572-fd35-42cc-8509-369021af63bd failed`

### Steps to reproduce

1. Clustered VertX instance Producer constantly send / publish "foo" to "address"
2. Clustered VertX instance Consumer consuming "address"
3. Having a **shutdown delay** in the consumer - delay for 5 seconds (we use this "trick" to remove our microservices from discovery - finish processing - then really closing vertx)
4. close Consumer instance

### Extra

OpenJDK 18 / Fedora CoreOS 35

DemonicTutor:
Signed-off-by: Markus Spika <markus.spika@gmail.com>

Motivation:

was looking for this over a year now! finally found it....

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/impl/VertxImpl.java
+++ b/src/main/java/io/vertx/core/impl/VertxImpl.java
@@ -581,14 +581,14 @@ public class VertxImpl implements VertxInternal, MetricsProvider {
         }
         haPromise.future().onComplete(ar2 -> {
           addressResolver.close(ar3 -> {
-            Promise<Void> ebClose = getOrCreateContext().promise();
-            eventBus.close(ebClose);
-            ebClose.future().onComplete(ar4 -> {
-              closeClusterManager(ar5 -> {
-                // Copy set to prevent ConcurrentModificationException
-                deleteCacheDirAndShutdown(completionHandler);
+              closeClusterManager(ar4 -> {
+                Promise<Void> ebClose = getOrCreateContext().promise();
+                eventBus.close(ebClose);
+                ebClose.future().onComplete(ar5 -> {
+                  // Copy set to prevent ConcurrentModificationException
+                  deleteCacheDirAndShutdown(completionHandler);
+                });
               });
-            });
           });
         });
       });
--- a/src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
+++ b/src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
@@ -516,4 +516,24 @@ public class ClusteredEventBusTest extends ClusteredEventBusTestBase {
     assertNotNull(nodeSelectorRef.get());
     assertFalse(nodeSelectorRef.get().wantsUpdatesFor(ADDRESS1));
   }
+
+  @Test
+  public void testClusterEventBusClosesItsInternalNetServer() throws Exception {
+    startNodes(2);
+
+    final MessageProducer<String> producer = vertices[0].eventBus().sender(ADDRESS1, new DeliveryOptions().setLocalOnly(false));
+
+    ((VertxInternal) vertices[1]).addCloseHook(
+      closeable -> vertx.setTimer(250, // delay to fail consistently - NetServer shutdown takes a bit
+        id -> producer.write("foo")
+          .onComplete(onSuccess(nothing -> testComplete()))
+          .onComplete(closeable) // completing closeable later is important
+      )
+    );
+    vertices[1].eventBus().consumer(ADDRESS1).handler(msg -> vertices[1].close());
+
+    producer.write("foo");
+
+    await();
+  }
 }
--- a/src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
+++ b/src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
@@ -11,9 +11,7 @@
 
 package io.vertx.core.eventbus;
 
-import io.vertx.core.Promise;
-import io.vertx.core.Vertx;
-import io.vertx.core.VertxOptions;
+import io.vertx.core.*;
 import io.vertx.core.impl.VertxInternal;
 import io.vertx.core.spi.cluster.NodeSelector;
 import io.vertx.core.spi.cluster.RegistrationUpdateEvent;
@@ -520,20 +518,15 @@ public class ClusteredEventBusTest extends ClusteredEventBusTestBase {
   @Test
   public void testClusterEventBusClosesItsInternalNetServer() throws Exception {
     startNodes(2);
-
     final MessageProducer<String> producer = vertices[0].eventBus().sender(ADDRESS1, new DeliveryOptions().setLocalOnly(false));
-
-    ((VertxInternal) vertices[1]).addCloseHook(
-      closeable -> vertx.setTimer(250, // delay to fail consistently - NetServer shutdown takes a bit
-        id -> producer.write("foo")
-          .onComplete(onSuccess(nothing -> testComplete()))
-          .onComplete(closeable) // completing closeable later is important
-      )
+    final Closeable onClosing = closeable -> vertx.setTimer(250, // delay to fail consistently - NetServer shutdown takes a bit
+      id -> producer.write("foo")
+        .onComplete(onSuccess(nothing -> testComplete()))
+        .onComplete(closeable) // completing closeable later is important
     );
+    ((VertxInternal) vertices[1]).addCloseHook(onClosing);
     vertices[1].eventBus().consumer(ADDRESS1).handler(msg -> vertices[1].close());
-
     producer.write("foo");
-
     await();
   }
 }
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
@@ -11,6 +11,7 @@
 
 package io.vertx.core.eventbus.impl.clustered;
 
+import io.vertx.core.Closeable;
 import io.vertx.core.Handler;
 import io.vertx.core.MultiMap;
 import io.vertx.core.Promise;
@@ -96,6 +97,7 @@ public class ClusteredEventBus extends EventBusImpl {
       clusterManager.setNodeInfo(nodeInfo, setPromise);
       return setPromise.future();
     }).onSuccess(v -> {
+      vertx.removeCloseHook((Closeable) server);
       started = true;
       nodeSelector.eventBusStarted();
     }).onComplete(promise);


src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -581,14 +581,14 @@ public synchronized void close(Handler<AsyncResult<Void>> completionHandler) {
...
-              closeClusterManager(ar5 -> {
-                // Copy set to prevent ConcurrentModificationException
-                deleteCacheDirAndShutdown(completionHandler);
+              closeClusterManager(ar4 -> {
+                Promise<Void> ebClose = getOrCreateContext().promise();
+                eventBus.close(ebClose);
+                ebClose.future().onComplete(ar5 -> {
+                  // Copy set to prevent ConcurrentModificationException
+                  deleteCacheDirAndShutdown(completionHandler);
+                });
```
DemonicTutor, line 590: currently ClusterEventBus is closed first which closes the NetServer and then ClusterManager which removes all MessageConsumer subscriptions.

maybe this is super naive of me but can this be just turned around?

**NOTE**: WIP commit

src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
```diff
@@ -516,4 +516,24 @@ public void init(Vertx vertx, NodeSelector nodeSelector) {
...
     assertFalse(nodeSelectorRef.get().wantsUpdatesFor(ADDRESS1));
   }
+
+  @Test
+  public void testClusterEventBusClosesItsInternalNetServer() throws Exception {
+    startNodes(2);
+
+    final MessageProducer<String> producer = vertices[0].eventBus().sender(ADDRESS1, new DeliveryOptions().setLocalOnly(false));
+
+    ((VertxInternal) vertices[1]).addCloseHook(
```
vietj, line 526: the lambda argument of `addCloseHook` should be kept in local variable of the `testClusterEventBusClosesItsInternalNetServer` otherwise it might be garbage collected and the test would not be correct

src/main/java/io/vertx/core/impl/VertxImpl.java
vietj: I think it makes sense if we are sure that the parts of the clustered event bus used by the event bus will work correctly when the cluster manager is closed, i.e the closed node selector will return a failure, currently it does not seem to.

@tsegismont WDYT ?

DemonicTutor: the interesting question is the order of shutdown... is there any "lifecycle" documented or implemented?

IMHO the ideal sequence would be:
1. discovery unpublish self
2. --optional delay--
3. shutoff servers / external consumers (httpserver, netserver, custer message consumers)
4. shutoff timers
5. verticle stop()
6. local message consumers
7. clients (httpclient, netclient, mongoclient, ....)
8. clustered maps / locks / counter

tsegismont:
>currently ClusterEventBus is closed first which closes the NetServer and then ClusterManager which removes all MessageConsumer subscriptions.

@DemonicTutor when the clustered eventbus is closed, it removes subscriptions from the cluster manager, which is why shutdown happens in this order.

When I look at the reproducer (test case), I can't see anything wrong. The `write` operation on the message producer has a _fire and forget_ behavior. In other words, the future it returns may signal completion even though many different bad things can happen.

DemonicTutor:
@tsegismont  are you referring to my first commit regarding the closehook?

currently the ClusterEventBus is closed first and this **immediately** closes the NetServer -> can no longer receive messages.
later when the ClusterManager is closed it triggers removal of the subscriptions.

i dont have a standalone reproducer right now but what i do to reproduce the issue is:

```
public static final class PingVerticle extends AbstractVerticle {

    @Override
    public Completable rxStart() {
      shutdown.register(super.vertx);

      super.vertx.setPeriodic(10, id -> super.vertx.eventBus().publish("request", Instant.now() + " ping"));
      ((VertxInternal)vertx.getDelegate()).getClusterManager().nodeListener(new NodeListener() {
        @Override
        public void nodeAdded(String nodeID) {
          System.err.println("add " + nodeID);
        }

        @Override
        public void nodeLeft(String nodeID) {
          System.err.println("remove " + nodeID);
        }
      });
      return Completable.complete();
    }
  }

  public static final class PongVerticle extends AbstractVerticle implements Closeable {

    private static final AtomicReference<Vertx> ref = new AtomicReference<>();

    public static void shutdown() {
      System.err.println(Instant.now() + " shutdown");
      ref.get().getDelegate().close();
    }

    private MessageConsumer<String> consumer;

    @Override
    public Completable rxStart() {
      ref.set(super.vertx);
      shutdown.register(super.vertx);

      ((VertxInternal)vertx.getDelegate()).addCloseHook(this);

      consumer = vertx.eventBus().<String>consumer("request").getDelegate();
      consumer.endHandler(nothing -> System.err.println("end handler"));
      consumer.completionHandler(nothing -> System.err.println("complete handler"));
      consumer.handler(msg -> { /* eat it */ });

      return Completable.complete();
    }

    @Override
    public Completable rxStop() {
      System.err.println("stopping");
      return Completable.complete();
    }

    @Override
    public void close(final Promise<Void> completion) {
      System.err.println("closing");
      completion.complete();
    }
  }
```

this test has some magic but basically it allows discovery + starting via vertx Launcher
```
  @Test
  public void shouldFormCluster(final TestContext context) throws Throwable {
    // GIVEN
    final var given = "test-config-ssl.json";
    // EXPECTATIONS
    mockConsulReturningFirstRegistration();

    whenLaunching(ClusteredLaunchTaskTest.class.getName() + "$" + PongVerticle.class.getSimpleName(), asTestConfigJsonFile(given)).accept(null);
    whenLaunching(ClusteredLaunchTaskTest.class.getName() + "$" + PingVerticle.class.getSimpleName(), asTestConfigJsonFile(given)).accept(null);

    Thread.sleep(5_000);
    PongVerticle.shutdown();

    Thread.sleep(15_000);
  }
```

and the output is:


```
[INFO] c.n.m.c.b.Launcher - Using 10.147.110.52 as serviceHost
[INFO] c.n.m.c.b.Launcher - Using 10.147.110.52 as serviceHost
[INFO] c.n.m.c.b.Launcher - Max Direct Memory: 7976MB
[INFO] c.n.m.c.b.Launcher - Max Direct Memory: 7976MB
[INFO] c.n.m.i.l.ClusteredLaunchTask - clustering enabled
[INFO] c.n.m.i.l.ClusteredLaunchTask - clustering enabled
[INFO] c.n.m.i.l.ConfigurableCommand - Starting clustering...
[INFO] c.n.m.i.l.ConfigurableCommand - Starting clustering...
complete handler
[INFO] i.v.c.i.l.c.VertxIsolatedDeployer - Succeeded in deploying verticle
[INFO] i.v.c.i.l.c.VertxIsolatedDeployer - Succeeded in deploying verticle
2022-04-11T18:30:22.177886999Z shutdown
closing
stopping
end handler
[WARN] i.v.c.e.i.c.ConnectionHolder - Connecting to server 96470f4b-f94a-4ff3-8c2d-fd83fa56f643 failed
[WARN] i.v.c.e.i.c.ConnectionHolder - Connecting to server 96470f4b-f94a-4ff3-8c2d-fd83fa56f643 failed
[WARN] i.v.c.e.i.c.ConnectionHolder - Connecting to server 96470f4b-f94a-4ff3-8c2d-fd83fa56f643 failed
[WARN] i.v.c.e.i.c.ConnectionHolder - Connecting to server 96470f4b-f94a-4ff3-8c2d-fd83fa56f643 failed
[WARN] i.v.c.e.i.c.ConnectionHolder - Connecting to server 96470f4b-f94a-4ff3-8c2d-fd83fa56f643 failed
remove 96470f4b-f94a-4ff3-8c2d-fd83fa56f643

Process finished with exit code 130 (interrupted by signal 2: SIGINT)
```

if the stop is delayed these connection failures get even more pronounced:

```
@Override
public Completable rxStop() {
      System.err.println("stopping");
      return
         // AsyncResultCompletable.toCompletable(Future.future(((VertxInternal)vertx.getDelegate()).getClusterManager()::leave))
         // .doOnComplete(() -> System.err.println("shutoff cluster"))
      Completable.complete()
        .delay(10, TimeUnit.SECONDS)
        .doOnComplete(() -> System.err.println("stopped"));
}
```

tsegismont:
@DemonicTutor I assumed this PR is related to #4326 and tried to understand the issue looking at the reproducer in this PR (the test case).

>currently the ClusterEventBus is closed first and this immediately closes the NetServer -> can no longer receive messages.
later when the ClusterManager is closed it triggers removal of the subscriptions.

Indeed, the `ClusteredEventBus` is closed first indeed. In practice, this means:

1. `EventBus.close` method is invoked; this method unregisters all handlers (and to do so, it needs the cluster manager to be up and running)
2. the `NetServer` is closed.

See https://github.com/eclipse-vertx/vert.x/blob/8d9bb2ba44ce51bf461e21217c501ea31aa019ae/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java#L104-L124

And https://github.com/eclipse-vertx/vert.x/blob/eebee2c31616a3a01b889ccf69dc590777173fcf/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java#L206-L218

So the order of shutdown seems good to me.

>i dont have a standalone reproducer right now but what i do to reproduce the issue is:

I looked at the code snippets and I can't see anything wrong. The error messages are shown because the lookup for recipient in the node that runs `PingVerticle` happened before the consumer was unregistered but the message is not sent yet when it has been unregistered.

zyclonite:
seems to be related: https://groups.google.com/g/vertx/c/Kn2g0F0taIk?hl=en-GB

DemonicTutor:
> seems to be related: https://groups.google.com/g/vertx/c/Kn2g0F0taIk?hl=en-GB

yep.. the closehook (that i remove in the first commit) comes before the *stop* of the verticle

zyclonite:
and maybe more https://groups.google.com/g/vertx/c/dBjEhThpr8o?hl=en-GB

tsegismont:
@zyclonite I believe this message is related to https://github.com/vert-x3/vertx-zookeeper/issues/128

zyclonite:
@tsegismont for me it sounded like he want's to shutdown vertx and delays one stop method of a verticle to give his application time to process the remaining work but as the netclient registered the shutdown hook with vertx, it get already closed at the beginning of the stop call...

that's what @DemonicTutor explained to me yesterday, happens in his case as well

tsegismont:
@zyclonite right. I read the message again and replied to the author

DemonicTutor:
@tsegismont glad it helped!

PR #4340, opened 2022-04-20
Support yaml files on the use of "-conf"

riojano0:
Motivation:

Try to provide support to the use of the flag "-conf" with YAML files

Eg: 
```java -jar target/my-project.jar -conf target/config.yaml```

The approach is similar to the one used on io.vertx.config.yaml.YamlProcessor

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,6 +134,10 @@
       <artifactId>jackson-databind</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-yaml</artifactId>
+    </dependency>
 
     <!-- Loggers -->
     <dependency>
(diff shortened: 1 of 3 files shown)


tsegismont:
@riojano0 thank you for your proposal. We want Vert.x core to have as few dependencies as possible and to keep functionality to simple level. If you need to handle config as YAML then `vertx-config` is the way to go.

tsegismont:
If you really want to support `-conf` pointing to `YAML` file, you could do the following:

- create a subclass of `RunCommand` which overrides `getJsonFromFileOrString()` method to support YAML
- override the `load()` method in your custom launcher in order to replace the `run` command with you implementation

riojano0:
Thanks that already worked. But how do you think that will be a nice way to make work outside the box only adding the dependency of vert-conf-yaml?

tsegismont:
It's not possible out of the box.

PR #4344, opened 2022-04-25
Introduce HttpClientInternal interface

tsegismont:
It is required to fix https://github.com/vert-x3/vertx-web/issues/2165

When the `WebClient` is configured to use a shared `HttpClient`, the client instance cannot be cast to `HttpClientImpl`.


src/main/java/io/vertx/core/http/impl/HttpClientInternal.java
```diff
@@ -0,0 +1,26 @@
...
+import io.vertx.core.http.HttpClient;
+import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.impl.VertxInternal;
+
+public interface HttpClientInternal extends HttpClient {
+
+  /**
+   * @return the vertx, for use in package related classes only.
+   */
+  VertxInternal getVertx();
```
vietj, line 23: name it "vertx()" instead

tsegismont: Will do and merge.

src/main/java/io/vertx/core/http/impl/HttpClientInternal.java
```diff
@@ -0,0 +1,26 @@
...
+
+  HttpClientOptions getOptions();
```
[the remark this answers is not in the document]
tsegismont: Will do and merge.

PR #4346, opened 2022-04-26
Update BareCommand to support argument processor on cli

riojano0:
New PR base from https://github.com/eclipse-vertx/vert.x/pull/4340

This will allow to each vertx-conf-{extension} to define another CliJsonProcessor(if is needed) to allow for example work with YAML files


tsegismont:
Closing, see the discussion in https://github.com/eclipse-vertx/vert.x/pull/4340

PR #4350, opened 2022-04-29
Local consumer should never get a message sent or published remotely

### LINKED ISSUE #4116: EventBus.localConsumer and EventBus.consumer interfer with each other
### Version

The problems occurs with versions 3.8.5, 3.9.9 and 4.1.4

### Context

When handlers are registered at the eventBus for the **same address** with `localConsumer()` and `consumer()`, the behaviour is not as expected:

1. When the `localConsumer()` is registered first, the `consumer()` will never get called for messages, which were published on other cluster nodes.
2. When the `consumer()` is registered first, the `localConsumer()` will always get called even for messages, which are not local, but were published on other cluster nodes.

The expected behaviour would be, that a `localConsumer()` only gets called for local messages and the `consumer()` gets called for all messages, even if registered for the same address.

### Reproducer

Message Receiver:
```
import java.util.concurrent.CountDownLatch;

import io.vertx.core.Vertx;
import io.vertx.core.VertxOptions;
import io.vertx.core.eventbus.EventBus;
import io.vertx.core.eventbus.EventBusOptions;
import io.vertx.spi.cluster.hazelcast.HazelcastClusterManager;

public class Receiver {
	public static Vertx vertx;

	public final static String FIRST_ADDRESS = "first.address";

	public final static String SECOND_ADDRESS = "second.address";

	public static void main(String[] args) throws InterruptedException {
		int port = 6151;
		CountDownLatch waitLatch = new CountDownLatch(1);

		VertxOptions vertxOptions = new VertxOptions();
		vertxOptions.setClusterManager(new HazelcastClusterManager());

		EventBusOptions eventbus = vertxOptions.getEventBusOptions();
		eventbus.setHost("0.0.0.0");
		eventbus.setPort(port);
		eventbus.setClusterPublicHost("0.0.0.0");
		eventbus.setClusterPublicPort(port);

		Vertx.clusteredVertx(vertxOptions, result -> {
			vertx = result.result();

			EventBus eventBus = vertx.eventBus();

			eventBus.localConsumer(FIRST_ADDRESS, message -> {
				System.out.println("Event to " + FIRST_ADDRESS + " handled with local consumer");
			});

			eventBus.consumer(FIRST_ADDRESS, message -> {
				System.out.println("Event to " + FIRST_ADDRESS + " handled with consumer");
			});

			eventBus.consumer(SECOND_ADDRESS, message -> {
				System.out.println("Event to " + SECOND_ADDRESS + " handled with consumer");
			});

			eventBus.localConsumer(SECOND_ADDRESS, message -> {
				System.out.println("Event to " + SECOND_ADDRESS + " handled with local consumer");
			});

			System.out.println("Awaiting messages");
		});

		waitLatch.await();
	}
}
```

Message Sender:
```
import java.util.concurrent.CountDownLatch;

import io.vertx.core.Vertx;
import io.vertx.core.VertxOptions;
import io.vertx.core.eventbus.EventBus;
import io.vertx.core.eventbus.EventBusOptions;
import io.vertx.spi.cluster.hazelcast.HazelcastClusterManager;

public class Sender {
	public static Vertx vertx;

	public final static String FIRST_ADDRESS = "first.address";

	public final static String SECOND_ADDRESS = "second.address";

	public static void main(String[] args) throws InterruptedException {
		int port = 6152;
		CountDownLatch waitLatch = new CountDownLatch(1);

		VertxOptions vertxOptions = new VertxOptions();
		vertxOptions.setClusterManager(new HazelcastClusterManager());

		EventBusOptions eventbus = vertxOptions.getEventBusOptions();
		eventbus.setHost("0.0.0.0");
		eventbus.setPort(port);
		eventbus.setClusterPublicHost("0.0.0.0");
		eventbus.setClusterPublicPort(port);

		Vertx.clusteredVertx(vertxOptions, result -> {
			vertx = result.result();

			EventBus eventBus = vertx.eventBus();

			System.out.println("Sending messages");

			eventBus.publish(FIRST_ADDRESS, null);
			eventBus.publish(SECOND_ADDRESS, null);
		});

		waitLatch.await();
	}
}
```

### Steps to reproduce:
1. Start the Message Receiver
2. Wait until the message receiver prints "Awaiting messages"
3. Start the sender
4. Sender will print "Sending messages" and then publish two messages
5. Receiver will handle the second message with both the localConsumer and the consumer, but will not handle the first message at all

tsegismont:
Fixes #4116

When a node has a mix of local and clustered consumers registered to a single address, a message sent from a remote node may be delivered to the local consumers, which is unexpected.

This happens because when the node receives the clustered message, it invokes deliverMessageLocally which does not inspect the type of handlers.

So, in case of point to point messaging, we need to iterate over the content of the sequence of handlers until we find one which is either a reply handler or not local only.
For pub/sub, we need to filter out local-only consumers.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -330,12 +330,13 @@ public class EventBusImpl implements EventBusInternal, MetricsProvider {
 
   protected ReplyException deliverMessageLocally(MessageImpl msg) {
     ConcurrentCyclicSequence<HandlerHolder> handlers = handlerMap.get(msg.address());
+    boolean messageLocal = isMessageLocal(msg);
     if (handlers != null) {
       if (msg.isSend()) {
         //Choose one
-        HandlerHolder holder = handlers.next();
+        HandlerHolder holder = nextHandler(handlers, messageLocal);
         if (metrics != null) {
-          metrics.messageReceived(msg.address(), !msg.isSend(), isMessageLocal(msg), holder != null ? 1 : 0);
+          metrics.messageReceived(msg.address(), !msg.isSend(), messageLocal, holder != null ? 1 : 0);
         }
         if (holder != null) {
           holder.handler.receive(msg.copyBeforeReceive());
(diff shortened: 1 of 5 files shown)


tsegismont:
Reviewed off-list with Julien

PR #4351, opened 2022-05-02
Enable using Supplier functions in sequential coordination of Futures

markmatney:
I sometimes find myself writing code like this:

```java
Future<Void> f = returnSomeVoidFuture();
f.compose(unused -> returnAnotherFuture());
```

The problem here is that when the unary `Function` argument is invoked, `unused` has the value `null`, so programmers must be careful not to reference it. To mitigate this danger, I have been using a naming convention for such unused parameters, but I think a better solution is to allow passing a nullary `Supplier` function instead:

```java
f.compose(() -> returnAnotherFuture());
```

This changeset adds support for that syntax, and similarly for `Future::eventually` (which has the same problem).

P.S. I think it would be nice if `Future::onComplete` and `Future::onSuccess` also supported this syntax, but that change seems much more difficult to implement in a backwards-compatible way.

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -20,6 +20,7 @@ import io.vertx.core.impl.future.SucceededFuture;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * Represents the result of an action that may, or may not, have occurred yet.
@@ -181,6 +182,13 @@ public interface Future<T> extends AsyncResult<T> {
   }
 
   /**
+   * Alias for {@link #compose(Supplier)}.
+   */
+  default <U> Future<U> flatMap(Supplier<Future<U>> supplier) {
+    return compose(supplier);
+  }
+
+  /**
    * Compose this future with a {@code mapper} function.<p>
    *
    * When this future (the one on which {@code compose} is called) succeeds, the {@code mapper} will be called with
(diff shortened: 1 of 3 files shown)


tsegismont:
Thank you for your proposal @markmatney .

I don't understand how, as programmers, we can come to reference a `Void` value. For years, Vert.x has been using, for example, `Handler<Void>` in API and internally, without any trouble.

Besides, even with the changes in this PR, nothing prevents from using `compose` with a `Function` on void future. And nothing prevents, for example, a Future<String> to hold a null value.

The solution is to be careful with the values we reference.

markmatney:
Thanks for the review @tsegismont. I will respect your decision if this is to remain "wontfix", but I would like to elaborate on why I think this is important.

For experienced users of Vert.x and those familiar with its internals, I'm sure the risk of referencing these null values can be considered nonexistent, but _I don't think that assumption should be made about all users_ (which is a much larger and more diverse group in terms of skill level, since it includes interns, novice programmers, etc.). I think every programmer has accidentally referenced a null value at least once, whether in production code or during development.

Considering the size and diversity of the user base (as a rough measurement, this repo has 12.9k stars and 1.9k "forks" as of this writing), I really want to emphasize that in order to call `eventually` on any `Future`, users are _effectively required_ to declare, in the callback function, a local variable with the value `null`. (The same is true for calling `compose` on `Future<Void>`, a type that is returned by lots of methods in the public API, not just in vertx-core but across the ecosystem.) This does not seem like a good practice to me, especially since it can be avoided by instead using the methods introduced by this PR.

tsegismont:
I understand your point of view, yet having these additional methods won't prevent anyone from invoking `compose`, `map`.

If you prefer to work with an API that distinguishes completions from future non-null values, consider using [`vertx-rx-java3`](https://vertx.io/docs/vertx-rx/java3/)

PR #4359, opened 2022-05-04
ClusterSerializable improvements

tsegismont:
Commits can be reviewed individually


src/main/java/io/vertx/core/shareddata/ClusterSerializable.java
```diff
@@ -0,0 +1,29 @@
...
+
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * Objects implementing this interface will be written to and read from a {@link Buffer} when respectively
+ * stored and read from an {@link AsyncMap}.
+ * <p>
+ * Implementations must have a public no-argument constructor.
+ */
+public interface ClusterSerializable extends io.vertx.core.shareddata.impl.ClusterSerializable {
```
vietj, line 22: I think it would be better the other way around : have the impl extends the non impl

tsegismont:
Approved by @vietj off-list

zyclonite:
@tsegismont seems this broke all cluster managers as well as the session serialization in the vertx-web
was this by accident because the deprecated interface can still be used?

tsegismont:
@zyclonite thanks for the heads-up. I had tried with Infinispan CM before the changes asked in [this comment
](https://github.com/eclipse-vertx/vert.x/pull/4359/files/b9196262da9459d9efed49392c049f52b1b4652f#diff-51cb1f897d162023a24e45b49179cb1ead7f0ce024c8acde88b7e0bb7a35b9b7)

I can't remember but that was maybe why in the first design the new interface extended the old one.

There is still another option: update all CMs. But perhaps it is better to allow smooth upgrades between nodes.

 I'll look into this and come back to you later today.

tsegismont:
I can confirm this is due to 40d9726.

@vietj I am going to revert this commit so that people can mix 4.2 nodes with 4.3 nodes while upgrading.

tsegismont:
@zyclonite @vietj see https://github.com/eclipse-vertx/vert.x/pull/4363

When I compile Vert.x core with this patch, the Infinispan CM build passes without any changes.

zyclonite:
@tsegismont thank you for the fast clarification

PR #4362, opened 2022-05-05
Codec improvements

tsegismont:
Commits can be reviewed individually.


src/main/java/io/vertx/core/eventbus/ClassNamePredicate.java
```diff
@@ -0,0 +1,38 @@
...
+ * @see EventBus#rejectSerializable(ClassNamePredicate)
+ * @see EventBus#allowSerializable(ClassNamePredicate)
+ * @see EventBus#rejectClusterSerializable(ClassNamePredicate)
+ * @see EventBus#allowClusterSerializable(ClassNamePredicate)
+ */
+@VertxGen
+@FunctionalInterface
+public interface ClassNamePredicate extends Predicate<String> {
+
+  /**
```
jponge, line 32: You might as well put not docs here 😉

src/main/java/io/vertx/core/impl/ClusterSerializableUtils.java
```diff
@@ -0,0 +1,36 @@
...
+    try {
+      ClusterSerializable copy = obj.getClass().getConstructor().newInstance();
+      copy.readFromBuffer(0, buffer);
+      return copy;
+    } catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private ClusterSerializableUtils() {
```
jponge, line 33: You might just do an interface with static methods.

jponge: (also for `SerializableUtils`)

jponge (commented):
I'm no expert, but LGTM overall.

src/main/java/io/vertx/core/eventbus/ClassNamePredicate.java
```diff
@@ -0,0 +1,38 @@
...
+ * This is used by the system codecs for the subtypes of {@link io.vertx.core.shareddata.ClusterSerializable} and {@link java.io.Serializable}.
+ *
+ * @see EventBus#rejectSerializable(ClassNamePredicate)
+ * @see EventBus#allowSerializable(ClassNamePredicate)
+ * @see EventBus#rejectClusterSerializable(ClassNamePredicate)
+ * @see EventBus#allowClusterSerializable(ClassNamePredicate)
+ */
+@VertxGen
+@FunctionalInterface
+public interface ClassNamePredicate extends Predicate<String> {
```
vietj, line 30: I want to avoid defining a new interface for this, instead we should simply use a `Function<String, Boolean>`

vietj:
instead of maintaining state in the codec manager about the allowed/rejected classes, we instead use a `Function<String, Boolean` (null would mean false) that indicates whether the class is allowed or rejected, users would simply use a lambda with a switch to return true/false or use their own set and call `eventBus.theFilter(set::contains)`

vietj (changes requested):
instead of maintaining state in the codec manager about the allowed/rejected classes, we instead use a `Function<String, Boolean` (null would mean false) that indicates whether the class is allowed or rejected, users would simply use a lambda with a switch to return true/false or use their own set and call `eventBus.theFilter(set::contains)`

src/main/java/io/vertx/core/eventbus/ClassNamePredicate.java
tsegismont: Yeah, but that was really a simple oversight :smile:

src/main/java/io/vertx/core/impl/ClusterSerializableUtils.java
tsegismont: There are habits you cannot quit after reading "Enforce non-instantiability with a private constructor" :wink:

src/main/java/io/vertx/core/eventbus/ClassNamePredicate.java
tsegismont: ok

tsegismont:
@vietj proceeding with the changes right away. I will go down the function route. It is more flexible than asking users to provided a  fixed set of names to check.

src/main/java/io/vertx/core/eventbus/EventBus.java
```diff
@@ -286,10 +288,32 @@ default <T> EventBus request(String address, @Nullable Object message, DeliveryO
...
+  /**
+   * Register a predicate to invoke when verifying if an object is forbidden to be encoded/decoded as {@link io.vertx.core.shareddata.ClusterSerializable}.
+   * <p>
+   * This is only used when Vert.x is clustered.
+   *
+   * @param classNamePredicate the predicate
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  EventBus clusterSerializableCheck(Function<String, Boolean> classNamePredicate);
```
vietj, line 306: rather Checker than Check

vietj (changes requested):
Remove BigDecimal and BigInteger codecs : with this PR we don't need codecs for BigDecimal and BigInteger since they implement Serializable, users will only need to allow them in the checker.

tsegismont:
>Remove BigDecimal and BigInteger codecs : with this PR we don't need codecs for BigDecimal and BigInteger since they implement Serializable, users will only need to allow them in the checker.

They are allowed in shared data maps and totally safe to share. For consistency it is better to keep them.

I've pushed a commit for the other changes.

vietj:
_They are allowed in shared data maps and totally safe to share. For consistency it is better to keep them._ can you elaborate ? I don't get what would be consistent

tsegismont:
@vietj consistency in user experience. With the codecs for big integer and big decimal, users don't have to do anything to be able to exchange instances on the bus, whether it is clustered or not.

If we remove these codecs, users will be able to exchange instances for local messages only (because this PR introduces support for any serializable, but they are all disabled in clustered messages by default).

On the other hand, big integer and big decimal can already be shared in async maps (clustered or local) without any issue.

We could perhaps remove the two codecs and change the policy for serializable so that any class that belongs to the Java platform is allowed by default (e.g. authorize any class which name starts with the `java.` or `javax.` prefix?).

I would rather keep the two codecs and our deny all policy for serializable by default.

tsegismont:
When we'll come to an agreement on the content of the PR, I will add a commit with updates to the docs.

vietj:
"On the other hand, big integer and big decimal can already be shared in async maps (clustered or local) without any issue." because the underlying grid handles it ?

vietj:
I think we need to find a way to serailize them as Serializable without adding an extra codec for it, since that's possible.

We can have a default serializable checker (that is public static) that allows only these two types (for now) and that user can override and reference, e.g setSerializableChecker(t -> EventBus.DEFAULT_SERIALIZABLE_CHECKER.apply(t) || otherCond(t)).

That is what we do currently for HttpClient redirection handler that user can override and reference the default redirection handler `redirectHandler()` (it's not a public static but it could be).

tsegismont:
It is possible without an extra codec, at the expense of some usability and consistency across APIs.

Let's move forward, I will remove them and add documentation.

tsegismont:
@vietj done, PTAL

src/main/java/io/vertx/core/eventbus/impl/DefaultSerializableChecker.java
```diff
@@ -0,0 +1,39 @@
...
+
+  INSTANCE;
+
+  private final Set<String> classNames;
+
+  DefaultSerializableChecker() {
+    classNames = Stream.of(
+      byte[].class,
+      Number.class,
+      BigDecimal.class,
```
vietj, line 31: BigDecimal and BigInteger are both Number

src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
```diff
@@ -583,4 +587,72 @@ public void handle(Message<Object> msg) {
...
+  private <T> void testRejectedNotSent(Class<T> clazz, T message) {
+    startNodes(2);
+    vertices[0].eventBus()
+      .clusterSerializableChecker(s -> Boolean.FALSE)
+      .serializableChecker(s -> Boolean.FALSE);
+    vertices[1].eventBus().consumer("foo", msg -> fail()).completionHandler(onSuccess(reg -> {
+      try {
+        vertices[0].eventBus().send("foo", message);
+        fail();
+      } catch (IllegalArgumentException e) {
```
vietj, line 615: here we expect the error to be asynchronous and fail the returned future instead of being synchronous

src/main/asciidoc/override/eventbus.adoc
```diff
@@ -22,3 +22,20 @@ You unregister a message codec with {@link io.vertx.core.eventbus.EventBus#unreg
 
 Message codecs don't always have to encode and decode as the same type. For example you can write a codec that
 allows a MyPOJO class to be sent, but when that message is sent to a handler it arrives as a MyOtherPOJO class.
+
+Vert.x has built-in codecs for certain data types:
+
+- basic types (e.g. strings, booleans, ... etc), or
+- some Vert.x data types (buffers, JSON array, JSON objects), or
+- types implementing the {@link io.vertx.core.shareddata.ClusterSerializable} interface, or
+- types implementing the `java.io.Serializable` interface.
```
vietj, line 31: instead of "types implementing" which implies everything at first read, we should phrase with something like "some of the types" or "a configurable subset of the types implementing serializable"

we should also mention which exact types are allowed by default( byte[] and Number), etc...

src/test/java/io/vertx/core/eventbus/ClusteredEventBusTest.java
tsegismont: This is not in the scope of this PR. Failure for missing codecs have always worked like this.

src/main/asciidoc/override/eventbus.adoc
tsegismont: >instead of "types implementing" which implies everything at first read, we should phrase with something like "some of the types" or "a configurable subset of the types implementing serializable"

All types implementing these interfaces are supported by default, and then there's a note below indicating the restriction in clustered mode.

tsegismont: >we should also mention which exact types are allowed by default( byte[] and Number), etc...

done

vietj: ok

vietj (approved):
LGTM

Issue #4370, opened 2022-05-11
Timers can fail to fire due to netty emitting trigger before timer is added to timeouts


gaganis:
### Version

4.1.2 & 4.2.7

### Context
We were seeing intermittent failures in our build pipelines upon investigation we suspected that timers might not be firing.

We suspected that timers that we were setting were actually failing to fire. In our debug we could see that the timer was in the timeouts field but it would never fire.

This is how it looks like:
![img](https://user-images.githubusercontent.com/6605129/167875050-11d8a5c7-9a01-4b6f-b838-8545d4cb6e5a.png)

Upon looking at the code
```java

 InternalTimerHandler(long timerID, Handler<Long> runnable, boolean periodic, long delay, ContextInternal context) {
      this.context = context;
      this.timerID = timerID;
      this.handler = runnable;
      this.periodic = periodic;
      EventLoop el = context.nettyEventLoop();
      if (periodic) {
        future = el.scheduleAtFixedRate(this, delay, delay, TimeUnit.MILLISECONDS);
      } else {
        future = el.schedule(this, delay, TimeUnit.MILLISECONDS);
      }
    }

    @Override
    public void run() {
      context.emit(this);
    }

    public void handle(Void v) {
      if (periodic) {
        if (timeouts.containsKey(timerID)) {
          handler.handle(timerID);
        }
      } else if (timeouts.remove(timerID) != null) {
        try {
          handler.handle(timerID);
        } finally {
          // Clean up after it's fired
          context.removeCloseHook(this);
        }
      }
    }

```

The netty scheduled task is added in the constructor of the `VertxImpl.InternalTimerHandler` and when
netty fires `handle(Void v)` is run where we try to remove the timer from `VertxImpl.timeouts`.  Thus a timer is expected 
to be in the map when the netty has fired.

This creates a race with `VertxImpl.scheduleTimeout()` where the timer is added to the `timeouts` map.

### Do you have a reproducer?

I have been unable to reproduce this in a project without our proprietary code.

### Extra

I will be attaching a PR wich fixes the problem for us. In this version we add the `InternalTimeHandler` instance to the map _before_ setting the netty scheduler.

In this implementation it could be possible that we might try to cancel a timer before the netty `future` has been initialized in the constructor.
We do not think this is a problem because a timer would be cancelled by an external client by the `timerID`. This is generated and returned by
[...]

PR #4371, opened 2022-05-11
Ensure timers are put in timeouts before scheduled in netty

gaganis:
Closes: #4370

Motivation:

Please see #4370 

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/impl/VertxImpl.java
+++ b/src/main/java/io/vertx/core/impl/VertxImpl.java
@@ -515,7 +515,10 @@ public class VertxImpl implements VertxInternal, MetricsProvider {
       throw new IllegalArgumentException("Cannot schedule a timer with delay < 1 ms");
     }
     long timerId = timeoutCounter.getAndIncrement();
-    InternalTimerHandler task = new InternalTimerHandler(timerId, handler, periodic, delay, context, timeouts);
+    InternalTimerHandler task = new InternalTimerHandler(timerId, handler, periodic, delay, context);
+    timeouts.put(timerId, task);
+    task.scheduleOnNettyEventLoop();
+
     if (context.isDeployment()) {
       context.addCloseHook(task);
     }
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -871,11 +870,17 @@ private class InternalTimerHandler implements Handler<Void>, Closeable, Runnable
...
+      Handler<Long> runnable,
+      boolean periodic,
+      long delay,
+      ContextInternal context,
+      ConcurrentMap<Long, InternalTimerHandler> timeouts) {
       this.context = context;
       this.timerID = timerID;
       this.handler = runnable;
       this.periodic = periodic;
+      timeouts.put(timerID, this);
```
vietj, line 883: I think this should be left out of the constructor and we should instead move the timer scheduling outside of the constructor, to avoid passing `this` in a constructor.

vietj:
do you have a test case for this ? e.g scheduling many timers outside the event loop and making sure all of them are fired

src/main/java/io/vertx/core/impl/VertxImpl.java
gaganis: I have posted a version that moves the timer scheduling outside the constructor.

gaganis:
> do you have a test case for this ? e.g scheduling many timers outside the event loop and making sure all of them are fired

Unfortunatelly no. 

We do have a test case in our proprietary codebase where this issue is the root cause and I have used to verify this solution which I am unable to share.

vietj:
did you check those timers would be setup by non event-loop threads ?

gaganis:
We do not use threads of our own. 
But I think the first timers might be setup by the test framework main thread.  
Is this a problem? We do this a lot, to call functions on a vertx instance from other threads as an entry point to run code then with vertx. 


Sent from Yahoo Mail on Android 
 
  On Thu, May 12, 2022 at 11:40, Julien ***@***.***> wrote:   


did you check those timers would be setup by non event-loop threads ?

gaganis:
Running this test in IntelliJ IDEA reproduces the issue.
JVM: corretto-11.0.14.1
OS: MacOS 11.6.5
[reproducer-vertx-4370.zip](https://github.com/eclipse-vertx/vert.x/files/8699624/reproducer-vertx-4370.zip)

vietj:
thanks for the reproducer, I wrote one as well :-)

vietj:
my reproducer is similar to your, I will let you review the PR as well.

vietj:
@gaganis please have a look https://github.com/eclipse-vertx/vert.x/pull/4377/files

PR #4373, opened 2022-05-13
Remove some lambda expressions

geoand:
These are causing `NoSuchMethodError` on application startup
(see https://bugs.openjdk.java.net/browse/JDK-8161588 for more details)
which can have a small impact on startup performance

N.B. I only replaced the code for the instances
that were causing the issue at the startup of a
Quarkus application that uses Vert.x

### DIFF
--- a/src/main/java/io/vertx/core/impl/DeploymentManager.java
+++ b/src/main/java/io/vertx/core/impl/DeploymentManager.java
@@ -188,34 +188,37 @@ public class DeploymentManager {
         vertx.createEventLoopContext(deployment, closeFuture, workerPool, tccl));
       VerticleHolder holder = new VerticleHolder(verticle, context, workerPool, closeFuture);
       deployment.addVerticle(holder);
-      context.runOnContext(v -> {
-        try {
-          verticle.init(vertx, context);
-          Promise<Void> startPromise = context.promise();
-          Future<Void> startFuture = startPromise.future();
-          verticle.start(startPromise);
-          startFuture.onComplete(ar -> {
-            if (ar.succeeded()) {
-              if (parent != null) {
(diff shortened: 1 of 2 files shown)


franz1981 (approved):
LGTM

tsegismont:
@geoand out of curiosity, what's the Quarkus related issue?

geoand:
It's not a functionality issue per-se, it's more of a minor performance nussance.

Essentially what happens is that the as part of the way the JDK implements some lambdas and method handles, it throws a `NoSuchMethodError` (any exception throwing is not good for performance) and also records a JFR event for it (which can send users on a wild goose chace - both myself and @franz1981 have been down this road...)

geoand:
> Is there any way to understand why these two in particular are a problem

I personally have not weeded through the JDK code that leads to this. Maybe @franz1981 has?

> Do you know if this problem will happen for every call to runOnContext

Not sure. If you have samples, I can try them and we can see in practice.

geoand:
> and it affect where there's some reflective lookup in place IIUC (it's our case @geoand ?)

No reflective access in our case

vietj:
I'm not in favour of dropping lambdas, unless that is an important performance hit.

can you provide benchmark that exhibit this ?

geoand:
The performance hit is not really measurable.
It only makes a small difference when there are a lot NSME thrown (as explained in the description) at application startup.
We are trying to cut these down in Quarkus

geoand:
Just to get back to this, here is what the flamegraph for the `NoSuchMethodError` exceptions thrown during startup looks like:

flame_graph

This change would remove most of them

geoand:
I should also mention that JMC flags these exceptions as Max Severity in its analysis of the JFR recording.

PR #4377, opened 2022-05-17
Timers created outside event loop might not fire

vietj:


src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -870,17 +871,18 @@ private class InternalTimerHandler implements Handler<Void>, Closeable, Runnable
...
+    private final AtomicBoolean disposed = new AtomicBoolean();
 
-    InternalTimerHandler(long timerID, Handler<Long> runnable, boolean periodic, long delay, ContextInternal context) {
+    InternalTimerHandler(long timerID, Handler<Long> runnable, boolean periodic, long delay, TimeUnit timeUnit, ContextInternal context) {
       this.context = context;
       this.timerID = timerID;
       this.handler = runnable;
       this.periodic = periodic;
       EventLoop el = context.nettyEventLoop();
       if (periodic) {
```
gaganis, line 882: The proposed solution does ensure that the handler will fire regardless of what happens with the entry in the `timeouts` map. This resolves my concerns I had about my solution of how ti would impact cancelling of the timers. Nice!

But the root cause of the issue that the InternalTimerHandler might see a hasmap without the entry when it tries to  remove it remains in this implementation. 
In this case the handler would work as expected but we would be leaving garbage entries in the `timeouts` map that would never be removed, having a very subtle memory leak.
I would suggest to also move the setting of the netty scheduling outside the constructor in `scheduleTimeout()` after the timer has been added to `timeouts` like I have implemented in the second version of my PR: #4371

src/test/java/io/vertx/core/TimerTest.java
```diff
@@ -446,4 +449,24 @@ public void handle(Long l) {
...
+  @Test
+  public void testRaceWhenTimerCreatedOutsideEventLoop() {
+    int numThreads = 1000;
+    int numIter = 1;
+    Thread[] threads = new Thread[numThreads];
+    AtomicInteger count = new AtomicInteger(numIter * numThreads);
+    for (int i = 0;i < numThreads;i++) {
+      Thread th = new Thread(() -> {
+        // We need something more aggressive than a millisecond for this test
+        ((VertxImpl)vertx).scheduleTimeout(((VertxImpl) vertx).getOrCreateContext(), false, 1, TimeUnit.NANOSECONDS, ignore -> {
```
gaganis, line 463: Heh! that was a nice trick to make the scheduling faster than 1ms to make the racing condition more likely :)

gaganis (changes requested):

src/main/java/io/vertx/core/impl/VertxImpl.java
vietj: good catch, I actually had local changes I forgot to commit, I forced pushed with all the changes.

PTAL

gaganis (approved):
LGTM

gaganis:
I have looked at the CI failure of this and I have concluded that it is failing because that specific test is flaky. I have posted https://github.com/eclipse-vertx/vert.x/pull/4378 with a fix.

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -891,10 +893,11 @@ public void run() {
 
     public void handle(Void v) {
```
franz1981, line 894: I suppose this is going to happen in the event loop, but `cancel` can happen from any thread.
If that's true I can see this case to happen:
- thread 1 schedule a periodic task on the event loop 
- the event loop is ready to fire it the first time and `disposed::get` return `false`: GC pause the event loop before calling `handler::handle(timerID)`
- thread N try cancel the timer and win the race cancelling it
- the event loop is resumed (damn GC!) and it's calling `handler::handle` on the already closed/racing closing task

If this scenario is fine to happen, I don't need to propose an alternative solution to this
ie using an hand-rolled spin lock-like shared int that takes care of tracking the handling state (handling -> idle) in a way that a `tryCancel` can decide to spin await (or give up, returning false) and `compareAndSet(idle, disposed)`.
And any update to the concurrent hash map could/should depend by the ability of correctly setting a new state

vietj: it can happen from any thread

vietj: cancelling a timer can fail (it returns a boolean), what is not fine is to have the task executed and have cancel return true

vietj: I think the disposed.get() is OK because the timer can fire several times and cancel will always return true.

the bug was with the other path with the CAS

gaganis: For me this scenario is fine. 
I do not expect being able to cancel a timer that has fired and has started running.

For me when `disposed::get` has been called, I consider that the time to run the timer has come and that we have commited to starting it -or- that it has already started.

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -869,19 +875,14 @@ private class InternalTimerHandler implements Handler<Void>, Closeable, Runnable
     private final boolean periodic;
     private final long timerID;
     private final ContextInternal context;
-    private final java.util.concurrent.Future<?> future;
+    private java.util.concurrent.Future<?> future;
```
franz1981, line 878: This need to be volatile, probably. 
You can use a `StoreStore` barrier (ie lazySet) or a full volatile here (safer).

vietj: I don't think it is necessary, the thread storing the field will then pass the timerID to another thread that will need to lookup the object in the timeout map.

PR #4380, opened 2022-05-18
Add Netty optional native dependencies for macos dns and unix-common

zyclonite:
Motivation:

allows adding
```
<dependency>
  <groupId>io.netty</groupId>
  <artifactId>netty-resolver-dns-native-macos</artifactId>
  <classifier>osx-${os.detected.arch}</classifier>
</dependency>
```
to prevent `DnsServerAddressStreamProviders - Can not find io.netty.resolver.dns.macos.MacOSDnsServerAddressStreamProvider in the classpath, fallback to system defaults. This may result in incorrect DNS resolutions on MacOS.`

as well as
```
<dependency>
  <groupId>io.netty</groupId>
  <artifactId>netty-transport-native-unix-common</artifactId>
  <classifier>${os.detected.name}-${os.detected.arch}</classifier>
</dependency>
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,6 +123,16 @@
       <artifactId>netty-transport-native-kqueue</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-transport-native-unix-common</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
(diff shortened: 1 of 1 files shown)


zyclonite:
@vietj I just stumbled over this, would you agree this would make sense to added?

vietj:
I think this should be done with a profile

zyclonite:
all fine but why keep the native transport optionals then? should we remove these in exchange?

vietj:
native transport are necessary to compile the project, in this case we are running tests so this dependency makes only sense when running tests on MacOS

PR #4387, opened 2022-05-26
Reduce InboundBuffer pending queue's allocation pressure

franz1981:
InboundBuffer's allocation create 16 pending entries by default while it could delay it when the initial size information
is available.
This change is going to lazy allocate it: it's less elegant due to the many null checks, but it's effective to reduce allocations in the hot paths.
Subsequent allocations will make uses of the ArrayDeque enlarging strategy but starting with a much lower number (double capacity, but starting from 1 instead of 16)

### DIFF
--- a/src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
+++ b/src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
@@ -17,6 +17,7 @@ import io.vertx.core.Vertx;
 import io.vertx.core.impl.ContextInternal;
 
 import java.util.ArrayDeque;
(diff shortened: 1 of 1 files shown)


franz1981:
image

This is screen from JFR allocation recording of Techempower's plaintext, showing this as one of the top allocators

vietj:
in which case does it improve allocation ?

franz1981:
In the cases of 0 -> 15 pending elements, but the most benefit come when there are none or very few

franz1981:
@vietj ping on this bud

vietj:
thanks for the contrib @franz1981

PR #4388, opened 2022-05-27
Update expired test certificates

tsaarni:
Some of the certificates in test resources had unintentionally expired. They may cause test errors if used in tests that expects valid certificates.

* Reformatted `ssl.txt` into a shell script `ssl.sh` which can be executed  to regenerate certificates and keystores.
* Updated all test certificates with 30 year expiration period (was 3 years).

One example of expired certificate:

```console
$ keytool -list -v -keystore ./src/test/resources/tls/server-keystore.jks -storepass wibble | grep ^Valid
Valid from: Sat May 26 14:36:49 EEST 2018 until: Tue May 25 14:36:49 EEST 2021
```

After this change:

```console
$ keytool -list -v -keystore ./src/test/resources/tls/server-keystore.jks -storepass wibble | grep ^Valid
Valid from: Fri May 27 20:06:02 EEST 2022 until: Sun May 19 20:06:02 EEST 2052
```


vietj:
where did you see that certificate expired ?

tsaarni:
> where did you see that certificate expired ?

I saw it while working on #4372, when using `NewSunX509` KeyManager.  It does not select expired certificates.

By the way, motivated by this issue, I just started a small library using Bouncy Castle for creating test certificates "on the fly" during test suite execution. But you might not be looking for new (test) dependencies. https://github.com/tsaarni/java-certy

vietj:
I believe an expired cert would have made the testsuite fail

tsaarni:
It did, with the caveat I mentioned above https://github.com/eclipse-vertx/vert.x/pull/4388#issuecomment-1169533794

> I saw it while working on https://github.com/eclipse-vertx/vert.x/pull/4372, when using NewSunX509 KeyManager. It does not select expired certificates.

But currently the code does not use it, so it does not matter even though they are expired.

tsaarni:
@vietj BTW I think the tests do not currently fail since the client explicitly trusts the expired server certificate https://github.com/eclipse-vertx/vert.x/blob/488b0b972696edadb820f5cb3dce798831b789e7/src/test/java/io/vertx/core/http/Http1xTLSTest.java#L57

It will not fail even if setting CA as trust anchor:

```diff
-    testTLS(Cert.NONE, Trust.SERVER_JKS, Cert.SERVER_JKS, Trust.NONE)
+    testTLS(Cert.NONE, Trust.SERVER_JKS_ROOT_CA, Cert.SERVER_JKS_ROOT_CA, Trust.NONE)
```

because the those certs are NOT expired yet.  Comparing the two:

```console
$ keytool -list -v -keystore ./src/test/resources/tls/server-keystore.jks -storepass wibble | grep ^Valid
Valid from: Sat May 26 14:36:49 EEST 2018 until: Tue May 25 14:36:49 EEST 2021
$ keytool -list -v -keystore src/test/resources/tls/server-keystore-root-ca.jks -storepass wibble | grep ^Valid
Valid from: Sat May 26 14:36:50 EEST 2018 until: Tue May 23 14:36:50 EEST 2028
```

Issue #4389, opened 2022-06-01
FileResolverImpl tries to unpack resources from bundle URL on each call for same file resolving


gaol:
When resolving resources from bundle URL, the cache is not used at all.

In the [FileResolverImpl.unpackFromBundleURL](https://github.com/eclipse-vertx/vert.x/blob/4.2.7/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java#L350) method, the `String file = url.getHost() + File.separator + url.getFile()` is used as the file name to cache to the cache directory, but next calls of `resolve(fileName)` tries to use the `fileName` to resolve the File in cache first, so it never finds it, which leads to it tries to unpack again and again.

### Version

Vertx version:

* 4.2.7
* 4.3.2.SNAPSHOT

Java Version:

```
[🎩 lgao@lins-p1 resolver-cache]$ java -version
java version "11.0.11" 2021-04-20 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.11+9-LTS-194)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.11+9-LTS-194, mixed mode)
```

### Do you have a reproducer?

Please refer to https://github.com/gaol/tests/tree/master/vertx/resolver-cache for more detail information.

PR #4390, opened 2022-06-01
[Issue-4389] FileResolverImpl tries to unpack resources from bundle URL on each call for same file resolving

gaol:
Motivation:

Fixes: https://github.com/eclipse-vertx/vert.x/issues/4389

Conformance:

### DIFF
--- a/src/test/java/io/vertx/core/file/impl/URLBundleFileResolverTest.java
+++ b/src/test/java/io/vertx/core/file/impl/URLBundleFileResolverTest.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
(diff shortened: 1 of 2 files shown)


src/test/java/io/vertx/core/file/impl/URLBundleFileResolverTest.java
```diff
@@ -0,0 +1,42 @@
...
+public class URLBundleFileResolverTest extends JarFileResolverTest {
+
+  @Test
+  public void testResolveURLBundle() {
+    String fileName = "java/lang/Object.class";
+    assertFalse(resolver.getFileCache().getFile(fileName).exists());
+    File file = resolver.resolveFile(fileName);
+    assertTrue(file.exists());
+    // cache.getFile should return the cached file.
+    assertTrue(resolver.getFileCache().getFile(fileName).exists());
```
vietj, line 35: I think this test should be improved like this:

1/ we try to resolve and get the file
2/ we updated the file content with java.io API
3/ we resolve again and check the file content is still the same (i.e it was not overwritten)

gaol: `FileResolverImpl.enableCaching` is default to `true`, which also indicates that it won't overwrite the file in cache. so the test by checking file content won't fail anyway. (The [FileAlreadyExistsException](https://github.com/eclipse-vertx/vert.x/blob/4.2.7/src/main/java/io/vertx/core/file/impl/FileCache.java#L203) is ignored.)

PR #4401, opened 2022-06-13
Add a convenient way to create JsonObject and JsonArray

liuchong:
Motivation:

Add a convenient way to create JsonObject and JsonArray, from a list of objects. Like Map.of(...) but is from an array of objects.


src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -88,6 +88,26 @@ public JsonObject(Buffer buf) {
...
   }
 
+  /**
+   * Create a JsonObject from some mapping pairs from arguments.
+   * Ignore the last argument if the number of arguments is not even.
+   *
+   * @param args The key/value pairs.
+   */
+  public static JsonObject of(Object ...args) {
+    JsonObject obj = new JsonObject();
```
pmlopes, line 98: Maybe some validation should be performed here? If the number of elements is not even, throw `IllegalArgumentException`?

src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -88,6 +88,26 @@ public JsonObject(Buffer buf) {
...
+
+    int length = args.length;
+
+    for (int i = 0; i + 1 < length; i += 2) {
+      String k = (String) args[i];
```
pmlopes, line 103: Perhaps instead of forcing a cast, always perform a `.toString()`? This prevents `ClassCastException` and works a bit like Jackson data format extensions. (For example, in Yaml/CBOR non string keys are converted to String and no errors are thrown)

src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -92,6 +92,21 @@ public JsonArray(Buffer buf) {
...
+   * @param args The objects into JsonArray.
+   */
+  public static JsonArray of(Object... args) {
+    JsonArray arr = new JsonArray();
+
+    for (Object o : args) {
+      arr.add(o);
+    }
+
+    return arr;
```
tsegismont, line 107: How about creating a list of the right size from the beginning and then copying efficiently (system arraycopy internally) instead of iterating over the values?

```suggestion
    if (args == null) {
      throw new NullPointerException();
    }
    return new JsonArray(new ArrayList(Arrays.asList(args)));
```

src/main/java/io/vertx/core/json/JsonObject.java
tsegismont: And instead of creating immediately the `JsonObject`, how about creating a map of the right size:

```java
Map map = new HashMap(length/2);
```

Populate the map and then eventually:

```java
return new JsonObject(map);
```

tsegismont (commented):
Thank you @liuchong

src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -88,6 +88,26 @@ public JsonObject(Buffer buf) {
     }
   }
 
+  /**
+   * Create a JsonObject from some mapping pairs from arguments.
+   * Ignore the last argument if the number of arguments is not even.
+   *
+   * @param args The key/value pairs.
+   */
+  public static JsonObject of(Object ...args) {
```
vietj, line 97: Map.of has signatures like:

Map<K, V> of(K k1, V v1, K k2, V v2, K k3, V v3, K k4, V v4, K k5, V v5,
                               K k6, V v6, K k7, V v7, K k8, V v8, K k9, V v9, K k10, V v10);

in our case it would give something similar but with K=String and V=Object instead.

The benefit is that it is more type safe (for the keys) and the argument count is always even.

The drawback is to have a noisy API, but JDK did it.

Shouldn't we follow the same pattern here ?

liuchong:
Another option, maybe we can just return empty instance when args is null:

```
if (args == null || args.length == 0) {
  return new JsonObject();
}

if (args == null || args.length == 0) {
  return new JsonArray();
}
```

When do you think, `return empty instance` or `throws NullPointerException` is better?

src/main/java/io/vertx/core/json/JsonObject.java
vietj: I think we should stick to what JDK do here, if somebody has more than this number of arguments then he should use the other way (i.e create an empty json object and add everything)

liuchong: Ok, I could change that if so. Below is some other information about it before I do make the change.

I thought again about it, and then I remembered another reason of why I didn't use the JDK `Map.of` style.
The Map<K,V> has a `K,V` type but our JsonObject does not like that. Here the `K` is always `String` and `V` is always uncertain and finally be a `Object`. The benefits of the noisy style is less than the `Map.of`.

In fact `Map.of` will create a `Map1` or `MapN`, and the constructor of `MapN` is `MapN(Object... input)` style:
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/ImmutableCollections.java#L1172

In summary, we could have options of the style:
1. current `JsonObject.of(Object... args)` style
2. Map.of(K k1, V v1...) style, like `JsonObject.of(String k1, Object, v1, String v2, Object, v2...)`
3. Or just remove the `of`, create another constructor JsonObject(Object... args)

Which do you prefer? @vietj 
Also looking forward to your suggestions @pmlopes @tsegismont

vietj: I think we should use 2. 

I think we want to avoid constructors since we already have the JsonObject(Map) that is often misleading ppl and we should discourage using it.

liuchong: Ok, let me do that then, thanks!

liuchong:
Just finished a version of noisy API as JDK pattern, please review 👀 
@vietj @pmlopes @tsegismont

src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -92,6 +92,238 @@ public JsonArray(Buffer buf) {
...
+  }
+
+  /** Create a JsonArray containing three values.
+   *
+   * @param v1 the first value
+   * @param v2 the second value
+   * @param v3 the third value
+   * @return a JsonArray containing the specified values.
+   */
+  public static JsonArray of(Object v1, Object v2, Object v3) {
```
vietj, line 138: I think for this we can use varrags, can't we ?

liuchong: yes, JsonArray does not need a `noisy API`, I changed it yesterday and just pushed 👀

liuchong: Another thing is for JsonObject maybe we could do this:
```
  protected JsonObject(String k1, Object v1, Object... more) {

    // implicit nullcheck of args
    if (more.length == 0) {
      map = new LinkedHashMap<>(1);
      put(k1, v1);
      return;
    }

    if ((more.length & 1) != 0) {
      throw new IllegalArgumentException("The more args length is odd");
    }

    map = new LinkedHashMap<>(1 + (more.length >> 2));
    put(k1, v1);
    for (int i = 0; i + 1 < more.length; i += 2) {
      put(more[i].toString(), more[i + 1]);
    }
  }

...

  public static JsonObject of(String k1, Object v1, String k2, Object v2) {
    return new JsonObject(k1, v1, k2, v2);
  }

  public static JsonObject of(String k1, Object v1, String k2, Object v2, String k3, Object v3) {
    return new JsonObject(k1, v1, k2, v2, k3, v3);
  }

... upto 15 or 20
```

liuchong: Please check @vietj 👀

src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -92,6 +92,20 @@ public JsonArray(Buffer buf) {
...
+   *
+   * @param values The objects into JsonArray.
+   * @throws NullPointerException if the args is null.
+   */
+  public static JsonArray of(Object... values) {
+    // implicit nullcheck of args
+    if (args.length == 0) {
+      return new JsonArray();
+    }
+    return new JsonArray(new ArrayList<>(Arrays.asList(values)));
```
vietj, line 106: we should avoid Arrays.asList() as it creates an intermediary list that is not necessary, instead we should add the values to the created JsonArray like we do for JsonObject

liuchong: Ok, fixed @vietj

vietj:
thanks for the contribution @liuchong

liuchong:
Thanks 🎉

PR #4410, opened 2022-06-20
Support reading PEM file containing EC Private Key

sophokles73:
vert.x currently does not support reading PEM files that contain a EC private key.
However, ECC based keys are much smaller and easier to process than RSA based keys and therefore
are getting used more and more often, in particular when implementing TLS secured network servers that
need to be accessed by constrained clients (e.g. IoT devices).

The KeyStoreHelper has therefore been extended to support reading EC private keys from PEM files
as described in https://datatracker.ietf.org/doc/html/rfc5915#section-4

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/pkcs1/PrivateKeyParser.java
+++ b/src/main/java/io/vertx/core/net/impl/pkcs1/PrivateKeyParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -11,14 +11,19 @@
 
 package io.vertx.core.net.impl.pkcs1;
 
-import io.vertx.core.VertxException;
-import io.vertx.core.buffer.Buffer;
-
 import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
-import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.GeneralSecurityException;
+import java.security.KeyPairGenerator;
+import java.security.interfaces.ECPublicKey;
+import java.security.spec.ECGenParameterSpec;
+import java.security.spec.ECParameterSpec;
+import java.security.spec.ECPrivateKeySpec;
 import java.security.spec.RSAPrivateCrtKeySpec;
 import java.util.Arrays;
+
+import io.vertx.core.VertxException;
+import io.vertx.core.buffer.Buffer;
 
 /**
  * This code is copies and modifies over from net.oauth.java.jmeter:ApacheJMeter_oauth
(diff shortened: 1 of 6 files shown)


pmlopes (commented):
I think this is a good addition, I just had a small remark to add extra validation. @vietj feel free to approve this if you don't think it's really required.

src/test/java/io/vertx/core/net/KeyStoreHelperTest.java
```diff
@@ -70,6 +68,23 @@ public void testKeyStoreHelperSupportsECPrivateKeys() throws Exception {
...
+   * from a PEM file.
+   *
+   * @throws Exception if the key cannot be read.
+   */
+  @Test
+  public void testKeyStoreHelperSupportsPKCS1ECPrivateKey() throws Exception {
+
+    Assume.assumeTrue("ECC is not supported by VM's security providers", isECCSupportedByVM());
+    PemKeyCertOptions options = new PemKeyCertOptions()
+            .addKeyPath("target/test-classes/tls/server-key-ec-pkcs1.pem")
```
pmlopes, line 82: Shouldn't this refer to the class path location instead? I assume that resources are loaded through vert.x `FileSystem` right? In that case: `tls/server-key-ec-pkcs1.pem` should suffice.

src/main/java/io/vertx/core/net/impl/pkcs1/PrivateKeyParser.java
```diff
@@ -79,6 +117,59 @@ public static String getPKCS8EncodedKeyAlgorithm(byte[] encodedKey) {
...
+
+    Asn1Object sequence = parser.read();
+    if (sequence.getType() != DerParser.SEQUENCE) {
+      throw new VertxException("Invalid DER: not a sequence");
+    }
+
+    // Parse inside the sequence
+    parser = sequence.getParser();
+
+    parser.read(); // Skip version
```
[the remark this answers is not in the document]
sophokles73: My understanding is that the DERParser.read() method being invoked will throw a VertxException if it cannot read an element from the byte array. Thus, FMPOV we already implicitly check that the sequence has the three expected elements ...

I am happy to check the version field value, though.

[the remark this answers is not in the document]
vietj: I think better error reporting for the user should be favored so a version check is better

src/test/java/io/vertx/core/net/KeyStoreHelperTest.java
vietj: yes it should be like that

vietj:
can you fix those @sophokles73 so it can be part of 4.3.2 ?

sophokles73:
@vietj @pmlopes 
I have pushed corresponding changes. I have also added a check so that the ECC based tests are only run if the VM actually supports ECC ...

sophokles73:
@vietj is it possible to also add this to the 4.2.x branch? I would like this to be available in Quarkus soon but it looks like they will stick to 4.2.x for some time to come...

src/test/java/io/vertx/test/core/TestUtils.java
```diff
@@ -512,4 +517,18 @@ public static TestLoggerFactory testLogging(Runnable runnable) {
...
+   * Checks if the JVM supports ECC algorithms.
+   *
+   * @return {@code true} if the JVM supports ECC.
+   */
+  public static boolean isECCSupportedByVM() {
+    try {
+      KeyFactory.getInstance("EC");
+      return true;
+    } catch (GeneralSecurityException e) {
+        return false;
```
vietj, line 531: wrong indent

sophokles73:
I have rebased to HEAD of master, don't know if that causes the tests to fail?

vietj:
@sophokles73 a few tests have intermitent failures in CI, we  are working on improving this constantly. I reran the tests and now it's fine

vietj:
thanks for the contribution @sophokles73

PR #4424, opened 2022-06-29
Introduced abstraction layer to base64 encoding

dominik-cnx:
Motivation:

Refer to the discussion in

https://github.com/eclipse-vertx/vert.x/pull/3123

The purpose of this PR is to get feedback on the idea to use 

org.apache.commons.codec.binary.Base64

to decode. The advantage is that it supports both, the "legacy" and "new" mode and thus makes transition to the "new" mode possible in certain scenarios.

The usage of the commons Base64 decoder is optional and can be configured through a system property. 

I realize that https://github.com/eclipse-vertx/vertx-codegen and a number of other projects will need to be adapted to make this work. I'm happy do so and hope to be able to figure out how to actually achieve this.

Note that I wasn't able to run the tests in my environment, yet to find out if the build will pass on the build server.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,7 +50,7 @@
     <assertj.version>3.4.1</assertj.version>
     <apacheds-protocol-dns.version>1.5.7</apacheds-protocol-dns.version>
     <generated.dir>${project.basedir}/src/main/generated</generated.dir>
-    <stack.version>4.3.2-SNAPSHOT</stack.version>
+    <stack.version>4.3.1</stack.version>
     <jmh.version>1.19</jmh.version>
     <vertx.testNativeTransport>false</vertx.testNativeTransport>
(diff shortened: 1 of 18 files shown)


vietj:
it is not clear why this is needed, can you elaborate ?

dominik-cnx:
I tried to explain it in here 

https://github.com/eclipse-vertx/vert.x/pull/3123

The relevant section  is
----
... we have a high number of clients that we cannot migrate at once. Even if we could, we would still have to deal with data in persistent queues that is encoded in the old mode for some time (probably a few weeks in our case).

Now, to avoid issues in the future (the legacy mode is possibly dropped?) we aim to switch to the new format (base64url), we are not in a hurry though.
---

Does this make it clear?

dominik-cnx:
@vietj do you need further explanation? Does it make sense or not really?

PR #4428, opened 2022-07-01
add TCP_USER_TIMEOUT config to EpollTransport

rbsrafa:
Motivation:

When using Vertx HttpClient with HTTP2 configured, the client gets stuck in an established TCP 
connection in case the FIN (finish) and RST (reset) packets are not transmitted/lost
from the server. 

Please see more at: [rbsrafa/test-vertx](https://github.com/rbsrafa/test-vertx) readme.md

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/net/TCPSSLOptions.java
+++ b/src/main/java/io/vertx/core/net/TCPSSLOptions.java
@@ -103,9 +103,11 @@ public abstract class TCPSSLOptions extends NetworkOptions {
   public static final boolean DEFAULT_TCP_QUICKACK = false;
 
   /**
-   * The default TCP_USER_TIMEOUT value in milliseconds = null
+   * The default TCP_USER_TIMEOUT value in milliseconds = 0
+   * <p/>
+   * When the default value of 0 is used, TCP will use the system default.
    */
-  public static final Integer DEFAULT_TCP_USER_TIMEOUT = null;
+  public static final int DEFAULT_TCP_USER_TIMEOUT = 0;
 
   /**
    * The default value of SSL handshake timeout = 10
@@ -138,7 +140,7 @@ public abstract class TCPSSLOptions extends NetworkOptions {
   private boolean tcpFastOpen;
   private boolean tcpCork;
   private boolean tcpQuickAck;
-  private Integer tcpUserTimeout;
+  private int tcpUserTimeout;
 
   /**
    * Default constructor
@@ -760,7 +762,7 @@ public abstract class TCPSSLOptions extends NetworkOptions {
    *
    * @return the {@code TCP_USER_TIMEOUT} value
    */
-  public Integer getTcpUserTimeout() {
+  public int getTcpUserTimeout() {
     return tcpUserTimeout;
   }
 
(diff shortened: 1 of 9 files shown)


src/main/java/io/vertx/core/net/TCPSSLOptions.java
```diff
@@ -102,6 +102,11 @@ public abstract class TCPSSLOptions extends NetworkOptions {
    */
   public static final boolean DEFAULT_TCP_QUICKACK = false;
 
+  /**
+   * The default TCP_USER_TIMEOUT value in milliseconds = null
+   */
+  public static final Integer DEFAULT_TCP_USER_TIMEOUT = null;
```
tsegismont, line 108: For consistency, can you please change this to `int` and use the default value of `0`?

tsegismont (changes requested):
@rbsrafa thank you for providing a PR. Small change required.

src/main/java/io/vertx/core/net/TCPSSLOptions.java
tsegismont: As you change it, can you please verify your git configuration for author and user emails? Make sure to sign off the commit and that you signed the Eclipse Contributor Agreement using the same address.

rbsrafa: Hi @tsegismont, thanks for the quick feedback. I'll do the necessary changes.

rbsrafa: Hi @tsegismont, I've fixed the default value to '0' and signed both commits and Eclipse Contributor Agreement. Thanks.

vietj: I think that when the value is > 0 then we should warn on the console that the user timeout will not be honoured when it is not epoll

rbsrafa: Hi @tsegismont and @vietj. Could you point me to where should I add this warning please? Also there are other options that would only work with Linux native: SO_REUSEPORT, TCP_QUICKACK, TCP_CORK and TCP_FASTOPEN, should I also add a warning for them?

I noticed that the tests for Ubuntu-latest, 8, with profile `testDomainSockets` has failed, I'm not sure why, as I run it in my local ubuntu and they passed.

Thanks.

tsegismont: To me it's fine without a warning because we don't do it for other props that are applied to native setup only.

It could be addressed in a follow-up PR if that's an important concern.

tsegismont: I've restarted the failed jobs, I don't believe the failure is related.

tsegismont (approved):

tsegismont:
I'll take care of this and the warning in a follow-up PR tomorrow

Le mar. 5 juil. 2022 à 22:41, Julien Viet ***@***.***> a
écrit :

src/main/java/io/vertx/core/net/TCPSSLOptions.java
rbsrafa: Thanks guys, I've added this info in the TCPSSLOptions setter: https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/net/TCPSSLOptions.java#L770

Please let me know if another PR is required for the javadocs.

Cheers.

PR #4430, opened 2022-07-06
Improve fairness of SimpleConnectionPool by yielding its combiner executor

franz1981:
Motivation: 

`SimpleConnectionPool` is backed by a `CombinerExecutor` that can be caught in a tight (potentially endless) loop while consuming (and executing) actions, occupying its current running event loop for too long.

This PR add both duration-based and max action count based yield strategies with some opportunistic action resume ie
if a resume continuation is emitted, any concurrent "lucky" resume/submit can still consume the existing actions backlog.

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
+++ b/src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
@@ -12,61 +12,149 @@ package io.vertx.core.net.impl.pool;
 
 import io.netty.util.internal.PlatformDependent;
 
+import java.time.Duration;
+import java.util.Objects;
 import java.util.Queue;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.LongSupplier;
 
 /**
  * Lock free executor.
- *
+ * <p>
  * When a thread submits an action, it will enqueue the action to execute and then try to acquire
  * a lock. When the lock is acquired it will execute all the tasks in the queue until empty and then
  * release the lock.
  */
 public class CombinerExecutor<S> implements Executor<S> {
+  @FunctionalInterface
+  interface YieldCondition {
+    boolean yield(long actions);
+
+    static YieldCondition withMaxDuration(final Duration duration) {
+      return withMaxDuration(duration, System::nanoTime);
+    }
+
+    static YieldCondition withMaxDuration(final Duration duration, final LongSupplier nanoTimeSupplier) {
+      return new YieldCondition() {
+        private final long durationInNs = Objects.requireNonNull(duration).toNanos();
+        private long start;
+
+        @Override
+        public boolean yield(long actions) {
+          assert actions >= 0;
+          if (actions == 0) {
+            start = nanoTimeSupplier.getAsLong();
+            return false;
+          }
+          final long elapsed = nanoTimeSupplier.getAsLong() - start;
+          assert elapsed >= 0;
+          return elapsed >= durationInNs;
+        }
+      };
+    }
+
+    static YieldCondition withMaxCount(final int maxActions) {
+      if (maxActions < 0) {
+        throw new IllegalArgumentException("maxActions must be >= 1");
+      }
+      return count -> count >= maxActions;
+    }
+  }
 
   private final Queue<Action<S>> q = PlatformDependent.newMpscQueue();
   private final AtomicInteger s = new AtomicInteger();
   private final S state;
 
-  public CombinerExecutor(S state) {
+  private final YieldCondition yieldCondition;
+
+  private final Continuation resume;
+
+  private final AtomicBoolean inflightContinuation = new AtomicBoolean();
+
+  public CombinerExecutor(final S state) {
+    this(state, null);
+  }
+
+  public CombinerExecutor(final S state, final YieldCondition yieldCondition) {
     this.state = state;
+    this.resume = () -> {
+      if (!inflightContinuation.compareAndSet(true, false)) {
+        throw new IllegalStateException("Continuation cannot be resumed twice or without a prior yield");
+      }
+      return pollAndExecute();
+    };
+    this.yieldCondition = yieldCondition;
+  }
+  public int actions() {
+    return q.size();
   }
 
   @Override
-  public void submit(Action<S> action) {
+  public Continuation submitAndContinue(final Action<S> action) {
+    Objects.requireNonNull(action);
     q.add(action);
-    if (s.get() != 0 || !s.compareAndSet(0, 1)) {
-      return;
-    }
-    Task head = null;
-    do {
-      try {
-        head = pollAndExecute(head);
-      } finally {
-        s.set(0);
-      }
-    } while (!q.isEmpty() && s.compareAndSet(0, 1));
-    while (head != null) {
-      head.run();
-      head = head.next;
-    }
+    return pollAndExecute();
   }
 
-  private Task pollAndExecute(Task head) {
-    Action<S> a;
-    while ((a = q.poll()) != null) {
-      Task action = a.execute(state);
-      if (action != null) {
-        if (head == null) {
-          head = action;
-          head.prev = head;
-        } else {
-          action.prev = head.prev;
-          head.prev.next = action;
-          head.prev = action;
-        }
-      }
+  private Continuation pollAndExecute() {
+    if (s.get() != 0 || !s.compareAndSet(0, 1)) {
+      return null;
     }
-    return head;
+    final YieldCondition condition = this.yieldCondition;
+    Task head = null;
+    Task tail = null;
+    Continuation resume = null;
+    long actions = 0;
+    do {
+      // single threaded
+      boolean requiresResume = false;
+      try {
+        while (true) {
+          final Action<S> a = q.poll();
+          if (a == null) {
+            break;
+          }
+          if (condition != null && actions == 0) {
+            // we can ignore the value here
+            condition.yield(0);
+          }
+          Task task = a.execute(state);
+          if (task != null) {
+            if (head == null) {
+              assert tail == null;
+              tail = task;
+              head = task;
+            } else {
+              tail = tail.next(task);
+            }
+          }
+          actions++;
+          if (condition != null && condition.yield(actions)) {
+            requiresResume = true;
+            break;
+          }
+        }
+      } finally {
+        s.set(0);
+        // no longer single threaded
+      }
+      // requiresResume == true doesn't mean 100% chances to resume: if others has already picked up
+      // the action backlog or there wasn't any new actions already, there's no need to resume
+      if (q.isEmpty()) {
+        break;
+      }
+      if (requiresResume) {
+        if (inflightContinuation.compareAndSet(false, true)) {
+          resume = this.resume;
+        }
+        break;
+      }
+    } while (s.compareAndSet(0, 1));
+    if (head != null) {
+      head.runNextTasks();
+    }
+    return resume;
   }
+
 }
--- a/src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
+++ b/src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
@@ -18,6 +18,7 @@ import io.vertx.core.http.ConnectionPoolTooBusyException;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.impl.EventLoopContext;
 
+import java.time.Duration;
 import java.util.AbstractList;
 import java.util.ArrayList;
 import java.util.Iterator;
@@ -26,6 +27,7 @@ import java.util.NoSuchElementException;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 
 /**
  * <p> The pool is a state machine that maintains a queue of waiters and a list of connections.
@@ -78,6 +80,11 @@ import java.util.function.Predicate;
  * a request.
  */
 public class SimpleConnectionPool<C> implements ConnectionPool<C> {
+
+  private static final long DEFAULT_EXECUTE_ACTIONS_TIMEOUT_MS = Integer.getInteger("combiner.execute.timeout.ms", 100);
+  private static final Supplier<CombinerExecutor.YieldCondition> DEFAULT_YIELD_CONDITION_FACTORY = DEFAULT_EXECUTE_ACTIONS_TIMEOUT_MS <= 0 ?
+    () -> null :
+    () -> CombinerExecutor.YieldCondition.withMaxDuration(Duration.ofMillis(DEFAULT_EXECUTE_ACTIONS_TIMEOUT_MS));
 
   private static final Future POOL_CLOSED = Future.failedFuture("Pool closed");
 
@@ -194,11 +201,14 @@ public class SimpleConnectionPool<C> implements ConnectionPool<C> {
   private final Waiters<C> waiters;
   private int requests;
 
-  SimpleConnectionPool(PoolConnector<C> connector, int[] maxSizes) {
-    this(connector, maxSizes, -1);
+  private final java.util.concurrent.Executor resumeExecutor;
+
+  SimpleConnectionPool(PoolConnector<C> connector, int[] maxSizes, java.util.concurrent.Executor resumeExecutor) {
+    this(connector, maxSizes, -1, resumeExecutor, null);
   }
 
-  SimpleConnectionPool(PoolConnector<C> connector, int[] maxSizes, int maxWaiters) {
+  SimpleConnectionPool(PoolConnector<C> connector, int[] maxSizes, int maxWaiters,
+                       java.util.concurrent.Executor resumeExecutor, CombinerExecutor.YieldCondition yieldCondition) {
 
     int[] capacities = new int[maxSizes.length];
     int maxCapacity = 1;
@@ -222,11 +232,23 @@ public class SimpleConnectionPool<C> implements ConnectionPool<C> {
     this.maxWaiters = maxWaiters;
     this.capacity = 0;
     this.maxCapacity = maxCapacity;
-    this.sync = new CombinerExecutor<>(this);
+    this.sync = new CombinerExecutor<>(this, validYieldCondition(resumeExecutor, yieldCondition));
     this.selector = (BiFunction) SAME_EVENT_LOOP_SELECTOR;
     this.fallbackSelector = (BiFunction) FIRST_AVAILABLE_SELECTOR;
     this.contextProvider = EVENT_LOOP_CONTEXT_PROVIDER;
     this.waiters = new Waiters<>();
+    this.resumeExecutor = resumeExecutor;
+  }
+
+  private static CombinerExecutor.YieldCondition validYieldCondition(java.util.concurrent.Executor resumeExecutor,
+                                                                     CombinerExecutor.YieldCondition yieldCondition) {
+    if (resumeExecutor == null) {
+      return null;
+    }
+    if (yieldCondition != null) {
+      return yieldCondition;
+    }
+    return DEFAULT_YIELD_CONDITION_FACTORY.get();
   }
 
   @Override
@@ -242,7 +264,27 @@ public class SimpleConnectionPool<C> implements ConnectionPool<C> {
   }
 
   private void execute(Executor.Action<SimpleConnectionPool<C>> action) {
-    sync.submit(action);
+    if (resumeExecutor == null) {
+      sync.submit(action);
+    } else {
+      final Executor.Continuation continuation = sync.submitAndContinue(action);
+      if (continuation != null) {
+        // single threaded
+        resumeExecutor.execute(() -> resumeAndContinue(continuation));
+      }
+    }
+  }
+
+  private void resumeAndContinue(final Executor.Continuation continuation) {
+    assert continuation != null;
+    assert resumeExecutor != null;
+    //single threaded
+    final Executor.Continuation furtherContinue = continuation.resume();
+    if (furtherContinue == null) {
+      return;
+    }
+    //single threaded
+    resumeExecutor.execute(() -> resumeAndContinue(furtherContinue));
   }
 
   public int size() {
@@ -860,9 +902,12 @@ public class SimpleConnectionPool<C> implements ConnectionPool<C> {
 
     List<PoolWaiter<C>> clear() {
       List<PoolWaiter<C>> lst = new ArrayList<>(size);
-      this.forEach(lst::add);
-      size = 0;
-      head.next = head.prev = head;
+      PoolWaiter<C> waiter;
+      // polling implies removing -> save from GC nepotism
+      while ((waiter = poll()) != null) {
+        lst.add(waiter);
+      }
+      assert size == 0;
       return lst;
     }
 
--- a/src/test/benchmarks/io/vertx/benchmarks/CombinerExecutorBenchmark.java
+++ b/src/test/benchmarks/io/vertx/benchmarks/CombinerExecutorBenchmark.java
@@ -35,7 +35,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 @State(Scope.Benchmark)
 @Warmup(iterations = 20, time = 200, timeUnit = TimeUnit.MILLISECONDS)
 @Measurement(iterations = 10, time = 200, timeUnit = MILLISECONDS)
-@Threads(2)
+@Threads(8)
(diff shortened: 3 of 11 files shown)


franz1981:
Current logic is not enforcing an issued resume to NOT be emitted again if the previous one has succeeded (or not) - that means that a "lucky" concurrent submit after a previous resume has been emitted, but not yet executed, can consume the backlong and emit/not emit any continuation "again" capturing the current state of backlong, instead of the current state of scheduled resume(s).

The main purpose of yield is to guarantee that a single submit cannot hog a single event loop for too much time (to be defined); but it doesn't mean that subsequent ones (on the same event loops/other event loops) cannot keep on executing actions, making some progress.


Conversely, the "yield" state can be turned in a proper combiner state that only a resume operation is allowed to unblock - meaning that any submit before it won't be allowed to pick-up the action backlog (maybe causing it to become too large!): these are 2 different ways to solve the same problem, each one with their own pros/cons.

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
```diff
@@ -12,61 +12,149 @@
...
  * Lock free executor.
- *
+ * <p>
  * When a thread submits an action, it will enqueue the action to execute and then try to acquire
  * a lock. When the lock is acquired it will execute all the tasks in the queue until empty and then
  * release the lock.
  */
 public class CombinerExecutor<S> implements Executor<S> {
+  @FunctionalInterface
+  public interface YieldCondition {
```
vietj, line 31: for now as this is only used in the same package, make this package private

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
```diff
@@ -12,61 +12,149 @@
...
+      return count -> count >= maxActions;
+    }
+  }
 
   private final Queue<Action<S>> q = PlatformDependent.newMpscQueue();
   private final AtomicInteger s = new AtomicInteger();
   private final S state;
 
-  public CombinerExecutor(S state) {
+  private final YieldCondition yieldCondition;
```
vietj, line 69: I think the yieldCondition should be an argument of the submitAndContinue instead of being a field.

franz1981: Good idea, less states inside the executor

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
franz1981: But if it becomes part of the submitAndContinue API, will need it to be public, as `Continuation` is

src/test/benchmarks/io/vertx/benchmarks/CombinerExecutorBenchmark.java
```diff
@@ -39,39 +46,49 @@
...
+  public static class ActionStatistics {
+
+    public long submitted;
+    public long completed;
+
+    private Executor.Action<Object> action;
+
+    @Setup
+    public void clean(CombinerExecutorBenchmark benchmark, Blackhole bh) {
+      submitted = completed = 0;
```
franz1981, line 76: Sadly `completed` cannot be a plain field, because will be read by the submitter thread (that's not by any mean the executor!!!) - but given that actions are accessed in serialized fashion I can just use setOpaque/lazySet in order to keep it lightweight

vietj: actually having it global is better so it applies consistently to all actions and we don't need to deal with different policiies submitted by different threads

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
vietj: I think it's fine keeping it like this, so we don't need to deal with potentially different conditions that would be randomly used

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
vietj: for now this API is only for SimpleConnectionPool. It is decoupled from the connection pool only for purpose of testing and benchmarking.

franz1981:
@vietj I've changed the benchmark and something that requires investigation has happened:
```
Benchmark                                               (actionWork)    (type)   Mode  Cnt   Score    Error   Units
CombinerExecutorBenchmark.submitFourThreads                        0      sync  thrpt   20  23.363 ±  1.548  ops/us
CombinerExecutorBenchmark.submitFourThreads:completed              0      sync  thrpt   20  23.365 ±  1.548  ops/us
CombinerExecutorBenchmark.submitFourThreads:submitted              0      sync  thrpt   20  23.365 ±  1.548  ops/us
CombinerExecutorBenchmark.submitFourThreads                        0  combiner  thrpt   20   9.419 ±  0.455  ops/us
CombinerExecutorBenchmark.submitFourThreads:completed              0  combiner  thrpt   20   9.423 ±  0.455  ops/us
CombinerExecutorBenchmark.submitFourThreads:submitted              0  combiner  thrpt   20   9.423 ±  0.455  ops/us
CombinerExecutorBenchmark.submitFourThreads                        1      sync  thrpt   20  23.940 ±  0.486  ops/us
CombinerExecutorBenchmark.submitFourThreads:completed              1      sync  thrpt   20  23.944 ±  0.486  ops/us
CombinerExecutorBenchmark.submitFourThreads:submitted              1      sync  thrpt   20  23.944 ±  0.486  ops/us
CombinerExecutorBenchmark.submitFourThreads                        1  combiner  thrpt   20   8.640 ±  0.578  ops/us
CombinerExecutorBenchmark.submitFourThreads:completed              1  combiner  thrpt   20   8.646 ±  0.578  ops/us
CombinerExecutorBenchmark.submitFourThreads:submitted              1  combiner  thrpt   20   8.646 ±  0.578  ops/us
CombinerExecutorBenchmark.submitFourThreads                       10      sync  thrpt   20  18.615 ±  0.569  ops/us
CombinerExecutorBenchmark.submitFourThreads:completed             10      sync  thrpt   20  18.617 ±  0.569  ops/us
CombinerExecutorBenchmark.submitFourThreads:submitted             10      sync  thrpt   20  18.617 ±  0.569  ops/us
CombinerExecutorBenchmark.submitFourThreads                       10  combiner  thrpt   20   9.274 ±  0.644  ops/us
CombinerExecutorBenchmark.submitFourThreads:completed             10  combiner  thrpt   20   9.279 ±  0.644  ops/us
CombinerExecutorBenchmark.submitFourThreads:submitted             10  combiner  thrpt   20   9.279 ±  0.644  ops/us
CombinerExecutorBenchmark.submitFourThreads                       20      sync  thrpt   20  12.842 ±  0.643  ops/us
CombinerExecutorBenchmark.submitFourThreads:completed             20      sync  thrpt   20  12.843 ±  0.643  ops/us
CombinerExecutorBenchmark.submitFourThreads:submitted             20      sync  thrpt   20  12.843 ±  0.643  ops/us
CombinerExecutorBenchmark.submitFourThreads                       20  combiner  thrpt   20   8.704 ±  0.127  ops/us
CombinerExecutorBenchmark.submitFourThreads:completed             20  combiner  thrpt   20   8.708 ±  0.125  ops/us
CombinerExecutorBenchmark.submitFourThreads:submitted             20  combiner  thrpt   20   8.708 ±  0.125  ops/us
CombinerExecutorBenchmark.submitOneThread                          0      sync  thrpt   20  32.403 ±  0.239  ops/us
CombinerExecutorBenchmark.submitOneThread:completed                0      sync  thrpt   20  32.403 ±  0.239  ops/us
CombinerExecutorBenchmark.submitOneThread:submitted                0      sync  thrpt   20  32.403 ±  0.239  ops/us
CombinerExecutorBenchmark.submitOneThread                          0  combiner  thrpt   20  23.212 ±  0.205  ops/us
CombinerExecutorBenchmark.submitOneThread:completed                0  combiner  thrpt   20  23.212 ±  0.205  ops/us
CombinerExecutorBenchmark.submitOneThread:submitted                0  combiner  thrpt   20  23.212 ±  0.205  ops/us
CombinerExecutorBenchmark.submitOneThread                          1      sync  thrpt   20  32.159 ±  0.906  ops/us
CombinerExecutorBenchmark.submitOneThread:completed                1      sync  thrpt   20  32.159 ±  0.906  ops/us
CombinerExecutorBenchmark.submitOneThread:submitted                1      sync  thrpt   20  32.159 ±  0.906  ops/us
CombinerExecutorBenchmark.submitOneThread                          1  combiner  thrpt   20  22.781 ±  0.127  ops/us
CombinerExecutorBenchmark.submitOneThread:completed                1  combiner  thrpt   20  22.781 ±  0.127  ops/us
CombinerExecutorBenchmark.submitOneThread:submitted                1  combiner  thrpt   20  22.781 ±  0.127  ops/us
CombinerExecutorBenchmark.submitOneThread                         10      sync  thrpt   20  25.222 ±  0.251  ops/us
CombinerExecutorBenchmark.submitOneThread:completed               10      sync  thrpt   20  25.222 ±  0.251  ops/us
CombinerExecutorBenchmark.submitOneThread:submitted               10      sync  thrpt   20  25.222 ±  0.251  ops/us
CombinerExecutorBenchmark.submitOneThread                         10  combiner  thrpt   20  18.778 ±  0.257  ops/us
CombinerExecutorBenchmark.submitOneThread:completed               10  combiner  thrpt   20  18.778 ±  0.257  ops/us
CombinerExecutorBenchmark.submitOneThread:submitted               10  combiner  thrpt   20  18.778 ±  0.257  ops/us
CombinerExecutorBenchmark.submitOneThread                         20      sync  thrpt   20  17.348 ±  0.622  ops/us
CombinerExecutorBenchmark.submitOneThread:completed               20      sync  thrpt   20  17.348 ±  0.622  ops/us
CombinerExecutorBenchmark.submitOneThread:submitted               20      sync  thrpt   20  17.348 ±  0.622  ops/us
CombinerExecutorBenchmark.submitOneThread                         20  combiner  thrpt   20  14.599 ±  0.064  ops/us
CombinerExecutorBenchmark.submitOneThread:completed               20  combiner  thrpt   20  14.599 ±  0.064  ops/us
CombinerExecutorBenchmark.submitOneThread:submitted               20  combiner  thrpt   20  14.599 ±  0.064  ops/us
CombinerExecutorBenchmark.submitThreeThreads                       0      sync  thrpt   20  23.608 ±  1.872  ops/us
CombinerExecutorBenchmark.submitThreeThreads:completed             0      sync  thrpt   20  23.609 ±  1.872  ops/us
CombinerExecutorBenchmark.submitThreeThreads:submitted             0      sync  thrpt   20  23.609 ±  1.872  ops/us
CombinerExecutorBenchmark.submitThreeThreads                       0  combiner  thrpt   20  11.596 ±  0.978  ops/us
CombinerExecutorBenchmark.submitThreeThreads:completed             0  combiner  thrpt   20  11.599 ±  0.978  ops/us
CombinerExecutorBenchmark.submitThreeThreads:submitted             0  combiner  thrpt   20  11.599 ±  0.978  ops/us
CombinerExecutorBenchmark.submitThreeThreads                       1      sync  thrpt   20  22.583 ±  1.232  ops/us
CombinerExecutorBenchmark.submitThreeThreads:completed             1      sync  thrpt   20  22.585 ±  1.232  ops/us
CombinerExecutorBenchmark.submitThreeThreads:submitted             1      sync  thrpt   20  22.585 ±  1.232  ops/us
CombinerExecutorBenchmark.submitThreeThreads                       1  combiner  thrpt   20  11.525 ±  0.867  ops/us
CombinerExecutorBenchmark.submitThreeThreads:completed             1  combiner  thrpt   20  11.529 ±  0.867  ops/us
CombinerExecutorBenchmark.submitThreeThreads:submitted             1  combiner  thrpt   20  11.529 ±  0.867  ops/us
CombinerExecutorBenchmark.submitThreeThreads                      10      sync  thrpt   20  20.801 ±  0.873  ops/us
CombinerExecutorBenchmark.submitThreeThreads:completed            10      sync  thrpt   20  20.802 ±  0.873  ops/us
CombinerExecutorBenchmark.submitThreeThreads:submitted            10      sync  thrpt   20  20.802 ±  0.873  ops/us
CombinerExecutorBenchmark.submitThreeThreads                      10  combiner  thrpt   20  11.474 ±  0.111  ops/us
CombinerExecutorBenchmark.submitThreeThreads:completed            10  combiner  thrpt   20  11.477 ±  0.111  ops/us
CombinerExecutorBenchmark.submitThreeThreads:submitted            10  combiner  thrpt   20  11.477 ±  0.111  ops/us
CombinerExecutorBenchmark.submitThreeThreads                      20      sync  thrpt   20  13.453 ±  0.592  ops/us
CombinerExecutorBenchmark.submitThreeThreads:completed            20      sync  thrpt   20  13.454 ±  0.592  ops/us
CombinerExecutorBenchmark.submitThreeThreads:submitted            20      sync  thrpt   20  13.454 ±  0.592  ops/us
CombinerExecutorBenchmark.submitThreeThreads                      20  combiner  thrpt   20  10.800 ±  0.693  ops/us
CombinerExecutorBenchmark.submitThreeThreads:completed            20  combiner  thrpt   20  10.803 ±  0.693  ops/us
CombinerExecutorBenchmark.submitThreeThreads:submitted            20  combiner  thrpt   20  10.803 ±  0.693  ops/us
CombinerExecutorBenchmark.submitTwoThreads                         0      sync  thrpt   20  13.214 ±  1.548  ops/us
CombinerExecutorBenchmark.submitTwoThreads:completed               0      sync  thrpt   20  13.215 ±  1.548  ops/us
CombinerExecutorBenchmark.submitTwoThreads:submitted               0      sync  thrpt   20  13.215 ±  1.548  ops/us
CombinerExecutorBenchmark.submitTwoThreads                         0  combiner  thrpt   20  21.587 ± 17.779  ops/us
CombinerExecutorBenchmark.submitTwoThreads:completed               0  combiner  thrpt   20  21.717 ± 18.038  ops/us
CombinerExecutorBenchmark.submitTwoThreads:submitted               0  combiner  thrpt   20  21.717 ± 18.038  ops/us
CombinerExecutorBenchmark.submitTwoThreads                         1      sync  thrpt   20  12.017 ±  1.040  ops/us
CombinerExecutorBenchmark.submitTwoThreads:completed               1      sync  thrpt   20  12.018 ±  1.040  ops/us
CombinerExecutorBenchmark.submitTwoThreads:submitted               1      sync  thrpt   20  12.018 ±  1.040  ops/us
CombinerExecutorBenchmark.submitTwoThreads                         1  combiner  thrpt   20   8.986 ±  1.087  ops/us
CombinerExecutorBenchmark.submitTwoThreads:completed               1  combiner  thrpt   20   8.989 ±  1.088  ops/us
CombinerExecutorBenchmark.submitTwoThreads:submitted               1  combiner  thrpt   20   8.989 ±  1.088  ops/us
CombinerExecutorBenchmark.submitTwoThreads                        10      sync  thrpt   20   9.651 ±  0.986  ops/us
CombinerExecutorBenchmark.submitTwoThreads:completed              10      sync  thrpt   20   9.652 ±  0.986  ops/us
CombinerExecutorBenchmark.submitTwoThreads:submitted              10      sync  thrpt   20   9.652 ±  0.986  ops/us
CombinerExecutorBenchmark.submitTwoThreads                        10  combiner  thrpt   20  10.229 ±  0.317  ops/us
CombinerExecutorBenchmark.submitTwoThreads:completed              10  combiner  thrpt   20  10.231 ±  0.317  ops/us
CombinerExecutorBenchmark.submitTwoThreads:submitted              10  combiner  thrpt   20  10.231 ±  0.317  ops/us
CombinerExecutorBenchmark.submitTwoThreads                        20      sync  thrpt   20   9.041 ±  0.612  ops/us
CombinerExecutorBenchmark.submitTwoThreads:completed              20      sync  thrpt   20   9.041 ±  0.613  ops/us
CombinerExecutorBenchmark.submitTwoThreads:submitted              20      sync  thrpt   20   9.041 ±  0.613  ops/us
CombinerExecutorBenchmark.submitTwoThreads                        20  combiner  thrpt   20  10.482 ±  0.209  ops/us
CombinerExecutorBenchmark.submitTwoThreads:completed              20  combiner  thrpt   20  10.490 ±  0.217  ops/us
CombinerExecutorBenchmark.submitTwoThreads:submitted              20  combiner  thrpt   20  10.490 ±  0.217  ops/us
```
TLDR the semaphore based one seems to perform better (or slightly worse) then the combiner impl: maybe there's something obvious I'm missing but this requires some investigation


Sadly AuxCounters are collected and merged by JMH, and cannot report per-thread statistics (that would be nice); I can send a PR to JMH for that because it would be nice to know how the executions of tasks is distributed between threads - here I can create a counter but I the reporting won't say anything useful

vietj:
I think we are not looking to compare this to the semaphore version because we are not comparing apples to apples. The semaphore version will create contention and potentially block all threads and the lock free version will not.

We should instead compare performance to the same implementation without a deadline to see that the changes will not increase too much the cost of the submission.

franz1981:
> I think we are not looking to compare this to the semaphore version because we are not comparing apples to apples. The semaphore version will create contention and potentially block all threads and the lock free version will not.

It's fair, and indeed the benchmark itself is not comparing the same behaviour nor kind of service we expect by the executor but still...it has highlighted an important bottleneck.

The semaphore version won't allow too many contended requests on the semaphore (because the winner will enter and the looser will park, without performing any CAS loop), while the JCTools chosen queue, instead, is using a CAS loop to submit the action; just switching the JCTools mpsc unbounded xadd q, performance will increase *a lot* given that who designed it (myself) tried hard to make it a better performer under high contention (on x86, while M1 Apple users and PowerPC won't benefit by this optimization). 

This same scenario can happen in a real world case if the event loop acting as combiner got other N concurrent event loops competing to add to the action queue; they will end up slowing each others because of the contention (!)

I can show some number if you think that's a valid use case

franz1981:
These are the numbers with the 'master` benchmark (using 2 threads):
```
Benchmark                        Mode  Cnt      Score      Error   Units
CombinerExecutorBenchmark.impl  thrpt   10  11484.804 ± 5100.280  ops/ms
```
These are by using the mpsc unbounded xadd q:
```
Benchmark                        Mode  Cnt      Score      Error   Units
CombinerExecutorBenchmark.impl  thrpt   10  17208.869 ± 1450.893  ops/ms
```

With 8 threads:
```
Benchmark                        Mode  Cnt     Score     Error   Units
CombinerExecutorBenchmark.impl  thrpt   10  6423.568 ± 361.666  ops/ms
```
8 threads using the mpsc unbounded xadd q:
```
Benchmark                        Mode  Cnt      Score      Error   Units
CombinerExecutorBenchmark.impl  thrpt   10  27310.143 ± 3168.112  ops/ms
```
The CAS loop of the original queue is causing a diminishing return on a burst of requests due to the contention, while the xadd one isn't affected; 
there are other effects in place (q::poll is much cheaper on the cas q) that can affect the benchmark but the behaviour under contention is quite real.

vietj:
> This same scenario can happen in a real world case if the event loop acting as combiner got other N concurrent event loops competing to add to the action queue; they will end up slowing each others because of the contention (!)

In practice I think this is the case of a benchmark, we are talking of event-loop threads which have plenty of tasks to perform, including IO and that access a database pool to perform interactions with the database.

src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
```diff
@@ -860,9 +902,12 @@ boolean remove(PoolWaiter<C> node) {
 
     List<PoolWaiter<C>> clear() {
       List<PoolWaiter<C>> lst = new ArrayList<>(size);
-      this.forEach(lst::add);
-      size = 0;
-      head.next = head.prev = head;
+      PoolWaiter<C> waiter;
```
franz1981, line 905: This is  solving a potential GC nepotism issue @vietj

src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
```diff
@@ -79,6 +81,11 @@
  */
 public class SimpleConnectionPool<C> implements ConnectionPool<C> {
 
+  private static final long DEFAULT_EXECUTE_ACTIONS_TIMEOUT_MS = Integer.getInteger("combiner.execute.timeout.ms", 100);
```
franz1981, line 84: I can lower the default or expose this differently too @vietj

src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
```diff
@@ -242,7 +264,27 @@ public ConnectionPool<C> contextProvider(Function<ContextInternal, EventLoopCont
...
 
   private void execute(Executor.Action<SimpleConnectionPool<C>> action) {
-    sync.submit(action);
+    if (resumeExecutor == null) {
+      sync.submit(action);
+    } else {
+      final Executor.Continuation continuation = sync.submitAndContinue(action);
+      if (continuation != null) {
+        // single threaded
+        resumeExecutor.execute(() -> resumeAndContinue(continuation));
```
franz1981, line 273: The comment here will be used to add counters (that can just use  `putOrdered/lazySet` given that's a single-threaded thing. @vietj 
I'm NOT happy to allocate a whole new lambda for this, because `Continuation` is captured :(
To save this we should NOT make any `Continuation` and just provide an API on the executor like

```
// return false implies calling a resume later
boolean executeAndContinue(...);

// throw IllegalStateException if not called after a previous returning false resume/executeAndContinue
boolean resume();
```
This means that `resumeAndContinue` on the pool doesn't need anymore the `continuation` parameter

src/test/java/io/vertx/core/net/impl/pool/ConnectionPoolTest.java
```diff
@@ -335,12 +365,21 @@ public void testRemoveSingleConnectionWithWaiter() throws Exception {
 
   @Test
   public void testConnectFailureWithPendingWaiter() throws Exception {
+    // This is necessary because:
```
franz1981, line 368: @vietj is it supposed to happen?

src/test/java/io/vertx/core/net/impl/pool/StressTest.java
```diff
@@ -218,6 +240,14 @@ protected void onSuccess(FakeConnection conn) {
       latch.countDown();
     });
     awaitLatch(latch);
+    if (resumeExecutor != null) {
+      // maybe there's still a suspended continuation waiting to be resumed:
+      // given that's a single threaded executor we just need to enqueue past to it
+      // and no new continuations should be in-flights, because after closing the pool
+      // there won't be any submitted actions
+      resumeExecutor.submit(() -> {}).get();
```
franz1981, line 248: that's a cheap trick IMO @vietj 
My concern is that even in the real use case, if the current continuation is still suspended ie yield happened, ready to be executed on some event loop
none knows about it! Nor the pool or the combiner too

src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
```diff
@@ -194,11 +201,14 @@ public long concurrency() {
   private final Waiters<C> waiters;
   private int requests;
 
-  SimpleConnectionPool(PoolConnector<C> connector, int[] maxSizes) {
-    this(connector, maxSizes, -1);
+  private final java.util.concurrent.Executor resumeExecutor;
```
vietj, line 204: in practice this will not work because executor is the current event-loop.

since this class is already married to the `EventLoopContext`, we should assume it and replace the `resumeExecutor` with the current `EventLoopContext#executor()`

franz1981: Perfect, will change it :+1:

franz1981: Looking at it twice I don't see how:
> since this class is already married to the EventLoopContext

EventLoopContext is used on the Slot by using
```java
EventLoopContext connectionContext = pool.contextProvider.apply(waiter.context);
```
Given that I dont't have a `ContextInternal` (the type of `waiter,context`) how can I obtain it on pool creation and/or
submit/resume?

src/test/java/io/vertx/core/net/impl/pool/ConnectionPoolTest.java
vietj: yes it can happen why ?

franz1981: In this case the test isn't written assuming that post actions can happen in different moments, modifying the state not atomically (look at the comments I've added past this `//`)

franz1981:
I would close this as not merged

PR #4432, opened 2022-07-07
fix(VertxOptions): QuickWin https://github.com/eclipse-vertx/vert.x/i…

srivatsa-cfp:
…ssues/4431#issue-1297779421

Motivation:

VertOptions has been incorrect exception messages for Quorum Size

Following the inputs:

Current 

public VertxOptions setQuorumSize(int quorumSize) {
    if (quorumSize < 1) {
      throw new IllegalArgumentException("quorumSize should be >= 1");
    }
    this.quorumSize = quorumSize;
    return this;
  }

setQuorumSize :
condition to throw exception: quorum < 1
message part of exception: workerPoolSize muse be >= 1

Proposed  condition  

public VertxOptions setQuorumSize(int quorumSize) {
    if (quorumSize <=0 ) {
      throw new IllegalArgumentException("quorumSize should be >= 1");
    }
    this.quorumSize = quorumSize;
    return this;
  }

Vertx version: Latest version

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/VertxOptions.java
+++ b/src/main/java/io/vertx/core/VertxOptions.java
@@ -412,7 +412,7 @@ public class VertxOptions {
    * @return a reference to this, so the API can be used fluently
    */
   public VertxOptions setQuorumSize(int quorumSize) {
-    if (quorumSize < 1) {
+    if (quorumSize <= 0) {
       throw new IllegalArgumentException("quorumSize should be >= 1");
     }
     this.quorumSize = quorumSize;


src/main/java/io/vertx/core/VertxOptions.java
```diff
@@ -412,7 +412,7 @@ public int getQuorumSize() {
    * @return a reference to this, so the API can be used fluently
    */
   public VertxOptions setQuorumSize(int quorumSize) {
-    if (quorumSize < 1) {
+    if (quorumSize <= 0) {
```
tsegismont, line 415:
```suggestion
    if (quorumSize < 1) {
```

tsegismont (changes requested):
Thank you @srivatsa-cfp , the check must to be changed

src/main/java/io/vertx/core/VertxOptions.java
srivatsa-cfp: @tsegismont If quorumSize is less than 1, then Exception message is incorrect here.
                           `throw new IllegalArgumentException("quorumSize should be >= 1");` 
 This will break if we pass 1 as message is expecting it can be 1.

tsegismont: Oh yeah, sorry :blush:

tsegismont (approved):

src/main/java/io/vertx/core/VertxOptions.java
tsegismont: Actually no, the test was fine. If `quorumSize` is set to 0 or less, the test fails and the message is printed.
If you give 1, then the test passes and the option field is set to 1.

srivatsa-cfp:
@tsegismont Is the provided solution incorrect for setting the quorum size? If the server count = 1, the following condition doesn't allow to set 1 and throw exception.

PR #4438, opened 2022-07-13
Netty 5 test wip

vietj:


src/main/java/io/vertx5/core/net/NetSocket.java
```diff
@@ -97,7 +114,16 @@ public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
       channelHandlerContext = null;
     }
     @Override
+    public void channelReadComplete(ChannelHandlerContext ctx) throws Exception {
+      readInProgress = false;
+      if (needsFlush) {
+        needsFlush = false;
+        ctx.flush();
```
franz1981, line 121: I would check, after the flush, if the channel is writable: if not, I would setAutoRead = false and register for the channelWritabilityChange before enabling it to true again.
In this way, if client cannot keep up with reading data (or other network related reasons), netty won't keep on ingesting requests which responses won't be sent back, risking to OOM the server

vietj: I will do that, that being said there is no OOM risk as user should register its own hook to start/stop writing in the vertx layer.

src/main/java/io/vertx5/core/net/NetSocket.java
```diff
@@ -104,41 +105,65 @@ public NetSocket closeHandler(Handler<Void> handler) {
...
+      netSocket.writable = ctx.channel().isWritable();
     }
+
     @Override
-    public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
-      channelHandlerContext = null;
+    public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
+      boolean writable = !netSocket.writable;
+      netSocket.writable = writable;
+      if (writable && netSocket.needsFlush) {
```
franz1981, line 125: it shouldn't load the writability from ctx.channel().isWritable?

vietj: I'm assuming that the event is triggered when the value changes, so using writable = !writable seems correct

Issue #4439, opened 2022-07-15
Noisy logging from TCPServerBase#listen during startup


MikeEdgar:
### Version

- 4.1.7 (via Quarkus 2.2.5)
- 4.2.4 (via Quarkus 2.7.5)

### Context

In a resource-constrained environment (e.g. Kubernetes with limited a CPU request/limit), the start-up of the HTTP server results in multiple blocked thread stack traces being logged. When deploying both TLS and plain text hosts this increases as well. 

```
2022-07-15 11:41:54,804 WARN  [io.ver.cor.imp.BlockedThreadChecker] (vertx-blocked-thread-checker) Thread Thread[vert.x-eventloop-thread-0,5,main] has been blocked for 2695 ms, time limit is 2000 ms: io.vertx.core.VertxException: Thread blocked
	at io.vertx.core.net.impl.TCPServerBase.listen(TCPServerBase.java:125)
	at io.vertx.core.net.impl.TCPServerBase.bind(TCPServerBase.java:100)
	at io.vertx.core.http.impl.HttpServerImpl.listen(HttpServerImpl.java:217)
	at io.vertx.core.http.impl.HttpServerImpl.listen(HttpServerImpl.java:149)
	at io.vertx.core.http.impl.HttpServerImpl.listen(HttpServerImpl.java:154)
	at io.quarkus.vertx.http.runtime.VertxHttpRecorder$WebDeploymentVerticle.setupTcpHttpServer(VertxHttpRecorder.java:1075)
	at io.quarkus.vertx.http.runtime.VertxHttpRecorder$WebDeploymentVerticle.start(VertxHttpRecorder.java:1029)
```

Scenario is also referenced in #1379

### Extra

Given the likelihood of this situation occurring and producing log noise, it seems it may be desirable to execute `listen` using internal worker threads. I've verified this works with my application [1], but would like to discuss here if this approach is correct more broadly within Vert.x.

[1] https://github.com/eclipse-vertx/vert.x/compare/master...MikeEdgar:vert.x:tcpserverbase-listen-blocking?expand=1

vietj:
binding a server is a non blocking interaction with the system, can you elaborate what do you think make it blocking ?

MikeEdgar:
[...]

PR #4441, opened 2022-07-21
Execute SSLHelper.validate as blocking

MikeEdgar:
Motivation:

Proposed to fix https://github.com/eclipse-vertx/vert.x/issues/4439

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/SSLHelper.java
+++ b/src/main/java/io/vertx/core/net/impl/SSLHelper.java
@@ -14,10 +14,13 @@ package io.vertx.core.net.impl;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.handler.ssl.*;
 import io.netty.util.Mapping;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
 import io.vertx.core.VertxException;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.ClientAuth;
 import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.impl.VertxInternal;
 import io.vertx.core.impl.logging.Logger;
 import io.vertx.core.impl.logging.LoggerFactory;
@@ -99,6 +102,8 @@ public class SSLHelper {
   private static final Logger log = LoggerFactory.getLogger(SSLHelper.class);
 
   private boolean ssl;
+  private volatile boolean validated = false;
+  private volatile Throwable validationError = null;
   private boolean sni;
   private long sslHandshakeTimeout;
   private TimeUnit sslHandshakeTimeoutUnit;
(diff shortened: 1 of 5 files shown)


src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -502,10 +507,37 @@ public SslContext getContext(VertxInternal vertx, String serverName, boolean use
...
+        return Future.failedFuture(validationError);
+      }
+      return Future.succeededFuture();
+    }
+
+    validated = true;
+
     if (ssl) {
-      getContext(vertx, null);
+      ContextInternal validateContext = vertx.getOrCreateContext();
```
vietj, line 521: the context should be provided as argument of the validate method instead of being access as a static.

vietj (changes requested):
a couple of changes to do

MikeEdgar:
I think the use of `CountDownLatch` in `TCPServerBase` is also going to be a problem. I'm still looking into that.

src/main/java/io/vertx/core/net/impl/TCPServerBase.java
```diff
@@ -120,102 +117,159 @@ private synchronized io.netty.util.concurrent.Future<Channel> listen(SocketAddre
...
-          shared = false;
-          bindAddress = localAddress;
+
+        if (main != null) {
+          try {
+            main.initialization.await();
+          } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return vertx.getAcceptorEventLoopGroup().next().newFailedFuture(e);
+          }
```
MikeEdgar, line 155: @vietj , since this effectively blocks the thread waiting for the initialization of the other server, I think it may also need to run with the blocking thread pool. Wdyt?

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -99,6 +102,8 @@ public static SSLEngineOptions resolveEngineOptions(TCPSSLOptions options) {
   private static final Logger log = LoggerFactory.getLogger(SSLHelper.class);
 
   private boolean ssl;
+  private volatile boolean validated = false;
```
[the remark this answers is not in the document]
MikeEdgar: Done

src/main/java/io/vertx/core/net/impl/SSLHelper.java
MikeEdgar: Done

src/main/java/io/vertx/core/net/impl/TCPServerBase.java
MikeEdgar: The PR is updated to await initialization via a blocking thread.

MikeEdgar:
@vietj, I'm curious to get your thinking on these changes. I realize there needs to be a certain amount of caution with changes to these classes/methods.

vietj:
it looks good overall but there are a few things that need to be addressed

MikeEdgar:
> it looks good overall but there are a few things that need to be addressed

Thanks, let me know what's needed when you have an opportunity. No rush from my perspective.

vietj:
I will actually reuse this in an general improvement of the SSLHlper because there are other SSLHelper related work that should be done.

- https://github.com/eclipse-vertx/vert.x/pull/4372
- https://github.com/eclipse-vertx/vert.x/issues/4452
- https://github.com/eclipse-vertx/vert.x/pull/4453

vietj:
@MikeEdgar please have a look at https://github.com/eclipse-vertx/vert.x/pull/4468

MikeEdgar:
@vietj , now that #4468 is the path forward on this, should we just close this one?

vietj:
yes, I'll credit you in the orignal issue.

PR #4442, opened 2022-07-22
Allow passing throwable in event bus message fail

tylerscoville:
Motivation:

I'm opening this pr mostly as a question with a possible solution.

I would like to be able to pass a `throwable` with the event bus `message.fail` method to preserve some of the context of the failure. Something like:
```
public class MyCustomThrowable extends Throwable {}

vertx.eventBus()
        .consumer("test", message -> message.fail(1, "failure", new MyCustomThrowable()));

vertx.eventBus().request("test", "test")
    .onFailure(h -> {
      if(h.getCause() instanceof MyCustomThrowable) {
        //Custom error handling here.
      }
    });
```
I'm not sure if there is a better way to handle this or if this is a tenable solution?

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/Message.java
+++ b/src/main/java/io/vertx/core/eventbus/Message.java
@@ -141,4 +141,18 @@ public interface Message<T> {
     reply(new ReplyException(ReplyFailure.RECIPIENT_FAILURE, failureCode, message));
   }
 
+  /**
+   * Signal to the sender that processing of this message failed.
+   * <p>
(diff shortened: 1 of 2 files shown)


tsegismont:
Actually you can use `msg.reply` to send a subclass of ReplyException. You need to create a custom codec for it.

See https://github.com/eclipse-vertx/vert.x/blob/a8d9a164df16c4f2f5eed5f176a5570c46f30d41/src/test/java/io/vertx/core/eventbus/EventBusTestBase.java#L779-L834

PR #4445, opened 2022-07-30
fix: ensure `DeliveryOptions.headers` are properly cloned

BFergerson:
Motivation:

I've run into situations where using re-using a  `*VertxEBProxy` (produced by vertx-service-proxy) sends messages to the wrong consumer because the `DeliveryOptions.headers` contains multiple `action` values.

This ensures the `DeliveryOptions.headers` are properly cloned.


vietj:
thanks for the contrib

PR #4447, opened 2022-08-03
Future terminal operation documentation

### LINKED ISSUE #4446: Future terminal operations clarification
In some parts of the stack, it is assumed that `Future` terminal operations (`onSuccess`, `onFailure`, `onComplete`) are executed in order of registration.

While in most cases this is true, it is not guaranteed and can lead to bugs like https://github.com/vert-x3/vertx-jdbc-client/issues/281

The actual behavior should be clarified in documentation and we should inspect the stack to find possible problems.

tsegismont:
Closes #4446 

Commits can be reviewed individually.


src/test/java/io/vertx/core/FutureTest.java
```diff
@@ -1627,4 +1628,91 @@ private void testListenersReportFailureOnContextAfterCompletion(Function<Context
     });
     waitUntil(() -> caught.size() == 1 && caught.get(0) == failure);
   }
+
+  @Test
+  @Repeat(times = 50)
```
vietj, line 1633: no need to repeat this test as there is no context involved so the current thread is always used

src/main/asciidoc/futures.adoc
```diff
@@ -0,0 +1,155 @@
...
+====
+Do not confuse _futures_ with _promises_.
+
+If futures represent the "read-side" of an asynchronous result, promises are the "write-side".
+They allow you to defer the action of providing a result.
+
+In most cases, you don't need to create promises yourself in a Vert.x application.
+<<_future_composition>> and <<_future_coordination>> provide you with the tools to transform and merge asynchronous results.
+
+However, if, in your codebase, you have legacy methods which use callbacks, you can leverage the fact that a promise extends {@link io.vertx.core.Handler io.vertx.core.Handler<io.vertx.core.AsyncResult>}:
```
vietj, line 29: we should omit this part, as this is something that might be removed in vertx 5

src/main/asciidoc/futures.adoc
```diff
@@ -0,0 +1,155 @@
...
+Terminal operations like `onSuccess`, `onFailure` and `onComplete` provide no guarantee whatsoever regarding the invocation order of callbacks.
+
+Consider a future on which 2 callbacks are registered:
+
+[source,$lang]
+----
+{@link examples.CoreExamples#promiseCallbackOrder}
+----
+
+It is possible that the second callback is invoked before the first one.
```
vietj, line 48: I think here we should mention that ordered side effect behaviour can be achieved with `andThen`

src/main/asciidoc/futures.adoc
tsegismont: That's something we added because we had many related questions on the user group. I would like to keep it.

vietj (approved):

vietj (changes requested):
the PR does not seem to compile

Issue #1623, opened 2016-09-12
Support classpath root nested in JAR


tsegismont:
This is the upstream issue for vert-x3/vertx-web#452 Allow templates to be loaded from classpath

Classpath root may be a folder inside a JAR, not the JAR itself.

For example, in Spring Boot, here's the list of root URLs:

```
0 = {URL@4064} "jar:file:/home/tsegismont/Temp/wv/target/testwb-1.0-SNAPSHOT.jar!/"
1 = {URL@4065} "jar:file:/home/tsegismont/Temp/wv/target/testwb-1.0-SNAPSHOT.jar!/BOOT-INF/classes!/"
2 = {URL@4066} "jar:file:/home/tsegismont/Temp/wv/target/testwb-1.0-SNAPSHOT.jar!/BOOT-INF/lib/vertx-web-templ-freemarker-3.4.0-SNAPSHOT.jar!/"
3 = {URL@4067} "jar:file:/home/tsegismont/Temp/wv/target/testwb-1.0-SNAPSHOT.jar!/BOOT-INF/lib/jackson-core-2.8.1.jar!/"
```

... and so on with all libs

Note the jar root, as well as BOOT-INF/classes

PR #4454, opened 2022-08-12
Support classpath root nested in JAR

gaol:
Motivation:

Trying to add the fix to https://github.com/eclipse-vertx/vert.x/issues/1623 back.

Conformance:

This also relates to https://github.com/vert-x3/vertx-web/issues/2183

### DIFF
--- a/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
+++ b/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
@@ -48,6 +48,7 @@ public class FileResolverImpl implements FileResolver {
   public static final String DISABLE_CP_RESOLVING_PROP_NAME = "vertx.disableFileCPResolving";
   public static final String CACHE_DIR_BASE_PROP_NAME = "vertx.cacheDirBase";
   private static final boolean NON_UNIX_FILE_SEP = File.separatorChar != '/';
+  private static final String JAR_URL_SEP = "!/";
 
   private final File cwd;
   private final boolean enableCaching;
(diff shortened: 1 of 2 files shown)


gaol:
@vietj @tsegismont would you please review? thanks

vietj:
is there a regression @gaol ? wasn't this supposed to be fixed ?

can you rebase on latest master to have the CI pass ?

src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
```diff
@@ -294,18 +296,25 @@ private File unpackFromJarURL(URL url, String fileName, ClassLoader cl) {
         File file = resolveFile(s);
         zip = new ZipFile(file);
       }
-
+      String inJarPath = path.substring(idx1 + 6);
+      String[] parts = JAR_URL_SEP_PATTERN.split(inJarPath);
```
vietj, line 300: rather use a Matcher for the pattern and build the prefix using matcher index and append(String, int, int)

gaol:
> is there a regression @gaol ? wasn't this supposed to be fixed ?

Yes, please take a look at issue: https://github.com/vert-x3/vertx-web/issues/2183

> can you rebase on latest master to have the CI pass ?

Sure, on my way :)

src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
gaol: Since we just need to split the path with `!/`, I would like use `String.indexOf(String, int)`, and leave Pattern away.

gaol:
@vietj CI passed, would you please review again ? thanks

src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
```diff
@@ -49,6 +49,7 @@ public class FileResolverImpl implements FileResolver {
   public static final String CACHE_DIR_BASE_PROP_NAME = "vertx.cacheDirBase";
   private static final boolean NON_UNIX_FILE_SEP = File.separatorChar != '/';
 
+  private static final String JAR_URL_SEP = "!/";
```
vietj, line 52: move this with the group of static final above instead of the class fields

src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
```diff
@@ -295,17 +296,28 @@ private File unpackFromJarURL(URL url, String fileName, ClassLoader cl) {
...
+        prefixBuilder.append(inJarPath, first, second).append("/");
+        first = second + len;
+      }
+      String prefix = prefixBuilder.toString();
       Enumeration<? extends ZipEntry> entries = zip.entries();
       while (entries.hasMoreElements()) {
         ZipEntry entry = entries.nextElement();
         String name = entry.getName();
-        if (name.startsWith(fileName)) {
+        if (name.startsWith(prefix.isEmpty() ? fileName : prefix + fileName)) {
```
vietj, line 313: the expression `prefix.isEmpty() ? fileName : prefix + fileName)` can be moved out of the while loop which avoids to compute prefix + fileName at each iteration

src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
gaol: Done by moving the line up.

src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
gaol: Done

gaol:
@vietj updated based on the feedback

vietj:
thanks @gaol

PR #4459, opened 2022-08-18
Do not log before being LoggerFactory is fully initialized

tsegismont:
Fixes reactiverse/reactiverse-contextual-logging#52

Otherwise, a logger extension might use Vert.x classes before they are fully initialized.


vietj:
is it possible to add an integration test for this ?

tsegismont:
@vietj added a test which does not pass without the fix

vietj (approved):

PR #4450, opened 2022-08-07
Feature: setPeriodic add argument initialDelay parameter

pigbayspy:
Motivation:

add a new parameter `initialDelay` in `setPeriodic` method of `Vertx` https://github.com/eclipse-vertx/vert.x/issues/4436

PR #4467, opened 2022-08-29
Add default implementation for new periodic timer methods in Vertx

tsegismont:
Avoids breaking implementations that do not have the method yet.

Follows-up on #4450


tsegismont:
cc @vietj  @cescoffier

vietj:
it's ugly as no one should implement this API, but yikes

PR #4471, opened 2022-09-07
Fix NPE in VertxImpl when the threadFactory is accessed before its assignment

cescoffier:
`createThreadFactory` references the `threadFactory` field. If a thread is created before the assignation, it leads to an NPE. By moving the assignment before the first call to `createThreadFactory`, we prevent that from happening.


cescoffier:
@vietj I looked at the reason why it happened in Quarkus, and wrote a unit test.

cescoffier:
@vietj Should we close this PR?

vietj:
I still need to have a look I think

vietj:
See https://github.com/eclipse-vertx/vert.x/pull/4539

Issue #4464, opened 2022-08-25
Allow vertx thread to be daemon


Apache9:
#### Describe the feature

Allow vertx thread to be daemon

#### Use cases

https://github.com/eclipse-vertx/vert.x/blob/2439c849549b08a1e7ffee8f09eae886c9fe88f6/src/main/java/io/vertx/core/impl/VertxImpl.java#L1144

Here we force setting the daemon to false when creating vertx thread. The reason is to prevent JVM exit for embedded user as said in the comment.
But actually, the default thread factory will create non daemon threads, if users set a special thread factory to create daemon threads, they must have a reason. It does not make sense to reset the thread to non daemon under this scenario.

#### Contribution

The change is very simple, just remove the `thread.setDaemon(false);` line. I can open a PR for this.

vietj:
that's a good point, however a thread can be a daemon when the thread that constructed it is a daemon itself, e.g

```java
  public static void main(String[] args) throws Exception {
    System.out.println("1 daemon=" + new Thread().isDaemon());
    Thread t1 = new Thread(() -> {
      System.out.println("2 daemon=" + new Thread().isDaemon());
    });
    t1.setDaemon(true);
    t1.start();
    System.in.read();
  }
```

we can provide a setting for configuring that though (e.g VertxOptions)

Apache9:
OK, check the code of VertxThreadFactory, the default implementation is not like JDK's DefaultThreadFactory...

```
  default VertxThread newVertxThread(Runnable target, String name, boolean worker, long maxExecTime, TimeUnit maxExecTimeUnit) {
    return new VertxThread(target, name, worker, maxExecTime, maxExecTimeUnit);
  }
```

```
        public Thread newThread(Runnable r) {
            Thread t = new Thread(group, r,
                                  namePrefix + threadNumber.getAndIncrement(),
                                  0);
            if (t.isDaemon())
                t.setDaemon(false);  // <===== here it will force set the daemon to false
            if (t.getPriority() != Thread.NORM_PRIORITY)
                t.setPriority(Thread.NORM_PRIORITY);
            return t;
        }
```

So I think maybe could change the default implementation of VertxThreadFactory, to force set thread to non daemon. In this way we do not need to introduce a new config then since we have already support users to specify a customized thread factory.

WDYT?

Thanks.

vietj:
I don't think so, that should be a configurable option, e.g `DefaultThreadFactory` make this configurable:

```java
[...]

PR #4474, opened 2022-09-08
Allow vertx thread to be daemon

Apache9:
Introduce a ThreadOptions for setting daemon, priority and uncaughtExceptionHandler for VertxThread

For #4464

### DIFF
--- a/src/main/java/io/vertx/core/VertxOptions.java
+++ b/src/main/java/io/vertx/core/VertxOptions.java
@@ -113,6 +113,13 @@ public class VertxOptions {
 
   public static final boolean DEFAULT_DISABLE_TCCL = Boolean.getBoolean(DISABLE_TCCL_PROP_NAME);
 
+  /**
+   * Set default value to false for aligning with the old behavior
+   * By default, Vert.x threads are NOT daemons - we want them to prevent JVM exit so embedded user
+   * doesn't have to explicitly prevent JVM from exiting.
+   */
+  public static final boolean DEFAULT_USE_DAEMON_THREAD = false;
+
   private int eventLoopPoolSize = DEFAULT_EVENT_LOOP_POOL_SIZE;
   private int workerPoolSize = DEFAULT_WORKER_POOL_SIZE;
   private int internalBlockingPoolSize = DEFAULT_INTERNAL_BLOCKING_POOL_SIZE;
@@ -135,6 +142,7 @@ public class VertxOptions {
   private TimeUnit warningExceptionTimeUnit = DEFAULT_WARNING_EXCEPTION_TIME_UNIT;
   private TimeUnit blockedThreadCheckIntervalUnit = DEFAULT_BLOCKED_THREAD_CHECK_INTERVAL_UNIT;
   private boolean disableTCCL = DEFAULT_DISABLE_TCCL;
+  private Boolean useDaemonThread = DEFAULT_USE_DAEMON_THREAD;
 
   /**
    * Default constructor
@@ -169,6 +177,7 @@ public class VertxOptions {
     this.blockedThreadCheckIntervalUnit = other.blockedThreadCheckIntervalUnit;
     this.tracingOptions = other.tracingOptions != null ? other.tracingOptions.copy() : null;
     this.disableTCCL = other.disableTCCL;
+    this.useDaemonThread = other.useDaemonThread;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


Apache9:
@vietj FYI.

src/main/java/io/vertx/core/VertxOptions.java
```diff
@@ -675,6 +675,20 @@ public VertxOptions setDisableTCCL(boolean disableTCCL) {
     return this;
   }
 
+  /**
+   * Return the {@link ThreadOptions} used to create {@link io.vertx.core.impl.VertxThread}.
+   */
+  public ThreadOptions getThreadOptions() {
```
vietj, line 681: I think only the `daemon` setting is necessary, priority can always be set by the thread factory and will not be overwritten

Apache9: So you mean we should only provide a daemon setting here, without touching the VertxThreadFactory, but instead changing the code in VertxImpl to reset the daemon field of the VertxThread?

vietj:
no I'm saying that the VertxOptions would have the setting to set a thread
to be a daemon or not.

Apache9:
OK, so you prefer we add a ThreadOptions which only contains a daemon field in it, or we just add a daemon flag to VertxOptions directly?

Thanks.

vietj:
only a flag in options

Apache9:
Got it. Let me update the PR.

Apache9:
I've updated the PR, PTAL. @vietj Thanks.

Apache9:
@vietj Any concerns here?

Thanks.

src/main/java/io/vertx/core/VertxOptions.java
```diff
@@ -675,6 +681,28 @@ public VertxOptions setDisableTCCL(boolean disableTCCL) {
...
+
+  /**
+   * Mark the vertx thread as daemon thread or user thread.
+   * <p/>
+   * For keeping the old behavior, the default value is {@code false} instead of {@code null}.
+   *
+   * @param daemon {@code true} means daemon, {@code false} means not daemon(user), {@code null}
+   *        means do not change the daemon option of the created vertx thread.
+   */
+  public void setUseDaemonThread(Boolean daemon) {
```
vietj, line 702: this should be a fluent setter like other setters.

src/main/java/io/vertx/core/VertxOptions.java
```diff
@@ -135,6 +135,12 @@ public class VertxOptions {
   private TimeUnit warningExceptionTimeUnit = DEFAULT_WARNING_EXCEPTION_TIME_UNIT;
   private TimeUnit blockedThreadCheckIntervalUnit = DEFAULT_BLOCKED_THREAD_CHECK_INTERVAL_UNIT;
   private boolean disableTCCL = DEFAULT_DISABLE_TCCL;
+  /**
+   * Set default value to false for aligning with the old behavior
+   * By default, Vert.x threads are NOT daemons - we want them to prevent JVM exit so embedded user
+   * doesn't have to explicitly prevent JVM from exiting.
+   */
+  private Boolean useDaemonThread = false;
```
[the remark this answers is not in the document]
Apache9: Fixed.

src/main/java/io/vertx/core/VertxOptions.java
Apache9: Fixed.

Apache9:
Thanks @vietj for reivewing! I've updated the PR according to the review comment. Please take a look again.

Thanks.

vietj (commented):
we are missing a test and then we are good to go.

Apache9:
Ah, when trying to implement a UT, I found that we will not call threadFactory,init(VertxBuilder)...

We will only call the init method on all VertxServiceProviders in initProviders method...

That means, the init method of VertxThreadFactory.INSTANCE will never be called...

There are several possible ways to call the init method of VertxThreadFactory.INSTANCE but all of them are a bit ugly I think...

And I also noticed another problem that, since we added a field in VertxThreadFactory.INSTANCE, it can not be shared by multiple vertx instances now, because different verx instance may want to set different daemon config...

For me I do not have a good solution in mind for this direction now, i.e, add an option in VertxOptions and make the default implementation to honor this option...

WDYT? @vietj 

Thanks.

src/main/java/io/vertx/core/impl/VertxImpl.java
```diff
@@ -1126,9 +1126,6 @@ private ThreadFactory createThreadFactory(long maxExecuteTime, TimeUnit maxExecu
     return runnable -> {
       VertxThread thread = threadFactory.newVertxThread(runnable, prefix + threadCount.getAndIncrement(), worker, maxExecuteTime, maxExecuteTimeUnit);
       checker.registerThread(thread, thread);
-      // Vert.x threads are NOT daemons - we want them to prevent JVM exit so embedded user doesn't
```
Apache9, line 1129: Or maybe we could just change the code here, to check the daemon option in VertxOptions, to reset the daemon flag of the thread. This could keep the old behavior if users do not change anything, and could also make the thread daemon if users set daemon to true in VertxOptions. WDYT? @vietj 

Thanks.

vietj:
just the flag, I can't see any other thread options we would like to
use in practice.

Apache9:
Thanks @vietj . The daemon option has already been fixed.

The problem now is the current implementation does not work in real world, as the init method will never be called...

I think a possible solution is to change the code in VertxImpl, where we used to reset the daemon flag to false. We can change it to reset the flag to the value of the daemon field in the VertxOptions.

https://github.com/eclipse-vertx/vert.x/blob/bbb4ddc417a604697d2ff4f5a8eda2e74113697e/src/main/java/io/vertx/core/impl/VertxImpl.java#L1131

WDYT?

Thanks.

Apache9:
@vietj Please take a look at the new approach? Just change the code where we used to force daemon to be false, to consider whether we want the thread to be daemon by checking the daemon flag in VertxOptions.

If you think this is acceptable, I will try to implement UTs then.

Thanks.

Apache9:
@vietj I've added some tests. PTAL?

Thanks.

vietj:
yeah

PR #4480, opened 2022-09-12
Reusing existing Netty types to speedup http encoding and save scalability issues

franz1981:
This can be completed only with:
- proper benchmarking
- https://github.com/netty/netty/pull/12796 is completed


src/main/java/io/vertx/core/http/impl/Http1xServerResponse.java
```diff
@@ -745,14 +744,14 @@ private Http1xServerResponse write(ByteBuf chunk, PromiseInternal<Void> promise)
         }
       }
       bytesWritten += chunk.readableBytes();
-      HttpObject msg;
       if (!headWritten) {
         prepareHeaders(-1);
-        msg = new AssembledHttpResponse(head, version, status, headers, chunk);
+        conn.writeToChannel(new VertxHttpResponse(head, version, status, headers));
```
franz1981, line 749: this is the sole case I'm worried about @vietj

vietj:
Netty has been bumped to 4.1.82.Final @franz1981

vietj:
can you update and make a proper PR ?

src/main/java/io/vertx/core/http/impl/Http1xServerResponse.java
vietj: could Netty have a similar default type ?

franz1981: With the `head` element in?

vietj: with the headers+chunk

franz1981:
Let me close this, no longer need it

PR #4490, opened 2022-09-20
ListenerArray throuhput and allocation improvements

tsegismont:
Use our own elements array with small initial size instead of extending `ArrayList` and using default capacity (10).

Running `FutureListenerBenchmark` before and after this change shows:

* 30% throughput increase and 15% or 12%  memory allocation decrease (per op)
  when adding 2 or 4 listeners
* 6% throughput increase and 6% memory allocation increase (per op) when
  adding 6 listeners
* 1.5% throughput decrease and 24% memory allocation increase (per op)
  when adding 8 listeners

Note that:

* the previous version was tested unchanged, i.e. with `ListenerArray`
  extending `ArrayList` and creating a list with the default capacity (10)
* in many cases, futures get two listeners (`onSuccess`/`onFailure`)

### DIFF
--- a/src/test/benchmarks/io/vertx/benchmarks/FutureListenerBenchmark.java
+++ b/src/test/benchmarks/io/vertx/benchmarks/FutureListenerBenchmark.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.benchmarks;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import org.openjdk.jmh.annotations.*;
+
+import static org.openjdk.jmh.annotations.CompilerControl.Mode.DONT_INLINE;
+
+@State(Scope.Thread)
+public class FutureListenerBenchmark extends BenchmarkBase {
+
+  @Param({"2", "4", "6", "8"})
+  int listenerCount;
+
+  @CompilerControl(DONT_INLINE)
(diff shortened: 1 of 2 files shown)


tsegismont:
cc @franz1981 (in case you got time to review this short PR)

src/test/benchmarks/io/vertx/benchmarks/FutureListenerBenchmark.java
```diff
@@ -0,0 +1,40 @@
...
+  @CompilerControl(DONT_INLINE)
+  public static void consume(AsyncResult<Void> ar) {
+  }
+
+  @Benchmark
+  public void addListeners() throws Exception {
+    Promise<Void> promise = Promise.promise();
+    Future<Void> future = promise.future();
+    for (int i = 0; i < listenerCount; i++) {
+      future.onComplete(FutureListenerBenchmark::consume);
```
franz1981, line 36: this could allocate consume

src/test/benchmarks/io/vertx/benchmarks/FutureListenerBenchmark.java
```diff
@@ -0,0 +1,40 @@
...
+import static org.openjdk.jmh.annotations.CompilerControl.Mode.DONT_INLINE;
+
+@State(Scope.Thread)
+public class FutureListenerBenchmark extends BenchmarkBase {
+
+  @Param({"2", "4", "6", "8"})
+  int listenerCount;
+
+  @CompilerControl(DONT_INLINE)
+  public static void consume(AsyncResult<Void> ar) {
```
franz1981, line 28: Or you can add a field to the benchmark

```java
@State(Scope.Thread)
public class FutureListenerBenchmark extends BenchmarkBase implements Handler<AsyncResult<Void>> {

private int consumed;

public void handle(Void event) {
    consumed++;   
}
```
and return it on `addListeners()`:
```java
@Benchmark
  public int addListeners() throws Exception {
    Promise<Void> promise = Promise.promise();
    Future<Void> future = promise.future();
    consumed = 0;
    for (int i = 0; i < listenerCount; i++) {
      future.onComplete(this);
    }
    promise.complete();
    return consumed;
  }
```
At this point, for clarity, you can define a separated per-thread `State` class to be passed to `addListeners` as argument; will make easier to understand that its a per-thread context (it will implements Handler too)

franz1981 (approved):

PR #4494, opened 2022-09-23
Add GlobalTrafficShapingHandler to server pipeline for bandwidth limiting when bandwidthLimited is set to True in server options

sarankk:
Co-authored-by: "Dinesh A. Joshi" <dinesh.joshi@apple.com>

Motivation:
We want to add traffic shaping capability to server created by createHttpServer API, with the help of `GlobalTrafficShapingHandler` already existing in Netty. By adding this feature, we are able to create outbound and inbound traffic shaped server, instead of having to do any external rate limiting. 

**Advantage of adding this feature**
This ensures server created with `createHttpServer` API will always have enough bandwidth to run efficiently. 

**Contribution**
To do this, our thought process is to add options to `HttpServerOptions` that will allow us to enable traffic shaping while creating server.


vietj (changes requested):
- it looks like traffic shaping has more configuration than only `outboundGlobalBandwidth` so we should have all the available options as a `TrafficShapingOptions` object instead.

- this option should be added to NetServer options too, so it works generically on TCP servers as well.

vietj:
ping @sarankk can you contribute the changes required ?

sarankk:
@vietj yes will update the PR by this week

vietj:
thanks

sarankk:
@vietj updated the PR with `TrafficShapingOptions` object and also added it to `NetServerOptions`. Do we do channel initialization for `NetServer` separately? I couldn't find the pipelines for it, to add `GlobalTrafficShapingHandler`. Thanks for the reviews!

vietj:
let me have a look @sarankk

src/main/java/io/vertx/core/net/NetServerOptions.java
```diff
@@ -70,6 +70,16 @@ public class NetServerOptions extends TCPSSLOptions {
    */
   public static final TimeUnit DEFAULT_PROXY_PROTOCOL_TIMEOUT_TIME_UNIT = TimeUnit.SECONDS;
 
+  /*
+   * Default inbound bandwidth limit in bytes/sec = 0 (0 implies unthrottled)
+   */
+  public static final long DEFAULT_INBOUND_GLOBAL_BANDWIDTH_LIMIT = 0;
```
vietj, line 76: this should be moved to `TrafficShapingOptions`

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -0,0 +1,75 @@
...
+  private long peakOutboundGlobalBandwidth;
+  private long maxDelayToWaitTime;
+  private long checkIntervalForStats;
+
+  public TrafficShapingOptions(long inboundGlobalBandwidth, long outboundGlobalBandwidth) {
+    this.inboundGlobalBandwidth = inboundGlobalBandwidth;
+    this.outboundGlobalBandwidth = outboundGlobalBandwidth;
+  }
+
+  public void setInboundGlobalBandwidth(long inboundGlobalBandwidth) {
```
vietj, line 33: should be fluent and return `TrafficShapingOptions` also missing javadoc. That is valid for other setters of this class.

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+@DataObject(generateConverter = true, publicConverter = false)
+public class TrafficShapingOptions {
+  private long inboundGlobalBandwidth;
+  private long outboundGlobalBandwidth;
+  private long peakOutboundGlobalBandwidth;
+  private long maxDelayToWaitTime;
+  private long checkIntervalForStats;
+
+  public TrafficShapingOptions(long inboundGlobalBandwidth, long outboundGlobalBandwidth) {
```
vietj, line 28: instead we use an empty constructor + a json constructor + a copy constructor

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -0,0 +1,75 @@
...
+    return peakOutboundGlobalBandwidth;
+  }
+
+  public long getMaxDelayToWaitTime() {
+    return maxDelayToWaitTime;
+  }
+
+  public long getCheckIntervalForStats() {
+    return checkIntervalForStats;
+  }
```
vietj, line 74: missing a toJson() method that can be delegated to the generated converter

vietj (changes requested):
a couple of minor changes and we are good to go for 4.4 I think

vietj:
can you rebase on latest master ? there is a conflict

sarankk:
@vietj Thanks for the comments! :), updated the PR

sarankk:
@vietj I resolved conflicts once again, when you have some time can you please take another look, Thanks

sarankk:
@vietj any updates?

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -0,0 +1,259 @@
...
+import io.vertx.core.buffer.Buffer;
+import io.vertx.test.core.TestUtils;
+
+@RunWith(Parameterized.class)
+public class HttpBandwidthLimitingTest extends Http2TestBase {
+  private static final int OUTBOUND_LIMIT = 64 * 1024;  // 64KB/s
+  private static final int INBOUND_LIMIT = 64 * 1024;   // 64KB/s
+  private static final int TEST_CONTENT_SIZE = 64 * 1024 * 4;   // 64 * 4 = 256KB
+
+  private final File sampleF = new File("src/test/resources/test_traffic.txt");
```
vietj, line 41: instead use the constant `TestUtils.MAVEN_TARGET_DIR` as base directory and then the resource should be in `classes/test_traffic.txt` , you should replace `/` by the file separator so test can run on windows

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -0,0 +1,259 @@
...
+          assertEquals(expected.length(), req.bytesRead());
+          testComplete();
+        });
+      };
+    }
+  }
+
+  static class Providers {
+    private static HttpServer http1Server(Vertx vertx, int inboundLimit, int outboundLimit) {
+      int openPort = findFreePort();
```
vietj, line 229: you can listen on port 0 that will give you a randomly available port, then the port can be retrieved using `server.actualPort()`

vietj:
@sarankk ping

sarankk:
@vietj thanks for the comments, updated patch

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -0,0 +1,153 @@
...
+package io.vertx.core.net;
+
+import io.netty.util.internal.ObjectUtil;
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Options describing how {@link io.netty.handler.traffic.GlobalTrafficShapingHandler} will handle traffic shaping.
+ */
+@DataObject(generateConverter = true, publicConverter = false)
```
vietj, line 21: this should be annotated with `io.vertx.codegen.annotations.Unstable` as this might change in the future. I can see several traffic shaping strategies and perhaps we should have the ability to configure also per channel option. We can do this in a follow up PR.

vietj:
The feature does not seem to support TCP server sharing (https://vertx.io/docs/vertx-core/java/#_server_sharing), that is we will have global traffic shaping per event loop. 

We should test this case and make sure that the same global traffic handler is shared among servers.

Here is an example of a such test that uses a verticle to use several event-loops:

```java
    int numEventLoops = 4; // We start a shared TCP server with 4 event-loops
    Future<String> listenLatch = vertx.deployVerticle(() -> new AbstractVerticle() {
      private HttpServer server;
      @Override
      public void start(Promise<Void> startPromise) {
        server = vertx.createHttpServer(new HttpServerOptions().setPort(DEFAULT_HTTP_PORT));
          // Handle request
        }).listen(testAddress).<Void>mapEmpty().onComplete(startPromise);
      }
    }, new DeploymentOptions().setInstances(numEventLoops));
    awaitFuture(listenLatch);
```

sarankk:
@vietj Thanks for pointing it out. Added support for traffic shaping with shared servers and tested it.

src/main/java/io/vertx/core/http/impl/HttpServerWorker.java
```diff
@@ -270,7 +277,7 @@ private void configureHttp1OrH2C(ChannelPipeline pipeline, SslChannelProvider ss
     if (options.isCompressionSupported()) {
       pipeline.addLast("deflater", new HttpChunkContentCompressor(compressionOptions));
     }
-    if (options.isSsl() || options.isCompressionSupported() || !vertx.transport().supportFileRegion()) {
+    if (options.isSsl() || options.isCompressionSupported() || !vertx.transport().supportFileRegion() || (options.getTrafficShapingOptions() != null && options.getTrafficShapingOptions().getOutboundGlobalBandwidth() > 0)) {
```
vietj, line 280: here also, I think we can test the nullity of the trafficShapingHandler field

src/main/java/io/vertx/core/net/impl/NetServerImpl.java
```diff
@@ -231,6 +236,9 @@ private void configurePipeline(Channel ch, SslChannelProvider sslChannelProvider
       } else {
         connected(ch, sslChannelProvider);
       }
+      if (options.getTrafficShapingOptions() != null && (options.getTrafficShapingOptions().getOutboundGlobalBandwidth() > 0 || options.getTrafficShapingOptions().getInboundGlobalBandwidth() > 0)) {
```
vietj, line 239: like for HTTP, I think we can check the trafficShapingHandler field

vietj (changes requested):
A couple of minor comments, but we also need documentation in `net.adoc` document for this

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Default outbound bandwidth limit in bytes/sec = 0 (0 implies unthrottled)
+   */
+  public static final long DEFAULT_OUTBOUND_GLOBAL_BANDWIDTH_LIMIT = 0;
+
+  private long inboundGlobalBandwidth;
+  private long outboundGlobalBandwidth;
+  private long peakOutboundGlobalBandwidth;
+  private long maxDelayToWaitTime;
```
vietj, line 38: we need a `maxDelayToWaitTimeUnit` as well that is a `TimeUnit`, e.g we do have in `TCPSSLOptions` `idleTimeout` and `idleTimeoutUnit`

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  public JsonObject toJson() {
+    JsonObject json = new JsonObject();
+    TrafficShapingOptionsConverter.toJson(this, json);
+    return json;
+  }
+
+  /**
+   * Set bandwidth limit for inbound connections
```
vietj, line 63: we should say that it is bytes

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * Default outbound bandwidth limit in bytes/sec = 0 (0 implies unthrottled)
+   */
+  public static final long DEFAULT_OUTBOUND_GLOBAL_BANDWIDTH_LIMIT = 0;
+
+  private long inboundGlobalBandwidth;
+  private long outboundGlobalBandwidth;
+  private long peakOutboundGlobalBandwidth;
+  private long maxDelayToWaitTime;
+  private long checkIntervalForStats;
```
vietj, line 39: we also need a `TimeUnit` as well here

src/main/java/io/vertx/core/net/impl/NetServerImpl.java
```diff
@@ -259,8 +267,8 @@ protected void initChannel(ChannelPipeline pipeline, boolean ssl) {
     if (options.getLogActivity()) {
       pipeline.addLast("logging", new LoggingHandler(options.getActivityLogDataFormat()));
     }
-    if (ssl || !vertx.transport().supportFileRegion()) {
-      // only add ChunkedWriteHandler when SSL is enabled or FileRegion isn't supported
+    if (ssl || !vertx.transport().supportFileRegion() || (options.getTrafficShapingOptions() != null && options.getTrafficShapingOptions().getOutboundGlobalBandwidth() > 0)) {
```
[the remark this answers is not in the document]
vietj: thanks

src/main/asciidoc/net.adoc
```diff
@@ -349,6 +349,16 @@ name. When using network activity logging there are a few things to keep in mind
 
 You should read the <<netty-logging>> section.
 
+=== Throttling inbound and outbound bandwidth of TCP connections
+
+TCP servers (Net/Http) can be configured with traffic shaping options to enable bandwidth limiting. Both inbound and outbound
+bandwidth can be limited through {@link io.vertx.core.net.TrafficShapingOptions}. For NetServers, traffic shaping options can be set
+through {@link io.vertx.core.net.NetServerOptions} and for HttpServers it can be set through {@link io.vertx.core.http.HttpServerOptions}.
```
vietj, line 356: typo HttpServer is the name of the class, so either HttpServer or HTTP servers

src/main/asciidoc/net.adoc
```diff
@@ -349,6 +349,16 @@ name. When using network activity logging there are a few things to keep in mind
 
 You should read the <<netty-logging>> section.
 
+=== Throttling inbound and outbound bandwidth of TCP connections
+
+TCP servers (Net/Http) can be configured with traffic shaping options to enable bandwidth limiting. Both inbound and outbound
+bandwidth can be limited through {@link io.vertx.core.net.TrafficShapingOptions}. For NetServers, traffic shaping options can be set
```
vietj, line 355: typo NetServer is the name of the class, so either NetServer or TCP servers

vietj (changes requested):
couple of minor issues

vietj (approved):

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -0,0 +1,289 @@
...
+    testServer.requestHandler(HANDLERS.bufferRead(expectedBuffer));
+    startServer(testServer);
+
+    watch.start();
+    HttpClient testClient = clientFactory.apply(vertx);
+    read(expectedBuffer, testServer, testClient);
+    await();
+    watch.stop();
+
+    assertTrue(watch.getTime() > expectedTimeMillis(TEST_CONTENT_SIZE, OUTBOUND_LIMIT));
```
vietj, line 100: we need here to use `Assert.assertTrue(...)` instead of `assert(...)` because `await()` was called before and the async test supposes the test is finished and no error can be reported

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -0,0 +1,289 @@
...
+    testServer.requestHandler(HANDLERS.bufferWrite(expectedBuffer));
+    startServer(testServer);
+
+    watch.start();
+    HttpClient testClient = clientFactory.apply(vertx);
+    write(expectedBuffer, testServer, testClient);
+    await();
+    watch.stop();
+
+    assertTrue(watch.getTime() > expectedTimeMillis(TEST_CONTENT_SIZE, INBOUND_LIMIT));
```
vietj, line 137: we need here to use `Assert.assertTrue(...)` instead of `assert(...)` because `await()` was called before and the async test supposes the test is finished and no error can be reported

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -0,0 +1,289 @@
...
+    testServer.requestHandler(HANDLERS.uploadFile(sampleF));
+    startServer(testServer);
+
+    watch.start();
+    HttpClient testClient = clientFactory.apply(vertx);
+    upload(testServer, testClient, sampleF);
+    await();
+    watch.stop();
+
+    assertTrue( watch.getTime() > 1000);
```
vietj, line 153: we need here to use `Assert.assertTrue(...)` instead of `assert(...)` because `await()` was called before and the async test supposes the test is finished and no error can be reported

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -0,0 +1,289 @@
...
+          HttpClientRequest result = req.result();
+          result.send(resp -> {
+            resp.result().bodyHandler(r -> waitForResponse.countDown());
+          });
+        });
+      }
+    });
+    awaitLatch(waitForResponse);
+    watch.stop();
+    assertTrue(watch.getTime() > 6000); // because there are simultaneous 4 requests
```
vietj, line 184: we need here to use `Assert.assertTrue(...)` instead of `assert(...)` because `await()` was called before and the async test supposes the test is finished and no error can be reported

vietj (changes requested):
compilation issues : the code seems to use callbacks

vietj:
@sarankk it seems that the CI for some configuration (native transports) because the time measured is a bit shorter than the expected time, I added debug logging to check it and ran the CI

https://github.com/vietj/vert.x/actions/runs/5130215652/jobs/9228941586

```
java.lang.AssertionError: Was expecting 997 > 1000
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at io.vertx.core.http.HttpBandwidthLimitingTest.sendFileIsThrottled(HttpBandwidthLimitingTest.java:121)
```

So we should consider either there is a bug or be more tolerant when doing the comparison.

vietj:
still failures happening in CI @sarankk

sarankk:
> still failures happening in CI @sarankk
@vietj Pushed a commit to decrease bandwidth, so that expected time is not as low as 1sec. This with being more tolerant with time range fixes the issue

vietj:
@sarankk I'll have a look myself

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -0,0 +1,299 @@
...
+   *
+   * @param size
+   * @param rate
+   * @return
+   */
+  private long expectedTimeMillis(long size, int rate) {
+    return (long) (TimeUnit.MILLISECONDS.convert(( size / rate), TimeUnit.SECONDS) * 0.5); // multiplied by 0.5 to be more tolerant of time pauses during CI runs
+  }
+
+  private void read(Buffer expected, HttpServer server, HttpClient client) {
```
vietj, line 212: this code is racy (can be observed on windows CI) because the HTTP response can be received after the response handler is set and lose buffers (see https://vertx.io/docs/vertx-core/java/#_request_and_response_composition), instead should be

```java
    client.request(HttpMethod.GET, server.actualPort(), DEFAULT_HTTP_HOST,"/buffer-read")
          .compose(req -> req
            .send()
            .andThen(onSuccess(resp -> assertEquals(200, resp.statusCode())))
            .compose(HttpClientResponse::body))
          .onComplete(onSuccess(body -> {
            assertEquals(expected.getBytes().length, body.getBytes().length);
            testComplete();
          }));
```

src/test/java/io/vertx/core/net/NetBandwidthLimitingTest.java
```diff
@@ -0,0 +1,308 @@
...
+    server.connectHandler(sock -> {
+      sock.handler(buf -> {
+        sock.write(expected);
+      });
+    });
+    Future<NetServer> result = server.listen(testAddress);
+    result.onComplete(ar -> {
+      assertTrue(ar.succeeded());
+      Future<NetSocket> clientConnect = client.connect(testAddress);
+      clientConnect.onComplete(ar2 -> {
```
vietj, line 89: we can simplify with the `onSuccess` static method of `AsyncTestBase`:

```java
clientConnect.onComplete(onSuccess(sock - > {
  ...
}));
```

vietj (changes requested):
A few change related to racy test (failed on windows) and also we can use `onSuccess` to simplify the tested code with the idiom

`onComplete(onSuccess(result -> ...))`

instead of

```
  ...onComplete(ar -> {
    assertTrue(ar.succeeded());
    T result = ar.result();
  });
```

sarankk:
> A few change related to racy test (failed on windows) and also we can use `onSuccess` to simplify the tested code with the idiom
> 
> `onComplete(onSuccess(result -> { ... }))`
> 
> instead of
> 
> ```
>   ...onComplete(ar -> {
>     assertTrue(ar.succeeded());
>     T result = ar.result();
>   });
> ```

Thanks, updated with your suggestion. The windows pipeline pass for me

vietj (changes requested):
Can you also rewrite the onComplete statements that must succeed with onSuccess ?

I still see a few onComplete/succeeded patterns.

sarankk:
> Can you also rewrite the onComplete statements that must succeed with onSuccess ?
> 
> I still see a few onComplete/succeeded patterns.

Thanks @vietj updated, also rebased to resolve conflicts and squashed commits.

sarankk:
> Can you also rewrite the onComplete statements that must succeed with onSuccess ?
> 
> I still see a few onComplete/succeeded patterns.
Thanks @vietj updated PR. The tests pass on my fork.

vietj:
Thanks for the contribution @sarankk and @dineshjoshi

PR #4533, opened 2022-11-10
Add convenient Trust/KeyCertOptions wrap() methods

alekkol:
TrustOptions class already contains a static `wrap()` method to create an instance using the provided `TrustManager` instance. The same is true for `KeyCertOptions.wrap(KeyManager)`. If we want to use several trust or key managers, we need to deal with `Trust/KeyManagerFactory`.
This PR contains 2 more convenient `wrap()` methods which receive `TrustManagerFactory` or `KeyManagerFactory`. The changes are pretty compact so far as`Trust/KeyManagerFactoryOptions` already have suitable package-private constructors.


vietj:
thanks for the contribution

Issue #4499, opened 2022-10-02
MimeMapping.getMimeTypeForExtension should be Case Insensitive


leotu:
### Issue

Static file response no content type if extension name contains upper case characters

[StaticHandlerImpl.java#L463](https://github.com/vert-x3/vertx-web/blob/efc35f2caeedc135779895b9008268cdcbdb9042/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/StaticHandlerImpl.java#L463)

### Version

4.2.7

### Do you have a reproducer?
[MimeMapping.java#L1016](https://github.com/eclipse-vertx/vert.x/blob/11ab144725e551be591118831377101015e1392c/src/main/java/io/vertx/core/http/impl/MimeMapping.java#L1016)

### Steps to reproduce

```java
@TestMethodOrder(OrderAnnotation.class)
public class MimeMappingTester {

    @Test
    @Order(1)
    void textLocalFile() {
        Path file = Path.of("/Users/leo/Pictures/IMG_4157.JPG");
        System.out.println(String.format("'%s', contentType: %s", file, MimeMapping.getMimeTypeForFilename(file.toString())));

        Path file2 = Path.of("/Users/leo/Pictures/IMG_4157.jpg");
        System.out.println(String.format("'%s', contentType: %s", file2, MimeMapping.getMimeTypeForFilename(file2.toString())));
    }

    @Test
    @Order(2)
    void textUrlFile() {
        try {
            URL url = new URL("https://xgos.com/Pictures/IMG_4157.JPG");
            System.out.println(String.format("'%s', contentType: %s", url, MimeMapping.getMimeTypeForFilename(url.toString())));

            URL url2 = new URL("https://xgos.com/Pictures/IMG_4157.jpg");
            System.out.println(String.format("'%s', contentType: %s", url2, MimeMapping.getMimeTypeForFilename(url2.toString())));
        } catch (MalformedURLException e) {
            e.printStackTrace();
        }
    }
}
```

### Console Output
```console
'/Users/leo/Pictures/IMG_4157.JPG', contentType: null
'/Users/leo/Pictures/IMG_4157.jpg', contentType: image/jpeg

'https://xgos.com/Pictures/IMG_4157.JPG', contentType: null
'https://xgos.com/Pictures/IMG_4157.jpg', contentType: image/jpeg
```

PR #4545, opened 2022-11-19
MimeMapping.getMimeType should be case insensitive 

pigbayspy:
Motivation:

- MimeMapping.getMimeType should be case insensitive  https://github.com/eclipse-vertx/vert.x/issues/4499
- add MimeMapping's test

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/MimeMapping.java
+++ b/src/main/java/io/vertx/core/http/impl/MimeMapping.java
@@ -1014,7 +1014,10 @@ public class MimeMapping {
   }
 
   public static String getMimeTypeForExtension(String ext) {
-    return m.get(ext);
+    if (ext == null) {
+      return null;
+    }
+    return m.get(ext.toLowerCase());
   }
   public static String getMimeTypeForFilename(String filename) {
     int li = filename.lastIndexOf('.');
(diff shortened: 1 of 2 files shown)


pmlopes (approved):

zll600:
@vietj Could you take a look?

src/main/java/io/vertx/core/http/impl/MimeMapping.java
```diff
@@ -1014,7 +1014,10 @@ public class MimeMapping {
   }
 
   public static String getMimeTypeForExtension(String ext) {
-    return m.get(ext);
+    if (ext == null) {
+      return null;
+    }
+    return m.get(ext.toLowerCase());
```
vietj, line 1020: use HttpUtils#toLowerCase here instead, it is more performant

src/test/java/io/vertx/core/http/MimeTypeTest.java
zll600, line 1: how about `MimeTypeCaseInsensitiveTest.java`

src/main/java/io/vertx/core/http/impl/MimeMapping.java
pigbayspy: get it

pigbayspy:
- changed to use `HttpUtils` to lowercase string
- rename test  class name

PR #4552, opened 2022-11-28
Openssl no longer has much better performance

julianladisch:
https: //www.oracle.com/java/technologies/javase/8-whats-new.html "Hardware intrinsics were added to use Advanced Encryption Standard (AES)."
https: //openjdk.org/jeps/164 "JEP 164: Leverage CPU Instructions for AES Cryptography"
https: //docs.oracle.com/javase/9/whatsnew/#d1373e94 "Improves performance ranging from 34x to 150x for AES/GCM/NoPadding"
https: //openjdk.org/jeps/246 "JEP 246: Leverage CPU Instructions for GHASH and RSA"


vietj:
thanks.

I think also we should improve the documentation and say that one of the interest to use OpenSSL is to delegate the SSL maintenant to the OS instead of the JDK.

Issue #4572, opened 2023-01-05
Invalid header name [X-B3-TraceId] when WebClient requests with HTTP2 and OpenTelemetry B3Propagator tracing options


hsrky:
### Issue

"Invalid header name [X-B3-TraceId]" exception thrown when the WebClient request a same URL for second time. The WebClient is configured with HTTP2 and OpenTelemetry with B3Propagator.

### Version

4.3.7

### Context

I am getting the "Invalid header name [X-B3-TraceId]" exception when the WebClient request a same URL for second time. The WebClient is configured with HTTP2 and OpenTelemetry with B3Propagator.

The HTTP2 protocol requires all the headers to be in lower case. First WebClient request able to convert the tracing propagator headers to lower case, however the WebClient failed to do so on second request onward.

Exceptions:

> io.netty.handler.codec.http2.Http2Exception: invalid header name [X-B3-TraceId]
> 	at io.netty.handler.codec.http2.Http2Exception.connectionError(Http2Exception.java:109)
> 	at io.netty.handler.codec.http2.DefaultHttp2Headers$2.validateName(DefaultHttp2Headers.java:71)
> 	at io.netty.handler.codec.http2.DefaultHttp2Headers$2.validateName(DefaultHttp2Headers.java:42)
> 	at io.netty.handler.codec.DefaultHeaders.validateName(DefaultHeaders.java:1012)
> 	at io.netty.handler.codec.http2.DefaultHttp2Headers.validateName(DefaultHttp2Headers.java:166)
> 	at io.netty.handler.codec.http2.DefaultHttp2Headers.validateName(DefaultHttp2Headers.java:33)
> 	at io.netty.handler.codec.DefaultHeaders.add(DefaultHeaders.java:329)
> 	at io.vertx.tracing.opentelemetry.HeadersPropagatorSetter.set(HeadersPropagatorSetter.java:24)
> 	at io.vertx.tracing.opentelemetry.HeadersPropagatorSetter.set(HeadersPropagatorSetter.java:17)
> 	at io.opentelemetry.extension.trace.propagation.B3PropagatorInjectorMultipleHeaders.inject(B3PropagatorInjectorMultipleHeaders.java:52)
> 	at io.opentelemetry.extension.trace.propagation.B3Propagator.inject(B3Propagator.java:111)
> 	at io.vertx.tracing.opentelemetry.OpenTelemetryTracer.sendRequest(OpenTelemetryTracer.java:130)
[...]

PR #4573, opened 2023-01-05
Tracer call for HTTP Client should use headers adaptors

tsegismont:
See #4572

Otherwise, the tracing implementation may try to add invalid header name/value directly to Netty headers.


tsegismont:
@vietj ok to merge? Waiting for this before backporting to the `4.3` branch.

vietj (approved):
LGTM

Issue #4582, opened 2023-01-10
Regression in Kotlin Compatibility of SLF4JLogDelegate from 4.3.4


hjed:
### Version

4.3.4 onwards (tested up to 4.3.7)

### Context

When using a kotlin implementation of org.slf4j.Logger from version 4.3.7 startup fails since vert.x passes null values to varargs which isn't allowed in Kotlin. 

I think the change that caused startup to fail is:
https://github.com/eclipse-vertx/vert.x/commit/6f71b5f2adfb440b554ec9979ff3fdfd7830e985

But the solution is that https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/logging/SLF4JLogDelegate.java#L163 ideally shouldn't pass an explicitly null value to the slf4j logger (an empty array would work). 

I realise that this is really an incompatibility in the SLF4J interface, but fixing this in vert.x would be great for anyone using a Koltin implementation of the SLF4J Logger API.

### Do you have a reproducer?
A noop logger written in kotlin like the one below would fail:

```kotlin 
package org.slf4j.impl

import org.slf4j.Logger
import org.slf4j.helpers.MarkerIgnoringBase
import org.slf4j.helpers.MessageFormatter
import java.io.PrintStream
import java.util.stream.Collectors

class NoopLogger(): Logger, MarkerIgnoringBase() {

    /** Are `trace` messages currently enabled?  */
    override fun isTraceEnabled(): Boolean {
        
    }

    /**
     * A simple implementation which logs messages of level TRACE according to
     * the format outlined above.
     */
    override fun trace(msg: String) {
        
    }

    /**
     * Perform single parameter substitution before logging the message of level
     * TRACE according to the format outlined above.
     */
    override fun trace(format: String, param1: Any?) {
        
    }

    /**
     * Perform double parameter substitution before logging the message of level
     * TRACE according to the format outlined above.
     */
    override fun trace(format: String, param1: Any?, param2: Any?) {
        
    }

    /**
     * Perform double parameter substitution before logging the message of level
     * TRACE according to the format outlined above.
     */
    override fun trace(format: String, vararg argArray: Any?) {
        
    }

    /** Log a message of level TRACE, including an exception.  */
    override fun trace(msg: String, t: Throwable?) {
        
    }

    /** Are `debug` messages currently enabled?  */
    override fun isDebugEnabled(): Boolean {
        return true
    }

    /**
[...]

PR #4586, opened 2023-01-11
Avoid invoking SLF4J logger API with null parameters

tsegismont:
See #4582

Otherwise, we cannot delegate to loggers written in Kotlin (which does not allow null variable arguments).


vietj:
can we have a test for that ?

tsegismont:
We already have, it's only the implementation which changes

vietj:
so tests were failing ?

vietj:
ok I see

tsegismont:
> so tests were failing ?

Only the windows build was failing, for some unrelated reason.

tsegismont:
> ok I see

Ok to merge and backport then?

vietj (approved):

PR #4592, opened 2023-01-25
Add maven wrapper

ardetrick:
Motivation:

Add the [Maven Wrapper](https://maven.apache.org/wrapper/) to the repository and update GitHub actions and documentation to use the Maven Wrapper instead of a system installed Maven.

Using the maven wrapper is a best practice and promotes consistent builds. This is because everyone running maven commands are using the same maven version (I have also updated the GitHub actions). It also reduces the number of dependencies required to run because there is no longer a need to install Maven on the system.

Note that this is not using the [Takari Maven Wrapper](https://github.com/takari/maven-wrapper) and that Apache has taken over that project.

See [this build](https://github.com/ardetrick/vert.x/actions/runs/4002406422) for passing tests.

Conformance:

- [X] You should have signed the Eclipse Contributor Agreement as explained in https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md
- [X] Please also make sure you adhere to the code style guidelines: https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines


vietj:
the main issue I see is the maintenance of the property file for this. The thing is that if we were going to do that, then it means also updating all the projects from the vertx stack to be consistent and then keeping up to date all projects at once.

ardetrick:
Hi Julien, thanks for your quick response. I'm wondering if you can expand on some of your concerns. Best practices have really shifted towards using a Maven/Gradle wrapper whenever possible and I want to make sure I have all the context with regards to the needs this project.

The maintenance cost of the properties file should be relatively low. In the event we ever want to update the maven version used we run a command like this: `mvn wrapper:wrapper -Dmaven=X.Y.Z` (I would be happy to include some documentation around this). 

In fact, having the maven version explicitly declared is an advantage. When I first cloned the repository I noticed tests were not passing on my machine because I had an old maven version. The documentation for the project in `BUILDING.md` suggested that any version `3.2.x` and above should work. When I looked at the GitHub action, I saw no declaration on what maven version the build was using, and digging into the GitHub documentation I couldn't find that explicitly defined anywhere either. Eventually I ran the GitHub action with a new step to log the maven version. That was quite a bit of effort required to simply run tests. With the wrapper, that concern is eliminated.

>  it means also updating all the projects from the vertx stack to be consistent and then keeping up to date all projects at once

Is there are a requirement that all Vertx project use the exact same version at all times? I understand that it would be ideal and I also understand that there would be overhead whenever a new maven version is desired, but the tradeoff of shifting towards more reproducible builds and best practices of our tooling seems worth it. At this moment, we are relying on a working maven version by coincidence not by design. What happens if maven 4 is released and the Vert.x build is not compatible but the GitHub runner upgrades to Maven 4? For better or worse, Maven releases are uncommon.

Thank you for taking the time to review my PR.

PR #4595, opened 2023-01-30
Connection flush policy 2

vietj:
Investigate alternate flush policy.

The flush policy performs a flush when all Netty messages have been read.

This attempts to flush when all Netty messages have been processed instead of being read. Flush also happens when the message processing has stalled (pause) or when the outbound buffer is full.


vietj:
superseded by outbound write queue

PR #4596, opened 2023-02-01
JsonParser should process fully the possible events when processing malformed input

vietj:
The JsonParser class does not progress as much as it could when the processing faulty input.

When the JsonParser process malformed input, it does halt the parsing and handle the error leaving unprocessed tokens in the underlying Jackson parser. These tokens will be processed when the JsonParser handle another input. When the Jackson parser fails we handle the error and then give up and reported behavior is a consequence.


adamdickmeiss:
This change changes behavior .. Seems like exceptionHandler can be called after endHandler. Thus, no clear indication on when a document is parsed and valid. Need a reproducer?

PR #4605, opened 2023-02-08
Add intervallic timer methods for Vertx

liuchong:
Motivation:

In many usages of my scenarios, I will prefer to a fixed delay timer instead of a fixed rate one.

I've added an implementation of it, please check and give some feedback.

I used the word `interval`, say the methods are `setInterval` and `intervalStream`, if we need change it to a better one, please give your suggestions.

Please check it @vietj @pmlopes @tsegismont 👀 If we decide to add these methods, I will add some tests before merge.

Thanks!

### DIFF
--- a/src/main/java/io/vertx/core/Vertx.java
+++ b/src/main/java/io/vertx/core/Vertx.java
@@ -324,6 +324,55 @@ public interface Vertx extends Measured {
   TimeoutStream periodicStream(long initialDelay, long delay);
 
   /**
+   * Set a interval timer to fire after {@code delay} milliseconds and after initial delay and after
+   * every of the previous handler finished, at which point {@code handler} will be called with
+   * the id of the timer.
+   *
+   * @param delay  the delay in milliseconds, after which the timer will fire
+   * @param handler  the handler that will be called with the timer ID when the timer fires
+   * @return the unique ID of the timer
+   */
+  default long setInterval(long delay, Handler<Long> handler) {
+    return setInterval(delay, delay, handler);
+  }
+
+  /**
+   * Set a interval timer to fire after {@code delay} milliseconds and after initial delay and after
+   * every of the previous handler finished, at which point {@code handler} will be called with
+   * the id of the timer.
+   *
+   * @param initialDelay the initial delay in milliseconds
+   * @param delay        the delay in milliseconds, after which the timer will
+   *                     fire
+   * @param handler      the handler that will be called with the timer ID when
+   *                     the timer fires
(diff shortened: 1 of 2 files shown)


liuchong:
WIthout the methods of this pull request, I used some other methods for this requirement.

I did not found a very good method for the async version, but I think we could convert a future into a sync one with calling the get method when we need.

Here are the code snippets:

```java
import io.vertx.core.Future;
import io.vertx.core.Handler;
import io.vertx.core.Vertx;

/** TimerUtil */
public class TimerUtil {

  /** Run a job with a delay after the previous one finished. */
  public static <T> void setInterval(Vertx vertx, long delay, Handler<Long> handler) {
    vertx.setTimer(
        delay,
        tid -> {
          try {
            handler.handle(tid);
          } catch (Throwable t) {
            // pass
          } finally {
            setInterval(vertx, delay, handler);
          }
        });
  }

  /** Run an async job with a delay after the previous one finished. */
  public static <T> void setIntervalAsync(Vertx vertx, long delay, AsyncHandler<Long, T> handler) {
    vertx.setTimer(
        delay,
        tid -> {
          try {
            Future<T> ret = handler.handle(tid);
            ret.onComplete(
                _ar -> {
                  setIntervalAsync(vertx, delay, handler);
                });

          } catch (Throwable t) {
            // pass
          } finally {
            setIntervalAsync(vertx, delay, handler);
          }
        });
  }

  @FunctionalInterface
  public interface AsyncHandler<E, R> {

    /**
     * Something has happened, so handle it async.
     *
     * @param event the event to handle async
     */
    Future<R> handle(E event);
  }
}
```

vietj:
what is the motivation for this ?

as far as I can see there is not much interest as the task is executed on the event loop and should not block, so `scheduleAtFixedRate` = `scheduleWithFixedDelay`

otherwise it could make sense for worker thread but in that case, it would not work as expected because tasks are triggered from event loop but executed then on a worker thread, which means that if a worker blocks, it will not impact the delay

liuchong:
How do you think the `setIntervalAsync` in the commit https://github.com/eclipse-vertx/vert.x/pull/4605#issuecomment-1422045217?

liuchong:
Hi @vietj , please check this version of `setInterval`. If it seems on your expected direction, I will do more work on it, thanks!

He-Pin:
It's currently using `scheduleWithFixedRate` but  not `scheduleWithFixedDelay` of netty's eventloop, may be we can support both on the vertx API?

PR #4612, opened 2023-02-18
VertxImpl init order for logger that reads Context

### LINKED ISSUE #4611: VertxImpl init order for logger that reads Context
### Version

All since 4.3.2

### Context

https://github.com/eclipse-vertx/vert.x/blob/4.3.8/src/main/java/io/vertx/core/impl/VertxImpl.java
has this code:

```
public class VertxImpl implements VertxInternal, MetricsProvider {
  private static final Logger log = LoggerFactory.getLogger(VertxImpl.class);
  ...
  static final ThreadLocal<ContextDispatch> nonVertxContextDispatch = new ThreadLocal<>();
  ...
}
```

The getLogger method creates a debug log message.

Log4j loggers can be configured to access the Vert.x Context to call getLocal.

This tries to read nonVertxContextDispatch that is still null when getLogger is called.

To fix it the order should be swapped:

```
public class VertxImpl implements VertxInternal, MetricsProvider {
  static final ThreadLocal<ContextDispatch> nonVertxContextDispatch = new ThreadLocal<>();
  private static final Logger log = LoggerFactory.getLogger(VertxImpl.class);
  ...
}
```

Example of an affected logger:

https://github.com/folio-org/okapi/blob/v5.0.0/okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java

```
Resolver failed to lookup FolioLoggingContext:requestid java.lang.NullPointerException: Cannot invoke "java.lang.ThreadLocal.get()" because "io.vertx.core.impl.VertxImpl.nonVertxContextDispatch" is null
	at io.vertx.core.impl.ContextInternal.current(ContextInternal.java:46)
	at io.vertx.core.Vertx.currentContext(Vertx.java:117)
	at org.folio.okapi.common.logging.FolioLoggingContext.lookup(FolioLoggingContext.java:67)
	at org.apache.logging.log4j.core.lookup.StrLookup.evaluate(StrLookup.java:119)
	at org.apache.logging.log4j.core.lookup.Interpolator.evaluate(Interpolator.java:198)
```

julianladisch:
Swap the init order of log and nonVertxContextDispatch so that the log init can access the Context.

Fixes #4611.

### DIFF
--- a/src/main/java/io/vertx/core/impl/VertxImpl.java
+++ b/src/main/java/io/vertx/core/impl/VertxImpl.java
@@ -87,6 +87,13 @@ import java.util.function.Supplier;
  */
 public class VertxImpl implements VertxInternal, MetricsProvider {
 
+  /**
+   * Context dispatch info for context running with non vertx threads (Loom).
+   */
+  static final ThreadLocal<ContextDispatch> nonVertxContextDispatch = new ThreadLocal<>();
+  // We need to initialize nonVertxContextDispatch before log because log may use the context on startup.
+  // https://github.com/eclipse-vertx/vert.x/issues/4611
+
   private static final Logger log = LoggerFactory.getLogger(VertxImpl.class);
 
   private static final String CLUSTER_MAP_NAME = "__vertx.haInfo";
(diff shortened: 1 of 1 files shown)


vietj:
can you rebase your PR on latest master to make the CI pass ?

tsegismont:
@vietj I think it's not necessary, I'll restart the CI jobs

tsegismont:
The failures due to Maven OSS repo being unavailable earlier this week are no longer present. There's one test failing now, which seems unrelated to this change.

@julianladisch it seems the rebase is needed after all. Thank you

vietj:
it is required because of the Netty upgrade that is a snapshot of
vertx-dependencies and not this project (which is a reason I don't like
much centralised dependencies as it fragilises the CI)

PR #4614, opened 2023-02-19
Await hadIdle in testThreadPoolMetricsWithInternalExecuteBlocking

julianladisch:
Sometimes the build fails with this error:

MetricsTest.testThreadPoolMetricsWithInternalExecuteBlocking:1010->AsyncTestBase.assertTrue:378 null

Switching from assertTrue to assertWaitUntil mitigates the race condition.


vietj:
thanks!

PR #4590, opened 2023-01-17
Http validation improvements

franz1981:
In most of our benchmarks, HTTP validation is not enabled and it doesn't make easy to spot which cost it has.
This PR is addressing this by adding special flags to both test it with/without ASCII values and it improves it by removing branches out of its validation path.

PR #4621, opened 2023-02-22
Favour inlining on http header value validation and specialize it per sequence type

franz1981:
This is addressing a small (relative) regression on header validation due to the improvement on Netty side.
The previous comparison with Netty was at https://github.com/eclipse-vertx/vert.x/pull/4590#issuecomment-1385021782

Before:
```
Benchmark                              (asciiNames)  (validate)   Mode  Cnt     Score     Error   Units
HeadersSetBenchmark.vertxSmall                 true        true  thrpt   20  3522.025 ±  42.544  ops/ms
HeadersSetBenchmark.vertxSmall                false        true  thrpt   20  2709.417 ± 138.243  ops/ms
```
This PR:
```
Benchmark                              (asciiNames)  (validate)   Mode  Cnt     Score    Error   Units
HeadersSetBenchmark.vertxSmall                 true        true  thrpt   20  5247.075 ± 86.684  ops/ms
HeadersSetBenchmark.vertxSmall                false        true  thrpt   20  3575.735 ± 80.029  ops/ms
```

With this patch we're now back near to Netty validation (that's less complex then our)

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpUtils.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpUtils.java
@@ -784,7 +784,10 @@ public final class HttpUtils {
   }
 
   public static void validateHeaderValue(CharSequence seq) {
-
+    if (seq instanceof AsciiString) {
+      validateHeaderAsciiValue((AsciiString) seq);
+      return;
+    }
     int state = 0;
     // Start looping through each of the character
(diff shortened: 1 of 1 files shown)


vietj:
we need a test for this

franz1981:
Yep, let me add it :)

vietj:
missing test :-)

PR #4626, opened 2023-02-22
Disable writeHandlers by default

### LINKED ISSUE #4625: Disable writeHandler registration for net sockets and websockets by default
When a net socket is created, it registers a local consumer which address is given by `io.vertx.core.net.NetSocket#writeHandlerID`.

Similarly, when a websocket is created, it registers local consumers which addresses are given by `io.vertx.core.http.WebSocketBase#binaryHandlerID` and `io.vertx.core.http.WebSocketBase#textHandlerID`.

This puts some pressure on cluster managers when Vert.x is clustered, even if most users don't need the feature.

We shall disable write handlers registration by default (as we did for SockJS handler in https://github.com/vert-x3/vertx-web/issues/1221).

This is a breaking (behavioral) change, but users will be able to switch the feature on with options.

tsegismont:
Closes #4625

### DIFF
--- a/src/main/generated/io/vertx/core/http/HttpServerOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/http/HttpServerOptionsConverter.java
@@ -110,6 +110,11 @@ public class HttpServerOptionsConverter {
             obj.setPerMessageWebSocketCompressionSupported((Boolean)member.getValue());
           }
           break;
+        case "registerWebSocketWriteHandlers":
+          if (member.getValue() instanceof Boolean) {
+            obj.setRegisterWebSocketWriteHandlers((Boolean)member.getValue());
+          }
+          break;
         case "tracingPolicy":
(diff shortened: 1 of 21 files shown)


tsegismont:
If this PR is approved, the doc must be updated, as well as the breaking changes page.

pmlopes (approved):

vietj (changes requested):
No tests to ensure that write handler id are null for WebSocket

tsegismont:
@vietj do you mean a test checking that `binaryHandlerId` and `textHandlerId` are null by default? Because `writeHandlerId` is for net sockets.

vietj:
I meant any handler I think

tsegismont:
Thanks for merging @vietj , I'll make a follow-up PR for the documentation

vietj:
you should also update the breaking change wiki page for 4.4

tsegismont:
Yep

PR #4627, opened 2023-02-23
Reproduce issue https://github.com/eclipse-vertx/vert.x/issues/4618

Sgitario:
Relates to https://github.com/eclipse-vertx/vert.x/issues/4618

### DIFF
--- a/src/test/java/io/vertx/core/http/Http2Test.java
+++ b/src/test/java/io/vertx/core/http/Http2Test.java
@@ -1047,4 +1047,37 @@ public class Http2Test extends HttpTest {
     }));
     await();
   }
+
+  /**
+   * Test to reproduce "HTTP/2 upgrade only happens once per connection"
+   * @see <a href="https://github.com/eclipse-vertx/vert.x/issues/4618">#4618</a>
+   */
+  @Test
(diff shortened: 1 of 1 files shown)


Sgitario:
About the options to fix the issue:
- Create a Http2ClientRequest for `HTTP/2`, and the existing HttpClientRequest for `HTTP/1.1`. So, if HttpClientRequest is received, the upgrade operation will always be performed. 
- Or always perform the upgrade operation if (for example) the headers are not HTTP/2 compliant

@vietj

vietj:
can I ask why you are setting the request URI to empty ?

it is not allowed and Netty will translate the empty URI into "/" when it happens. (see https://www.rfc-editor.org/rfc/rfc2616#section-5.1.2 : _Note that the absolute path
   cannot be empty; if none is present in the original URI, it MUST be
   given as "/" (the server root)._)

Netty does not do it for HTTP/2, so I believe that Vert.x should automatically translate the empty URI into "/" to avoid this situation and have the same behaviour.

vietj:
ah you are setting to empty to reproduce the situation, I think vertx should translate it to "/"

vietj:
fixed

PR #4629, opened 2023-02-23
Save expensive interface type checks (related JDK-8180450 and https://github.com/netty/netty/issues/12708)

franz1981:
JDK-8180450 has forced Netty http encoding/decoding to order type checks in order to save a huge scalability hit.
And still this is not enough, perf-wise: vertx, as a Netty users, still consume Netty types and cannot ignore such ordering, or would have a similar performance hit.

Moreover, there are cases (outlined on https://github.com/TechEmpower/FrameworkBenchmarks/pull/7942) where the existing JDK mechanism for interface type checks waste CPU resources, but the "constrained" vertx http types can save it to happen both performing exact class type checks and/or instanceof vs concrete and known types.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
     <apacheds-protocol-dns.version>1.5.7</apacheds-protocol-dns.version>
     <generated.dir>${project.basedir}/src/main/generated</generated.dir>
     <stack.version>4.4.0-SNAPSHOT</stack.version>
-    <jmh.version>1.19</jmh.version>
+    <jmh.version>1.36</jmh.version>
     <vertx.testNativeTransport>false</vertx.testNativeTransport>
     <vertx.testDomainSockets>false</vertx.testDomainSockets>
     <jar.manifest>${project.basedir}/src/main/resources/META-INF/MANIFEST.MF</jar.manifest>
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
@@ -130,21 +130,24 @@ public class Http1xServerConnection extends Http1xConnectionBase<ServerWebSocket
   }
 
   public void handleMessage(Object msg) {
-    if (msg instanceof HttpRequest) {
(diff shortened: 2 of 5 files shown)


franz1981:
I've run the existing JMH benchmark `HttpServerHandlerBenchmark` but:
- sadly it's limited to just one side of the http handling ie `ertxHttpRequestDecoder` but **not** `VertxHttpResponseDecoder`
- metrics are enabled (!) and it is badly hitting a SUPER slow path type check on `DefaultHttpRequest` on https://github.com/eclipse-vertx/vert.x/blob/5ea7bc2fe2cf17ab78b57c5ac68386821bb841b5/src/main/java/io/vertx/core/http/impl/Http1xConnectionBase.java#L195-L226

I'll modify the benchmark in order to make evident the benefits

franz1981:
Before this PR:
```
Benchmark                                      Mode  Cnt    Score   Error   Units
HttpServerHandlerBenchmark.vertx              thrpt   20  614.545 ± 7.614  ops/ms
HttpServerHandlerBenchmark.vertxOpt           thrpt   20  777.142 ± 7.255  ops/ms
HttpServerHandlerBenchmark.vertxOptMetricsOn  thrpt   20  759.129 ± 7.972  ops/ms
```
These are the results post PR:

```
Benchmark                                      Mode  Cnt    Score    Error   Units
HttpServerHandlerBenchmark.vertx              thrpt   20  640.783 ±  4.690  ops/ms
HttpServerHandlerBenchmark.vertxOpt           thrpt   20  834.770 ± 3.142  ops/ms
HttpServerHandlerBenchmark.vertxOptMetricsOn  thrpt   20  832.764 ±  4.015  ops/ms
````
While Netty (now optimized in order to save expensive type checks as well):
```
Benchmark                                      Mode  Cnt    Score    Error   Units
HttpServerHandlerBenchmark.netty              thrpt   20  917.029 ±  3.416  ops/ms
```

In short:
- this PR improve http performance ~7-10% 
- even with such improvement, at our best, we're ~10% slower then Netty

Issue #4630, opened 2023-02-24
Client using HTTP/2.2 does not generate the spans of kind CLIENT when upgrading (first request)


Sgitario:
### Version

4.3.7

### Context

Using the Vert.x client and the protocol to HTTP/2, the span of kind `CLIENT` that is made by Opentelemetry is ignored in the first request. 

This is because the connection is upgraded to become an HTTP/2 connection via (this is done by [Http2UpgradeClientConnection](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/Http2UpgradeClientConnection.java)). 

Then, the `Http2UpgradeClientConnection` implementation will call the method [current.createStream](https://github.com/eclipse-vertx/vert.x/blob/011d271bf470600989a59a605dee05ae92034170/src/main/java/io/vertx/core/http/impl/Http2UpgradeClientConnection.java#L758) where `current` is an instance of [Http1xClientConnection](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java) in the first request. So, a stream of type [Http1xClientConnection.StreamImpl](https://github.com/eclipse-vertx/vert.x/blob/011d271bf470600989a59a605dee05ae92034170/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java#L413) is created and all the CLIENT traces will be added in this stream (see [here](https://github.com/eclipse-vertx/vert.x/blob/011d271bf470600989a59a605dee05ae92034170/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java#L276)). 

And the problem is that when ending the first request (and producing the traces), the connection has been already upgraded, so it's using one of type `Http2ClientConnection` which uses a new of type [Http2ClientConnection.StreamImpl](https://github.com/eclipse-vertx/vert.x/blob/011d271bf470600989a59a605dee05ae92034170/src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java#L390) which does not have the CLIENT traces that was created in Http1xClientConnection.StreamImpl.

At the second request, the connection has been already upgraded, so the `current` variable in Http2UpgradeClientConnection is a `Http2ClientConnection` instance, so everything works as expected.

PR #4632, opened 2023-02-24
Propagate trace when upgrading connections

Sgitario:
Fix https://github.com/eclipse-vertx/vert.x/issues/4630

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -395,6 +395,10 @@ public class Http1xClientConnection extends Http1xConnectionBase<WebSocketImpl> 
       return metric;
     }
 
+    Object trace() {
+      return trace;
+    }
+
     abstract void handleContinue();
     abstract void handleEarlyHints(MultiMap headers);
     abstract void handleHead(HttpResponseHead response);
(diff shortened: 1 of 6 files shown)


Sgitario:
fyi @vietj 
I tried to add a Vert.x test, but I could not reproduce it (the upgrade operation was not committed). However, I tried these changes using the Quarkus test and it worked fine now.

vietj:
thanks @Sgitario what do you mean by not committed ?

Sgitario:
> what do you mean by not committed ?

Where did you read "what do you mean by not committed" from me?

Sgitario:
> thanks @Sgitario what do you mean by not committed ?

I mean that was not done. I found that if we restarted both the client and the server in the test, then the upgrade operation is done in the first request. Still, let me know if you want me to add a test and I will figure it out.

vietj:
we should have a test @Sgitario can you provide that soon ?

Sgitario:
> we should have a test @Sgitario can you provide that soon ?

PR updated with a test that reproduces this issue.
If you comment [this line](https://github.com/eclipse-vertx/vert.x/pull/4632/files#diff-fe5c92fbccf3b4aafaaea26628a7084280ef3f58a6076eeb042e67765f71f74eR146), the test will fail as expected.

vietj:
@Sgitario nice! just in time for 4.4

PR #4633, opened 2023-02-24
HTTP client should rewrite the empty path to / for HTTP/2

vietj:
see https://github.com/eclipse-vertx/vert.x/pull/4627


Sgitario:
Many thanks!

PR #4635, opened 2023-02-27
Optimise connection metrics reporting

vietj:
Metrics size of connection messages are computed even though no metrics is configured to report to. We can refactor the internal API a bit to only compute the message size when it shall be effectively reported.


franz1981 (approved):
LGTM we're still getting a perf hit, but much less, so it's ok, well done!

PR #4636, opened 2023-02-27
Favour inlining without metrics reporting

franz1981:
It allow to have the same performance of no metrics configured, improving the inlining chances when no metrics reporting is configured.


franz1981:
I've run the benchmark of https://github.com/eclipse-vertx/vert.x/pull/4629 and indeed I get the same level of performance, as expected.

PR #4659, opened 2023-04-03
Update the JSON toString() to allow better debugging

pmlopes:
Motivation:

Vert.x JsonObject/JsonArray `toString()` methods call the codec encode method. While this is handy for debugging it has several drawbacks. When dealing with very large objects, a log statement or debug can incur on some performance penalty. Worse, when debugging, IDEs will try to show you a string representation of the object so it can impact your productivity.

When debugging circular references, you will notice that the debugger will constantly throw `StackOverflowError` slowing your IDE to a non use state.

This PR introduces a breaking change in the `toString()` methods of `JsonObject` and `JsonArray`, note that this should not impact most users as JSON encoding should always be done with `encode()` or `encodePrettilly()`.

The changes try to follow what other runtimes do. Like with V8 or Python `inspect`/`dirs`, the proposed `toString()` methods will return a JSON like representation of the object/array. It will enforce a few limits:

1. max nested objects/arrays 3 (this means that deep objects will be shown as ellipsis, but calling `toString()` from a nested object will start counting from that node)
2. Arrays and Objects will only list the 1st 100 elements
3. String values will be truncated to the 1st 10000 characters
4. Circular references are detected and identified by the system hashcode of the json wrapper container.

The reason to avoid using the container directly is that, although `Map` and `List` will detect `self` references, they would fail to detect other circular references, leading again to `StackOverflowError`.

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -621,7 +621,7 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
 
   @Override
   public String toString() {
-    return encode();
+    return inspect(this);
   }
 
   @Override
--- a/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/src/main/java/io/vertx/core/json/JsonObject.java
@@ -1155,7 +1155,7 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
 
   @Override
   public String toString() {
-    return encode();
+    return inspect(this);
   }
 
   @Override
--- a/src/main/java/io/vertx/core/json/impl/Inspector.java
+++ b/src/main/java/io/vertx/core/json/impl/Inspector.java
@@ -1,0 +1,218 @@
+/*
+ * Copyright (c) 2023 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.json.impl;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+import java.util.List;
+import java.util.Map;
+
+/**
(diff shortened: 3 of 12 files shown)


Yaytay:
When logging large objects having the output be JSON is particularly useful because we can stick it somewhere else, reformat it and read it easily.
I really dislike the Java map output because it makes that a lot more awkward.
So I definitely prefer the ":".

tsegismont (approved):
LGTM

pmlopes:
I've added a flag `vertx.json.tostring=legacy` to get the <5.0.0 behavior (`toString` is an alias to `encode`).

pmlopes:
@Yaytay this PR will not change the JSON encoding, you can still encode it with `obj.encode()` which is what `obj.toString()` was doing. What the PR changes is that `toString()` shall now return a debug friendly string that will not bring down your VM if you made a circular reference somewhere.

For example on a large object `toString()` will return:

```
{created: \"2020-05-12T15:10:37Z\", device: {device_info: {device_fw: 204, device_sn: \"06-02133\", device_trait: 2, device_type: 190}, timeseries: [{...}]}}
```

Note the `{...}` which means that that object is deep, so if you need to print it you do:

```
// using dot notation to avoid writing all the getJsonObject() calls...
obj.device.device_info.timeseries.toString()
```

BUT if you do:

```
obj.encode()
```

You still see the full thing:

```json
{"created":"2020-05-12T15:10:37Z","device":{"device_info":{"device_fw":204,"device_sn":"06-02133","device_trait":2,"device_type":190},"timeseries":[{"configuration":{"sensors":[{"measurements":["BATTERY","BATTERY_MV"],"port":7,"sensor_bonus_value":"Unavailable","sensor_firmware_ver":"Unavailable","sensor_number":133,"sensor_sn":"Unavailable"},{"measurements":["REFERENCE_KPA","TEMPC_LOGGER"],"port":8,"sensor_bonus_value":"Unavailable","sensor_firmware_ver":"Unavailable","sensor_number":134,"sensor_sn":"Unavailable"}],"valid_since":"2018-08-11T21:45:00Z","values":[[1534023900,0,19,[{"description":"Battery Percent","error":false,"units":"%","value":100},{"description":"Battery Voltage","error":false,"units":" mV","value":7864}],[{"description":"Reference Pressure","error":false,"units":" kPa","value":100.62},{"description":"Logger Temperature","error":false,"units":" °C","value":28.34}]]]}}]}}
```

pk-work (approved):

vietj:
this PR has been merged to rapidly, it broke the build of vertx-sql-client at least (https://github.com/eclipse-vertx/vertx-sql-client/actions/runs/4643518385/jobs/8218213912) and it might have breaken more.

I will revert it for now.

PR #4661, opened 2023-04-06
fix #4660: incrementing the fetch for ping/pong

### LINKED ISSUE #4660: WebSocket fetch and ping / pong
### Version

4.4.1

### Context

In looking at the fabric8 integration of vert.x we are using pause / fetch on websockets to control the rate at which messages are processed.  However it appears that ping / pong messages from the server count towards the fetch.  We can workaround this by not sending pings ourselves, or if we ever need to, use a pongHandler to increase the fetch.  However if the server sends a ping since the ping handling is baked-in there's nothing we can do to increase the fetch.

### Do you have a reproducer?

```
    Vertx vertx = Vertx.vertx(new VertxOptions());
    HttpServer server = vertx.createHttpServer();

    server.webSocketHandler(sws -> {
      sws.writePing(Buffer.buffer("ping"));
      sws.writeTextMessage("some message");
    }).listen(8080);

    CompletableFuture<String> message = new CompletableFuture<String>();
    client.webSocket(8080, "localhost", "/").onSuccess(ws -> {
      ws.pause();

      // demand is used for the ping instead
      ws.fetch(1);

      ws.textMessageHandler(message::complete);
    });

    // never completes
    message.get();
```

If we remove the pause / fetch, this will complete as expected.

cc @vietj the current api server implementation does not send a ping, so this isn't an issue for fabric8 yet.  However the likely resolution of the upstream issue https://github.com/kubernetes/kubernetes/pull/116778 will include using ping / pong as the heartbeat mechanism for websockets, so we'll probably have to account for this eventually.

shawkins:
Motivation:

Addresses #4660

This is the simplest change - always increase the fetch whenever there's a ping or pong.  However the pong case is debatable since there is the ability to set a pong handler - if someone is already handling this situation in their pong handler, then this change will further increase the fetch.  

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
+++ b/src/main/java/io/vertx/core/http/impl/WebSocketImplBase.java
@@ -463,6 +463,10 @@ public abstract class WebSocketImplBase<S extends WebSocketBase> implements WebS
           context.dispatch(frame.binaryData(), handler);
         }
         break;
+      case PING:
+      case PONG:
+        fetch(1);
+        break;
     }
   }
 
--- a/src/test/java/io/vertx/core/http/WebSocketTest.java
+++ b/src/test/java/io/vertx/core/http/WebSocketTest.java
@@ -2402,6 +2402,31 @@ public class WebSocketTest extends VertxTestBase {
     }));
     await();
   }
+  
+  @Test
+  public void testWebSocketPausePing() throws InterruptedException {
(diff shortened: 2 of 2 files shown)


vietj:
@shawkins can you sign the Eclipse Contributor Agreement ?

vietj:
can you provide a PR for 4.x branch @shawkins ? thanks

PR #4664, opened 2023-04-13
use entries().size() when encoding headers to fix missing headers

dhermanson:
**Motivation:**

This commit fixes an issue with message header encoding/decoding when using a clustered event bus. Without this change, headers have the potential to be lost when decoding. The reason for the header loss is that when encoding headers, the header length that gets written is based upon the `.size()` method. This method is based on the number of _unique_ header names. If a `MultiMap` has a header that contains more than one value, this causes a problem. The problem manifests itself during decoding. During decoding, the number of headers is read and then subsequently iterated over:

```
      ...

      int numHeaders = wireBuffer.getInt(headersPos);
      headersPos += 4;
      headers = MultiMap.caseInsensitiveMultiMap();
      for (int i = 0; i < numHeaders; i++) {

      ...
```

If there are any headers with more than one value, then the header(s) at the end of the list will not be decoded.

This commit fixes this by declaring the header size based on the entries, instead of the number of unique header names. A test has been added to assert the desired behavior.

**Conformance:**

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredMessage.java
@@ -200,7 +200,7 @@ public class ClusteredMessage<U, V> extends MessageImpl<U, V> {
     if (headers != null && !headers.isEmpty()) {
       int headersLengthPos = buffer.length();
       buffer.appendInt(0);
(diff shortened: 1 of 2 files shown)


vietj:
good catch thanks!

vietj:
can you provide a PR for master branch as well @dhermanson

PR #4665, opened 2023-04-14
Remove uses of classes in javax.security.cert

eirbjo:
Motivation:

Efforts are underway in the OpenJDK project to remove the long deprecated-for-removal classes in the package `javax.security.cert`. These classes were introduced for backwards compatibility concerns with the unbundled JSSE release for JDK 1.2/1.3, but their use have been discouraged since they were introduced.

It would be good to update Vert.x to not depend on / use these archaic APIs.

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/http/HttpConnection.java
+++ b/src/main/java/io/vertx/core/http/HttpConnection.java
@@ -20,7 +20,6 @@ import io.vertx.core.net.SocketAddress;
 
 import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
-import javax.security.cert.X509Certificate;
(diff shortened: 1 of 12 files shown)


eirbjo:
> can you check other projects in the vertx stack don't use those ? I think if we merge this PR as is, it will break other projects

Thanks!

Now is the time to admit that I'm completely new to Vert.x  ecosystem development, so I have no idea of downstream projects. Which projects would this be?

There is an ongoing OpenJDK effort to remove these APIs, so projects will most probably need to deal with this, sooner or later.

I'm mostly here to help reduce the risk of removing these APIs in the OpenJDK. I'll admit it is a bit of a chicked-and-egg problem, so any help and cooperation from projects like Vert.x is highly appreciated.

PR #4672, opened 2023-04-15
Upgrade assertJ version to 3.24.2

bmscomp:
Update the version of assertJ dependency to the latest stable one 3.24.2


src/test/java/io/vertx/core/BlockedThreadCheckerTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
```
vietj, line 2: can you remove this change ?

bmscomp: Yes of course I'll do it right now

bmscomp: @vietj  done , thanks for the review

Issue #4680, opened 2023-04-24
Add `remove` to connection pool's `Lease` interface.


kdubb:
#### Describe the feature

Add a `remove` method to `io.vertx.core.net.impl.pool.Lease` that allows leaseholders to remove the leased object instead of recycling it to the pool. A client would call `remove` instead of `recycle` when they determine the leased object should no longer be used.

The goal is to allow clients of the pool to make deterministic decisions on wether a leased object should be returned or recycled.

#### Use cases

`SqlConnectionPool` evictions are now ([#1298](https://github.com/eclipse-vertx/vertx-sql-client/pull/1298)) tested for idle time and max lifetime. Additionally, the eviction process is periodic and only checks "slots" that are not actively in use. Generally this works fine.

This leaves the problem of pools under heavy load. In this scenario there is no guarantee that eviction will ever be able to check for max lifetime because all connections might be in use when the eviction check is run.

During the `SqlConnectionPool.PooledConnection` "close" we know if the connection has exceeded its maximum lifetime but there is no way to hold the lease (so it's not recycled) and close/remove the connection from the pool.

With an added `remove` method, `PooledConnection` could communicate this removal requirement to the pool as it encounters it.

#### Contribution

I am willing to implement the feature or pass it off to @tsegismont or @vietj.

PR #4681, opened 2023-04-24
Add `remove` to connection pool `Lease`

kdubb:
Implements #4680

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/pool/Lease.java
+++ b/src/main/java/io/vertx/core/net/impl/pool/Lease.java
@@ -25,4 +25,9 @@ public interface Lease<T> {
    */
   void recycle();
 
+  /**
+   * Remove the object.
+   */
+  void remove();
+
 }
--- a/src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
+++ b/src/main/java/io/vertx/core/net/impl/pool/SimpleConnectionPool.java
@@ -715,6 +715,9 @@ public class SimpleConnectionPool<C> implements ConnectionPool<C> {
       slot.pool.recycle(this);
     }
 
+    @Override
+    public void remove() { slot.pool.remove(this.slot); }
+
     void emit() {
       Future<Lease<C>> fut = slot.context.succeededFuture(this);
       fut.onComplete(handler);


vietj:
it seems that this operation can break pool invariants, e.g if the connection has several leases, when another lease calls `recycle()` then no operation should be done since the slot might hold a different connection, e.g it could call `slot.usage--` and we assume this is possible because it has increase the slot usage before with the same connection

vietj:
as I understand you would like to mark a connection as `stale` so it should never be borrowed again from the pool and eventually becomes removed from the pool to let another connection to be created ?

kdubb:
This probably should have been a draft because I don't quite understand all the inner workings of the pool slots.

> as I understand you would like to mark a connection as stale so it should never be borrowed again from the pool and eventually becomes removed from the pool to let another connection to be created ?

Essentially yes, we need a way to communicate to the pool that the connection should be removed.

vietj:
I think then we need a new `shutdown` operation on `ConnectionPool` that performs this that will set a shutdown flag on the slot and have the pool not check to lease this again when the flag is true, pretty much like:

`void shutdown(Predicate<C> pred, Supplier<C> closer)`

It would be called like `pool.shutdown(conn -> conn.isValid(), conn -> conn.close())`

when a connection is marked as shutdown, the slot.usage can only decrease and when it reaches 0 then it will be simply removed from the pool an then closed, another connection can be created to satisfy the waiters then.

chandramouli-r:
@vietj @tsegismont I tried to implement this idea as a pre-cursor to maintaining minIdle connections in the pool.

Basically,
1) Add a maxTttlMs option to the pool
2) Add a softEvict flag to the slot (boolean), and createTimeMs field to the slot.
3) createTimeMs is initialized when a connection is created in a slot
4) In recycle, check if the maxTtlMs has expired for the slot. If yes, and if the slot usage is 0, remove the connection and set softEvict = true. If not, only set softEvict = true.
5) In acquire, add a check to the Selector to not select any slot that has softEvict = true

Then, during pool initialization and during remove, we need to create a method that refreshes the pool with minIdle connections (not part of the gist).

If this looks good to you, I can create a PR. Please let me know.

https://gist.github.com/chandramouli-r/f8adf3b765919ceb54dea9126a2d97f2

chandramouli-r:
@vietj @tsegismont create a draft PR for this: https://github.com/eclipse-vertx/vert.x/pull/4736 (the added API can be used in SqlConnectionPool for checking maxLifetime when a connection is returned back to the pool).

PR #4684, opened 2023-05-01
CompositeFuture methods should use generic Future<?> type instead of Future raw type

vietj:


EmadAlblueshi:
Awesome! 👍🏻

vietj:
Awesome but remains a breaking change, check the entire commit.

PR #4689, opened 2023-05-05
Remove read stream implementations that are not relevant anymore.

vietj:


vietj:
Original 4.x methods should be deprecated.

PR #4694, opened 2023-05-11
Create SNI KeyManagerFactory in KeyCertOptions instead of SslContextProvider so it can be cached

vietj:
When using SNI, SslContextProvider will always convert the mapped key manager to a key manager factory, that can be an expensive operation.

This conversion of the key manager to a factory has been moved to the KeyCertOptions interface as a default method, to let implementations of this interface override it with caching. The default Vert.x implementation based on KeyStoreOptionsBase caches the created factories.

### DIFF
--- a/src/main/java/io/vertx/core/net/KeyCertOptions.java
+++ b/src/main/java/io/vertx/core/net/KeyCertOptions.java
@@ -12,6 +12,7 @@
 package io.vertx.core.net;
 
 import io.vertx.core.Vertx;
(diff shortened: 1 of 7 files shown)


NilsRenaud:
I tried Vert.x 4.4.2 and have one issue with this fix: we can't avoid the deprecation warning since `keyManagerMapper` must be redefined in KeyCertsOptions implementations.

Would it be possible to have: 
```java
default Function<String, X509KeyManager> keyManagerMapper(Vertx vertx) throws Exception {
    return null;
}
```

PR #4699, opened 2023-05-15
Attempt to provide metrics for vertx bytebuf allocator

vietj:


vietj:
ping @franz1981

src/main/java/io/vertx/core/buffer/impl/VertxHeapByteBuf.java
```diff
@@ -53,4 +50,17 @@ public boolean release() {
   public boolean release(int decrement) {
```
franz1981, line 50: Given that the logic for the reference count seems to be disabled here (see `::release` code pointed here)
how do we know that `freeArray` will be correctly fired?
I haven't downloaded this branch locally, but I'm not sure this is going to work

vietj: I did like what is in `UnpooledByteBufAllocator.InstrumentedUnpooledHeapByteBuf` class

franz1981: yeah but our implementations doesn't make free to be called, try with a debugger (maybe I'm wrong, but it looks like that)

vietj: you are right, I think we should handle this through finalize/cleaner instead ?

vietj: or maybe don't account for these buffers ?

franz1981: I prefer this option, unless we are ready to pay the price for reference counting and "activate" free for such buffers as well

vietj: I do agree with you, so this PR is actually not needed at all.

PR #4704, opened 2023-05-16
Introduce generic version of CompositeFuture static methods

vietj:
Deprecate `CompositeFuture` static methods and provide alternatives on the `Future` type that declares generic types using wildcard instead of using raw types. Raw types were necessary because of a limitation with the code generator when `Future` / `CompositeFuture` were generated types. Since this limitation is gone we can provide better declarations.

Deprecated methods are removed in Vert.x 5 : https://github.com/vert-x3/wiki/wiki/4.4.3-Deprecations-and-breaking-changes

### DIFF
--- a/src/main/asciidoc/futures.adoc
+++ b/src/main/asciidoc/futures.adoc
@@ -81,7 +81,7 @@ Beyond this, {@link io.vertx.core.Future} offers more: `map`, `recover`, `otherw
 Coordination of multiple futures can be achieved with Vert.x {@link io.vertx.core.Future futures}.
 It supports concurrent composition (run several async operations in parallel) and sequential composition (chain async operations).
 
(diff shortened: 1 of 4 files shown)


src/main/asciidoc/futures.adoc
```diff
@@ -81,7 +81,7 @@ Beyond this, {@link io.vertx.core.Future} offers more: `map`, `recover`, `otherw
 Coordination of multiple futures can be achieved with Vert.x {@link io.vertx.core.Future futures}.
 It supports concurrent composition (run several async operations in parallel) and sequential composition (chain async operations).
 
-{@link io.vertx.core.CompositeFuture#all} takes several futures arguments (up to 6) and returns a future that is
+{@link io.vertx.core.Future#all} takes several futures arguments (up to 6) and returns a future that is
```
He-Pin, line 84: epic!, thanks.

Issue #4707, opened 2023-05-18
Unable to load io.netty.resolver.dns.macos.MacOSDnsServerAddressStreamProvider on MacOS with Apple Silicon (M1 / M2 chip)


rampler:
### Questions
I am getting error like that on MacOS with M2 Chip:
```
[vert.x-eventloop-thread-6] ERROR io.netty.resolver.dns.DnsServerAddressStreamProviders - Unable to load io.netty.resolver.dns.macos.MacOSDnsServerAddressStreamProvider, fallback to system defaults. This may result in incorrect DNS resolutions on MacOS.
java.lang.reflect.InvocationTargetException: null
```
Found out that adding this profile in `pom.xml` should do the job:
```
    <profile>
      <id>mac-m1</id>
      <activation>
        <os>
          <family>mac</family>
          <arch>aarch64</arch>
        </os>
      </activation>
      <dependencies>
        <dependency>
          <groupId>io.netty</groupId>
          <artifactId>netty-resolver-dns-native-macos</artifactId>
          <classifier>osx-aarch_64</classifier>
          <scope>test</scope>
        </dependency>
      </dependencies>
    </profile>
```

### Version

4.2.3, but I see that problem would be on 4.4.2 too.

### Context

The exception is thrown at startup of application.

### Do you have a reproducer?

Any vert.x project run on MacOS with M1/M2 chip.

### Steps to reproduce

1. Run vert.x application on MacOS with Apple Silicon.

### Extra

* MacOS with M2 chip
* JVM - openJDK 17

PR #4708, opened 2023-05-18
Added profile in pom.xml for MacOS M1/M2 chip to use native Netty DNS…

rampler:
Implements #4707

### DIFF
--- a/.github/workflows/ci-4.x.yml
+++ b/.github/workflows/ci-4.x.yml
@@ -12,10 +12,10 @@ jobs:
             profile: ''
           - os: ubuntu-latest
             jdk: 8
-            profile: '-PtestNativeTransport'
+            profile: '-DaddMacSupport -PtestNativeTransport'
           - os: ubuntu-latest
             jdk: 8
-            profile: '-PtestDomainSockets'
+            profile: '-DaddMacSupport -PtestDomainSockets'
           - os: ubuntu-latest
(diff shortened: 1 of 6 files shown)


vietj:
can you sign the Eclipse ECA ?

vietj (changes requested):
can you also add support for OSX native KQueue ?

rampler:
Done. Unfortunately I needed to add some maven property to activate arch-based dependency. If you have idea how to do it more elegant, just tell.

.github/workflows/ci-4.x.yml
```diff
@@ -12,10 +12,10 @@ jobs:
             profile: ''
           - os: ubuntu-latest
             jdk: 8
-            profile: '-PtestNativeTransport'
+            profile: '-DaddMacSupport -PtestNativeTransport'
```
vietj, line 15: I don't think we need this as we only test on Linux and Windows for now

vietj:
why is the property necessary ? I don't get that

.github/workflows/ci-4.x.yml
rampler: Changes from workflows reverted

vietj:
thanks I took inspiration from your PR and made the change myself https://github.com/eclipse-vertx/vert.x/pull/4714

PR #4710, opened 2023-05-25
Event bus streaming mockup

vietj:
A refactoring that decouples a few things + framing introduction

- decouple the registration handlers from current interactions (e.g remove the reply handler knowledge from the registration handler)
- a registration handler now process generic frames
- a message is a frame
- clustering are responsible to send frames to peers very much like they are sending messages (which is now a specific frame)

Thanks to frames we keep existing interactions as well as allow new interactions to support bidi streaming. Very similar to TCP we have syn/fin frames between peers to control stream boundaries.

See also https://github.com/eclipse-vertx/vert.x/pull/4712 that does the same but keep the existing protocol using codecs to implement the protocol instead of adding frames.

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/Frame.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/Frame.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2011-2023 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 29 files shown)


vietj:
@tsegismont FYI

src/main/java/io/vertx/core/eventbus/impl/Frame.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.vertx.core.buffer.Buffer;
+
+public interface Frame {
+
+  String address();
+
+  Buffer encodeToWire();
+
+  boolean isFromWire();
```
He-Pin, line 21: Add an Enum `FrameSource`?

PR #4712, opened 2023-05-29
Event bus streaming mockup 2

vietj:
Another take on https://github.com/eclipse-vertx/vert.x/pull/4710 that reuse codecs and remains compatible with the existing event bus transport protocols


src/main/java/io/vertx/core/eventbus/MessageStream.java
```diff
@@ -0,0 +1,27 @@
...
+@VertxGen
+public interface MessageStream {
+
+  void handler(Handler<Message<String>> handler);
+
+  void endHandler(Handler<Void> handler);
+
+  void write(String msg);
+
+  void end();
```
He-Pin, line 25: What's the relation of this and `WriteStream`

vietj: none for now

vietj: we might have later back pressure support but that's not related to this PR

PR #4720, opened 2023-06-01
Add annotations which helps code highlight in IntelliJ

imanushin:
Motivation: Vert.X has method expecting json-formatted String. IntelliJ Idea IDE has an ability to highlight language syntax for such strings. The goal of this change is to have syntax highlight of Strings pages as inputs to JsonObject.

Note, that this annotation is optional (we don’t need to add real dependency) and it only highlights syntax in the specific IDE.


vietj:
thanks but cannot accept this extra dependency in the project

PR #4735, opened 2023-06-12
Use class declared constructor

yeikel:
Motivation:

Quoting from the Javadocs 

> Deprecated
> This method propagates any exception thrown by the nullary constructor, including a checked exception. Use of this method effectively bypasses the compile-time exception checking that would otherwise be performed by the compiler. The Constructor.newInstance method avoids this problem by wrapping any exception thrown by the constructor in a (checked) InvocationTargetException.
> The call
>  clazz.newInstance()
> can be replaced by
>  clazz.getDeclaredConstructor().newInstance()
> The latter sequence of calls is inferred to be able to throw the additional exception types InvocationTargetException and NoSuchMethodException. Both of these exception types are subclasses of ReflectiveOperationException.


tsegismont (approved):
LGTM, thank you

PR #4754, opened 2023-06-23
Use fast thread local and fix GC nepotism (Fixes #4749) (#4750)

### LINKED ISSUE #4749: ThreadLocal::remove performance impacts

franz1981:
* Use fast thread local and fix GC nepotism (Fixes #4749)
* Replace order of reentrant task's merge

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
+++ b/src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
@@ -29,7 +29,7 @@ public class CombinerExecutor<S> implements Executor<S> {
   private final AtomicInteger s = new AtomicInteger();
   private final S state;
 
-  protected final class InProgressTail {
+  protected static final class InProgressTail {
     Task task;
   }
 
@@ -49,18 +49,20 @@ public class CombinerExecutor<S> implements Executor<S> {
     Task tail = null;
     do {
       try {
-        final Action<S> a = q.poll();
-        if (a == null) {
-          break;
-        }
-        Task task = a.execute(state);
-        if (task != null) {
-          if (head == null) {
-            assert tail == null;
-            tail = task;
-            head = task;
-          } else {
-            tail = tail.next(task);
+        for (; ; ) {
+          final Action<S> a = q.poll();
+          if (a == null) {
+            break;
+          }
+          Task task = a.execute(state);
+          if (task != null) {
+            if (head == null) {
+              assert tail == null;
+              tail = task;
+              head = task;
+            } else {
+              tail = tail.next(task);
+            }
           }
         }
       } finally {
--- a/src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
+++ b/src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
@@ -10,6 +10,7 @@
  */
 package io.vertx.core.net.impl.pool;
 
+import io.netty.util.concurrent.FastThreadLocal;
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
```diff
@@ -27,7 +28,12 @@ public class CombinerExecutor<S> implements Executor<S> {
   private final Queue<Action<S>> q = PlatformDependent.newMpscQueue();
   private final AtomicInteger s = new AtomicInteger();
   private final S state;
-  private final ThreadLocal<Task> current = new ThreadLocal<>();
+
+  protected final class InProgressTail {
```
He-Pin, line 32: Can it be `static`?

franz1981: Uh!! Missed that you are right! Please open a pr, we love PRs

franz1981: And many thanks to lookin at it!

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
```diff
@@ -40,50 +46,46 @@ public void submit(Action<S> action) {
...
+        Task task = a.execute(state);
+        if (task != null) {
+          if (head == null) {
+            assert tail == null;
+            tail = task;
+            head = task;
+          } else {
+            tail = tail.next(task);
+          }
+        }
```
He-Pin, line 65: The old `pollAndExecute` seems will poll and execute all actions and it will only execute one now, is that expected?

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
He-Pin: Glad that help, I'm using Vertx at work too, very great project!

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
franz1981: Good catch, I have cherry picked this from and old branch of mine and probably dropped the for, are you ok sending a pr to fix what you have found?

franz1981:
@He-Pin sent some new commit, please check ;)

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
He-Pin: But you already have this pr, I think you can address it in this pr too♡(*´∀｀*)人(*´∀｀*)♡

franz1981: Already done, we have another one already merged for master too

He-Pin: i see, i can try a pr tomorrow then.

PR #4759, opened 2023-06-27
Use double-locking and lazySet to speedup init and steady-state ctx map loading

franz1981:
Given that users can often retrieve and modify both local or not ctx maps, relying on optimized/cheap `synchronized` access isn't a good idea, especially after https://openjdk.org/jeps/374.

In theory we could have used a `getVolatile`/`compareAndSet` approach here, with no harm (if not that some wasteful empty CHMs can be allocated), but this one should be more known (see https://en.wikipedia.org/wiki/Double-checked_locking).
If you prefer me to just use the atomic reference variant I can do it @vietj (code will be slightly less).

I've introduced a variation on such (see http://psy-lob-saw.blogspot.com/2012/12/atomiclazyset-is-performance-win-for.html and https://shipilev.net/blog/2014/safe-public-construction/) which "relax" the total store ordering of the initialized map, but should always allow a safe-publication of the CHM instance (meaning, not half constructed maps).

### DIFF
--- a/src/main/java/io/vertx/core/impl/ContextBase.java
+++ b/src/main/java/io/vertx/core/impl/ContextBase.java
@@ -23,6 +23,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 /**
  * A base class for {@link Context} implementations.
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/impl/ContextBase.java
```diff
@@ -36,15 +37,19 @@ public abstract class ContextBase implements ContextInternal {
...
   private final VertxInternal owner;
   private final JsonObject config;
   private final Deployment deployment;
   private final CloseFuture closeFuture;
   private final ClassLoader tccl;
   private final EventLoop eventLoop;
-  private ConcurrentMap<Object, Object> data;
-  private ConcurrentMap<Object, Object> localData;
+  private volatile ConcurrentMap<Object, Object> data;
+  private volatile ConcurrentMap<Object, Object> localData;
```
geoand, line 52: I am wondering why you have made these fields volatile here while not doing the same for `DuplicatedContext`

franz1981: Because I've missed it for sure, thanks!

franz1981:
I've instead pushed https://github.com/eclipse-vertx/vert.x/pull/5125 to addressed the performance data we have obtaine in Quarkus for 4.x. I'll re-work on `master` in case the 4.x version is accepted.

PR #4769, opened 2023-07-06
 Remove CLI usage in Vert.x Core (examples and code)

tsegismont:
Commits can be reviewed individually


vietj:
it seems this PR removes more than just examples, so I'm guessing we need to wait until this is resurrected in vertx-launcher to merge it to avoid vertx-shell breaking

src/main/java/io/vertx/core/AbstractVerticle.java
```diff
@@ -87,10 +88,13 @@ public JsonObject config() {
 
   /**
    * Get the arguments used when deploying the Vert.x process.
+   *
```
vietj, line 91: the Java doc should be changed to : "returns an empty list"

src/main/java/io/vertx/core/Context.java
```diff
@@ -155,10 +155,11 @@ static boolean isOnVertxThread() {
   @Nullable JsonObject config();
 
   /**
-   * The process args
+   * @deprecated As of version 5, Vert.x is no longer tightly coupled to the CLI
```
vietj, line 158: should also say that now it returns an empty list

vietj (changes requested):
Couple of minor changes, but LGTM

tsegismont:
@vietj done

PR #4775, opened 2023-07-11
HTTP/2 server does not aggregate response flush for full response when there is no read in progress

vietj:


vietj:
@cescoffier FYI

cescoffier:
Thanks!

Issue #4787, opened 2023-07-20
SSLOptions object incorrectly states objects are equal when SSL handshake timeouts are different


Kyefer:
### Version

4.4.3

### Context

I am trying to update the SSL options of a running `HttpServer` object using the `updateSSLOptions(SSLOptions)` method and have changed the SSL Handshake timeout. I expected the SSL settings to have been be recreated but it seems that it did not. Inspecting the code, I noticed that the `equals()` method on the `SSLOptions` class returned `true` even when the SSL handshake timeout was different.

Looking into the code, I noticed that the `equals` method uses the same `sslHandshakeTimeout` variable on both sides of the equality check instead of using the `that`'s variable.

The line in issue: https://github.com/eclipse-vertx/vert.x/blob/032f2def78d5ab334b6090c77ce35ccf0ae1cea5/src/main/java/io/vertx/core/net/SSLOptions.java#L334

### Steps to reproduce
```
SSLOptions first = new SSLOptions().setSslHandshakeTimeout(10);
SSLOptions second = new SSLOptions().setSslHandshakeTimeout(11);

assert first.equals(second) == false;
```

### Extra
Note that I actually don't care about updating the SSL Handshake Timeout variable, but instead I am looking to "trick" vertx into update the SSL context since the keystore at the given path has updated but the path has stayed the same, so vertx doesn't update the SSL context. I was hoping to simply change the handshake timeout but ran into this issue

vietj:
can you provide a PR that fixes implements this ? initially the purpose of updating SSL options is mostly for certificates and not other settings

Kyefer:
OK, I will make the change

Kyefer:
Please see https://github.com/eclipse-vertx/vert.x/pull/4788

vietj:
See https://github.com/eclipse-vertx/vert.x/pull/4789

PR #4788, opened 2023-07-20
Fix SSLOptions equals method to use sslHandshakeTimeout from argument

Kyefer:
Motivation:

As explained in https://github.com/eclipse-vertx/vert.x/issues/4787, I am interested in updating the keystore used by an `HttpServer` object. However, the keystore at the path given in the `JksOptions` doesn't change; the keystore itself changes. So when attempting to call `updateSSLOptions(SSLOptions)` on the `HttpServer` object, vert.x thinks that the options have not changed, and thus wont reread the certificate. I was hoping to force vertx to reread the keystore by changing the `sslHandshakeTimeout` variable, as that is really the only one that can be modified without security concerns. However, in doing so, I noticed that the `equals()` method on the `SSLOptions` class incorrectly returned `true` when the SSL handshake timeout was different.

### DIFF
--- a/src/main/java/io/vertx/core/net/SSLOptions.java
+++ b/src/main/java/io/vertx/core/net/SSLOptions.java
@@ -331,7 +331,7 @@ public class SSLOptions {
     }
     if (obj instanceof SSLOptions) {
       SSLOptions that = (SSLOptions) obj;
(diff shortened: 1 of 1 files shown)


Kyefer:
If this gets approved and merged, would it be possible to backport it to the 4.X version?

vietj:
Sure @Kyefer

PR #4797, opened 2023-08-02
Buffer adding support for float le and double le

Fenrur:
**Title:** Add Little Endian Support for Floats and Doubles in Buffer Class

**Body:**

This pull request aims to add support for Little Endian representation in floats and doubles within the `Buffer` class.

**New Methods:**

In `Buffer`:
- `double getDoubleLE(int pos);`
- `float getFloatLE(int pos);`
- `Buffer appendFloatLE(float f);`
- `Buffer appendDoubleLE(double d);`
- `Buffer setDoubleLE(int pos, double d);`
- `Buffer setFloatLE(int pos, float f);`

Corresponding changes have been made to the `BufferInternal`, `BufferImpl` classes, and `BufferTest` to accommodate these new methods.

---

### DIFF
--- a/src/main/java/io/vertx/core/buffer/impl/BufferInternal.java
+++ b/src/main/java/io/vertx/core/buffer/impl/BufferInternal.java
@@ -108,7 +108,13 @@ public interface BufferInternal extends Buffer {
   BufferInternal appendFloat(float f);
 
   @Override
+  Buffer appendFloatLE(float f);
+
+  @Override
   BufferInternal appendDouble(double d);
+
+  @Override
+  Buffer appendDoubleLE(double d);
 
   @Override
   BufferInternal appendString(String str, String enc);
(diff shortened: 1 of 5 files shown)


src/main/java/io/vertx/core/buffer/impl/BufferInternal.java
```diff
@@ -149,9 +155,15 @@ static BufferInternal buffer(byte[] bytes) {
   @Override
   BufferInternal setDouble(int pos, double d);
 
+  @Override
+  Buffer setDoubleLE(int pos, double d);
```
vietj, line 159: it should return BufferInternal instead of Buffer

Fenrur: I just fix that in new commit ced5ae3c034717ba5d610284c28b3e748737c32a

PR #4809, opened 2023-08-11
Provide a lock free implementation of ConnectionBase write message

vietj:
The `ConnectionBase` class implementation has a few drawbacks

- `ConnectionBase#writeToChannel` currently uses an implicit queue (`writeInProgress` field + event-loop scheduling)
- use locks assuming biased locking
-back-pressure is not correctly handled by usage outside of its event-loop (e.g a pooled HTTP client connection) requiring work around to handle back-pressure

This PR use a new queue to solve these problems:
- reduce the number of scheduled tasks when writing from another thread than the event-loop (producer thread)
- use lock free structure
- connection back-pressure is now based on channel writability and the number of inflight messages

In addition it aims to improve producer message flush consolidation thanks to the queue: when a producer writes several messages and then schedules a drain, the drain consolidates the writes.

The introduced `OutboundWriteQueue` is a multi producer queue with a single consumer event-loop thread (MPSC) optimised for the ideal case (event-loop thread). The internal is inspired from https://akarnokd.blogspot.com/2015/05/operator-concurrency-primitives_11.html and reuse the MPSC queue + WIP atomic.

Write operations now returns a boolean indicating the writability, which should be used in the future for stream pipe like operations to replace the `write` _then_ `writeQueueFull` which can be racy when used outside of the event-loop. It is not yet used but is reserved for a future usage.

Note to reviewers: this PR is structured as is

- `OutboundWriteQueue` is a lock free queue that can be reviewed independently
- `ConnectionBase` changes replace the use of internal state + message rescheduling by the `OutboundWriteQueue`
- the other commits are update the various usage of `ConnectionBase`
  - HTTP/1.1 server schedules HTTP messages to be written and update `Http1xConnectionBase` state using those messages
  - HTTP/1.1 client schedule custom messages to update the state of the connection and perform a direct message write
- There is still some cleanup to do in `OutboundWriteQueue` tests (try avoiding hooks if possible + commented/ignored tests), this will be done of course


vietj:
@franz1981 can you have look ?

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
```diff
@@ -0,0 +1,361 @@
...
+   */
+  public static int numberOfUnwritableSignals(int value) {
+    return (value & ~0XF) >> 4;
+  }
+
+  /**
+   * When the masked bit is set, the queue became unwritable, this triggers only when the queue transitions
+   * from the <i>writable</i>> state to the <i>unwritable</i>> state.
+   */
+  public static int QUEUE_UNWRITABLE_MASK = 0x01;
```
franz1981, line 112: All the masks should be made static final

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
```diff
@@ -0,0 +1,361 @@
...
+  // Immutable
+
+  private final Predicate<E> consumer;
+  private final int highWaterMark;
+  private final int lowWaterMark;
+
+  // Concurrent part accessed by any producer thread
+
+  private final Queue<E> queue = PlatformDependent.newMpscQueue();
+  private final AtomicInteger wip = new AtomicInteger();
```
franz1981, line 145: I would consider using atomic int field updater: in subsequent iterations it could allow padding or just save allocating a wrapper Atomic, in both cases we can save this.
Another important point: given that we have no control on the amount of concurrent, accumulates writes and the OS pauses due to scarse CPU resources, is often better to avoid integer overflow by using a long here. 
External concurrent requests could be a lot, in pathological cases and addressing numeric overflow is a PITA.

Most HW arch words are 8 bytes anyway so...no need to save bits

jponge (changes requested):
I've done a first pass focusing on `OutboundWriteQueue`, see my remarks and suggested edits.

The queue logic looks ok to me, although I will do another pass.

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
```diff
@@ -0,0 +1,361 @@
...
+
+  // Consumer thread only
+
+  // The element refused by the consumer (null <=> overflow)
+  private E overflow;
+  // The number of times the queue was observed to be unwritable
+  private int writeQueueFull;
```
franz1981, line 152: As ditto as before: long?

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
```diff
@@ -0,0 +1,361 @@
...
+  // Note : we can optimize this by passing pending as argument of this method to avoid the initial
+  private int drainLoop() {
+    int pending = wip.get();
+    if (pending == 0) {
+      throw new IllegalStateException();
+    }
+    do {
+      int consumed;
+      for (consumed = 0; consumed < pending; consumed++) {
+        E elt = queue.poll();
```
franz1981, line 298: peek, given that JCTools is pretty efficient for single consumer, is still very effective. I would suggest to save overflow to exist/being handled and perform a 2 phase op: peek (with consume) + poll.
Or split the consumption in a predicated and consumption too (still using peek/poll)

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
```diff
@@ -0,0 +1,361 @@
...
+        overflow = element;
+        return DRAIN_REQUIRED_MASK | (wip.get() == highWaterMark ? QUEUE_UNWRITABLE_MASK : 0);
+      }
+      if (consume(1) == 0) {
+        return 0;
+      }
+      return drainLoop();
+    } else {
+      queue.add(element);
+      return wip.incrementAndGet() == highWaterMark ? QUEUE_UNWRITABLE_MASK : 0;
```
franz1981, line 221: Mmmm I should read it again where...why we don't drain here if we win the race?
Between the compareAndSet and the subsequent incrementAndGet the would could have change enough that we are still eligible for becoming the one to drain...

franz1981 (commented):
Currently in PTO till 4 sept, so just few comments

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
```diff
@@ -0,0 +1,361 @@
...
+ * <p>After a drain if the internal queue has shrunk under {@link #lowWaterMark}, the queue is considered as {@code writable}.
+ * {@link #add}/{@link #drain} methods return {@link #QUEUE_WRITABLE_MASK} to signal producers should start emitting. Note
+ * the consumer thread handle this signal and should transmit it to the producers.</p>
+ *
+ * <p>When {@link #QUEUE_WRITABLE_MASK} is signalled, the number of {@link #QUEUE_UNWRITABLE_MASK} signals emitted is encoded
+ * in the flags. This number allows the producer flow controller to correctly account the producer writability:
+ * {@link #QUEUE_WRITABLE_MASK}/{@link #QUEUE_UNWRITABLE_MASK} signals are observed by different threads, therefore a good
+ * practice to compute the queue writability is to increment/decrement an atomic counter for each signal received.</p>
+ */
+public class OutboundWriteQueue<E> {
```
franz1981, line 96: A point about the bitmasks. I am a big fan of them, but not to expose them externally because they reveal of things are handled internally and bound you to never hange them.
If we can expose enums or other things that decide which states leaks to become part of the API would be better, IMO. 
At the same time, I understand that if we need to return set of states we have no other options, to save allocations...

vietj: one of the mask returns an positive value encoded

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
vietj: winning is implied by wip 0 -> 1 transition

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
vietj: can you suggest a change ?

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
vietj: good point, forgot it :-)

src/main/java/io/vertx/core/streams/impl/OutboundWriteQueue.java
franz1981: Yeah but if incrementAndGet == 0 it still means you have won....you can win in both ways (via compare and set or incrementAndGet) no?

vietj: Actually here we just signal back-pressure to the consumer (it crossed the high-water mark), not related to the queue ownership.

franz1981: Yeah but if we have been the one to win it (by accident), no others will be able to take ownership anymore, unless fully drained. But draining happen only if someone win the race...

vietj: ah right, this should be retried we can miss the signal here, good catch!!!

vietj: I see what you mean but actually I don't think so because only the event-loop thread is allowed

- to call this method
- drain the queue (decrease the wip)

so if we don't get the ownership, it means another thread (producer) increased it and the value is greater than 1, consequently we cannot get ownership at this point.

franz1981: Nopez because these are two separate atomic ops and the world can change in many ways between them. If you fail the first cas(0, 1) you can succeed the next increment to still perform the same transition (0->1). 
Meaning that no others can perform it again and get the chance to drain, because there were no other concurrent requests while it has happened.

vietj: ah so you mean that we can fail the first cas(0, 1) and wip still = 0 ?

vietj: so

```java
      queue.add(element);
      long v = WIP_UPDATER.incrementAndGet(this);
      if (v == 1) {
        return drainLoop();
      } else {
        return v == highWaterMark ? QUEUE_UNWRITABLE_MASK : 0;
      }
```

would be correct then

franz1981: Exactly. The attempt for the compareAndSet(0, 1) is to avoid using the queue,. While te subsequent one can still fall in the same transition but the queue submission happens before (literally). In both cases draining is still required

vietj:
thanks @jponge @franz1981 for the review, it has been really helpful and improved the quality of this PR, I believe those changes will have a great impact on the internals of Vert.x 5

PR #4824, opened 2023-08-29
Event bus internal improvements

vietj:
See also https://github.com/eclipse-vertx/vert.x/pull/4712

# Decouple message emission from the interceptor chain

EventBus streaming requires to send messages that does not go through the interceptor chain, e.g. establishing a stream requires to send such messages. The sendOrPub method and the downstream chain of calls (sendLocally, sendOrPublishFailed, sendToNode, sendToNodes) are coupled to OutboundDeliveryContext by convenience. The EventBus#sendOrPub is refactored to carry the messages and the delivery options instead of the OutboundDeliveryContext.

Main changes

- `EventBusImpl#sendOrPub(OutboundDeliveryContext) -> EventBusimpl#sendOrPub(ContextInternal,MessageImpl,DeliveryOptions,Promise)`
- same for `sendLocally`, `sendOrPublishFailed`, `sendToNode`, `sendToNodes`

# Decouple the RegistrationHandler from the repliedAddress knowledge

The clustered event bus is aware of the nature of the `HandlerHolder`: when a `HandlerHolder` is registered it does broadcast the registration when the `HandlerHolder` is a reply handler. The `HandlerHolder#isReplyHandler()` method is removed, instead `EventBusImpl#addRegistration` replaces the `replyHandler` boolean by a `broadcast` boolean, when `broadcast` is set to false the `onLocalRegistration` method is not called avoiding the clustered event bus to perform the check.

Main changes:

- `HandlerRegistration#register(String repliedAddress, boolean localOnly, Promise<Void> promise)` -> `HandlerRegistration#register(boolean broadcast, boolean localOnly, Promise<Void> promise)`
- `HandlerHolder` does not carry anymore the `replyHandler` boolean
- `MessageConsumerImpl` calls register with `broadcast` = true
- `ReplyHandler` calls register with `broadcast` = false
- `ClusteredEventBus` processes a local registration whatsoever as it is not called for reply handlers

# NodeSelector cleanup

The `NodeSelector` SPI is coupled to event-bus message API, instead it should only care about the message address, this simplifies the API and decouple the `NodeSelector` SPI from the message.

Main changes:

- `NodeSelector#selectForSend(Message, Promise<String>)` -> `NodeSelector#selectForSend(String, Promise<String>)`
- etc...

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -329,14 +329,14 @@ public class EventBusImpl implements EventBusInternal, MetricsProvider {
   }
 
   protected <T> void sendOrPub(ContextInternal ctx, MessageImpl<?, T> message, DeliveryOptions options, Promise<Void> writePromise) {
-    sendLocally(message, options, writePromise);
+    sendLocally(message, writePromise);
   }
 
   protected <T> void sendOrPub(OutboundDeliveryContext<T> sendContext) {
     sendOrPub(sendContext.ctx, sendContext.message, sendContext.options, sendContext);
   }
 
-  protected <T> void sendLocally(MessageImpl<?, T> message, DeliveryOptions options, Promise<Void> writePromise) {
+  protected <T> void sendLocally(MessageImpl<?, T> message, Promise<Void> writePromise) {
     ReplyException failure = deliverMessageLocally(message);
     if (failure != null) {
       writePromise.tryFail(failure);
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -116,7 +116,7 @@ public class EventBusImpl implements EventBusInternal, MetricsProvider {
   @Override
   public EventBus send(String address, Object message, DeliveryOptions options) {
     MessageImpl msg = createMessage(true, address, options.getHeaders(), message, options.getCodecName());
-    sendOrPubInternal(msg, options, null, null);
+    sendOrPubInternal(msg, options, null);
     return this;
   }
 
@@ -124,7 +124,7 @@ public class EventBusImpl implements EventBusInternal, MetricsProvider {
   public <T> Future<Message<T>> request(String address, Object message, DeliveryOptions options) {
     MessageImpl msg = createMessage(true, address, options.getHeaders(), message, options.getCodecName());
     ReplyHandler<T> handler = createReplyHandler(msg, true, options);
-    sendOrPubInternal(msg, options, handler, null);
+    sendOrPubInternal(msg, options, handler);
     return handler.result();
   }
 
@@ -161,7 +161,7 @@ public class EventBusImpl implements EventBusInternal, MetricsProvider {
 
   @Override
   public EventBus publish(String address, Object message, DeliveryOptions options) {
-    sendOrPubInternal(createMessage(false, address, options.getHeaders(), message, options.getCodecName()), options, null, null);
+    sendOrPubInternal(createMessage(false, address, options.getHeaders(), message, options.getCodecName()), options, null);
     return this;
   }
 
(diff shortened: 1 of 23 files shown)


tsegismont:
> The `NodeSelector` SPI is coupled to event-bus message API,

It was designed like this so that users can create their own node selector and choose a server depending on metadata provided by messager headers (e..g. implementing sharing on the event bus)

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
```diff
@@ -58,13 +58,13 @@ void receive(MessageImpl msg) {
...
-  synchronized void register(String repliedAddress, boolean localOnly, Promise<Void> promise) {
+  synchronized void register(boolean broadcast, boolean localOnly, Promise<Void> promise) {
     if (registered != null) {
       throw new IllegalStateException();
     }
-    registered = bus.addRegistration(address, this, repliedAddress != null, localOnly, promise);
+    registered = bus.addRegistration(address, this, broadcast, localOnly, promise);
     if (bus.metrics != null) {
-      metric = bus.metrics.handlerRegistered(address, repliedAddress);
+      metric = bus.metrics.handlerRegistered(address, null /* regression */);
```
tsegismont, line 67: Have you got any ideas already about how to address this regression?

Perhaps can we use a constant for now as a signal for a reply handler?

For the long term we shall pass the `broadcast` param to the SPI implementation.

FYI, Micrometer SPI impl does not use this value.

tsegismont (commented):
Thanks @vietj

src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
```diff
@@ -329,39 +327,39 @@ public void handle(Buffer buff) {
...
-      sendRemote(sendContext, replyDest, message);
+      sendRemote(replyDest, message, options, writePromise);
     } else {
-      super.sendOrPub(sendContext);
+      super.sendOrPub(ebContext, message, options, writePromise);
     }
   }
 
-  private void sendRemote(OutboundDeliveryContext<?> sendContext, String remoteNodeId, MessageImpl message) {
+  private void sendRemote(String remoteNodeId, MessageImpl<?, ?> message, DeliveryOptions options, Promise<Void> writePromise) {
```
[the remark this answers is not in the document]
tsegismont: I noticed this has been addressed in a following commit

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
vietj: I think we should simply remove it from the metrics as it seems to be limiting the protocol.

vietj: I need to check dropwizard and if that's so we simply remove it

src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -414,10 +418,12 @@ public <T> void sendOrPubInternal(OutboundDeliveryContext<T> senderCtx) {
     senderCtx.next();
   }
 
-  public <T> void sendOrPubInternal(MessageImpl message, DeliveryOptions options,
-                                    ReplyHandler<T> handler, Promise<Void> writePromise) {
+  public <T> Future<Void> sendOrPubInternal(MessageImpl message, DeliveryOptions options,
```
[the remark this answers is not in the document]
vietj: it's used by `MessageProducerImpl`

src/main/java/io/vertx/core/eventbus/impl/HandlerRegistration.java
vietj: it is actually not used in dropwizard nor micrometer, so it can be safely removed

src/main/java/io/vertx/core/eventbus/impl/clustered/ConnectionHolder.java
```diff
@@ -41,12 +41,21 @@ class ConnectionHolder {
...
   private final EventBusMetrics metrics;
 
-  private Queue<OutboundDeliveryContext<?>> pending;
+  private Queue<SomeTask> pending;
   private NetSocket socket;
   private boolean connected;
   private long timeoutID = -1;
   private long pingTimeoutID = -1;
 
+  private static class SomeTask {
```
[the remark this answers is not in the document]
vietj: done as `MessageWrite`

PR #4826, opened 2023-08-30
Remove vertx.cwd system property support

tsegismont:
This is undocumented and was only used in vertx-examples before Vert.x 5


tsegismont:
@vietj I've rebased the PR and the failure (which looked unrelated) is gone

vietj (approved):

PR #4830, opened 2023-08-31
feat: Add onComplete method with dual handler signature

### LINKED ISSUE #4829: Add two mapper onComplete method to Future.java
#### Describe the feature

Currently the compose method has a two mapper signature:
```
<U> Future<U> compose(Function<T, Future<U>> successMapper, Function<Throwable, Future<U>> failureMapper);
```
Where you supply the success and failure mappers separately.

It would be useful if there was a similar signature for the `onComplete` method, rather than calling `onSuccess` and `onFailure` separately.

My proposal is to add a new `onComplete` signature here:
https://github.com/eclipse-vertx/vert.x/compare/master...samuel-hawker:vert.x:poc?expand=1
That simply calls onSuccess and onFailure like so:
```
 default Future<T> onComplete(Handler<T> successHandler, Handler<Throwable> failureHandler) {
      return onSuccess(successHandler).onFailure(failureHandler);
  }
```

#### Use cases
The use case would be when a user wants to provide two separate handlers, but does not want to use the compose method as there is no need to return anything.

#### Contribution

Who should implement this feature ? me (see POC change above)

samuel-hawker:
Motivation:

Add onComplete method to Future class with onSuccess and onFailure mappers
Closes #4829

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -268,6 +268,26 @@ public interface Future<T> extends AsyncResult<T> {
   Future<T> onComplete(Handler<AsyncResult<T>> handler);
 
   /**
+   * Add handlers to be notified on succeeded result and failed result.
+   * <p>
+   * <em><strong>WARNING</strong></em>: this is a terminal operation.
+   * If several {@code handler}s are registered, there is no guarantee that they will be invoked in order of registration.
+   *
+   * @param successHandler the handler that will be called with the succeeded result
+   * @param failureHandler the handler that will be called with the failed result
+   * @return a reference to this, so it can be used fluently
+   */
+  default Future<T> onComplete(Handler<T> successHandler, Handler<Throwable> failureHandler) {
+      return onComplete(ar -> {
+        if (successHandler != null && ar.succeeded()) {
+          successHandler.handle(ar.result());
+        } else if (failureHandler != null && ar.failed()) {
+          failureHandler.handle(ar.cause());
+        }
+      });
+  }
+
+  /**
    * Add a handler to be notified of the succeeded result.
    * <p>
    * <em><strong>WARNING</strong></em>: this is a terminal operation.
@@ -278,11 +298,7 @@ public interface Future<T> extends AsyncResult<T> {
    */
   @Fluent
   default Future<T> onSuccess(Handler<T> handler) {
-    return onComplete(ar -> {
-      if (ar.succeeded()) {
-        handler.handle(ar.result());
-      }
-    });
+    return onComplete(handler, null);
   }
 
   /**
@@ -296,11 +312,7 @@ public interface Future<T> extends AsyncResult<T> {
    */
   @Fluent
   default Future<T> onFailure(Handler<Throwable> handler) {
-    return onComplete(ar -> {
-      if (ar.failed()) {
-        handler.handle(ar.cause());
-      }
-    });
+    return onComplete(null, handler);
   }
 
   /**
--- a/src/main/java/io/vertx/core/impl/future/FutureImpl.java
+++ b/src/main/java/io/vertx/core/impl/future/FutureImpl.java
@@ -128,6 +128,41 @@ class FutureImpl<T> extends FutureBase<T> {
   }
 
   @Override
+  public Future<T> onComplete(Handler<T> successHandler, Handler<Throwable> failureHandler) {
+    addListener(new Listener<T>() {
+      @Override
+      public void onSuccess(T value) {
+        try {
+          if (successHandler != null) {
+            successHandler.handle(value);
+          }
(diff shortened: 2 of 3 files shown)


src/main/java/io/vertx/core/Future.java
```diff
@@ -267,6 +267,10 @@ static <T> Future<T> failedFuture(String failureMessage) {
   @Fluent
   Future<T> onComplete(Handler<AsyncResult<T>> handler);
 
+  default Future<T> onComplete(Handler<T> successHandler, Handler<Throwable> failureHandler) {
+      return onSuccess(successHandler).onFailure(failureHandler);
```
vietj, line 271: actually this should be implemented instead as

```java
onComplete(ar -> {
  if (ar.succeeded()) {
  } else {
  }
});
```

as it is more efficient

samuel-hawker: Implemented!

src/test/java/io/vertx/core/FutureTest.java
```diff
@@ -221,11 +221,11 @@ public void testCompleteCause() {
     Promise<Throwable> promise = Promise.promise();
     AtomicReference<Boolean> r1 = new AtomicReference<>();
     AtomicReference<Boolean> r2 = new AtomicReference<>();
-    promise.future().onSuccess(v -> r1.set(true)).onFailure(v -> r1.set(false));
+    promise.future().onComplete(v -> r1.set(true), v -> r1.set(false));
```
vietj, line 224: you should not change this existing test

src/test/java/io/vertx/core/FutureTest.java
```diff
@@ -221,11 +221,11 @@ public void testCompleteCause() {
...
     Checker<Throwable>  checker = new Checker<>(promise.future());
     promise.complete(object);
     checker.assertSucceeded(object);
-    promise.future().onSuccess(v -> r2.set(true)).onFailure(v -> r2.set(false));
+    promise.future().onComplete(v -> r2.set(true), v -> r2.set(false));
```
vietj, line 228: you should not change this existing test

samuel-hawker:
Apologies, since it used the onSuccess and onFailure, I reasoned it was a good place to switch to the new method, but I reverted it as my change has nothing to do with that test.

vietj:
Although I like it can be implemented as a default method, I think we need an implementation in `FutureImpl` class that uses `io.vertx.core.impl.future.Listener` that avoids to create an intermediary java object, e.g here is `FutureImpl#onSuccess ` implementation:

```java
  public Future<T> onSuccess(Handler<T> handler) {
    Objects.requireNonNull(handler, "No null handler accepted");
    addListener(new Listener<T>() {
      @Override
      public void onSuccess(T value) {
        try {
          handler.handle(value);
        } catch (Throwable t) {
          if (context != null) {
            context.reportException(t);
          } else {
            throw t;
          }
        }
      }
      @Override
      public void onFailure(Throwable failure) {
      }
    });
    return this;
  }
```

samuel-hawker:
@vietj Thanks for the advice. Implemented as you suggested.

Note that in re-using `onComplete` for onSuccess and onFailure, I had to add extra checks for null, i.e.:
```
default Future<T> onComplete(Handler<T> successHandler, Handler<Throwable> failureHandler) {
      return onComplete(ar -> {
        if (successHandler != null && ar.succeeded()) {
          successHandler.handle(ar.result());
        } else if (failureHandler != null && ar.failed()) {
          failureHandler.handle(ar.cause());
        }
      });
  }
```
otherwise we risk calling handle on null and getting an NPE, hope this makes sense, I believe these extra checks should be trivial performance wise.

src/main/java/io/vertx/core/impl/future/FutureImpl.java
```diff
@@ -127,6 +127,39 @@ public void onFailure(Throwable failure) {
     return this;
   }
 
+  @Override
+  public Future<T> onComplete(Handler<T> successHandler, Handler<Throwable> failureHandler) {
+    Objects.requireNonNull(successHandler, "No null handler accepted");
```
vietj, line 132: I think that prevents onFailure/onSuccess to work

samuel-hawker: Are you sure? Since this is the FutureImpl, I think this does not relate to the onComplete written in Future?
This method isn't being used in Future is it?
(i.e. the onSuccess and onFailure for FutureImpl are separate?)

If that isn't the case I can remove the null checks from here and instead call them in the addListener onSuccess and onFailure overrides?

vietj (approved):

vietj:
@samuel-hawker thanks for the contribution, can you port the PR to the 4.x branch as well ?

Issue #4840, opened 2023-09-07
Don’t capture a stacktrace when creating a ReplyException


magicprinc:
I suppose this is a mistake

```io.vertx.core.eventbus.Message#fail```
creates ```ReplyException``` with stack trace.

In this case, it is not "real" exception, but error code and error message,
so I think the stack trace disabling was simply forgotten.

How to reproduce:
```java
  @Test void testMessageReplyBug () throws InterruptedException{
    var q = new LinkedBlockingQueue<Message<String>>();

    eventBus.consumer("x.y.z", (Message<String> msg) ->q.offer(msg));

    var f = eventBus.request("x.y.z", "Hello!");

    Message<String> msg = q.take();
    assertEquals("x.y.z", msg.address());
    assertNotNull(msg.replyAddress());
    msg.fail(404, "not found");

    f.onComplete(ar -> {
      assertNotNull(ar.cause());
      assertEquals(0, ar.cause().getStackTrace().length);
// ERROR [vert.x-eventloop-thread-0] io.vertx.core.impl.ContextBase::Unhandled exception
//org.opentest4j.AssertionFailedError: expected: <0> but was: <72>
    });
  }
```

Why is it bad?

1) Throwable#fillInStackTrace() is VERY heavy operation (takes long time)
2) ReplyExceptionMessageCodec doesn't use stack trace (doesn't serialize or deserialize them)

So we lose time to collect stack trace in vain.

PR with fix:
https://github.com/eclipse-vertx/vert.x/pull/4844

magicprinc:
Updated fix after review
https://github.com/magicprinc/vert.x/commit/673b7c78bd70bbd7b61157b7807e260c5b65c409

magicprinc:
BTW I have signed Eclipse Contributor Agreement

magicprinc:
finally a PR
https://github.com/eclipse-vertx/vert.x/pull/4844

tsegismont:
Fixed by #4844

magicprinc:
Dear reader!

If you have problems and read this, then know:

My original idea was

/** Collect stack trace (heavy operation!). Default false */
public static boolean WITH_STACK_TRACE = isTrueStr(System.getProperty("vertx.stacktrace"));
...

public ReplyException(ReplyFailure failureType, int failureCode, String message) {
  super(message, !WITH_STACK_TRACE);
  this.failureType = failureType;
  this.failureCode = failureCode;
}
[...]

PR #4841, opened 2023-09-07
Fix reply stack trace

magicprinc:
https://github.com/eclipse-vertx/vert.x/issues/4840


magicprinc:
two PR where merged :-(

PR #4844, opened 2023-09-08
#4840 bugfix Disable ReplyException stacktrace

magicprinc:
bugfix: https://github.com/eclipse-vertx/vert.x/issues/4840

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/ReplyException.java
+++ b/src/main/java/io/vertx/core/eventbus/ReplyException.java
@@ -27,6 +27,19 @@ public class ReplyException extends VertxException {
   private final int failureCode;
 
   /**
+   * Create a ReplyException with all attributes.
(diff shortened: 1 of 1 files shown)


tsegismont (approved):
Look good to me, thank you @magicprinc 

Indeed, in most cases the stacktrace will not be necessary.

magicprinc:
BTW, my original idea was

```java
/** Collect stack trace (heavy operation!). Default false */
public static boolean WITH_STACK_TRACE = isTrueStr(System.getProperty("vertx.stacktrace"));
...

public ReplyException(ReplyFailure failureType, int failureCode, String message) {
  super(message, !WITH_STACK_TRACE);
  this.failureType = failureType;
  this.failureCode = failureCode;
}
```

so user can enable them in some rare cases if needed

PR #4845, opened 2023-09-09
feat: onComplete method new signature

samuel-hawker:
Add Future implementation
Add FutureImpl implementations
Refactor to re-use new code for onSuccess and onFailure

Motivation:

Add onComplete method to Future class with onSuccess and onFailure mappers
Closes https://github.com/eclipse-vertx/vert.x/issues/4829
Cherry-pick of #4830


yeikel:
As stated in the email list, this is a subtle breaking change : "Reference to 'onComplete' is ambiguous, both 'onComplete(Handler<AsyncResult<T>>)' and 'onComplete(Handler<T>, Handler<Throwable>)' match"

Issue #4851, opened 2023-09-11
ConcurrentHashSet is redundant with ConcurrentHashMap#newKeySet()


magicprinc:
It seems that ```io.vertx.core.impl.ConcurrentHashSet``` is a copy of JDK's ```Collections.newSetFromMap```.
The Best Code is No Code At All.
It is probably a good idea to review and replace this class with two utility methods:

```java
public static <E> Set<E> concurrentHashSet (){
  return Collections.newSetFromMap(new ConcurrentHashMap<>());
}

public static <E> Set<E> concurrentHashSet (int initialSize){
  return Collections.newSetFromMap(new ConcurrentHashMap<>(initialSize));
}
```

PR: https://github.com/eclipse-vertx/vert.x/pull/4852

He-Pin:
Better to use ConcurrentHashMap.newKeySet

tsegismont:
You're welcome to contribute a PR for this, as well as in the impacted projects:

- https://github.com/search?q=org%3Avert-x3+ConcurrentHashSet&type=code
- https://github.com/search?q=org%3Aeclipse-vertx+ConcurrentHashSet&type=code

magicprinc:
I would like to, but I still have failed tests when I build vert.x-core locally on my Windows10/JDK17.
So I am not ready yet... :-(

BTW, I would enjoy if you find my other PR useful:
https://github.com/eclipse-vertx/vert.x/pull/4844
https://github.com/eclipse-vertx/vert.x/pull/4847
https://github.com/eclipse-vertx/vert.x/pull/4850

They have more benefits than ConcurrentHashSet removial.

He-Pin:
If you don't mind, i can do this too:), @magicprinc ,a nice find!

magicprinc:
@tsegismont 
Anyway. Without tests :-)
https://github.com/eclipse-vertx/vert.x/pull/4852

@He-Pin Could you test it? And add test for concurrentHashSet implementations too?

magicprinc:
1) I have replaced method with its body.
2) Fix in vertx-service-discovery: https://github.com/vert-x3/vertx-service-discovery/pull/172
3) Fix in vertx-mqtt: https://github.com/vert-x3/vertx-mqtt/pull/243

Done!

tsegismont:
@magicprinc there's one more usage in vertx-mqtt, can you please take care of it before we merge this PR?

magicprinc:
[...]

PR #4852, opened 2023-09-11
#4851 ConcurrentHashSet→Utils.concurrentHashSet

magicprinc:
Issue: https://github.com/eclipse-vertx/vert.x/issues/4851

It seems that io.vertx.core.impl.ConcurrentHashSet is a copy of JDK's Collections.newSetFromMap.
The Best Code is No Code At All.
It is probably a good idea to review and replace this class with two utility methods.


src/main/java/io/vertx/core/impl/Utils.java
```diff
@@ -45,4 +49,11 @@ public static boolean isWindows() {
     return isWindows;
   }
 
+  public static <E> Set<E> concurrentHashSet (){
```
tsegismont, line 52: It is not necessary to add new methods.

Where a `ConcurrentHashSet` is needed, we can use `java.util.concurrent.ConcurrentHashMap#newKeySet()`

src/test/java/io/vertx/core/ComplexHATest.java
```diff
@@ -40,11 +37,11 @@
  */
 public class ComplexHATest extends VertxTestBase {
 
-  protected ClusterManager getClusterManager() {
+  @Override protected ClusterManager getClusterManager() {
```
tsegismont, line 40: Please configure the formatter of your IDE or text editor. The annotation should be on a separate line above.

src/test/java/io/vertx/core/eventbus/LocalEventBusTest.java
```diff
@@ -41,7 +41,7 @@ public class LocalEventBusTest extends EventBusTestBase {
   private EventBusInternal eb;
   private boolean running;
 
-  public void setUp() throws Exception {
+  @Override public void setUp() throws Exception {
```
tsegismont, line 44: Please configure the formatter of your IDE or text editor. The annotation should be on a separate line above.

src/test/java/io/vertx/core/eventbus/LocalEventBusTest.java
```diff
@@ -52,7 +52,7 @@ public void setUp() throws Exception {
     running = true;
   }
 
-  protected void tearDown() throws Exception {
+  @Override protected void tearDown() throws Exception {
```
tsegismont, line 55: Please configure the formatter of your IDE or text editor. The annotation should be on a separate line above.

tsegismont (changes requested):
Thank you @magicprinc 

Can you please make a few changes?

src/main/java/io/vertx/core/impl/Utils.java
He-Pin: Yes, use this method, I used it in my pekko pr too.

magicprinc:
@tsegismont fixed. Please see.

Changed implementation to 
```java
public static <E> Set<E> concurrentHashSet (){
  return ConcurrentHashMap.newKeySet();// we can change implementation
}
```
or do you mean replace Util.concurrentHashSet with its body.

I think this separate method is a more robust and readable solution.

magicprinc:
@He-Pin Could you fix other Vert.x modules (if any) which are using this - deleted - class

He-Pin:
@magicprinc Hi, thanks for the ping, I think it would be better for you to do it, and I don't have a write access to this project, then I have to make a PR to yours and that will make the thing a little complicated :)

magicprinc:
@He-Pin @tsegismont I can't find any other ConcurrentHashSet usages besides the corrected ones, so we probably have managed it :-)

src/main/java/io/vertx/core/impl/Utils.java
tsegismont: Please remove this method, it is not necessary.

tsegismont:
@magicprinc before we merge this PR we also need updates for the projects that depend on this class (see https://github.com/eclipse-vertx/vert.x/issues/4851#issuecomment-1713554985)

magicprinc:
@tsegismont All your tasks are done! 👍
1) I have replaced method with its body.
2) Fix in vertx-service-discovery: https://github.com/vert-x3/vertx-service-discovery/pull/172
3) Fix in vertx-mqtt: https://github.com/vert-x3/vertx-mqtt/pull/243

src/main/java/io/vertx/core/impl/DeploymentManager.java
```diff
@@ -122,24 +121,24 @@ public Future<Void> undeployAll() {
...
-    context.runOnContext(v -> {
-      try {
-        completionHandler.handle(result);
-      } catch (Throwable t) {
-        log.error("Failure in calling handler", t);
-        throw t;
-      }
-    });
-  }
+//  private <T> void reportFailure(Throwable t, ContextInternal context, Promise<T> completionHandler) {
```
tsegismont, line 124: Can you please remove the methods instead of commenting?

tsegismont (commented):
@magicprinc one last comment, thank you

src/main/java/io/vertx/core/impl/DeploymentManager.java
magicprinc: Deleted!

tsegismont: Thank you

tsegismont (approved):

magicprinc:
@He-Pin Could you review this too? 🙏

He-Pin (commented):
Lgtm

PR #4853, opened 2023-09-11
Http client improvements

vietj:
A set of improvements and deprecations preparing for Vert.x 5 changes

### Goal 1

The `HttpClient` interface aims in Vert.x 5 to contains only HTTP interactions related methods, unfortunately this interface contains other methods which need to be deprecated and exposed in a more suitable API

- pool related methods such as `redirectHandler`, `updateSSLOptions` and `connectionHandler`, these methods actually act on the pool of the `HttpClient` : it has been chosen to deprecate these methods and expose them in a new `HttpClientPool` interface (which extends `HttpClient`).
- WebSocket methods are moved to a new `WebSocketClient`, actually the `HttpClient` has currently two connection manager: HTTP connection manager and WebSocket connection manager. `WebSocketClient` only holds the WebSocket connection manager without the extra burden of the HTTP connection manager.

In Vert.x 5 we want to have the opportunity to map an existing connection to an `HttpClient` interface:

```java
// Speculative API
pool.getConnection(conn -> {
  // An HTTP client connection could be an HttpClient API
  HttpClient client = conn;

  // Schedule a request on this particular connection
  // which might be HTTP/1 (queued) or HTTP/2 (multiplexed)
  Future<Buffer> fut = client.request(HttpMethod.GET, "/requestUri", request -> request.send().compose(resp -> resp.body));
});

```

### Goal 2

Prepare for address resolver (Vert.x 5 feature), the resolver only makes sense for HTTP services and not WebSocket. The split between HTTP and WebSocket makes the API more consistent with this respect and does prevent the expectation that the resolver will act on WebSocket.

```java
// Instead of using a SocketAddress we use instead another kind of Address
// Address has been introduced in Vert.x 5 to extends address resolution beyond DNS hostname resolution
Address address = ServiceAddress.create("quote-of-the-day-service");

// Use an HTTP client capable of resolving address to the actual endpoint
Future<HttpClientRequest> request = client.request(address, HttpMethod.GET, "/quote");
```

### Goal 3

Provide a non racy API for connecting WebSocket from a non vertx thread.

### WebSocketClient

A new `WebSocketClient` replaces the `HttpClient#webSocket` methods

```java
// Before
HttpClient httpClient = vertx.createHttpClient();
Future<WebSocket> fut2 = httpClient.webSocket(connectOptions);

// After
WebSocketClient wsClient = vertx.createWebSocketClient();
Future<WebSocket> fut2 = wsClient.connect(connectOptions);
```

### HttpClientPool

A new `HttpClientPool` extends `HttpClient` and exposes connection related methods

- `redirectHandler`
- `connectionHandler`
- `updateSSLOptions`

Corresponding methods in `HttpClientOptions` are deprecated in Vert.x 4 and removed in Vert.x 5

```java
// Existing code in Vert.x 4, nothing wrong here

// However this should not be done anymore because this assumes we are dealing with a pool
// it would not make sense with an existing connection
// this should be avoided in Vert.x 4 and certainly breaks with Vert.x 5 that removed these methods
client.connectionHandler(conn -> ...); 
client.updateSSLOptions(newSSLOptions);

// Actually now createHttpClient() returns an HttpClientPool
// HttpClientPool extends HttpClient 
// In Vert.x 4 this interface overrides the deprecated methods and remove the deprecation
// In Vert.x 5 this interface defines these methods since they are removed from HttpClient

// So the bottom line is to change your code to use HttpClientPool if you need to use such interfaces
// otherwise you can continue using HttpClient instead
HttpClientPool client = vertx.createHttpClient();

// And it makes sense to have the opportunity to handle connection when using a pool
client.connectionHandler(conn -> ...);

// Same for updating SSL options
client.updateSSLOptions(newSSLOptions);
```

### ClientWebSocket

A new `ClientWebSocket` allows to create a WebSocket in a un-connected state with explicit `connect` methods.

This allows to set handlers before connecting which is useful when connecting WebSocket from a non Vert.x thread.

```java
client
  .webSocket()
  .textMessageHandler(msg -> ...)
  .connect(connectOptions);
```

### DIFF
--- a/src/main/generated/io/vertx/core/http/WebSocketClientOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/http/WebSocketClientOptionsConverter.java
@@ -40,6 +40,16 @@ public class WebSocketClientOptionsConverter {
             obj.setCompressionRequestServerNoContext((Boolean)member.getValue());
           }
           break;
+        case "defaultHost":
+          if (member.getValue() instanceof String) {
+            obj.setDefaultHost((String)member.getValue());
+          }
+          break;
+        case "defaultPort":
+          if (member.getValue() instanceof Number) {
+            obj.setDefaultPort(((Number)member.getValue()).intValue());
+          }
+          break;
         case "maxConnections":
           if (member.getValue() instanceof Number) {
             obj.setMaxConnections(((Number)member.getValue()).intValue());
@@ -98,6 +108,10 @@ public class WebSocketClientOptionsConverter {
     json.put("compressionAllowClientNoContext", obj.getCompressionAllowClientNoContext());
     json.put("compressionLevel", obj.getCompressionLevel());
     json.put("compressionRequestServerNoContext", obj.getCompressionRequestServerNoContext());
+    if (obj.getDefaultHost() != null) {
+      json.put("defaultHost", obj.getDefaultHost());
+    }
+    json.put("defaultPort", obj.getDefaultPort());
     json.put("maxConnections", obj.getMaxConnections());
     json.put("maxFrameSize", obj.getMaxFrameSize());
     json.put("maxMessageSize", obj.getMaxMessageSize());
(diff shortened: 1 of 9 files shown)


juddgaddie:
Is there any way to share comon client options between HTTP and websockets ?
e.g. setTrustAll.

Or do I have to just ensure when interacting with the same host e.g. www.example.com
I set setTrustAll(true) on both httpClientOptions and then webSocketClientOptions and maintian both.

setTrustAll is just 1 example, there are many other common options.

Issue #4854, opened 2023-09-12
Replace NoStackTraceThrowable with NoStackTraceException


magicprinc:
Vert.x historically uses NoStackTraceThrowable (extends Throwable) for lightweight exception without heavy stack trace collection process.

But:
1) Since 2023 there is NoStackTraceException whith the same meaning which extends VertxException (RuntimeException).
2) Extending Throwable is considered a bad practice.

magicprinc:
PR: https://github.com/eclipse-vertx/vert.x/pull/4855

senyaak:
Any change it will be merged?

kolstae:
This doesn't seem to be included for 5.0 - that's quite disappointing 😞

vietj:
I believe for Vert.x 5 we should keep the NoStackTraceThrowable and make it extends VertxException since the most important issue is that catching a RuntimeException with virtual threads does not produce the expected result.

The exception is still marked as deprecated for removal.

vietj:
See https://github.com/eclipse-vertx/vert.x/issues/5012 which report the actual issue

vietj:
we might be able to remove NoStackTraceException though because it is not often used as much as NoStackTraceException used to be. 

Keeping the throwable, deprecating it and making it extend VertxException solves the virtual thread / Kotlin issue and allows users to use VertxException instead since it will extend this type. We would remove that in Vert.x 6 most likely.

kolstae:
If we can remove the catch `Throwable` code, that's perfect thanks!

vietj:
@kolstae please have a look https://github.com/eclipse-vertx/vert.x/pull/5507

Issue #4082, opened 2021-08-26
Use of NoStackTraceThrowable


kolstae:
When using kotlin coroutines the use of `NoStackTraceThrowable` causes problems that are really hard to find.

This code will not catch the error

```kotlin
    try {
        pool.preparedQuery("select true = $1").execute(Tuple.of(12)).await()
    } catch (e: Exception) {
        LOG.error("Expected: {}", e.toString())
    }
```
to catch the error we would need to catch `Throwable` which forces us to check if it's an `Error`.
When using futures directly, this isn't as much of a problem I guess.

kolstae:
See https://github.com/eclipse-vertx/vert.x/pull/5507

PR #4855, opened 2023-09-12
Replace NoStackTraceThrowable with NoStackTraceException #4854

magicprinc:
See https://github.com/eclipse-vertx/vert.x/issues/4854

### DIFF
--- a/src/main/java/io/vertx/core/impl/NoStackTraceThrowable.java
+++ b/src/main/java/io/vertx/core/impl/NoStackTraceThrowable.java
@@ -12,11 +12,12 @@
 package io.vertx.core.impl;
 
 /**
+ * @deprecated Use {@link NoStackTraceException} instead.
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
-public class NoStackTraceThrowable extends Throwable {
-
-  public NoStackTraceThrowable(String message) {
-    super(message, null, false, false);
+@Deprecated
+public class NoStackTraceThrowable extends NoStackTraceException {
+  public NoStackTraceThrowable (String message){
+    super(message);
   }
 }
--- a/src/main/java/io/vertx/core/Promise.java
+++ b/src/main/java/io/vertx/core/Promise.java
@@ -13,7 +13,7 @@ package io.vertx.core;
 import io.vertx.codegen.annotations.CacheReturn;
 import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.VertxGen;
-import io.vertx.core.impl.NoStackTraceThrowable;
+import io.vertx.core.impl.NoStackTraceException;
 import io.vertx.core.impl.future.PromiseImpl;
 
 /**
@@ -137,7 +137,7 @@ public interface Promise<T> extends Handler<AsyncResult<T>> {
    * @return false when the future is already completed
    */
   default boolean tryFail(String message) {
-    return tryFail(new NoStackTraceThrowable(message));
+    return tryFail(new NoStackTraceException(message));
   }
 
   /**
--- a/src/main/java/io/vertx/core/dns/DnsException.java
+++ b/src/main/java/io/vertx/core/dns/DnsException.java
@@ -11,7 +11,7 @@
 
 package io.vertx.core.dns;
 
-import io.vertx.core.impl.NoStackTraceThrowable;
+import io.vertx.core.impl.NoStackTraceException;
 import java.util.Objects;
 
 /**
@@ -20,10 +20,10 @@ import java.util.Objects;
  *
  * @author <a href="mailto:nmaurer@redhat.com">Norman Maurer</a>
  */
-public final class DnsException extends NoStackTraceThrowable {
+public final class DnsException extends NoStackTraceException {
 
   private static final String ERROR_MESSAGE_PREFIX = "DNS query error occurred: ";
-  private DnsResponseCode code;
+  private final DnsResponseCode code;
 
   public DnsException(DnsResponseCode code) {
     super(ERROR_MESSAGE_PREFIX + code);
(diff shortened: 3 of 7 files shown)


magicprinc:
NoStackTraceThrowable is marked as @Deprecated with javadoc ```@deprecated Use {@link NoStackTraceException} instead.```, but not deleted.
It allows fixing other vert.x modules eventually without breaking or fixing all at once.

src/main/java/io/vertx/core/impl/NoStackTraceThrowable.java
```diff
@@ -12,11 +12,12 @@
...
 
 /**
+ * @deprecated Use {@link NoStackTraceException} instead.
  * @author <a href="http://tfox.org">Tim Fox</a>
  */
-public class NoStackTraceThrowable extends Throwable {
-
-  public NoStackTraceThrowable(String message) {
-    super(message, null, false, false);
+@Deprecated
```
vietj, line 18: why do we need to deprecate that ?

magicprinc: Do you mean:
1) delete it
or
2) keep it without detraction?
-
1) Delete: Great Idea! I completely agree! But this class is used in two other modules:
https://github.com/search?q=org%3Aeclipse-vertx+NoStackTraceThrowable&type=code
(vertx-junit5 and vertx-sql-client)
And I don't have time to fix them too 🙏
With the deprecation, you can defer this fix.

2) Keep it
But why?
a) It duplicates the fuctionality
b) Direct Throwable subclass is bad. Evenn here in this project there were already problems with it
https://github.com/eclipse-vertx/vert.x/issues/4082

magicprinc: Extending Throwable directly is generally considered a bad practice in Java for several reasons:

1) **Hierarchy of Exceptions**: In Java, Throwable is the superclass of all errors and exceptions. It has two direct subclasses: Error and Exception [baeldung.com](https://www.baeldung.com/java-catch-throwable-bad-practice). Error and its subclasses represent serious problems that a reasonable application should not try to catch because they indicate conditions that an application usually cannot or should not recover from, such as OutOfMemoryError or StackOverflowError. On the other hand, Exception and its subclasses represent conditions that a reasonable application might want to catch and handle, such as FileNotFoundException or NullPointerException [baeldung.com](https://www.baeldung.com/java-catch-throwable-bad-practice). By extending Throwable directly, you are bypassing this hierarchy and potentially creating confusion about the nature and severity of the condition that your class represents.

2) **Exception Handling**: Exceptions in Java are meant to be caught and handled appropriately. The general rule in handling exceptions is that the try-catch block must be as specific as possible in catching exceptions. That is, a catch-all scenario must be avoided [baeldung.com](https://www.baeldung.com/java-catch-throwable-bad-practice). By extending Throwable, you might be creating a type that can be caught by a general catch (Throwable t) block, leading to inappropriate handling of different types of errors and exceptions.

3) **Code Clarity**: Many applications assume only Exception and Error exist as subclasses of Throwable [devtut.github.io](https://devtut.github.io/java/java-pitfalls-exception-usage.html). Creating a new class that extends Throwable directly can be confusing for other developers who might expect your class to be a subclass of Exception or Error. It's more clear and conventional to extend Exception or one of its subclasses when you need to create a custom checked exception, and to extend RuntimeException or one of its subclasses when you need to create a custom unchecked exception.

4) Checked vs Unchecked: Classes directly extending Throwable are checked, as Exception is, so the compiler will enforce that this Throwable is declared to be thrown or caught. This can lead to verbose and cluttered code if the Throwable is used frequently [blog.vanillajava.blog](https://blog.vanillajava.blog/2021/12/unusual-java-stacktrace-extends.html).

**In conclusion**, while Java allows you to extend Throwable directly, it's generally better to extend Exception, RuntimeException, or one of their subclasses when you need to create a custom exception. This approach respects the established hierarchy of exceptions in Java, leads to clearer and more maintainable code, and allows for more appropriate and specific exception handling.

vietj: I mean that removing it or changing its inheritance can be a un whishable breaking change for user applications.

magicprinc: I have returned the inheritance hierarchy: we will not break anything. All 

Annotation ```@Deprecated``` (with detailed javadoc) doesn't harm any project using this class, but clearly shows a better solution.

magicprinc: I have finally found the best solution:
```VertxException``` → ```NoStackTraceThrowable``` → ```NoStackTraceException```

If some code catches NoStackTraceThrowable, it won't break (NoStackTraceException is NoStackTraceThrowable), but shows deprecated waring.
After all code migrates to NoStackTraceException, NoStackTraceThrowable can be removed:
```VertxException``` → ```NoStackTraceException```.
Nobody will miss changing parent of NoStackTraceThrowable from Throwable to RuntimeException (VertxException).

I can't imagine situation, where it can break code

- catch (NoStackTraceThrowable e) - works
- catch (NoStackTraceThrowable e) - works
- catch (Throwable e) - works

vietj: VertxException will not extend NoStackTraceThrowable as seducing as it can look

magicprinc: The point is:
- extending Throwable is bad practice.
- there were and will be problems with such practice.
- this old exception duplicates functionality of a new one (NoStackTraceThrowable ≈ NoStackTraceException)

I hope you know the better solution. this decision is entirely up to you 🙏👍

vietj: ok let's deprecate NoStrackTraceThrowable and redirect to use NoStackTraceException instead for now

vietj: I think we should not deprecate NoStackTraceThrowable itself, but instead its constructors

magicprinc: A great idea! Done (PR is updated)

Issue #4861, opened 2023-09-16
EventBus swallows original cause and only reports failureCode and message


glains:
The current implementation of `Message` swallows the original `Exception` instead of wrapping them into the `ReplyException`.
I am using the `EventBus` locally to route commands in my application and plan to add an exception mapper for REST.

The code and message currently used is not suficcient for my requirements.
```
default void fail(int failureCode, String message) {
    reply(new ReplyException(ReplyFailure.RECIPIENT_FAILURE, failureCode, message));
}
```

Workaround:
* Wrap an `.onFailure()` call around consumers and replace the body of the message with the exception
* Verify if the body is indeed that exception and proceed accordingly

Would it be reasonable to add the cause of the exception to the `ReplyException`?

tsegismont:
The `ReplyException` codec does not encode the cause/stacktrace into account (we wouldn't be able to decode it back into an Exception).

A good practice is too log the exception with an id that you send as as part of the failure message

glains:
Yeah, I already thought that decoding the message is not possible, but I am using a local codec here. That's why I thought it was possible to add it to the cause.

tsegismont:
Then you can create a subclass of `ReplyException` and register a custom codec for it with `io.vertx.core.eventbus.impl.CodecManager#registerDefaultCodec`.

magicprinc:
@tsegismont Here you are
https://github.com/eclipse-vertx/vert.x/pull/4865

jakub-bochenski:
Why is this only targeted for 5.0.0? Couldn't this be merged to the 4.x branch?

tsegismont:
@jakub-bochenski this can be backported, indeed. I will take care of it.

tsegismont:
@jakub-bochenski https://github.com/eclipse-vertx/vert.x/pull/5336

ben1222:
Before 4.5.11, the `ReplyException` called `VertxException(String message)` instead of the `VertxException(String message, boolean noStackTrace)` so it was possible to call `initCause` afterwards to set the cause...
Now it actually backported both #4840 and this fix to the 4.x, which makes the `initCause` cannot be used.

That said, I do not have a better solution that can both remove the stacktrace and keep the `initCause` working... not sure why `Throwable` does not allow such use case...

PR #4865, opened 2023-09-18
EventBus swallows original cause and only reports failureCode and message #4861

magicprinc:
Add ctor with cause

https://github.com/eclipse-vertx/vert.x/issues/4861

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/ReplyException.java
+++ b/src/main/java/io/vertx/core/eventbus/ReplyException.java
@@ -41,13 +41,14 @@ public class ReplyException extends VertxException {
 
   /**
    * Create a ReplyException with all attributes, including cause.<br>
-   * Default {@link io.vertx.core.eventbus.impl.codecs.ReplyExceptionMessageCodec} doesn't support Cause!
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/core/eventbus/ReplyException.java
```diff
@@ -39,6 +39,20 @@ public ReplyException(ReplyFailure failureType, int failureCode, String message,
...
 
+  /**
+   * Create a ReplyException with all attributes, including cause.<br>
+   * Default {@link io.vertx.core.eventbus.impl.codecs.ReplyExceptionMessageCodec} doesn't support Cause!
+   *
+   * @param failureType  the failure type
+   * @param failureCode  the failure code (e.g. 404)
+   * @param message  the failure message
+   */
+  public ReplyException(ReplyFailure failureType, int failureCode, String message, Throwable cause, boolean noStackTrace) {
```
tsegismont, line 50: This constructor should have `protected` visibility instead of `public` and we shall document that it is meant to be used for extension, together with a custom codec.

tsegismont (commented):
Thank you @magicprinc

src/main/java/io/vertx/core/eventbus/ReplyException.java
magicprinc: Fixed

tsegismont (approved):

PR #4888, opened 2023-10-06
Fix traffic shaping options

frankgh:
Motivation:

When setting non-zero values for `io.vertx.core.net.TrafficShapingOptions` `maxDelayToWait` and `checkIntervalForStats` properties, a NullPointerException is encountered because `options.getCheckIntervalForStatsTimeUnit()` returns `null`, as it never is set.

Additionally, the time units expected in netty’s `io.netty.handler.traffic.GlobalTrafficShapingHandler` are milliseconds, however we are incorrectly passing seconds while initializing the `GlobalTrafficShapingHandler` object.

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/net/TrafficShapingOptions.java
+++ b/src/main/java/io/vertx/core/net/TrafficShapingOptions.java
@@ -11,6 +11,7 @@
 
 package io.vertx.core.net;
 
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 import io.netty.util.internal.ObjectUtil;
@@ -24,7 +25,7 @@ import io.vertx.core.json.JsonObject;
 @Unstable
 @DataObject(generateConverter = true, publicConverter = false)
 public class TrafficShapingOptions {
-  /*
+  /**
    * Default inbound bandwidth limit in bytes/sec = 0 (0 implies unthrottled)
    */
   public static final long DEFAULT_INBOUND_GLOBAL_BANDWIDTH_LIMIT = 0;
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -127,7 +155,7 @@ public TrafficShapingOptions setCheckIntervalForStats(long checkIntervalForStats
    * @return a reference to this, so the API can be used fluently
    */
   public TrafficShapingOptions setCheckIntervalForStatsTimeUnit(TimeUnit checkIntervalForStatsTimeUnit) {
-    this.maxDelayToWaitTimeUnit = checkIntervalForStatsTimeUnit;
```
frankgh, line 130: this assignment is incorrect, and is causing NPEs when the check interval is set

src/main/java/io/vertx/core/net/impl/TCPServerBase.java
```diff
@@ -106,12 +106,12 @@ private GlobalTrafficShapingHandler createTrafficShapingHandler(EventLoopGroup e
     }
     GlobalTrafficShapingHandler trafficShapingHandler;
     if (options.getMaxDelayToWait() != 0 && options.getCheckIntervalForStats() != 0) {
-      long maxDelayToWaitInSeconds = options.getMaxDelayToWaitTimeUnit().toSeconds(options.getMaxDelayToWait());
-      long checkIntervalForStatsInSeconds = options.getCheckIntervalForStatsTimeUnit().toSeconds(options.getCheckIntervalForStats());
-      trafficShapingHandler = new GlobalTrafficShapingHandler(eventLoopGroup, options.getOutboundGlobalBandwidth(), options.getInboundGlobalBandwidth(), checkIntervalForStatsInSeconds, maxDelayToWaitInSeconds);
+      long maxDelayToWaitInMillis = options.getMaxDelayToWaitTimeUnit().toMillis(options.getMaxDelayToWait());
+      long checkIntervalForStatsInMillis = options.getCheckIntervalForStatsTimeUnit().toMillis(options.getCheckIntervalForStats());
```
frankgh, line 110: these options are expected in millis, however we are passing seconds here. Fixed it to pass millis instead

vietj:
thanks for the contribution

PR #4898, opened 2023-10-11
HTTP/2 RST frame flood protection configuration

vietj:
Netty 4.1.100.Final added an HTTP/2 RST frame flood protection, this makes it configurable.


tsegismont (commented):
LGTM

PR #4901, opened 2023-10-11
Ensure that the task is not in the tasks queue when it is rejected from the executor

### LINKED ISSUE #4900: TasksQueue doesn't remove entry from tasks on RejectedExecutionException
### Version

`master` branch.

### Context

### Do you have a reproducer?

```
        @Override
        }
    };

    assertThatThrownBy(() -> {

    // this fails in main
```

### Steps to reproduce

See test.

### Extra

ahus1:
Closes #4900

Motivation:

This is a PR to address #4900. It is rather crude, and uses package visibility. While this does the job to make the assertions, I'd be happy to receive advice on how to change it to match the Vert.x's project way.

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/impl/TaskQueue.java
+++ b/src/main/java/io/vertx/core/impl/TaskQueue.java
@@ -125,7 +125,6 @@ public class TaskQueue {
    */
   public void execute(Runnable task, Executor executor) {
     synchronized (tasks) {
-      tasks.add(new ExecuteTask(task, executor));
       if (currentExecutor == null) {
         currentExecutor = executor;
         try {
@@ -135,6 +134,18 @@ public class TaskQueue {
           throw e;
         }
       }
+      // Add the task after the runner has been accepted to the executor
+      // to cover the case of a rejected execution exception.
+      tasks.add(new ExecuteTask(task, executor));
+    }
+  }
+
+  /**
+   * Test if the task queue is empty and no current executor is running anymore.
+   */
+  public boolean isEmpty() {
+    synchronized (tasks) {
+      return tasks.isEmpty() && currentExecutor == null;
     }
   }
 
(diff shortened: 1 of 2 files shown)


ahus1:
Updated the PR to ensure proper encapsulation in `TaskQueue` even for tests by adding an `isEmpty()` method.

vietj:
can you elaborate when/how this happens ? if a task is rejected it usually means that vertx is closed, unless I am wrong ?

ahus1:
For me, Vert.x is used as part of Quarkus, which is then the base of Keycloak. At the moment I'm running some load tests around Keycloak, and I want to have a first naive load shedding in place.

In a Keycloak setup, Quarkus replaces the executor pool of Vert.x with its own thread pool I can ask Quarkus to limit the backlog of the executor thread pool using a configuration `QUARKUS_THREAD_POOL_QUEUE_SIZE=100`. Once I do this, the thread doesn't allow for an unlimited queue size. If under high load the queue size is exceeded, the requests are rejected. 

I'm discussing more elaborate means of load shedding and not relying on the thread pool queue site. Until that is in place, this is all I have. 

I hope this adds the context the issue is missing.

vietj:
thanks for the context

src/main/java/io/vertx/core/impl/TaskQueue.java
```diff
@@ -125,19 +125,30 @@ public Consumer<Runnable> unschedule() {
    */
   public void execute(Runnable task, Executor executor) {
     synchronized (tasks) {
-      tasks.add(new ExecuteTask(task, executor));
+      ExecuteTask executeTask = new ExecuteTask(task, executor);
+      tasks.add(executeTask);
```
vietj, line 129: I think instead since we are in a synchronized block, we could move this addition to the bottom of the block after we are sure the executor accepted the task instead of adding/removing it

ahus1:
@vietj - thank you for the review. I've moved the statement, and also added a comment to state why this is the right place. Please give it another review when you have the time.

vietj:
thanks for the contribution, do you mind porting this to the 4.x branch ?

ahus1:
@vietj - thank you for reviewing and merging this PR. Here's the backport for this change to 4.x: #4904

PR #4904, opened 2023-10-12
[4.x] Ensure that the task is not in the tasks queue when it is rejected from the executor

### LINKED ISSUE #4900: TasksQueue doesn't remove entry from tasks on RejectedExecutionException

ahus1:
Closes #4900

This is an adapted backport of #4901 that adds the also the handling of rejected executions which was previously only available in the 5.x branch.


ahus1:
Keeping this in draft state until the parent PR is discussed.

vietj (approved):

vietj:
thanks for the contribution @ahus1

PR #4908, opened 2023-10-13
Fix IllegalStateException in DnsClient when DNS server responds with DNAME record

mnylen:
Motivation:

When trying to resolve any record type using `DnsClient`, if the DNS server replies back with a DNAME record, the handler passed to `DnsClient#resolve*()` methods is never called. This is because internally, `DnsClientImpl` uses `RecordDecoder` which currently throws `IllegalStateException` if it encounters a record type it does not know how to decode. `DnsClientImpl` does not catch this exception, causing the query to never be completed.

If you try to do this:

```
dnsClient.resolveMX("some-domain-with-dname-record.com", System::out::println);
```

The handler that would print the result is never called, instead this is logged:

```
WARNING: An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
java.lang.IllegalStateException: Unsupported resource record type [id: DNAME(39)].
    at io.vertx.core.dns.impl.decoder.RecordDecoder.decode(RecordDecoder.java:223)
    at io.vertx.core.dns.impl.DnsClientImpl$Query.handle(DnsClientImpl.java:336)
    at io.vertx.core.dns.impl.DnsClientImpl$1.channelRead0(DnsClientImpl.java:89)
    at io.vertx.core.dns.impl.DnsClientImpl$1.channelRead0(DnsClientImpl.java:83)
    at ...
```

In this PR, this is fixed by making `RecordDecoder` throw `DecoderException` when the DNS server replies back with any record type it doesn't know how to decode. Then, `DnsClientImpl` is made to catch this and fail the query when this exception is encountered during decoding. Another potential fix would be add DNAME record decoder to the list of supported decoders in `RecordDecoder`, but even though it would fix this specific issue, I figured it's better to have the `try..catch` as that would allow handling any other unsupported record types in similar fashion.

Note that DNAME (Delegation Name) records are valid DNS records and are used as a means of redirecting the whole domain (including all of it's subdomains) to another domain. So very similar to CNAME, but with CNAME the alias applies to only single subdomain, whereas with DNAME it applies to the name and all of its subnames. Properly handling these would likely involve following those redirects. This PR does not attempt to do that, it just fixes the issue where the error is not propagated to the caller.

Unfortunately testing this scenario requires extending `DnsMessageEncoder` from apacheds-protocol-dns in a way I'm not happy about, as the encoder map used by the class is final and also unmodifidable. The class does not provide any usable extension points either, so subclasses would need to rewrite significant portions. I briefly checked the option of upgrading to newer 2.X versions, but the issue remains there as well. I'll welcome any suggestions if there is a better way than outright copying the class to vert.x as I've done here.

There are also potential other issues with `DnsClientImpl` that I didn't have time to look at too deeply. For example, if the DNS server returns malformed supported record type that the `RecordDecoder` can't parse, the exception is catched and logged, but `RecordDecoder` will in this case return null. `DnsClientImpl` does not check for null value, so there is potential for it to throw `NullPointerException` in this scenario. This would lead to a similar issue where the handler would never be called. Unfortunately extending `FakeDNSServer` to support returning malformed records (as it uses standards conforming apacheds-protocol-dns underneath) is not trivial, so fixing this would require either taking a leap of faith with no tests, or perhaps rewriting portions of `FakeDNSServer`.

Conformance:

### DIFF
--- a/src/test/java/io/vertx/test/fakedns/DnsMessageEncoder.java
+++ b/src/test/java/io/vertx/test/fakedns/DnsMessageEncoder.java
@@ -1,0 +1,233 @@
+package io.vertx.test.fakedns;
+
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ *
+ */
+
+
+  import java.io.IOException;
+  import java.util.*;
+
+  import org.apache.directory.server.dns.io.encoder.*;
+  import org.apache.directory.server.dns.messages.DnsMessage;
+  import org.apache.directory.server.dns.messages.MessageType;
+  import org.apache.directory.server.dns.messages.OpCode;
+  import org.apache.directory.server.dns.messages.QuestionRecord;
+  import org.apache.directory.server.dns.messages.RecordType;
+  import org.apache.directory.server.dns.messages.ResourceRecord;
+  import org.apache.directory.server.dns.messages.ResponseCode;
+  import org.apache.directory.server.i18n.I18n;
+  import org.apache.mina.core.buffer.IoBuffer;
+  import org.slf4j.Logger;
+  import org.slf4j.LoggerFactory;
+
+
+/**
+ * An encoder for DNS messages.  The primary usage of the DnsMessageEncoder is
+ * to call the <code>encode(ByteBuffer, DnsMessage)</code> method which will
+ * write the message to the outgoing ByteBuffer according to the DnsMessage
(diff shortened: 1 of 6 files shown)


src/test/java/io/vertx/test/fakedns/DnsMessageEncoder.java
```diff
@@ -0,0 +1,233 @@
...
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ *
+ */
+
+
+  import java.io.IOException;
```
vietj, line 24: why do we need to fork this class ?

mnylen: I explained this in the commit message where this was done: https://github.com/eclipse-vertx/vert.x/pull/4908/commits/50d7d82964b17436bdee94fce312dd69493bee44 Plus also noted this in the PR description.

TL;DR: `DnsMessageEncoder` does not know how to encode `DNAME` records at the moment,. So in order to reproduce the bug using `FakeDNSServer`, the `DnsMessageEncoder` class needs to be extended.

Unfortutely the design of `DnsMessageEncoder` does not lend itself to extension. The `DEFAULT_ENCODERS` map is final and uses `Collections.unmodifiableMap`, and the method that is used to look up the encoder from `DEFAULT_ENCODERS` has `private` access modifier, so doing a subclass wouldn't work either.

vietj:
Merged it will be back-ported to 4.5.0, note there is a reimplement of the 
`DnsClient` in 5.0 with the `DnsNameResolver` of Netty, that will change the behaviour : DNAME records will be decoded but simply ignored

vietj:
thanks for the contribution @mnylen

Issue #4906, opened 2023-10-12
Password-less internal keystore forbids start of Vert.x on Semeru 17 JVM in FIPS mode


Nicolas-Duboc-IBM:
### Version

Reproduced on 4.4.4 and 4.4.5

### Context

When trying to run Quarkus with TLS enabled on a [Semeru 17 JVM](https://developer.ibm.com/languages/java/semeru-runtimes/downloads/) with [FIPS mode enabled](https://www.ibm.com/support/pages/node/6612693), the Quarkus application fails to start with the stacktrace below.

This is because the FIPS mode of Semeru requires that private keys in keystores have a long password (minimum 16 character long) but the [`KeyStoreHelper`](https://github.com/eclipse-vertx/vert.x/blob/7e1dc81371b21b1205c3cbe09da11ef89c15de44/src/main/java/io/vertx/core/net/impl/KeyStoreHelper.java#L162C83-L162C83) class of Vertx creates an intermediate in-memory keystore with a null password.

The stacktrace is 

```
2023-10-12 08:55:36,121 ERROR [io.qua.run.Application] (main) Failed to start application (with profile [prod]): java.lang.RuntimeException: Failed to start quarkus
	at io.quarkus.runner.ApplicationImpl.doStart(Unknown Source)
	at io.quarkus.runtime.Application.start(Application.java:101)
	at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:111)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:71)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:44)
	at io.quarkus.runtime.Quarkus.run(Quarkus.java:124)
	at io.quarkus.runner.GeneratedMain.main(Unknown Source)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at io.quarkus.bootstrap.runner.QuarkusEntryPoint.doRun(QuarkusEntryPoint.java:61)
[...]

PR #4910, opened 2023-10-16
Use a long password on generated in-memory KeyStore (4.x branch)

Nicolas-Duboc-IBM:
Motivation:

Fix for https://github.com/eclipse-vertx/vert.x/issues/4906 on 4.x branch

Hopefully trivial enough to have no wrong side-effect.


Nicolas-Duboc-IBM:
Thanks,

PR #4912, opened 2023-10-17
[4.4.x] Ensure that the task is not in the tasks queue when it is rejected from the executor

### LINKED ISSUE #4900: TasksQueue doesn't remove entry from tasks on RejectedExecutionException

ahus1:
Closes #4900

This is backport of the PR #4904 as the Keycloak project and Quarkus' LTS 3.2 are still on Vert.x 4.4.x. 

It would be great if this could could eventually make it into a 4.4.x release. Not sure when and if the Quarkus team would upgrade, I will need to check that separately.

@vietj - I'd be happy if you could have a  look as you've reviewed the original PRs. Thanks!

### DIFF
--- a/src/main/java/io/vertx/core/impl/TaskQueue.java
+++ b/src/main/java/io/vertx/core/impl/TaskQueue.java
@@ -16,6 +16,7 @@ import io.vertx.core.impl.logging.LoggerFactory;
 
 import java.util.LinkedList;
 import java.util.concurrent.Executor;
+import java.util.concurrent.RejectedExecutionException;
 
 /**
  * A task queue that always run all tasks in order. The executor to run the tasks is passed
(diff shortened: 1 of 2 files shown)


ahus1:
Update: Quarkus confirmed they will stay on 4.4.x for a while, so I'd appreciate this backport to 4.4.x and eventually a release.

vietj:
@ahus1 we are not yet ready to work on a 4.4 release because we plan to release 4.5 within a couple of weeks, how does that sound ?

ahus1:
This sounds fine, let's merge now and wait until some more bits accumulate. As I wrote earlier, the Quarkus team said they might not immediately go for 4.5.

PR #4920, opened 2023-10-26
Use a virtual threads friendly pool with Jackson

mariofusco:
This pull request is a follow up of the work done to make the [internal pool used by Jackson configurable and pluggable](https://github.com/FasterXML/jackson-core/pull/1064). It is not intended to be merged until Jackson 2.16 won't be released. Also this is the porting of the [same pull request sent to quarkus](https://github.com/quarkusio/quarkus/pull/35816).

### DIFF
--- a/src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
+++ b/src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
@@ -1,0 +1,243 @@
+package io.vertx.core.json.jackson;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+import java.util.function.Predicate;
+import java.util.stream.IntStream;
+
+import com.fasterxml.jackson.core.util.BufferRecycler;
+import com.fasterxml.jackson.core.util.JsonRecyclerPools;
+import com.fasterxml.jackson.core.util.RecyclerPool;
+
+/**
+ * This is a custom implementation of the Jackson's {@link RecyclerPool} intended to work equally well with both
+ * platform and virtual threads. This pool works regardless of the version of the JVM in use and internally uses
+ * 2 distinct pools one for platform threads (which is exactly the same {@link ThreadLocal} based one provided
+ * by Jackson out of the box) and the other designed for being virtual threads friendly. It switches between
+ * the 2 only depending on the nature of thread (virtual or not) requiring the acquisition of a pooled resource,
+ * obtained via {@link MethodHandle} to guarantee compatibility also with old JVM versions. The pool also guarantees
+ * that the pooled resource is always released to the same internal pool from where it has been acquired, regardless
+ * if the releasing thread is different from the one that originally made the acquisition.
+ * <p>
+ * The virtual thread friendly inner pool is implemented with N striped linked lists using a simple lock free
+ * algorithm based on CAS. The striping is performed shuffling the id of the thread requiring to acquire a pooled
+ * resource with a xorshift based computation. The resulting of this computation is also stored in the pooled resource,
+ * bringing the twofold advantage of always releasing the resource in the same bucket from where it has been taken
+ * regardless if the releasing thread is different from the one that did the acquisition and avoiding the need of
+ * recalculating the position of that bucket also during the release. The heads of the linked lists are hold in an
+ * {@link AtomicReferenceArray} where each head has a distance of 16 positions from the adjacent ones to prevent
+ * the false sharing problem.
+ */
+public class HybridJacksonPool implements RecyclerPool<BufferRecycler> {
+
+  private static final HybridJacksonPool INSTANCE = new HybridJacksonPool();
+
+  private static final Predicate<Thread> isVirtual = VirtualPredicate.findIsVirtualPredicate();
+
+  private final RecyclerPool<BufferRecycler> nativePool = JsonRecyclerPools.threadLocalPool();
+
+  private static class VirtualPoolHolder {
+    // Lazy on-demand initialization
+    private static final StripedLockFreePool virtualPool = new StripedLockFreePool(Runtime.getRuntime().availableProcessors());
+  }
+
+  private HybridJacksonPool() {
+    // prevent external instantiation
+  }
+
+  public static HybridJacksonPool getInstance() {
+    return INSTANCE;
+  }
+
+  @Override
+  public BufferRecycler acquirePooled() {
+    return isVirtual.test(Thread.currentThread()) ?
+      VirtualPoolHolder.virtualPool.acquirePooled() :
+      nativePool.acquirePooled();
+  }
+
+  @Override
+  public void releasePooled(BufferRecycler bufferRecycler) {
+    if (bufferRecycler instanceof VThreadBufferRecycler) {
+      // if it is a PooledBufferRecycler it has been acquired by a virtual thread, so it has to be release to the same pool
+      VirtualPoolHolder.virtualPool.releasePooled(bufferRecycler);
+    }
+    // the native thread pool is based on ThreadLocal, so it doesn't have anything to do on release
+  }
+
+  static class StripedLockFreePool implements RecyclerPool<BufferRecycler> {
+
+    private static final int CACHE_LINE_SHIFT = 4;
+
+    private static final int CACHE_LINE_PADDING = 1 << CACHE_LINE_SHIFT;
+
+    private final XorShiftThreadProbe threadProbe;
+
+    private final AtomicReferenceArray<Node> topStacks;
+
+    private final int stripesCount;
+
+    public StripedLockFreePool(int stripesCount) {
+      if (stripesCount <= 0) {
+        throw new IllegalArgumentException("Expecting a stripesCount that is larger than 0");
+      }
+
+      this.stripesCount = stripesCount;
+      int size = roundToPowerOfTwo(stripesCount);
+      this.topStacks = new AtomicReferenceArray<>(size * CACHE_LINE_PADDING);
+
+      int mask = (size - 1) << CACHE_LINE_SHIFT;
+      this.threadProbe = new XorShiftThreadProbe(mask);
+    }
+
+    public int size() {
+      return stackSizes().sum();
+    }
+
+    public int[] stackStats() {
+      return stackSizes().toArray();
+    }
+
+    private IntStream stackSizes() {
+      return IntStream.range(0, stripesCount).map(i -> {
+        Node node = topStacks.get(i * CACHE_LINE_PADDING);
+        return node == null ? 0 : node.level;
+      });
+    }
+
+    @Override
+    public BufferRecycler acquirePooled() {
+      int index = threadProbe.index();
+
+      Node currentHead = topStacks.get(index);
+      while (true) {
+        if (currentHead == null) {
+          return new VThreadBufferRecycler(index);
+        }
+
+        if (topStacks.compareAndSet(index, currentHead, currentHead.next)) {
+          currentHead.next = null;
+          return currentHead.value;
+        } else {
+          currentHead = topStacks.get(index);
+        }
+      }
+    }
+
+    @Override
+    public void releasePooled(BufferRecycler recycler) {
+      VThreadBufferRecycler vThreadBufferRecycler = (VThreadBufferRecycler) recycler;
+      Node newHead = new Node(vThreadBufferRecycler);
+
+      Node next = topStacks.get(vThreadBufferRecycler.slot);
+      while (true) {
+        newHead.level = next == null ? 1 : next.level + 1;
+        if (topStacks.compareAndSet(vThreadBufferRecycler.slot, next, newHead)) {
+          newHead.next = next;
+          return;
+        } else {
+          next = topStacks.get(vThreadBufferRecycler.slot);
+        }
+      }
+    }
+
+    private static class Node {
+      final VThreadBufferRecycler value;
+      Node next;
+      int level = 0;
+
+      Node(VThreadBufferRecycler value) {
+        this.value = value;
+      }
+    }
+  }
+
+  private static class VThreadBufferRecycler extends BufferRecycler {
+    private final int slot;
+
+    VThreadBufferRecycler(int slot) {
+      this.slot = slot;
+    }
+  }
+
+  private static class VirtualPredicate {
+    private static final MethodHandle virtualMh = findVirtualMH();
+
+    private static MethodHandle findVirtualMH() {
+      try {
+        return MethodHandles.publicLookup().findVirtual(Thread.class, "isVirtual",
+          MethodType.methodType(boolean.class));
+      } catch (Exception e) {
+        return null;
+      }
+    }
+
+    private static Predicate<Thread> findIsVirtualPredicate() {
+      if (virtualMh != null) {
+        return new Predicate<Thread>() {
+          @Override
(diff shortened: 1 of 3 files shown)


vietj:
merge is on hold until 2.16.0 is released

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,208 @@
...
+
+  static final RecyclerPool INSTANCE = new HybridJacksonPool();
+
+  private static final Predicate<Thread> isVirtual = VirtualPredicate.findIsVirtualPredicate();
+
+  private final RecyclerPool<BufferRecycler> nativePool = JsonRecyclerPools.threadLocalPool();
+
+  static class VirtualPoolHolder {
+    // Lazy on-demand initialization
+    private static final RecyclerPool<BufferRecycler> virtualPool = new StripedLockFreePool(4);
```
franz1981, line 42: use the Runtime.getRuntime().availableProcessors here

mariofusco: Done, thanks for reminder.

vietj:
@mariofusco @franz1981 we will release 4.5.0 this week, I'm assuming we cannot yet merge this PR ?

mariofusco:
> @mariofusco @franz1981 we will release 4.5.0 this week, I'm assuming we cannot yet merge this PR ?

The final release of Jackson 2.16 is not out yet ( see https://central.sonatype.com/artifact/com.fasterxml.jackson.core/jackson-core/versions ) so, unless it is ok for you to depend on that 2.16.0-rc1, I'm afraid that this PR cannot be merged yet.

vietj:
I will move this issue to 4.5.1

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,208 @@
...
+            try {
+              return (boolean) virtualMh.invokeExact(thread);
+            } catch (Throwable e) {
+              throw new RuntimeException(e);
+            }
+          }
+        };
+      }
+
+      return new Predicate<Thread>() {
```
magicprinc, line 161: why not simply: `return thread->false;`

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,208 @@
...
+ * resource with a xorshift based computation. The resulting of this computation is also stored in the pooled resource,
+ * bringing the twofold advantage of always releasing the resource in the same bucket from where it has been taken
+ * regardless if the releasing thread is different from the one that did the acquisition and avoiding the need of
+ * recalculating the position of that bucket also during the release. The heads of the linked lists are hold in an
+ * {@link AtomicReferenceArray} where each head has a distance of 16 positions from the adjacent ones to prevent
+ * the false sharing problem.
+ */
+public class HybridJacksonPool implements RecyclerPool<BufferRecycler> {
+
+  static final RecyclerPool INSTANCE = new HybridJacksonPool();
```
magicprinc, line 34: Any chance to make it public? 🙏

magicprinc:
As you know, Jackson 2.16.0 is released
https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core/2.16.0

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
mariofusco: I made the field private and added a public `getInstance()` method.

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
mariofusco: We decided to avoid using lambdas here for performance reasons.

mariofusco:
@vietj I think that this is ready to be merged now. I bumped the jackson version with this other PR https://github.com/vert-x3/vertx-dependencies/pull/157

franz1981:
@mariofusco Please add some tests to be sure that it works as expected; by mocking the thread id(s) and verifying it hits/distribute the requests and/or reuse the same pool, if expected

franz1981 (changes requested):
@mariofusco Please add some tests to be sure that it works as expected (including not getting any weird array index out of bound exception); by mocking the thread id(s) and verifying it hits/distribute the requests and/or reuse the same pool, if expected.

Additionally, add something (I would use a counter into the node(s)) to report the current size of the pool; 
this type of telemetry will help to troubleshoot errors.

Decide if it makes sense to add a fastthreadlocal-only version to replace the current existing one too, and give vertx users a boost without relying on the existing implementation from Jackson (which uses SoftRef too, bleah!)

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,212 @@
...
+
+  private static class StripedLockFreePool implements RecyclerPool<BufferRecycler> {
+
+    private static final int CACHE_LINE_SHIFT = 4;
+
+    private static final int CACHE_LINE_PADDING = 1 << CACHE_LINE_SHIFT;
+
+    private final XorShiftThreadProbe threadProbe;
+
+    private final AtomicReferenceArray<Node> heads;
```
franz1981, line 72: replace with `topStack`

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    @Override
+    public void releasePooled(BufferRecycler recycler) {
+      VThreadBufferRecycler vThreadBufferRecycler = (VThreadBufferRecycler) recycler;
+      Node newHead = new Node(vThreadBufferRecycler);
+
+      Node next = heads.get(vThreadBufferRecycler.slot);
+      while (true) {
+        if (heads.compareAndSet(vThreadBufferRecycler.slot, next, newHead)) {
```
franz1981, line 112: right before this: `newHead.level = next.level + 1` - level doesn't need to be volatile, bacause is correctly acquired/released by reading the `Node`'s instance and performing this `compareAndset`.

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,212 @@
...
+          return;
+        } else {
+          next = heads.get(vThreadBufferRecycler.slot);
+        }
+      }
+    }
+
+    private static class Node {
+      final VThreadBufferRecycler value;
+      Node next;
```
franz1981, line 123: Add `int level = 0` field

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,212 @@
...
+        throw new IllegalArgumentException("Expecting a stripesCount that is larger than 0");
+      }
+
+      int size = roundToPowerOfTwo(stripesCount);
+      this.heads = new AtomicReferenceArray<>(size * CACHE_LINE_PADDING);
+
+      int mask = (size - 1) << CACHE_LINE_SHIFT;
+      this.threadProbe = new XorShiftThreadProbe(mask);
+    }
+
```
franz1981, line 85: add a method `int size` which round on all the topStacks and sum `topStack.ge().level` for each

mariofusco:
@franz1981 I implemented what you suggested and added a couple of simple tests. Please let me know if there is something more that I could check or improve.

mariofusco:
@vietj @franz1981 is this ok to be merged?

src/test/java/io/vertx/core/json/jackson/HybridJacksonPoolTest.java
```diff
@@ -0,0 +1,68 @@
...
+  public void testVirtualThreadPoolWithMultipleThreads() {
+    int stripesCount = 4;
+    HybridJacksonPool.StripedLockFreePool virtualPool = new HybridJacksonPool.StripedLockFreePool(stripesCount);
+    int nThreads = 100;
+    BufferRecycler[] resources = new BufferRecycler[nThreads];
+    CountDownLatch latch = new CountDownLatch(nThreads);
+
+    for (int i = 0; i < nThreads; i++) {
+      int threadIndex = i;
+      new Thread(() -> {
```
franz1981, line 37: This cat kill the CI, let's find an alternative way to achieve the same (expose a method which use the Id and always hit the striped variant?)

src/test/java/io/vertx/core/json/jackson/HybridJacksonPoolTest.java
```diff
@@ -0,0 +1,68 @@
...
+    assertEquals(0, virtualPool.size());
+
+    for (int i = 0; i < nThreads; i++) {
+      virtualPool.releasePooled(resources[i]);
+    }
+
+    // check that all resources have been released back to the pool
+    assertEquals(nThreads, virtualPool.size());
+
+    int avgResourcesNrPerStripe = nThreads / stripesCount;
```
franz1981, line 57: I love it, nice!!!

src/test/java/io/vertx/core/json/jackson/HybridJacksonPoolTest.java
franz1981: Or assume virtual threads are available and use them via reflection?

src/test/java/io/vertx/core/json/jackson/HybridJacksonPoolTest.java
```diff
@@ -26,6 +29,11 @@ public void testVirtualThreadPoolWithSingleThread() {
 
   @Test
   public void testVirtualThreadPoolWithMultipleThreads() {
+    if (!VirtualThreadRunner.hasVirtualThread()) {
```
franz1981, line 32: you can use assume for this

mariofusco:
@vietj @franz1981 I modified my test case to use virtual threads if they are available and completely skip the test otherwise. Please let me know if there is something else to be done here. If not I don't see any further reason to hold this pull request.

src/test/java/io/vertx/core/json/jackson/HybridJacksonPoolTest.java
```diff
@@ -65,4 +73,29 @@ public void testVirtualThreadPoolWithMultipleThreads() {
       assertTrue(poolStats[i] <= maxResourcesNrPerStripe);
     }
   }
+
+  private static class VirtualThreadRunner {
```
franz1981, line 77: nice!

franz1981 (changes requested):

franz1981 (approved):
LGTM well done @mariofusco !

mariofusco:
Hi @vietj is there any other outstanding issue why you're holding this PR? (not rushing, just checking if I missed to do something)

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,238 @@
...
+ * The virtual thread friendly inner pool is implemented with N striped linked lists using a simple lock free
+ * algorithm based on CAS. The striping is performed shuffling the id of the thread requiring to acquire a pooled
+ * resource with a xorshift based computation. The resulting of this computation is also stored in the pooled resource,
+ * bringing the twofold advantage of always releasing the resource in the same bucket from where it has been taken
+ * regardless if the releasing thread is different from the one that did the acquisition and avoiding the need of
+ * recalculating the position of that bucket also during the release. The heads of the linked lists are hold in an
+ * {@link AtomicReferenceArray} where each head has a distance of 16 positions from the adjacent ones to prevent
+ * the false sharing problem.
+ */
+public class HybridJacksonPool implements RecyclerPool<BufferRecycler> {
```
vietj, line 33: I think we can make this package private instead of being public

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
```diff
@@ -0,0 +1,238 @@
...
+      return new Predicate<Thread>() {
+        @Override
+        public boolean test(Thread thread) {
+          return false;
+        }
+      };
+    }
+  }
+
+  private static class XorShiftThreadProbe {
```
vietj, line 200: Some documentation is welcome :-)

vietj (changes requested):
A couple of minor improvements

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
mariofusco: The idea was to use the same pool also in Quarkus, if I make it package private this will be no longer possible.

mariofusco:
> A couple of minor improvements

@vietj I addressed one and replied to the other.

src/main/java/io/vertx/core/json/jackson/HybridJacksonPool.java
vietj: ah ok this makes sense, then just add a slight comment that this class is reused elsewhere so we won't refactor it if that's not needed in order to preserve compatiblity

mariofusco: Done.

vietj:
@mariofusco can you squash everything in a single commit with a nice commit message :-) ?

vietj:
@mariofusco I see some activity on this PR, is it safe to merge ?

mariofusco:
> @mariofusco I see some activity on this PR, is it safe to merge ?

@vietj I squashed everything as you asked, but now I'm also testing the Quarkus integration as suggested by @franz1981 ( see https://github.com/quarkusio/quarkus/pull/38196 ). Please just give me a bit of time to check that it is also working as expected.

mariofusco:
@vietj @franz1981 I'm sorry, but I tested this integration and found a serious bug in `jackson-databind`, see https://github.com/quarkusio/quarkus/pull/38196#issuecomment-1892430695

I'm afraid that this pull request cannot be merged yet, till the problem won't be fixed in jackson and for sure we cannot use jackson 2.16.1 for this.

vietj:
thanks for raising this @mariofusco so for now it will be on hold

mariofusco:
@vietj @franz1981 @geoand I further investigated the problem, found the root cause and with my [last commit](https://github.com/eclipse-vertx/vert.x/pull/4920/commits/766f2326f3c3f7f85c657d3f8f9c2e2ba258946b) made my hybrid pool compatible with jackson-databind behaviour. I also test that with this fix all Quarkus integration tests in the jackson extension are green. 

I'm [still discussing](https://github.com/FasterXML/jackson-databind/issues/4321#issuecomment-1893504676) with @cowtowncoder if what jackson-databind does is correct or not, or if we can find a way to improve it. Anyway with my fix that discussion is now unrelated with this pool implementation and now I believe that it is totally safe to merge this as it is. 

@vietj if it is ok for you, please feel free merging this pull request.

mariofusco:
Hi @vietj, the problem that I found is now fixed in Jackson, but as I wrote in my former comment, with my last commit I found a solution that works even without that fix and I checked that this behaves correctly also on the quarkus integration side. Is there any outstanding work here or any other reason why you're still holding this pull request?

vietj (approved):

mariofusco:
> @mariofusco merged, thanks for the contribution, can you port this to master branch ?

@vietj see https://github.com/eclipse-vertx/vert.x/pull/5084

vietj:
Awesome thanks!

PR #4928, opened 2023-11-02
Save creating (and validating) a peer INET address if not required

franz1981:
This is the twin of https://github.com/eclipse-vertx/vert.x/pull/4927


franz1981:
in order to make this more effective I should change `EndpointKey` too and NOT force non-null peer address in case ssl is disabled, because right now is enforcing this:

```java
  EndpointKey(boolean ssl, ProxyOptions proxyOptions, SocketAddress serverAddr, SocketAddress peerAddr) {
    if (serverAddr == null) {
      throw new NullPointerException("No null server address");
    }
    if (peerAddr == null) {
      throw new NullPointerException("No null peer address");
    }
    this.ssl = ssl;
    this.proxyOptions = proxyOptions;
    this.peerAddr = peerAddr;
    this.serverAddr = serverAddr;
  }
```
let me know if I can proceed and push this forward till than

PR #4933, opened 2023-11-08
Expose HTTP client control over creation of un-pooled HTTP connections

vietj:
In some scenario (e.g. load-testing), it is preferable to use un-pooled HTTP connection.

This expose at the internal API level, the creation of HTTP connection and HTTP requests. This will eventually be exposed in the Vert.x API but the current API is still not mature enough to do that and require more work and might require breaking changes.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
@@ -89,6 +89,11 @@ class Http2ClientConnection extends Http2ConnectionBase implements HttpClientCon
   }
 
   @Override
+  public long activeStreams() {
+    return handler.connection().numActiveStreams();
+  }
+
+  @Override
   boolean onGoAwaySent(GoAway goAway) {
     boolean goneAway = super.onGoAwaySent(goAway);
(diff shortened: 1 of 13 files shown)


vietj:
ping @franz1981

franz1981:
Give that thanks to this we now have control at the connection level, I can expect some existing public API to verify the number of opened stream too?
For the load testing use case I would need to track it myself in case I want to "limit" it

vietj:
I think we could add this, although you can track it yourself as well

vietj:
@franz1981 added a getter for the number of active streams in HttpClientConnection

src/main/java/io/vertx/core/http/impl/Http2ClientConnection.java
```diff
@@ -88,6 +88,11 @@ public long concurrency() {
...
   }
 
+  @Override
+  public long activeStreams() {
+    return handler.connection().numActiveStreams();
+  }
+
   @Override
   boolean onGoAwaySent(GoAway goAway) {
     boolean goneAway = super.onGoAwaySent(goAway);
```
franz1981, line 98: lovely, yep

Issue #4947, opened 2023-11-14
HostAndPort fails to parse a nip host


tsegismont:
[nip](https://nip.io/) hosts are often used in automated tests or development / testing environments.

For example, `127.0.0.1.nip.io` maps to `127.0.0.1`.

`io.vertx.core.net.impl.HostAndPortImpl#parseHostAndPort` returns `null` when parsing such a host+port string, because it assumes the host is `127.0.0.1` and expects the next character to be `:`

```java
  public static HostAndPortImpl parseHostAndPort(String s, int schemePort) { // s is 127.0.0.1.nip.io:8443
    int pos = parseHost(s, 0, s.length()); // pos is 9 because parseHost computes the host to be 127.0.0.1
    if (pos == s.length()) {
      return new HostAndPortImpl(s, schemePort);
    }
    if (pos < s.length() && s.charAt(pos) == ':') { // expected colon but finds dot
      String host = s.substring(0, pos);
      int port = 0;
      while (++pos < s.length()) {
        int digit = parseDigit(s, pos, s.length());
        if (digit == -1) {
          return null;
        }
        port = port * 10 + digit;
      }
      return new HostAndPortImpl(host, port);
    }
    return null; // returns null
  }
```

tsegismont:
Fixed by b407a30aa519b4263c28eba18bc0ea8da713d2a2

PR #4948, opened 2023-11-14
HostAndPort fails to parse a nip host

tsegismont:
See #4947

If `parseIPv4Address` finds an IP address, and the string still has characters which do not begin with `:`, this is not an IP address.

In other words, `10.0.0.1.nip.io` should not be parsed as IP address.

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/HostAndPortImpl.java
+++ b/src/main/java/io/vertx/core/net/impl/HostAndPortImpl.java
@@ -30,7 +30,7 @@ public class HostAndPortImpl implements HostAndPort {
         return -1;
       }
     }
-    return from;
+    return from < to && s.charAt(from + 1) != ':' ? -1 : from;
   }
 
   static int parseDecOctet(String s, int from, int to) {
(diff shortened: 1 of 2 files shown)


cescoffier (approved):

vietj:
I am not certain this is the correct fix to do

vietj:
The fix makes reg names parsed as IPV4 addresses when they are prefixed by an IPV4 string which makes the impl of `parseIPv4Address` wrong as it accepts more than it should and `testParseIPV4Address` has been added incorrect assertions.

Instead the code using `parseIPV4Address` should check when this method returns a positive value, than the returned position is actually the end of the string or followed by `:`.

The correct thing to do should be

```java
  static int parseHost(String val, int from, int to) {
    int pos;
    if ((pos = parseIPLiteral(val, from, to)) != -1) {
      return pos;
    } else if ((pos = parseIPv4Address(val, from, to)) != -1 && (pos == to || val.charAt(pos + 1 ) == ':') ) {
      return pos;
    } else if ((pos = parseRegName(val, from, to)) != -1) {
      return pos;
    }
    return -1;
  }
```

PR #4952, opened 2023-11-15
- allows for local only codec lookup when in clustered mode

Reef3rm4n:
Motivation:

When operating in clustered mode I want to allow local messages to perform the codec lookup as local.

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
@@ -119,14 +119,18 @@ public class EventBusImpl implements EventBusInternal, MetricsProvider {
 
   @Override
   public EventBus send(String address, Object message, DeliveryOptions options) {
-    MessageImpl msg = createMessage(true, address, options.getHeaders(), message, options.getCodecName());
+    MessageImpl msg = createMessage(true, options.isLocalOnly(), address, options.getHeaders(), message, options.getCodecName());
     sendOrPubInternal(msg, options, null);
     return this;
   }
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/eventbus/impl/EventBusImpl.java
```diff
@@ -253,9 +257,13 @@ public EventBusMetrics<?> getMetrics() {
     return metrics;
   }
 
-  public MessageImpl createMessage(boolean send, String address, MultiMap headers, Object body, String codecName) {
+  public MessageImpl createMessage(boolean send, boolean localOnly, String address, MultiMap headers, Object body, String codecName) {
     Objects.requireNonNull(address, "no null address accepted");
-    MessageCodec codec = codecManager.lookupCodec(body, codecName, true);
+    boolean local = true;
+    if (vertx.isClustered()) {
```
vietj, line 263: I think instead this should not be decided by this method and it should be passed by the caller, in other words the createMessage should just be dumb and apply what is said

Reef3rm4n: agreed, I will make the change

Reef3rm4n:
@vietj  is this ok now ?

tsegismont (approved):
LGTM, thank you @Reef3rm4n

tsegismont:
@Reef3rm4n If you'd like to have this in the next bugfix version, would you mind creating another PR to the `4.x` branch please?

Reef3rm4n:
@tsegismont will do

PR #4958, opened 2023-11-19
Future fix NPE, fast path for completed, fewer lines

magicprinc:
1) ```
} catch (InterruptedException e) {
      Utils.throwAsUnchecked(e.getCause());
```
could cause NPE

2) if Future.isComplete() simply return result

3)
```
Utils.throwAsUnchecked(cause());
return null;
```
looks ugly, can be easily optimized

### DIFF
--- a/src/main/java/io/vertx/core/Future.java
+++ b/src/main/java/io/vertx/core/Future.java
@@ -627,20 +627,26 @@ public interface Future<T> extends AsyncResult<T> {
    * @throws IllegalStateException when called from an event-loop thread or a non Vert.x thread
    */
   default T await() {
+    if (isComplete()) {
+      if (succeeded()) {
+        return result();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by magicprinc** [comment]
@vietj NPE is real. There is a test in PR

**Comment by vietj** [comment]
thanks for reporting the NPE @magicprinc

**Comment by magicprinc** [comment]
@vietj Any change to have in addition
`public static <E extends Throwable> RuntimeException throwAsUnchecked(Throwable t) throws E {`
instead of
`public static <E extends Throwable> void throwAsUnchecked(Throwable t) throws E {`

Please... 🙏😅

**Comment by magicprinc** [comment]
@vietj and "happy path" for already completed Futures

```
default T await() {
   if (isComplete()) {
      return ...
    }
```

Issue #4811, opened 2023-08-17
IndexOutOfBoundsException in RecordParser with HttpclientResponse


poryee:
### when streaming vertx httpclientresponse to record parser im faced with following kind of error

 ERROR [io.ver.cor.imp.ContextBase] (vert.x-eventloop-thread-1) Unhandled exception [Error Occurred After Shutdown]:
java.lang. Index0utOfBoundsException: writerIndex(40) + minWritableBytes(8192) exceeds maxCapacity(40): UnpooledslicedByteBuf(ridx: 0, widx: 48,
cap: 40/48, unwrapped: UnpooledByteBufAllocator$InstrumentedunpooledUnsafeHeapByteBuf(ridx: 42, widx: 42, cap: 64))
at io.netty.buffer.AbstractByteBuf.ensure₩ritableø(AbstractByteBuf. java:294)
at io.netty.buffer.AbstractByteBuf.ensureWritable (AbstractByteBuf. java:280)
at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf. java:1103)
at io.vertx.core.buffer.impl.BufferImpl.appendBuffer (BufferImpl. java:268)
I
at io.vertx.core.parsetools.impl.RecordParserImpl. handle (RecordParserImpl. java:283)
at io.vertx.core.parsetools.impl.RecordarserImpl. handle(RecordParserImpl. java:27)
at io.vertx.core.impl.ContextInternal.dispatch(ContextInternal. java:264)
at io.vertx.core.http.impl.HttpEventHandler.handleChunk(HttpEventHandler. java:51)
at io.vertx.core.http.impl.HttpclientResponseImpl. handleChunk(HttpClientResponseImpl. java:239)
at io.vertx.core.http.impl.HttplxClientConnection$streamImpl.lambdaŞnew$0(HttplxClientConnection. iava:452)
at io.vertx.core.streams.impl.InboundBuffer.handleEvent(InboundBuffer. java:255)
at io.vertx.core.streams.impl.InboundBuffer.write(InboundBuffer.java:134)
at io.vertx.core.http.impl.HttplxClientConnection$streamImpl.handleChunk(HttplxClientConnection. java:704)
at io.vertx.core.impl.EventLoopContext.execute(EventLoopContext. java:76)
at io.vertx.core.impl.ContextBase.execute(ContextBase. java:232)

I believe a similar ticket was resolved in version 3.9.3
#3546 

### Version
[...]

PR #4960, opened 2023-11-20
RecordParser may grow the initial buffer until the first event is emitted

tsegismont:
See #4811

This can be a problem if the received [Buffer](url) is backed by a Netty `ByteBuff` with low capacity.

Also, we shouldn't assume we can modify input.


tsegismont:
The test failures are due to:

```
java.lang.UnsupportedOperationException: OpenJdkSelfSignedCertGenerator not supported on the used JDK version
```

:man_facepalming:

tsegismont:
As agreed off-list, will merge this with an extra commit adding a comment in the `RecordParser` implementation.

PR #4967, opened 2023-11-22
[Feature] Future.await with timeout

magicprinc:
Future has await() method without timeout

It is easy to implement and nice to have await(timeout,TimeUnit) method


vietj:
that is not needed since it possible to use timeout on a future, i.e `await(future.timeout(timeout))`, instead we should document this usage in the documentation.

magicprinc:
@vietj Then https://github.com/eclipse-vertx/vert.x/pull/4958 
NullPointerException is real (see test)
And quick path for already completed futures is also a good idea

PR #4974, opened 2023-11-26
The Future await method should rethrow the ThreadInterruptedException without an NPE when it happens

vietj:


magicprinc:
more meat
https://github.com/eclipse-vertx/vert.x/pull/4958

Issue #4982, opened 2023-11-28
Racy behavior when writing HTTP/1.1 or HTTP/2.0 request from a worker and event loop thread


NilsRenaud:
### Version

This happens with Vert.x 4.4.6

### Context

When writing a request content from 2 threads (one being an event loop), the event loop write always happen **before** the write from the other thread. This leads to unordered writing on the underlying connection.

This happened to me while working with vertx-rx and offloading the content publication to another thread. In this case, the backpressure management uses the Event-loop thread alongside with the worker thread (see https://github.com/vert-x3/vertx-rx/issues/300)

### Reproducer

```java
import java.util.concurrent.CountDownLatch;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.atomic.AtomicBoolean;

import io.vertx.core.Future;
import io.vertx.core.Vertx;
import io.vertx.core.http.HttpClientRequest;
import io.vertx.core.http.HttpMethod;

public class SmallestReproducer {
    private static final Vertx VERTX = Vertx.vertx();
    private static final ExecutorService EXECUTOR = Executors.newFixedThreadPool(1);

    public static void main(String[] args) {
        createHttpServerWhichLogsReceivedBuffers();

        VERTX.createHttpClient()
             .request(HttpMethod.POST, 8080, "localhost", "/")
             .onSuccess(req -> send(req))
             .flatMap(req -> req.response())
             .flatMap(response -> response.body())
             .onSuccess(body -> System.out.println(body.toString()));
    }

    private static void send(HttpClientRequest req) {
        req.setChunked(true);
        Future<Void> future = req.sendHead();
        while (!future.isComplete()) { } // Wait for the header to be sent.
        AtomicBoolean latch = new AtomicBoolean(false);
        EXECUTOR.submit(() -> {
            req.write("1111");
            latch.set(true); // Release Event-loop thread
        });
        while (!latch.get()) { } // Active wait for the event to be published
        req.write("222222222");
        req.end();
        System.out.println(Thread.currentThread().getName() + " - ended");
    }

    // Nothing interesting here
    private static void createHttpServerWhichLogsReceivedBuffers() {
        CountDownLatch latch = new CountDownLatch(1);
        VERTX.createHttpServer()
             .requestHandler(req -> {
                 System.out.println("Received request");
                 req.bodyHandler(buf -> System.out.println("Received: " + buf.toString()));
                 req.endHandler(o -> {
[...]

PR #4983, opened 2023-11-28
Fix issue #4982 racy http writing when a writing from a non event loop thread concurrently

NilsRenaud:
Motivation:

See https://github.com/eclipse-vertx/vert.x/issues/4982

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -398,6 +398,7 @@ public class Http1xClientConnection extends Http1xConnectionBase<WebSocketImpl> 
     private Handler<MultiMap> endHandler;
     private Handler<Void> drainHandler;
     private Handler<Void> continueHandler;
+    private int writeInProgress = 0;
 
     private Handler<MultiMap> earlyHintsHandler;
     private Handler<Throwable> exceptionHandler;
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -570,12 +571,7 @@ private void writeBuffer(ByteBuf buff, boolean end, FutureListener<Void> listene
       } else {
         l = listener;
       }
-      EventLoop eventLoop = conn.context.nettyEventLoop();
-      if (eventLoop.inEventLoop()) {
-        conn.writeBuffer(this, buff, end, l);
-      } else {
-        eventLoop.execute(() -> writeBuffer(buff, end, l));
-      }
+      conn.writeBuffer(this, buff, end, l);
```
vietj, line 574: we need to actually keep if/else event loop block because we really want the method `writeBuffer` to always be executed on the event-loop and not be concurrency

NilsRenaud:
@vietj are you OK with this last version ?

src/test/java/io/vertx/core/http/Http2Test.java
```diff
@@ -1092,4 +1095,35 @@ public void testStreamResetErrorMapping() throws Exception {
     }));
     await();
   }
+
+  @Test
+  public void testConcurrentWrite() throws Exception {
```
vietj, line 1100: can we merge those two tests in a single in HttpTest class ?

NilsRenaud: yep, done, although there is a slight difference  between Http 1 and 2

NilsRenaud:
The failing tests looks unrelated to my changes 🤔

vietj:
> @vietj are you OK with this last version ?

it looks great :-)

vietj:
@NilsRenaud can you port the test to master branch ? I think it should not be buggy though (unless I'm wrong of course)

NilsRenaud:
here is the backported PR: https://github.com/eclipse-vertx/vert.x/pull/5009

NilsRenaud:
@vietj I would be interested in a backport of this fix in 4.4.x version. Would this be possible ?
The 4.5.0 release introduced breaking changes around WebSocket options so we won't be able to simply upgrade to 4.5.1 for our already released version :/

vietj:
which breaking changes are you referring to ?

NilsRenaud:
🤔 You're right they are not breaking changes but deprecations, so that's fine, we will be able to deal with that :)

PR #4984, opened 2023-11-28
Reduce garbage with small jackson encoded data (Fixes #4977)

### LINKED ISSUE #4977: JacksonCodec should better guess the allocated buffer
https://github.com/eclipse-vertx/vert.x/blob/1f016f43e5bca522670f3681994259b8eed1d415/src/main/java/io/vertx/core/json/jackson/JacksonCodec.java#L108 relies on default allocation, which is very small (256 bytes, but can be very big based on the actual need!). Subsequent reallocations, if any, will happen based on Netty default strategy, which is less then optimal, given that can quickly converge to very big values(see https://github.com/franz1981/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/AbstractByteBufAllocator.java#L277 for what is doing now).

There are 4 different approaches here:
- use a thread local tmp buffer (or something similar to the striped pool implemented for Loom - Jackson buffer pool)
- use a chunked approach which have smaller chunks and perform a final copy into an exact sized Netty one based on the list
- like the previous, but creating a Composite ByteBuf out of the list of chunks (which is dangerous in case we have too many abstract implementations of Netty buffers flying around in core and hot path methods)
- perform a better guess for interesting types (eg JsonArray/JsonObject) - but can be easily O(N)

I don't know how much important is the codec I have mentioned but is the default choice from what I can tell.
I'll go deeper in the code in the next days and maybe there is a better way to improve this.

franz1981:
Some numbers:

baseline:
```
Benchmark                                                  Mode  Cnt     Score    Error   Units
JsonEncodeBenchmark.tinyBufferJackson                      avgt   10     0.363 ±  0.001   us/op
JsonEncodeBenchmark.tinyBufferJackson:·gc.alloc.rate.norm  avgt   10   704.000 ±  0.001    B/op
```
With ba66dfa628ada94d9dac1d589255b0bcd080fdd0:
```
Benchmark                                                  Mode  Cnt     Score    Error   Units
JsonEncodeBenchmark.tinyBufferJackson                      avgt   10     0.350 ±  0.001   us/op
JsonEncodeBenchmark.tinyBufferJackson:·gc.alloc.rate.norm  avgt   10   512.000 ±  0.001    B/op
```
Performance is slightly increased (and will get better with highly concurrent allocations), but most importantly,
the GC pressure is highly decreased. 
It's not a case I've used the same json content we use for Techempower's JSON tests.

### DIFF
--- a/src/main/java/io/vertx/core/buffer/impl/VertxHeapByteBuf.java
+++ b/src/main/java/io/vertx/core/buffer/impl/VertxHeapByteBuf.java
@@ -13,15 +13,24 @@ package io.vertx.core.buffer.impl;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.UnpooledHeapByteBuf;
-import io.netty.buffer.UnpooledUnsafeHeapByteBuf;
 
 /**
  * An un-releasable, un-pooled, un-instrumented heap {@code ByteBuf}.
  */
 final class VertxHeapByteBuf extends UnpooledHeapByteBuf {
 
+  private static final byte[] EMPTY = new byte[0];
+
   public VertxHeapByteBuf(ByteBufAllocator alloc, int initialCapacity, int maxCapacity) {
     super(alloc, initialCapacity, maxCapacity);
+  }
+
+  @Override
+  protected byte[] allocateArray(int initialCapacity) {
+    if (initialCapacity == 0) {
+      return EMPTY;
+    }
+    return super.allocateArray(initialCapacity);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


franz1981:
I've added an additional change, playing a bit dirty with OpenJDK JIT behaviour to overcome a JIT limit related inlining, and getting

```
Benchmark                                                  Mode  Cnt     Score    Error   Units
JsonEncodeBenchmark.tinyBufferJackson                      avgt   10     0.328 ±  0.002   us/op
JsonEncodeBenchmark.tinyBufferJackson:·gc.alloc.rate.norm  avgt   10   512.000 ±  0.001    B/op
```
which is a global 10% improvement, in throughput, from the initial baseline.
I'll re-run all benchmarks in there, because probably the overall improvement is even bigger, given that having all the encoding mechanism now able to inline till `MaxInlineLevel` depth (which is now 15 by default) vs the previous `1` of `MaxRecursiveInlineLevel` is a big step forward.

franz1981:
need further analysis, because the commit at 62dc39abd48769d669298f11e18a9e4be8d5c18d is pretty unstable, if we perform some randomization on the json encoding in the warmup phase.

vietj:
can you backport to 4.x branch ?

PR #4995, opened 2023-12-02
JsonArray optimizations

magicprinc:
- Better stream() and spliterator() implementations
- ctor with initialCapacity
- better copy ctor

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -67,6 +67,13 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
   }
 
   /**
+   * Create an empty instance with initialCapacity
+   */
+  public JsonArray(int initialCapacity) {
+    list = new ArrayList<>(initialCapacity);
+  }
+
+  /**
    * Create an instance from a List. The List is not copied.
    *
    * @param list the underlying backing list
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -65,16 +66,27 @@ public JsonArray() {
...
+  }
+
   /**
    * Create an instance from a List. The List is not copied.
    *
    * @param list the underlying backing list
    */
-  public JsonArray(List list) {
-    if (list == null) {
+  @SuppressWarnings({"rawtypes", "unchecked"})
```
vietj, line 81: that is an avoidable breaking change, we need to keep the List constructor

vietj:
why this PR is also applying what your other PR about json object contains ?

we are not going to review the same code twice.

vietj:
in other words this PR seems to be focusing on JsonArray, it should not care about JsonObject unless there is a relationship

src/main/java/io/vertx/core/json/JsonArray.java
magicprinc: Does it really break?

I just thought widening is not a problem 🤷‍♀️

magicprinc:
I close this PR, sorry, no time for rebase and force-push - deadline at job.
This PR has no much value. So:
https://github.com/eclipse-vertx/vert.x/pull/5002

PR #4997, opened 2023-12-03
Emit tags suitable for OpenTelemetry 1.18 Semantic Conventions

Olamshin:
### Motivation:
Ensure that spans have the right attributes to be classified correctly in an OpenTelemetry backend. OpenTelemetry conventions should override OpenTracing conventions eventually since OpenTracing has been absorbed into OpenTelemetry.

### Approach:
Semantic conventions for [HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.18.0/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions) & [Messaging](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.18.0/specification/trace/semantic_conventions/messaging.md#messaging-attributes) show some required attributes for OpenTelemetry spans. Extra tags have been added and none removed to ensure backwards compatibility. Added tags seems to be satisfactory for telemetry backends like Uptrace.

This is an upstream fix for https://github.com/eclipse-vertx/vertx-tracing/issues/67

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/MessageTagExtractor.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/MessageTagExtractor.java
@@ -27,21 +27,31 @@ class MessageTagExtractor implements TagExtractor<Message<?>> {
 
   @Override
   public int len(Message<?> obj) {
-    return 1;
+    return 3;
   }
 
   @Override
   public String name(Message<?> obj, int index) {
-    if (index == 0) {
-      return "message_bus.destination";
+    switch (index) {
+      case 0:
+        return "message_bus.destination";
+      case 1:
+        return "messaging.system";
+      case 2:
+        return "messaging.operation";
     }
     throw new IndexOutOfBoundsException("Invalid tag index " + index);
   }
 
   @Override
   public String value(Message<?> obj, int index) {
-    if (index == 0) {
-      return obj.address();
+    switch (index) {
+      case 0:
+        return obj.address();
+      case 1:
+        return "vertx-eventbus";
+      case 2:
+        return "publish";
     }
     throw new IndexOutOfBoundsException("Invalid tag index " + index);
   }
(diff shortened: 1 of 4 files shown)


tsegismont:
@vietj why do you think it is a breaking change? The PR only adds tags.

Olamshin:
@vietj I believe this PR is not a breaking change, I did not change or remove tags. May I have your reasoning?

I would be opening a PR in vertx-tracing to update tests in that repo. I wanted to get buy-in with this PR and have it possibly merged. I am doing this so that I have passing tests when i open that PR.

vietj:
@Olamshin right, then use the same convention : `messaging.system` -> message.system

tsegismont:
> @Olamshin right, then use the same convention : `messaging.system` -> `message.system`

I don't understand, `messaging.system` is a new tag in this PR, that complies with the otel specification. Which spec do you refer to with `message.system`?

Btw, the opentracing spec is deprecated and superseded by otel (see https://github.com/opentracing/specification/issues/163)

I think we should update the comment in `TagExtractor`

Olamshin:
@vietj `messaging.system` is specific to OpenTelemetry spec and the reason for this PR. `message.system` would only make sense if adhering to a Vert.x standard/spec that is not OpenTracing. OpenTracing has been deprecated and the last version of its spec does not include `message.system`.

vietj:
> @vietj `messaging.system` is specific to OpenTelemetry spec and the reason for this PR. `message.system` would only make sense if adhering to a Vert.x standard/spec that is not OpenTracing. OpenTracing has been deprecated and the last version of its spec does not include `message.system`.

as I explained, use the same convention and translate the tag in vertx Otel implementation. We can consider using the new tag names in Vert.x 5.0

Olamshin:
> > @vietj `messaging.system` is specific to OpenTelemetry spec and the reason for this PR. `message.system` would only make sense if adhering to a Vert.x standard/spec that is not OpenTracing. OpenTracing has been deprecated and the last version of its spec does not include `message.system`.
> 
> as I explained, use the same convention and translate the tag in vertx Otel implementation. We can consider using the new tag names in Vert.x 5.0

Understood, thanks @vietj & @tsegismont!

PR #4994, opened 2023-12-01
JsonObject optimization

magicprinc:
1) use JDK Map.entry(key,value) instead of self-made Entry class 

One caveat: Map.entry disallows null keys and values (Attempts to create Map.Entry using a null key or value result in NullPointerException), but it is used in JsonObject iterator only and the iterator doesn't wrap null values in new Map.Entry

2) (Linked)HashMap has DEFAULT_LOAD_FACTOR = 0.75f: if you create it with initialCapacity=X and put X items into it, then HashMap resizes (with new internal array allocation and re-hash). One has to create HashMap with initialCapacity=X/0.75
e.g. for 10 keys initialCapacity must be at least 14 

3) better inheritance: equals 
`if (getClass() != o.getClass()) return false` rejects subclasses 
`if (!(o instanceof JsonObject)) return false` allows subclasses and checks for `null`

PR #5001, opened 2023-12-04
JsonObject optimizations

magicprinc:
1) remove two wrapper methods fromJson and fromBuffer 
2) use JDK Map.entry(key,value) instead of self-made Entry class 
3) better initialCapacity
4) better .stream() with flags and less lambdas

Replacement for https://github.com/eclipse-vertx/vert.x/pull/4994


magicprinc:
New approach: micro steps (with few lines of code)
https://github.com/eclipse-vertx/vert.x/pull/5097

PR #5033, opened 2023-12-14
Proposal for HttpClientOptions ConnectTimeout to be used in connection acquisition.

kziomek:
Motivation:

This PR proposes using HttpClientOptions.connectTimeout for connection acquisition when RequestOptions.connectTimeout is not set.

Background:

AFAIR in vert.x 3 request timeout started ticking before connection acquisition.
In vert.x 4 it has changed. Request timeout starts ticking only after connection is sent. Hence, understandably there is need for explicit connectTimeout parameter.
I can see vert.x 4.5.0 introduced new connectTimeout parameter in RequestOptions so separate from request timeout value is used for connection acquisition.
I can see the connectTimeout also exists in  HttpClientOptions but it's not used in HttpClientImpl.

Curently the only way to impose connectTimeout is to set it explicitly on RequestOptions for every request.
However, there is no way to pass connectTimeout to client.request methods that take parameters other than RequestOptions.

Caveat:
I can see ConnectTimeout exists in HttpClientOptions for a long time, at least from vert.x 3.x, and I might not be fully aware of it's purpose if its used deeper i.e. Netty code. 

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
@@ -312,6 +312,9 @@ public class HttpClientImpl extends HttpClientBase implements HttpClientInternal
     }
     long connectTimeout = 0L;
     long idleTimeout = 0L;
+    if (options.getConnectTimeout() >= 0) {
+      connectTimeout = options.getConnectTimeout();
+    }
     if (request.getTimeout() >= 0L) {
       connectTimeout = request.getTimeout();
       idleTimeout = request.getTimeout();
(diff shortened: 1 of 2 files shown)


vietj:
thanks that's an interesting proposition and I wish we could do that.

However `HttpClientOptions.connectTimeout`  is meant for TCP time to connect to the server and not related to getting a connection from a client that uses a pool. The `connect` word is overloaded and has two different meaning according to the context.

Maybe the mistake was to have HttpClientOptions extends `TCPSSLOptions` instead of having it embedded when it was designed.

Perhaps we should deprecate the connect timeout in RequestOptions and instead name it "acquireTimeout" or something like this.

kziomek:
Thanks for clarification. I'll close this PR.

"acquireTimeout" or "acquireConnectionTimeout" names sound reasonable to me.

Would  you consider adding acquire timeout to HttpClientOptions in near future or you would prefer sticking to RequestOptions for now?

PR #4991, opened 2023-12-01
Configurable server port for the vertx test using HttpTestBase.DEFAULT_HTTP_PORT

percyashu:
Motivation: [discussion](https://github.com/vert-x3/vertx-web/pull/2513)

HttpTestBase.DEFAULT_HTTP_PORT could be loaded from a system property or be 8080 when there is no such system property. So it can be used instead of hard coding port 8080 in tests.

Conformance:

You should have signed the Eclipse Contributor Agreement as explained in https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md
Please also make sure you adhere to the code style guidelines: https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines

PR #5037, opened 2023-12-18
Configurable server port for the vertx test using HttpTestBase.DEFAULT_HTTP_PORT 4.x

percyashu:
Motivation: https://github.com/eclipse-vertx/vert.x/pull/4991

HttpTestBase.DEFAULT_HTTP_PORT could be loaded from a system property or be 8080 when there is no such system property. So it can be used instead of hard coding port 8080 in tests.

Conformance:

### DIFF
--- a/src/test/java/io/vertx/core/http/impl/EndPointKeyTest.java
+++ b/src/test/java/io/vertx/core/http/impl/EndPointKeyTest.java
@@ -10,7 +10,6 @@
  */
 package io.vertx.core.http.impl;
 
-import io.vertx.core.http.HttpTestBase;
 import io.vertx.core.net.ProxyOptions;
 import io.vertx.core.net.SocketAddress;
 import org.junit.Test;
@@ -25,7 +24,7 @@ public class EndPointKeyTest {
 
   @Test
   public void testEndPointKey() {
-    final SocketAddress addr = SocketAddress.inetSocketAddress(HttpTestBase.DEFAULT_HTTP_PORT, "localhost");
+    final SocketAddress addr = SocketAddress.inetSocketAddress(8080, "localhost");
     EndpointKey key1 = new EndpointKey(false, new ProxyOptions(), addr, addr);
     EndpointKey key2 = new EndpointKey(false, new ProxyOptions(), addr, addr);
     assertEquals(key1, key2);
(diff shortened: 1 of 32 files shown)


src/test/java/io/vertx/core/net/impl/HostAndPortTest.java
```diff
@@ -1,5 +1,6 @@
 package io.vertx.core.net.impl;
 
+import io.vertx.core.http.HttpTestBase;
```
vietj, line 3: those tests don't use a server, so they should use literal values instead, that makes them less readable

src/test/java/io/vertx/core/net/SocketAddressTest.java
```diff
@@ -14,80 +15,80 @@ public class SocketAddressTest extends VertxTestBase {
 
   @Test
   public void testInetSocketAddressFromUnresolvedAddress() {
-    SocketAddress addr = SocketAddress.inetSocketAddress(InetSocketAddress.createUnresolved("localhost", 8080));
+    SocketAddress addr = SocketAddress.inetSocketAddress(InetSocketAddress.createUnresolved("localhost", HttpTestBase.DEFAULT_HTTP_PORT));
```
vietj, line 18: those tests don't use a server, so they should use literal values instead, that makes them less readable

src/test/java/io/vertx/core/http/impl/EndPointKeyTest.java
```diff
@@ -24,7 +25,7 @@ public class EndPointKeyTest {
 
   @Test
   public void testEndPointKey() {
-    final SocketAddress addr = SocketAddress.inetSocketAddress(8080, "localhost");
+    final SocketAddress addr = SocketAddress.inetSocketAddress(HttpTestBase.DEFAULT_HTTP_PORT, "localhost");
```
vietj, line 28: those tests don't use a server, so they should use literal values instead, that makes them less readable

vietj:
@percyashu some tests that do not use a server should instead use the literal values instead of constants there is no need here because we actually do not care about the port, can you revert those ? I added comments

percyashu:
> @percyashu some tests that do not use a server should instead use the literal values instead of constants there is no need here because we actually do not care about the port, can you revert those ? I added comments

Done

vietj:
thanks, I will merge that tomorrow

PR #5038, opened 2023-12-19
HTTP/2 server headers validation improvements

vietj:
The HTTP/2 server request handling first validates the request headers and then creates the request object. This leads to un-necessary work since same headers are lookup several times for the validation and the request creation.

In addition the HTTP/2 server malformed request check uses java URI to validate the request authority. We can reuse the HostAndPort parse method to validate the authority and reuse the created HostAndPort instance to avoid doing the same again when the HTTP/2 server stream is created.

We can split the request creation into creation/initialization, allowing to move the validation phase in between. The new sequence is creation of the request from the headers, pseudo headers are sanitized from the headers object instead of being done in the HttpServerRequest implementation. Then the request is validated against the its own state instead of the headers state, finally the request is initialized (e.g.g registering the vertx stream as a payload of the netty stream).

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerConnection.java
@@ -15,12 +15,8 @@ import io.netty.channel.EventLoop;
 import io.netty.handler.codec.compression.CompressionOptions;
 import io.netty.handler.codec.http.HttpContentCompressor;
 import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http2.DefaultHttp2Headers;
-import io.netty.handler.codec.http2.Http2CodecUtil;
-import io.netty.handler.codec.http2.Http2Error;
-import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.handler.codec.http2.*;
 import io.netty.handler.codec.http2.Http2Settings;
-import io.netty.handler.codec.http2.Http2Stream;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.FutureListener;
 import io.vertx.core.AsyncResult;
@@ -30,10 +26,9 @@ import io.vertx.core.Promise;
 import io.vertx.core.http.*;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.net.HostAndPort;
+import io.vertx.core.net.impl.HostAndPortImpl;
 import io.vertx.core.spi.metrics.HttpServerMetrics;
 
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.ArrayDeque;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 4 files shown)


vietj:
cc @franz1981

src/main/java/io/vertx/core/http/impl/Http2ServerConnection.java
```diff
@@ -170,15 +162,16 @@ VertxHttp2Stream<?> stream(int id) {
   protected synchronized void onHeadersRead(int streamId, Http2Headers headers, StreamPriority streamPriority, boolean endOfStream) {
     VertxHttp2Stream stream = stream(streamId);
     if (stream == null) {
-      if (isMalformedRequest(headers)) {
+      HostAndPort authority = isMalformedRequest(headers);
```
franz1981, line 165: This is still not preventing the authority to be validated again, but just making a lighter approach on validation: the ideal would be to check if the new stream matches with an existing last received valid authority and skip it.
I have verified that there are no hooks to detect if the authority is received fresh new or just cached by HPACK, so it means that we should do it ourself..

src/main/java/io/vertx/core/http/impl/Http2ServerConnection.java
```diff
@@ -86,44 +87,35 @@ public HttpServerMetrics metrics() {
     return metrics;
   }
 
-  private static boolean isMalformedRequest(Http2Headers headers) {
+  private static HostAndPort isMalformedRequest(Http2Headers headers) {
     if (headers.method() == null) {
-      return true;
+      return INVALID_REQUEST_SENTINEL;
     }
     String method = headers.method().toString();
```
franz1981, line 94: `*.method()/*.authority() ` etc etc looks like plain field access but they are not: are always lookup into a map, which is not cheap and we can just save in local fields them

vietj: can't we have an header implementation that does that ?

franz1981 (commented):
this is looking good to me!!

src/main/java/io/vertx/core/http/impl/Http2ServerStream.java
```diff
@@ -60,26 +62,35 @@ class Http2ServerStream extends VertxHttp2Stream<Http2ServerConnection> {
...
 
-  Http2ServerStream(Http2ServerConnection conn, ContextInternal context, Http2Headers headers, String serverOrigin, TracingPolicy tracingPolicy, boolean halfClosedRemote) {
+  Http2ServerStream(Http2ServerConnection conn, ContextInternal context, Http2Headers headers, TracingPolicy tracingPolicy, boolean halfClosedRemote) {
     super(conn, context);
 
-    String host = headers.get(":authority") != null ? headers.get(":authority").toString() : null;
-    if (host == null) {
-      int idx = serverOrigin.indexOf("://");
-      host = serverOrigin.substring(idx + 3);
+    CharSequence schemeHeader = headers.getAndRemove(":scheme");
```
franz1981, line 76: I know this commit was made a PoC to quantify the improvement, but I'm adding this notes to not forget...
All of these can be made faster by using the proper `AsciiString` values' on the Netty PseudoHeader's enums: differently, by using `String` literals, both the hash code computation and equals is going to be performed char-by-char and never cached.

franz1981 (changes requested):

src/main/java/io/vertx/core/http/impl/Http2ServerStream.java
```diff
@@ -60,26 +62,35 @@ class Http2ServerStream extends VertxHttp2Stream<Http2ServerConnection> {
...
+    HostAndPort authority = null;
+    String authorityString = null;
+    CharSequence authorityHeader = headers.getAndRemove(":authority");
+    if (authorityHeader != null) {
+      authorityString = authorityHeader.toString();
+      authority = HostAndPortImpl.parseHostAndPort(authorityString, -1);
     }
+    CharSequence pathHeader = headers.getAndRemove(":path");
```
franz1981, line 84: As said at https://github.com/eclipse-vertx/vert.x/pull/5038#discussion_r1431325625 for the others pseudoheaders

PR #5041, opened 2023-12-19
HTTP/2 pseudo header lookup optimisations

vietj:
Define in HttpHeaders constants for HTTP/2 pseudo headers and use them instead of mere strings to optimise headers lookup


vietj:
thanks @franz1981 for pointing this out

franz1981 (approved):
well done!

PR #5048, opened 2023-12-21
Context local data SPI

vietj:
A proper SPI for interacting with local context data.

The existing implementation relies on a concurrent hashmap guarded by the context instance.

The new implementation relies on prior knowledge of the set of context keys, so the context allocates upfront the required space to hold the local context data, reducing synchronisation and simplifying lookup that uses allocated index instead of hashed string lookups.


src/main/java/io/vertx/core/impl/ContextBase.java
```diff
@@ -0,0 +1,51 @@
...
+    return (T) res;
+  }
+
+  public final <T> void putLocal(ContextKey<T> key, T value) {
+    ContextKeyImpl<T> internalKey = (ContextKeyImpl<T>) key;
+    int index = internalKey.index;
+    if (index >= locals.length) {
+      throw new IllegalArgumentException();
+    }
+    LOCALS_UPDATER.setVolatile(locals, index, value);
```
franz1981, line 49: This can uses setRelease which is way cheaper

vietj:
cc @tsegismont

vietj:
cc @cescoffier

src/main/java/io/vertx/core/impl/ContextBase.java
```diff
@@ -0,0 +1,76 @@
...
+    while (true) {
+      res = LOCALS_UPDATER.getVolatile(locals, index);
+      if (res != null) {
+        break;
+      }
+      Object initial = supplier.get();
+      if (initial == null) {
+        throw new IllegalStateException();
+      }
+      if (LOCALS_UPDATER.compareAndSet(locals, index, null, initial)) {
```
franz1981, line 60: We don't need atomicity here, but the same semantic of putLocal, which doesn't care of races, imo.
If we need that level of atomicity we can have a new method which ensure an atomic change and use the compare and set.

src/main/java/io/vertx/core/Context.java
```diff
@@ -244,6 +246,42 @@ default List<String> processArgs() {
...
+
+  /**
+   * Get some local data from the context.
+   *
+   * @param key  the key of the data
+   * @param <T>  the type of the data
+   * @return the data
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  <T> T getLocal(ContextKey<T> key, Supplier<T> supplier);
```
tsegismont, line 267: The javadoc for the supplier is missing

src/main/java/io/vertx/core/Context.java
```diff
@@ -244,6 +246,42 @@ default List<String> processArgs() {
...
+   * This can be used to share data between different handlers that share a context
+   *
+   * @param key  the key of the data
+   * @param value  the data
+   */
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  <T> void putLocal(ContextKey<T> key, T value);
+
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
+  default <T> void removeLocal(ContextKey<T> key) {
```
tsegismont, line 281: Missing Javadoc

src/test/java/io/vertx/test/faketracer/FakeTracer.java
```diff
@@ -30,8 +32,7 @@
  */
 public class FakeTracer implements VertxTracer<Span, Span> {
 
-  private static final String ACTIVE_SCOPE_KEY = "active.scope";
-
+  private final ContextKey<Object> scopeKey = ContextKey.registerKey(Object.class);
```
tsegismont, line 35:
```suggestion
  private final ContextKey<Scope> scopeKey = ContextKey.registerKey(Scope.class);
```

src/test/java/io/vertx/test/faketracer/FakeTracer.java
```diff
@@ -53,7 +54,7 @@ public Span activeSpan() {
   }
 
   public Span activeSpan(Context data) {
-    Scope scope = data.getLocal(ACTIVE_SCOPE_KEY);
+    Scope scope = (Scope) data.getLocal(scopeKey);
```
tsegismont, line 57:
```suggestion
    Scope scope = data.getLocal(scopeKey);
```

src/test/java/io/vertx/test/faketracer/FakeTracer.java
```diff
@@ -62,9 +63,9 @@ public Scope activate(Span span) {
   }
 
   public Scope activate(Context context, Span span) {
-    Scope toRestore = context.getLocal(ACTIVE_SCOPE_KEY);
+    Scope toRestore = (Scope) context.getLocal(scopeKey);
```
tsegismont, line 66:
```suggestion
    Scope toRestore = context.getLocal(scopeKey);
```

src/test/java/io/vertx/core/spi/tracing/LocalEventBusTracerTest.java
```diff
@@ -32,15 +32,15 @@ public void testInboundInterceptor() throws Exception {
     tracer = new VertxTracer() {};
     vertx2.eventBus().addInboundInterceptor(deliveryCtx -> {
       ContextInternal ctx = (ContextInternal) vertx.getOrCreateContext();
-      ctx.localContextData().put("key", "val");
+//      ctx.putLocal(FakeTracer.ACTIVE_SCOPE_KEY, "val");
```
tsegismont, line 35: What's wrong here?

src/test/java/io/vertx/core/spi/tracing/LocalEventBusTracerTest.java
```diff
@@ -32,15 +32,15 @@ public void testInboundInterceptor() throws Exception {
...
       deliveryCtx.next();
     });
     Context receiveCtx = vertx2.getOrCreateContext();
     CountDownLatch latch = new CountDownLatch(1);
     receiveCtx.runOnContext(v -> {
       vertx2.eventBus().consumer("the_address", msg -> {
-        Object val = ((ContextInternal) vertx.getOrCreateContext()).localContextData().get("key");
-        assertEquals("val", val);
+//        Object val = vertx.getOrCreateContext().getLocal(FakeTracer.ACTIVE_SCOPE_KEY);
```
tsegismont, line 42: Same question

src/test/java/io/vertx/core/spi/tracing/HttpTracerTestBase.java
```diff
@@ -218,7 +225,7 @@ public Object sendRequest(Context context, SpanKind kind, TracingPolicy policy,
       @Override
       public void receiveResponse(Context context, Object response, Object payload, Throwable failure, TagExtractor tagExtractor) {
         assertSame(val, context.getLocal(key));
-        assertTrue(context.removeLocal(key));
+        context.putLocal(key, null);
```
tsegismont, line 228: For convenience, how about adding a `removeLocal(key)` method?

src/test/java/io/vertx/core/impl/ContextKeyHelper.java
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.impl;
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+public class ContextKeyHelper {
+
+  public static void reset() {
```
tsegismont, line 18: Maybe comment this is needed to keep the `reset` method package-protected in `KeySeq`?

src/main/java/io/vertx/core/impl/ContextBase.java
```diff
@@ -0,0 +1,76 @@
...
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+import java.util.function.Supplier;
+
+/**
+ * Base class for context.
+ *
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+class ContextBase {
```
tsegismont, line 24: Out of curiosity, why is it necessary to introduce a new base class?

src/main/java/io/vertx/core/impl/ContextBase.java
tsegismont: If I understand correctly, the goal of this method is to initialize the value if not present.

In this case:

- we can't do it with the same semantics as `putLocal`, can we?
- how about changing `compareAndSet` with `compareAndExchange` and return the value directly? This would avoid having to invoke `getVolatile` again

tsegismont (commented):
Good job @vietj !

vietj:
See that rather as a compute operation.

src/test/java/io/vertx/core/spi/tracing/HttpTracerTestBase.java
vietj: I think I added it but I haven't used it

src/main/java/io/vertx/core/impl/ContextBase.java
vietj: to hold the common state and not repeat the code in `DuplicatedContext`/`ContextImpl`

src/test/java/io/vertx/core/spi/tracing/LocalEventBusTracerTest.java
vietj: test is actually ignored :-) so previous code will not work anymore when the deprecated code will be removed, I sketched what should be correct more or less

src/test/java/io/vertx/core/spi/tracing/LocalEventBusTracerTest.java
vietj: test is actually ignored :-) so previous code will not work anymore when the deprecated code will be removed, I sketched what should be correct more or less

src/main/java/io/vertx/core/impl/ContextBase.java
vietj: I think I just applied a common pattern here that initialises the state when it is null otherwise discard the value obtained from the supplier if it was set by another thread concurrently, e.g in reactiverse contextual logging, this will be initialised with a concurrent hash map:

```java
Map<String, Object> map = getLocal(key, () -> new ConcurrentHashMap<>());
```

It is similar to `Map#computeIfAbsent(K, Function<K, V>)`

vietj: @tsegismont compareAndExchange only exist since Java 9, I do intent to back port this to 4.x branch for deprecation purpose (and actually the existing concurrent map would be retrofit as a context key)

cescoffier:
In Quarkus we don't know the key that may be used. We can imagine knowing a few of them at build time, but we will have an unknown set at runtime.

I would do something like:

- at build time, request extension to give the keys
- at runtime, allocate with this set of keys +1 
- the +1 is a concurrent hash map for all keys that have not being declared.

WDYT?

franz1981:
It is a nice approach @cescoffier which will give performance and flexibility based on the usage, I like it!

vietj:
@cescoffier note you can have your extension that uses a map (or whatever structure you need) as a slot with a unique key, giving you opportunity to proceed as you want, which is what reactiverse context logging will do, e.g. it could be a `ContextKey<Map>` and the `Map` would be managed by Quarkus.

src/main/java/io/vertx/core/Context.java
```diff
@@ -244,6 +246,47 @@ default List<String> processArgs() {
...
+
+  /**
+   * Get some local data from the context.
+   *
+   * @param key  the key of the data
+   * @param <T>  the type of the data
+   * @return the data
+   */
+  @GenIgnore
+  <T> T getLocal(ContextKey<T> key, Supplier<? extends T> supplier);
```
Ladicek, line 267: Documentation for `supplier` missing :-) Also it would be a good idea to mention that `supplier` may be called multiple times, which is sometimes important to know.

vietj: yes, it is actually pretty much like `Map#compute(Function)` that says _When multiple threads attempt updates, map operations and the remapping function may be called multiple times._

src/main/java/io/vertx/core/impl/KeySeq.java
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.impl;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+class KeySeq {
```
franz1981, line 18: I think that this one should be an assigned field to a context, because not all contexes requires the same "key space", or we risk to have every context to always allocate N keys even if they requires very different number of keys.

The other option, instead is to make things smarter working in chunks (and lazy):
- no `KeySeq`, but a chunked approach with N (eg 8) elements per chunk saved in a CopyOnWriteArrayList<KeyChunk> keys;
- a lookup in position N will hash N to find the proper bucket in the key space int chunkIndex = N / CHUNKS_CAPACITY and keyOffset = N % CHUNKS_CAPACITY
- each chunk is allocated lazily based on which index has been accessed

franz1981:
I have some mixed feeling after re-reading what @cescoffier rightly said, and remind me about something I've worked recently eg https://github.com/quarkusio/quarkus/pull/38761 which has a very similar problem ie at build time it can collect what are the "supposely" used keys (in the PR I've referenced, were bean ids, but is the same really...), but it:
- will risk to make every local/non-local data capacity the space to hold *all* of them while at runtime maybe only a subset is used
- where uncknown ones risk to go?

In shorts, constant lookup access and simple storage is trade against the cost of overestimating capacity, which is not good here. In https://github.com/quarkusio/quarkus/pull/38761 i've tried to keep the allocated cost at it(s) minimum by having lazily allocated data, and the performance advantage of this approach could be re-used for this.
I suggest to add some microbenchmark to compare about this approach as it is, but using different key used/total key generated ratios vs a "normal" concurrent hash map, because, for small utilization ratios (<= 1/10) the cost of allocating a full blown array vs using a CHM could be not that different.

Issue #5049, opened 2023-12-22
Pipe's completion handler possibly executed by the "wrong thread", leading to race conditions


sauthieg:
### Version

Vert.x 4.4.6, but likely earlier versions have the issue as well.

### Context

It started with a suspicious NPE in our home-made Vert.x metrics plugin:

```java
java.lang.NullPointerException: null
    at org.forgerock.http.vertx.monitoring.VertxHttpClientMetrics$HttpClientMetricsImpl$1.requestEnd(VertxHttpClientMetrics.java:120)
    at org.forgerock.http.vertx.monitoring.VertxHttpClientMetrics$HttpClientMetricsImpl$1.requestEnd(VertxHttpClientMetrics.java:94)
    at io.vertx.core.http.impl.Http1xClientConnection.endRequest(Http1xClientConnection.java:320)
    at io.vertx.core.http.impl.Http1xClientConnection.writeBuffer(Http1xClientConnection.java:307)
    at io.vertx.core.http.impl.Http1xClientConnection.access$1100(Http1xClientConnection.java:108)
    at io.vertx.core.http.impl.Http1xClientConnection$StreamImpl.writeBuffer(Http1xClientConnection.java:603)
    at io.vertx.core.http.impl.Http1xClientConnection$StreamImpl.writeBuffer(Http1xClientConnection.java:566)
    at io.vertx.core.http.impl.HttpClientRequestImpl.doWrite(HttpClientRequestImpl.java:506)
    at io.vertx.core.http.impl.HttpClientRequestImpl.write(HttpClientRequestImpl.java:471)
    at io.vertx.core.http.impl.HttpClientRequestImpl.end(HttpClientRequestImpl.java:417)
    at io.vertx.core.streams.impl.PipeImpl.handleSuccess(PipeImpl.java:116)
    at io.vertx.core.streams.impl.PipeImpl.lambda$to$2(PipeImpl.java:102)
    at io.vertx.core.impl.future.FutureImpl$3.onSuccess(FutureImpl.java:141)
    at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:60)
    at io.vertx.core.impl.future.FutureImpl.addListener(FutureImpl.java:196)
    at io.vertx.core.impl.future.PromiseImpl.addListener(PromiseImpl.java:23)
    at io.vertx.core.impl.future.FutureImpl.onComplete(FutureImpl.java:164)
[...]

PR #5050, opened 2023-12-22
5049 fix pipe race condition

NilsRenaud:
Fixes the issue https://github.com/eclipse-vertx/vert.x/issues/5049


NilsRenaud:
There is still a broken test, because a test explicitly calls `testStream.write()` after `testStream.fail()` I thought it wasn't possible...
Is this an error or are stream failures expected to happen multiple times on the same streams, even between 2 elements ?

NilsRenaud:
@vietj what do you think of my previous comment, is `RecordParserTest.testWrapReadStream`buggy ?

PR #5056, opened 2024-01-09
Core system properties cleanup

vietj:
- [x] introduce an enum representing core system properties that can be accounted for
- [x] `vertx.json.base64` removal
- [x] remove cluster manager system property
- [x] remove properties having a corresponding option which are not documented
  - [x] `vertx.javaCompilerOptions`
  - [x] `vertx.disableTCCL`
- [x] deprecate system properties  having a corresponding option which are documented
  - [x] discuss with @tsegismont the handling of file system properties as explained in the documentation
- [x] consider reifying system properties as env variable like `vertxweb.environment`
- [x] remaining properties analysis

### DIFF
--- a/src/main/java/io/vertx/core/Context.java
+++ b/src/main/java/io/vertx/core/Context.java
@@ -167,7 +167,7 @@ public interface Context {
    * Is the current context an event loop context?
    * <p>
    * NOTE! when running blocking code using {@link io.vertx.core.Vertx#executeBlocking(Callable)} from a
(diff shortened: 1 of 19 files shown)


vietj:
@tsegismont do you mind having a look at the system properties we do have ? I'm thinking at the cluster manager declaration at least that could go away in 5.0

tsegismont (approved):
It's fine to remove the cluster manager related sysprops, I think they only made sense when several cluster manager implementations where available on the CP (e.g. CLI with full profile, but the CLI is going away)

PR #5057, opened 2024-01-09
More robust Json types deserialization

tsegismont:
Possible fix for #5058 

Users can register a Jackson Databind module that modifies container types, such as the Scala module. In this case, Vert.x wasn't able to deserialize properly a `JsonObject`, `JsonArray`, or a POJO with fields having one these types. The reason is that Vert.x expected instances of `Map` and `List` before wrapping to `JsonObject` and `JsonArray`, respectively.

In this change, we introduce two methods on the `JsonCodec` type, which allow to decode a JSON string or buffer without prior knowledge of the expected Java type. For backwards compatibility, the default implementation delegates to the existing method, using the `Object` type. In the `DatabindCodec` implementation, these new methods delegate parsing to the `JacksonCodec`. Consequently, we can be sure no Jackson Databind module can interfere with the deserialization of `JsonObject` and `JsonArray`.

Similarly, the `JsonArrayDeserializer` and `JsonObjectDeserializer` now delegate parsing to the `JacksonCodec`. This is required to properly deserialize Vert.x Json objects embedded in a POJO.

Side fix: the `DatabindCodec.createParser` method takes a `Buffer` parameter again, instead of a `BufferInternal.` The method was no longer used. Instead, the `createParser` method of the `JacksonCodec` parent class was used.

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -53,6 +53,9 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
       throw new NullPointerException();
     }
     fromJson(json);
+    if (list == null) {
+      throw new DecodeException("Invalid JSON array: " + json);
+    }
   }
 
   /**
@@ -84,6 +87,9 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
       throw new NullPointerException();
     }
     fromBuffer(buf);
+    if (list == null) {
+      throw new DecodeException("Invalid JSON array: " + buf);
+    }
   }
 
   /**
@@ -702,20 +708,14 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
     return pos + length + 4;
   }
 
+  @SuppressWarnings("unchecked")
   private void fromJson(String json) {
-    JsonArray decoded = (JsonArray) Json.CODEC.fromString(json);
-    if (decoded == null) {
-      throw new DecodeException("Invalid JSON array: " + json);
-    }
-    list = decoded.list;
+    list = (List<Object>) Json.CODEC.fromString(json, false);
   }
 
+  @SuppressWarnings("unchecked")
   private void fromBuffer(Buffer buf) {
-    JsonArray decoded = (JsonArray) Json.CODEC.fromBuffer(buf);
-    if (decoded == null) {
-      throw new DecodeException("Invalid JSON array: " + buf);
-    }
-    list = decoded.list;
+    list = (List<Object>) Json.CODEC.fromBuffer(buf, false);
   }
 
   private static class Iter implements Iterator<Object> {
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2024 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -53,9 +53,6 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
       throw new NullPointerException();
     }
     fromJson(json);
-    if (list == null) {
-      throw new DecodeException("Invalid JSON array: " + json);
-    }
   }
 
   /**
@@ -87,9 +84,6 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
       throw new NullPointerException();
     }
     fromBuffer(buf);
-    if (list == null) {
-      throw new DecodeException("Invalid JSON array: " + buf);
-    }
   }
 
   /**
@@ -709,11 +703,19 @@ public class JsonArray implements Iterable<Object>, ClusterSerializable, Shareab
   }
 
   private void fromJson(String json) {
-    list = Json.CODEC.fromString(json, List.class);
+    JsonArray decoded = (JsonArray) Json.CODEC.fromString(json);
+    if (decoded == null) {
+      throw new DecodeException("Invalid JSON array: " + json);
+    }
+    list = decoded.list;
   }
 
   private void fromBuffer(Buffer buf) {
-    list = Json.CODEC.fromBuffer(buf, List.class);
+    JsonArray decoded = (JsonArray) Json.CODEC.fromBuffer(buf);
+    if (decoded == null) {
+      throw new DecodeException("Invalid JSON array: " + buf);
+    }
+    list = decoded.list;
   }
 
   private static class Iter implements Iterator<Object> {
--- a/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/src/main/java/io/vertx/core/json/JsonObject.java
@@ -49,6 +49,9 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
       throw new NullPointerException();
     }
     fromJson(json);
+    if (map == null) {
+      throw new DecodeException("Invalid JSON object: " + json);
+    }
   }
 
   /**
@@ -80,6 +83,9 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
       throw new NullPointerException();
     }
     fromBuffer(buf);
+    if (map == null) {
+      throw new DecodeException("Invalid JSON object: " + buf);
+    }
   }
 
   /**
@@ -1240,20 +1246,14 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
     return pos + length + 4;
   }
 
+  @SuppressWarnings("unchecked")
   private void fromJson(String json) {
-    JsonObject decoded = (JsonObject) Json.CODEC.fromString(json);
-    if (decoded == null) {
-      throw new DecodeException("Invalid JSON object: " + json);
-    }
-    map = decoded.map;
+    map = (Map<String, Object>) Json.CODEC.fromString(json, false);
   }
 
+  @SuppressWarnings("unchecked")
   private void fromBuffer(Buffer buf) {
-    JsonObject decoded = (JsonObject) Json.CODEC.fromBuffer(buf);
-    if (decoded == null) {
-      throw new DecodeException("Invalid JSON object: " + buf);
-    }
-    map = decoded.map;
+    map = (Map<String, Object>) Json.CODEC.fromBuffer(buf, false);
   }
 
   private static class Iter implements Iterator<Map.Entry<String, Object>> {
--- a/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/src/main/java/io/vertx/core/json/JsonObject.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2022 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2024 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
@@ -49,9 +49,6 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
       throw new NullPointerException();
     }
     fromJson(json);
-    if (map == null) {
-      throw new DecodeException("Invalid JSON object: " + json);
-    }
   }
 
   /**
@@ -83,9 +80,6 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
       throw new NullPointerException();
     }
     fromBuffer(buf);
-    if (map == null) {
-      throw new DecodeException("Invalid JSON object: " + buf);
-    }
   }
 
   /**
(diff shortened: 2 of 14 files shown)


tsegismont:
Windows failure is unrelated

vietj:
it is not clear how the issue happens

are we talking here about the specific Vert.x databind codec mapper of which the mapping of Map is changed to another type that does not extend Map ?

vietj:
the intent seems good but I am afraid this is not something we want to support, we expect the vertx mapper to map a map to a java.util.Map, if something else is required, then we need to think about another solution

src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -1247,11 +1241,19 @@ public int readFromBuffer(int pos, Buffer buffer) {
   }
 
   private void fromJson(String json) {
-    map = Json.CODEC.fromString(json, Map.class);
+    JsonObject decoded = (JsonObject) Json.CODEC.fromString(json);
```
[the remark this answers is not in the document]
tsegismont: Yes, that's a drawback. It's more efficient to create a `JsonObject` from a string with `Json.decodeValue(str)`. Or, from a buffer, with `Buffer.toJsonObject`.

Perhaps we can add a `fromString(String str, boolean wrapContainers)` method to the `JsonCodec` interface.

In this case, the wrapping `JsonObject` or `JsonArray` wouldn't be created

tsegismont:
> it is not clear how the issue happens
> 
> are we talking here about the specific Vert.x databind codec mapper of which the mapping of Map is changed to another type that does not extend Map ?

This happens when you configure the mapper with type modifiers that convert List/Map to other types. The modifiers are not invoked in all cases: if you ask the object mapper to deserialize something as a List or a Map, it returns an object with the corresponding type. But if the json payload contains nested objects or arrays, they could be deserialized as something else than a Map or a List.

Also, if you deserialize something using the Objec type hint, the type modifiers will be used.

vietj:
I would like to understand clearly whether the use case advocated here is to configure the mapper used by Vertx to deserialize list/map collections to another type than java.util.List and java.util.Map or if I'm incorrectly interpreting the case.

trym-moeller:
@vietj , 

I am not sure I should participate in this discussion. But I was having a problem (which is described in detail in the referenced #5058 issue), using vert.x in a Scala setup. I think that the suggested fix, is an enhancement of vert.x, so it supports more use cases. Do you see the fix introducing any limitations to the vert.x core or what are your concerns?

Thanks.

Best regards Trym

vietj:
@trym-moeller for now I would like to understand whether the List/Map mapping are changed and if it does, for which reason

tsegismont:
The Jackson [module for Scala](https://github.com/FasterXML/jackson-module-scala) adds support for Scala types, like `Sequence` and `Map`, which are not subtypes of Java `List` and `Map`.

By default, the Scala module adds an `UntypedObjectDeserializerResolver` which selects a target type for the json value to deserialize when it is known.

For example, if you ask Jackson to deserialize a Json object to a Java Map, it will deserialize the content to a Java Map. But if you let it choose (by using the `Object.class` parameter), then the Scala module's `UntypedObjectDeserializerResolver` is invoked and selects a deserializer to Scala Map, which is incompatible with our `JsonObject`.

This module doesn't make some strange mappings, it makes mappings that are idiomatic to the Scala language.

The proposed change makes it possible to use Vert.x with Jackson Databind in such an environment

vietj:
I still don't understand why this module is registered against the vertx mapper, what is happening here ? are we doing some kind of discovery for that ?

tsegismont:
In the case reported, it is the Quarkus Scala extension that registers the Jackson Scala module.

This is mandatory for users to map, for example, web client response bodies to their classes.

src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -709,11 +703,19 @@ public int readFromBuffer(int pos, Buffer buffer) {
   }
 
   private void fromJson(String json) {
-    list = Json.CODEC.fromString(json, List.class);
+    JsonArray decoded = (JsonArray) Json.CODEC.fromString(json);
```
vietj, line 706: why can't we specify here the `List` type anymore ? it seems fine to ask for a `List`

src/main/java/io/vertx/core/spi/json/JsonCodec.java
```diff
@@ -14,12 +14,25 @@
...
 public interface JsonCodec {
 
+  /**
+   * Decode a given JSON string.
+   *
+   * @param json the JSON string.
+   * @return a JSON element which can be a {@link JsonArray}, {@link JsonObject}, {@link String}, ...etc. if the string contains an array, object, string, ...etc
+   * @throws DecodeException when there is a parsing or invalid mapping.
+   */
+  default Object fromString(String json) throws DecodeException {
```
vietj, line 32: is it a required change to the PR or is it only an accommodation change ? I am not sure of this

tsegismont: It is a required change. This new method complies to the contract of `io.vertx.core.json.Json#decodeValue(java.lang.String)`, which is returning a `JsonArray`, `JsonObject`, `String, ...etc. if the content is an array, object, string, ...etc. in all cases.

The existing methods of `DatabindCodec` do not comply to the contract if, for example, the `ObjectMapper` is configured with the Jackson Scala module.

src/main/java/io/vertx/core/json/JsonArray.java
tsegismont: If the object mapper is configured with the Jackson Scala module, and you run:

```java
var json = """
[1,2,{"foo":"bar}]
""";
JsonArray arr = Json.CODEC.fromString(json, List.class);
JsonObject obj = arr.getJsonObject(2); // boom
```

You will get a `ClassCastException` because the object nested in the array will be deserialized as a Scala Map, not a Java Map.

vietj: so you are saying that the following code in `DatabindCodec`:

```java
      value = DatabindCodec.mapper.readValue(parser, type);
```

called with type = `Map.class` will throw this CCE ?

src/main/java/io/vertx/core/spi/json/JsonCodec.java
vietj: why isn't `io.vertx.core.json.Json#decodeValue(java.lang.String)` calling `fromString(Object.class)` incorrect ?

src/main/java/io/vertx/core/json/JsonArray.java
tsegismont: Assuming the object mapper has been configured with the Jackson Scala module, if you have a `parser` for a valid json object string or buffer, and invoke:

```
      value = DatabindCodec.mapper.readValue(parser, type);
```

with the `type` of `Map`, then the `value` will be an instance of Java `Map`, because, Jackson knows the target type for the top-level object. But if you have nested objects or arrays, they will deserialized to Scala Map or Sequence, because Jackson doesn't know which type they should be deserialized to.

src/main/java/io/vertx/core/spi/json/JsonCodec.java
tsegismont: Because, with the DatabindCodec, if the object mapper has been configured with the Jackson Scala Module, the type resolved for a json string or json array is not a Java Map or List, but a Scala Map or Sequence.

src/main/java/io/vertx/core/json/JsonArray.java
vietj: have you checked we cannot override the default collection type to map when vertx takes control of the mapper ?

vietj:
I looked at the Scala / Quarkus project and did not find evidence of this module modifying the Vert.x mapper, @tsegismont do you know where this happens ?

src/main/java/io/vertx/core/spi/json/JsonCodec.java
vietj: it seems that it only then works because of this 

```java
  @Override
  public Object fromString(String str) throws DecodeException {
    return JacksonCodec.fromParser(createParser(str), Object.class);
  }
```

which makes this `fromString` behavior very much different than the other `fromString` overloads in the `DatabindCodec` class

tsegismont:
> I looked at the Scala / Quarkus project and did not find evidence of this module modifying the Vert.x mapper,

There is the [Quarkus databind codec](https://github.com/quarkusio/quarkus/blob/main/extensions/vertx/runtime/src/main/java/io/quarkus/vertx/runtime/jackson/QuarkusJacksonJsonCodec.java) and Quarkus users can [customize](https://quarkus.io/guides/rest-json#jackson) the `ObjectMapper` with a `io.quarkus.jackson.ObjectMapperCustomizer`.

And then the `ScalaProcessor` in the Quarkus Scala extension [registers](https://github.com/quarkusio/quarkus/blob/781060d174ecbc1bfa462e8ccc4b6726bd33f1a3/extensions/scala/deployment/src/main/java/io/quarkus/scala/deployment/ScalaProcessor.java#L30) the `DefaultScalaModule` if the library is on the classpath.

src/main/java/io/vertx/core/spi/json/JsonCodec.java
tsegismont: Right, the behavior it's different, it's the only one which has the contract of returning a valid Vert.x JSON object: `String`, `Boolean`, `JsonObject`, `JsonArray`.

Users are not supposed to work directly with codec objects, they are supposed to work with the `Json` class. This one has two different methods (and more): `decodeValue(String)` and `decodeValue(String, Class<T>)`. The first one has the contract of returning a valid Vert.x JSON object, the other just says it will try to deserialize something with the type of parameter. It works as we expect with the `JacksonCodec` but, with the `DatabindCodec`, it may not work if the object mapper is configured in a way we hadn't anticipated.

vietj:
There are (at least) 2 different cases to address

1. Direct interactions with the JsonObject to decode a JSON string/buffer, e.g. `new JsonObject("{...}")`
2. Indirect interactions with the JsonObject to decode a JSON string/buffer, e.g. `Json.decodeValue(json, Pojo.class)` where `Pojo` uses `JsonObject`

It seems that using the JacksonCodec (as this PR does) is a reliable solution, but this solution should be confined within the private usage of `JsonObject` / `JsonArray`.

For the case 1. we should use the JacksonCodec and not `io.vertx.core.json.Json#CODEC`, this should not interfere with databding

For the case 2. this uses a Vert.x deserializer, e.g. `JsonObjectDeserializer`, this deserializer currently does

```java
  public JsonObject deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
    return new JsonObject(p.<Map<String, Object>>readValueAs(TYPE_REF));
  }
```

Instead of delegating to the parser to parse it to a `Map` (since this cannot be trusted due to possible interferences), it should reuse the `JacksonCodec` and pass it the parser the Jackson is providing:

```java
  public JsonObject deserialize(JsonParser p, DeserializationContext ctxt) {
    Object res;
    try {
      res = JacksonCodec.parseAny(p);
    } catch (IOException e) {
      throw new DecodeException(e.getMessage(), e);
    }
    return (JsonObject) JacksonCodec.cast(res, Object.class);
  }
```

src/main/java/io/vertx/core/json/JsonArray.java
tsegismont: Yes, and to my knowledge it's not possible.

vietj:
> 
> We can't invoke the Jackson Codec directly from `JsonObject` or `JsonArray` because even the Jackson core library could be excluded by the user and replaced by something else (e.g. gSON).

have we evidence anyone has one this ? this _feature_ was done for Quarkus which never used it. I am willing to believe that we can do it. If we do have evidence then there are ways to do that differently

> I'm not sure what you mean. This PR already changes the JsonObjectDeserializer and JsonArrayDeserializer to invoke the Jackson codec.

yeah that's right

vietj:
It appears we can add a new `internalCodec` to `JsonFactory` that returns a `JsonCodec` that will be `JacksonCodec` for which mapping a `java.util.Map`/`java.util.List` is honoured (recursively) and such codec remains privately used by Vert.x JSON classes (and the deserializers).

tsegismont:
If we make Jackson core mandatory again, what's the difference between this new `internalCodec` and the existing `JacksonCodec`?

tsegismont:
> also this forces to instantiate `JsonObject` twice when using `new JsonObject(str)`

I commented previously it should be possible to improve this.

vietj:
the internal codec gives the opportunity to the provide an
implementation with the JsonFacotry

tsegismont:
@vietj I got another idea for improvement: I've changed the signature of the new Json.fromString and Json.fromBuffer methods, adding a `boolean wrap` parameter that lets the caller decide if a `Map`/`List` should be converted to `JsonObject`/`JsonArray`.

This change allows to avoid creating a redundant `JsonObject` or `JsonArray` when invoking:

- `new JsonObject(String)` or `new JsonObject(Buffer)`
- `new JsonArray(String)` or `new JsonArray(Buffer)`

I think this was a major and legitimate concern.

The change comes at the expense of a behavior change of `JacksonCodec#fromParser` and `DatabindCodec#fromParser`,
that will no longer wrap `Map`/`List` to `JsonObject`/`JsonArray` automatically when invoked with the `Object` type parameter.

While this is technically a breaking change, it seems acceptable as both methods are meant for internal use.

tsegismont:
@vietj I've rebase the PR after the latest changes by @mariofusco

tsegismont:
Closing due to lack of consensus on the proposal. Possibly resolved in Vert.x 5 with a refactoring of codec SPI

vietj:
Actually I am keen having the codec SPI change in 4.x

PR #5059, opened 2024-01-09
Remove DatabindCodec pretty mapper

tsegismont:
This change relieves users from configuring two object mappers if they have customizations.

For Vert.x, it is not less efficient to use a writer with pretty printer when encoding json.


tsegismont:
@vietj this is a proposal for Vert.x 5.  If we agree on this, I will deprecate the pretty mapper in `4.x`.

vietj:
what is the rationale behind that ? I might have missed it

tsegismont:
The rationale is simplification. There is no need to keep two object mappers if a single one can be used to encode json in compact and pretty form.

vietj:
that makes sense

vietj (approved):

PR #5060, opened 2024-01-10
Deprecate DatabindCodec pretty mapper

tsegismont:
There is no need to keep two object mappers if a single one can be used to encode json in compact and pretty form. This change relieves users from configuring two object mappers if they have customizations.

### DIFF
--- a/src/main/java/io/vertx/core/json/jackson/DatabindCodec.java
+++ b/src/main/java/io/vertx/core/json/jackson/DatabindCodec.java
@@ -16,7 +16,6 @@ import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
-import com.fasterxml.jackson.databind.module.SimpleModule;
 import io.netty.buffer.ByteBufInputStream;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.DecodeException;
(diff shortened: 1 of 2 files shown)


mikand13:
Hey @tsegismont 

We are seeing a regression on this, or maybe misunderstanding in implications.

With this change I would expect we wouldnt need to configure prettyMapper anymore, but that does not seem to be the case. With going from 4.5.1 to 4.5.2 Json.encodePrettily now does not use the configuration of mapper. We still need to do the following:

```
    DatabindCodec.mapper()
        .registerKotlinModule()
        .registerModule(JavaTimeModule())
        .configure(WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS, false)
    @Suppress("DEPRECATION")
    DatabindCodec.prettyMapper()
        .registerKotlinModule()
        .registerModule(JavaTimeModule())
        .configure(WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS, false)
```

tsegismont:
Hi @mikand13 

I wanted to avoid breaking applications which rely on a specific configuration of the pretty printer. By your comment, I understand the deprecation signal is confusing since users can't take any action to prevent breakage when upgrading to 5.0.

So I've sent PR #5094 to get rid of the pretty mapper internally. If you could test it it would be awesome. Thanks

mikand13:
Can do @tsegismont , do you have a snapshot version for it on osshr or the like?

tsegismont:
Unfortunately, you have to build my branch to get this snapshot

mikand13:
No problem, can do.

PR #5063, opened 2024-01-12
Apply the queueing pattern to the initial message written by the HTTP client connection to preserve ordering of subsequent messages

vietj:
The HTTP client connection handle buffer writes from a non vertx thread and will re-schedule any message sent by the event-loop thread to preserve ordering. However this is not applied to the initial message is sent and therefore it is possible to have the initial message written after the subsequent HTTP messages.


vietj:
cc @NilsRenaud @sauthieg

PR #5070, opened 2024-01-15
Implementation of an HTTP client connection API.

vietj:
Introduce an HTTP client connection API.

The client can create an un-pooled client connection (`HttpClientConnection`) to a server, such connection is fully managed by the user. This provides full control over client connections which can matters in some cases (e.g. writing load tests).

The `HttpClient` interface now only exposes an HTTP interaction API, other methods are moved to a new interface `HttpClientAgent` which takes responsibility for managing HTTP connection pools as well as queuing requests until connections are available.

Both `HttpClientAgent` and `HttpClientConnection` extends `HttpClient` interface. It allows clients based on HTTP client to wrap either a pooling client or a mere connection, e.g. `WebClient` or `GrpcClient`.

### DIFF
--- a/src/main/asciidoc/http.adoc
+++ b/src/main/asciidoc/http.adoc
@@ -1581,6 +1581,8 @@ The {@link io.vertx.core.http.HttpClientConnection} can create {@link io.vertx.c
 {@link examples.HTTPExamples#connectAndGet}
 ----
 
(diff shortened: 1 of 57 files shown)


franz1981 (approved):
LGTM thanks Julien!

PR #5071, opened 2024-01-16
HTTP/2 configurable stream byte distributor

vietj:
A replacement of the HTTP/2 priority handling with a uniform stream byte distributor.

The default stream byte distributor use stream priorities to determine the amount of bytes to be sent for each stream, this change allows to use a strategy that does not use stream priorities and yields better performance since it consumes less CPU.

The default implementation remains the WeightedFairQueueByteDistributor, it can be replaced by UniformStreamByteDistributor with the system property `vertx.h2StreamByteDistributor` set to `uniform`. Note that this system property is evaluated when an HTTP server or an HTTP client is created and is not cached statically.


vietj:
@cescoffier how does sound the usage of a system property for this ? I'm reluctant using an option because it is quite low level and also coupled to the Netty implementation details

cescoffier:
Yes, a system property would be the best. I'm not sure we want to expose a property to do that switch.

vietj:
@cescoffier @franz1981 please have a look

cescoffier (approved):

Issue #5074, opened 2024-01-18
WebSocket client does not set underlying connection's error handler resulting in an unwanted error log


NilsRenaud:
### Version

I use Vert.x 4.5.2

### Context

I saw the error log statement below on the WebSocket client application while performing tests on Websocket connection between 2 components (explicitely closing/resetting it on server side).
```log
ERROR | vert.x-eventloop-thread-22 | i.v.c.n.i.ConnectionBase | @system | Connection reset 
```

Interestingly this only happens on Windows, never on Linux.

### Investigation

I found out the `ConnectionBase` error handler seems to never been set by the WebSocket

NilsRenaud:
Here is a reproducer: 
```java

import java.util.concurrent.CountDownLatch;

import io.vertx.core.http.HttpServerOptions;
import io.vertx.core.http.HttpServerRequest;

public class Test2 {

  public static void main(String[] args) throws InterruptedException {
    HttpServerOptions options = new HttpServerOptions();
    options.setPort(8080);
    CountDownLatch latch = new CountDownLatch(1);
    Vertx.vertx()
         .createHttpServer(options)
         .requestHandler(Test2::handleRequest)
         .exceptionHandler(t -> t.printStackTrace())
         .listen()
         .onSuccess(a -> {
           System.out.println("SERVER : started.");
           latch.countDown();
         });
    latch.await();

    Vertx.vertx().createWebSocketClient().connect(8080, "localhost", "/")
         .onSuccess(ws -> {
           ws.exceptionHandler(t -> {
             System.out.println("CLIENT : Exception - " + t.getMessage());
             //t.printStackTrace();
           });
           ws.endHandler(__ -> System.out.println("CLIENT : WS ended:" + ws.closeStatusCode() + " " + ws.closeReason()));
           ws.writeTextMessage("message")
             .onSuccess(__ -> System.out.println("CLIENT : message sent"))
             .onFailure(t -> System.out.println("CLIENT : message error" + t.getMessage()));
         })
         .onFailure(t -> {
           System.out.println("CLIENT : failed to connect");
           t.printStackTrace();
         });
  }

  private static void handleRequest(HttpServerRequest req) {
    req.toWebSocket()
       .onSuccess(ws -> {
         System.out.println("SERVER : request received");
         ws.accept();
         ws.handler(b -> {
           System.out.println("SERVER : message received, close WS");
           ws.close(Short.parseShort("1004")); // Invalid status code
         });
       });
  }
}
[...]

PR #5075, opened 2024-01-18
5074 Fix the WebSocket exception handling strategy

NilsRenaud:
Linked to this issue : https://github.com/eclipse-vertx/vert.x/issues/5074

The ConnectionBase exception handler was not set when `webSocket.exceptionHandler(...)` is called. This was leading `ConnectionBase` to consider the exception as uncaught and to print it, ignoring the exception handler set at the WebSocketImplBase level.

Moreover, this WebSocket handler was also called after the call to `ConnectionBase#handleException(...)`.

So I decided to set the `ConnectionBase` exception handler with the WebSocketImplBase.

The replacement of `Http1xConnectionBase conn` with `Http1xConnectionBase<?> conn` was needed otherwise the compiler complains.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -1229,17 +1229,12 @@ public class Http1xClientConnection extends Http1xConnectionBase<WebSocketImpl> 
   }
 
   @Override
-  protected void handleException(Throwable e) {
+  public void handleException(Throwable e) {
     super.handleException(e);
-    WebSocketImpl ws;
     LinkedHashSet<Stream> allStreams = new LinkedHashSet<>();
     synchronized (this) {
-      ws = webSocket;
       allStreams.addAll(requests);
       allStreams.addAll(responses);
-    }
-    if (ws != null) {
-      ws.handleException(e);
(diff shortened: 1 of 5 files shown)


vietj:
Your reasoning seems correct to me, however there are a few things we need to improve in this contribution

1. you should address also the HTTP server case that I think will behave the same (looking at `Http1xServerConnection`)

2. we should test that the related `HttpConnection` from where it is originating from is not reported the exception, i.e. any exception handler set before upgrading the `HttpConnection` to a `WebSocket` will never be called

3. setting up the `WebSocketImplBase#handleException` as exception handler of `HttpConnection` will prevent uncaught exception from being logged, so instead of doing so, we could remove the `WebSocketImplBase#exceptionHandler` field and instead set the exception handler on `WebSocketImplBase#conn` when it is called (i.e. `exceptionHandler(Handler<Throwable> handler)` implementation becomes `conn.exceptionHandler(handler)`. Doing this way we need to reroute internal calls reimplement `WebSocketImplBase#handleException` to delegate to `conn#handleException`

vietj:
Great contrib @NilsRenaud ! can you port to master branch ?

NilsRenaud:
Thanks @vietj I'll do this

Issue #5078, opened 2024-01-25
Memory Leak related to io.netty.util.concurrent.FastThreadLocal in Vertx code


bolbat:
### Version

Vertx: bug since  4.4.x, available on 4.5.1
Netty: version doesn't matter, tested with latest 4.1.106.Final
Java: tested on 17 and 21, but could be reproduced on any supported version

### Context

FastThreadLocal feature designed to be faster than basic ThreadLocal and internally use InternalThreadLocalMap which internally have:
- **static** _nextIndex_ field - to provide unique variable index for each _FastThreadLocal_ instance
- **instance** _indexedVariables_ field - to hold variables for specific _FastThreadLocal_ instance

Size of _indexedVariables_ field depends on a **static** _nextIndex_ field
If someone create a lot of _FastThreadLocal_ instance - this global static index will be increased by one for each instance
As result for all _InternalThreadLocalMap_ instances each invocation of _setIndexedVariable_ method will resize _indexedVariables_  array if index is bigger than array size (resizing _indexedVariables_ arrays in all instance because of **global static** index)

Consequently all _FastThreadLocalThread_ instances (including _VertxThread_) and FastThreadLocal fields will have continuously increasing _indexedVariables_ arrays in their InternalThreadLocalMap instances till JVM will be restarted

### Our case
On our production load one instance serving tens thousands requests per second and doing much more outgoing requests
Each instance use 64 physical cores and use 128 threads for Netty event-pool
For around 8h uptime each of this 128 threads have own _indexedVariables_ array instance with length 8_000_000+ and use 60MB+ memory
99.999% values in each array instance is empty and filled with _InternalThreadLocalMap.UNSET_ object
Usually each thread / array store only 3-15 values from features that use FastThreadLocal in a right way

### Heap dump details
<img width="1401" alt="Screenshot 2024-01-25 at 16 45 44" src="https://github.com/eclipse-vertx/vert.x/assets/323609/0ed453d9-82a3-4752-9821-bd055a78e366">
<img width="1402" alt="Screenshot 2024-01-25 at 16 46 23" src="https://github.com/eclipse-vertx/vert.x/assets/323609/afb53cc7-ba74-4af0-ad91-3cad3b10eccb">

###  What caused this?
After some investigation I've found that there some instances of _FastThreadLocal_ stored as not static fields and could be created a lot of times if parent object also created a lot of times

###  How to find exactly which code does this?
[...]

PR #5080, opened 2024-01-26
Use a static FastThreadLocal in CombinerExecutor instead of a instance field

vietj:
The CombinerExecutor class creates an instance of FastThreadLocal for each combiner executor leading to an increase of the InternalThreadLocalMap index. Consequently each thread local map of FastThreadLocalThread will get a new map sized accordingly, leading to an eventual memory leak.
    
Use a static FastThreadLocal in CombinerExecutor instead of a instance field, the data structure stored in this thread local map keeps track of the CombinerExecutor running in order to allow interleaved execution of CombinerExecutor post tasks without interfering each other. The structure is optimized for the most frequent case.

Fixes https://access.redhat.com/security/cve/CVE-2024-1023

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
+++ b/src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
@@ -13,6 +13,8 @@ package io.vertx.core.net.impl.pool;
 import io.netty.util.concurrent.FastThreadLocal;
 import io.netty.util.internal.PlatformDependent;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
```diff
@@ -72,23 +82,42 @@ public void submit(Action<S> action) {
...
-        inProgress.task = tail;
-
+        if (inProgress.combiner == this) {
+          Task oldNextTail = inProgress.task.replaceNext(head);
+          assert oldNextTail == null;
+          inProgress.task = tail;
+        } else {
+          Map<CombinerExecutor<S>, Task> map = inProgress.others;
+          if (map == null) {
+            map = inProgress.others = new HashMap<>();
```
franz1981, line 104: make it as small as possible, like capacity 2 or 1, depends on the default load factor

src/main/java/io/vertx/core/net/impl/pool/CombinerExecutor.java
```diff
@@ -72,23 +82,42 @@ public void submit(Action<S> action) {
...
       } else {
-        assert inProgress.task != null;
-        Task oldNextTail = inProgress.task.replaceNext(head);
-        assert oldNextTail == null;
-        inProgress.task = tail;
-
+        if (inProgress.combiner == this) {
+          Task oldNextTail = inProgress.task.replaceNext(head);
+          assert oldNextTail == null;
+          inProgress.task = tail;
```
franz1981, line 100: lovely!

franz1981 (approved):
LGTM just for "fun" try running the test suite adding an assert in case of "others"  is allocated (disable the test which make it happen eh!)

franz1981:
This is related https://github.com/eclipse-vertx/vert.x/issues/5078

PR #5088, opened 2024-02-01
Allow dynamic update of read and write limits set for Servers

sarankk:
Motivation:

Allowing dynamic updates of read and write limits set for HttpServer and NetServers created through Vert.x would be helpful during rate limit threshold setting tests. Currently dynamic updates are not possible, hence servers created have to be bounced and recreated. This could cause downtime. With this patch server thresholds can be updated dynamically.

### DIFF
--- a/src/main/java/examples/NetExamples.java
+++ b/src/main/java/examples/NetExamples.java
@@ -748,12 +748,8 @@ public class NetExamples {
                                      .setOutboundGlobalBandwidth(128 * 1024); // unchanged
     server
       .listen(1234, "localhost")
-      .onComplete(res -> {
-        if (res.succeeded()) {
-          // wait until traffic shaping handler is created for updates
-          server.updateTrafficShapingOptions(update);
-        }
-      });
+      // wait until traffic shaping handler is created for updates
+      .onSuccess(v -> server.updateTrafficShapingOptions(update));
   }
 
   public void configureTrafficShapingForHttpServer(Vertx vertx) {
@@ -781,11 +777,7 @@ public class NetExamples {
                                      .setOutboundGlobalBandwidth(128 * 1024); // unchanged
     server
       .listen(1234, "localhost")
-      .onComplete(res -> {
-        if (res.succeeded()) {
-          // wait until traffic shaping handler is created for updates
-          server.updateTrafficShapingOptions(update);
-        }
-      });
+      // wait until traffic shaping handler is created for updates
+      .onSuccess(v -> server.updateTrafficShapingOptions(update));
   }
 }
--- a/src/main/java/examples/NetExamples.java
+++ b/src/main/java/examples/NetExamples.java
@@ -735,6 +735,27 @@ public class NetExamples {
     NetServer server = vertx.createNetServer(options);
   }
 
+  public void dynamicallyUpdateTrafficShapingForNetServer(Vertx vertx) {
+    NetServerOptions options = new NetServerOptions()
+                                 .setHost("localhost")
+                                 .setPort(1234)
+                                 .setTrafficShapingOptions(new TrafficShapingOptions()
+                                                             .setInboundGlobalBandwidth(64 * 1024)
+                                                             .setOutboundGlobalBandwidth(128 * 1024));
+    NetServer server = vertx.createNetServer(options);
+    TrafficShapingOptions update = new TrafficShapingOptions()
+                                     .setInboundGlobalBandwidth(2 * 64 * 1024) // twice
+                                     .setOutboundGlobalBandwidth(128 * 1024); // unchanged
+    server
+      .listen(1234, "localhost")
+      .onComplete(res -> {
+        if (res.succeeded()) {
+          // wait until traffic shaping handler is created for updates
+          server.updateTrafficShapingOptions(update);
+        }
+      });
+  }
+
   public void configureTrafficShapingForHttpServer(Vertx vertx) {
     HttpServerOptions options = new HttpServerOptions()
       .setHost("localhost")
@@ -744,5 +765,27 @@ public class NetExamples {
         .setOutboundGlobalBandwidth(128 * 1024));
 
     HttpServer server = vertx.createHttpServer(options);
+  }
+
+
+  public void dynamicallyUpdateTrafficShapingForHttpServer(Vertx vertx) {
+    HttpServerOptions options = new HttpServerOptions()
+                                  .setHost("localhost")
+                                  .setPort(1234)
+                                  .setTrafficShapingOptions(new TrafficShapingOptions()
+                                                              .setInboundGlobalBandwidth(64 * 1024)
+                                                              .setOutboundGlobalBandwidth(128 * 1024));
+    HttpServer server = vertx.createHttpServer(options);
+    TrafficShapingOptions update = new TrafficShapingOptions()
+                                     .setInboundGlobalBandwidth(2 * 64 * 1024) // twice
+                                     .setOutboundGlobalBandwidth(128 * 1024); // unchanged
+    server
+      .listen(1234, "localhost")
+      .onComplete(res -> {
+        if (res.succeeded()) {
+          // wait until traffic shaping handler is created for updates
+          server.updateTrafficShapingOptions(update);
+        }
+      });
   }
 }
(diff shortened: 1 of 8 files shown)


src/main/java/io/vertx/core/net/impl/TCPServerBase.java
```diff
@@ -162,6 +162,26 @@ public Future<Boolean> updateSSLOptions(ServerSSLOptions options, boolean force)
...
+    }
+    if (trafficShapingHandler == null) {
+      throw new IllegalArgumentException("Unable to update traffic shaping options because the server was not configured " +
+                                         "to use traffic shaping during startup");
+    }
+    TCPServerBase server = actualServer;
+    if (server != null && server != this) {
+      server.updateTrafficShapingOptions(options);
+    } else {
+      if (options.getCheckIntervalForStatsTimeUnit() != null) {
```
frankgh, line 177: time unit cannot be null, so we should probably remove the else branch

src/main/java/io/vertx/core/http/HttpServer.java
```diff
@@ -130,6 +129,14 @@ default Future<Boolean> updateSSLOptions(ServerSSLOptions options) {
    */
   Future<Boolean> updateSSLOptions(ServerSSLOptions options, boolean force);
 
+  /**
+   * Update traffic shaping options {@code options}, the update happens if valid values are passed for traffic
+   * shaping options. This update happens synchronously and at best effort for rate update to take effect immediately.
+   *
+   * @param options the new traffic shaping options
+   */
+  void updateTrafficShapingOptions(TrafficShapingOptions options);
```
frankgh, line 138: Should we add usage sample in `examples.NetExamples`?

src/main/java/io/vertx/core/net/impl/TCPServerBase.java
```diff
@@ -162,6 +162,26 @@ public Future<Boolean> updateSSLOptions(ServerSSLOptions options, boolean force)
...
+        trafficShapingHandler.configure(options.getOutboundGlobalBandwidth(), options.getInboundGlobalBandwidth(), options.getCheckIntervalForStats());
```
frankgh, line 178: This seems incorrect, we should probably compute the millis value for the check interval for stats property:

```
long checkIntervalForStatsInMillis = options.getCheckIntervalForStatsTimeUnit().toMillis(options.getCheckIntervalForStats());
```

I think we are missing updating the `maxDelayToWaitInMillis` value as well as the `maxGlobalWriteSize` 

```
    if (options.getPeakOutboundGlobalBandwidth() != 0) {
      trafficShapingHandler.setMaxGlobalWriteSize(options.getPeakOutboundGlobalBandwidth());
    }
```

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
```diff
@@ -139,14 +139,14 @@ public TrafficShapingOptions setMaxDelayToWaitUnit(TimeUnit maxDelayToWaitTimeUn
...
-   * Set the delay between two computations of performances for channels or 0 if no stats are to be computed
+   * Set the delay between two computations of performances for channels
    *
    * @param checkIntervalForStats delay between two computations of performances
    * @return a reference to this, so the API can be used fluently
    */
   public TrafficShapingOptions setCheckIntervalForStats(long checkIntervalForStats) {
     this.checkIntervalForStats = checkIntervalForStats;
-    ObjectUtil.checkPositive(this.checkIntervalForStats, "checkIntervalForStats");
+    ObjectUtil.checkPositiveOrZero(this.checkIntervalForStats, "checkIntervalForStats");
```
frankgh, line 149: looking at the netty code, nothing prevents us from having a negative value for this field. Any reason we are restricting it to zero or positive?

```
    public void configure(long newCheckInterval) {
        long newInterval = newCheckInterval / 10 * 10;
        if (checkInterval.getAndSet(newInterval) != newInterval) {
            if (newInterval <= 0) {
                stop();
                // No more active monitoring
                lastTime.set(milliSecondFromNano());
            } else {
                // Restart
                stop();
                start();
            }
        }
    }
```

frankgh (changes requested):
I think there are a few places where we need to more carefully look at the underlying netty options and try to respect the underlying behavior. Additionally, we are not updating all the `TrafficShapingOptions`, so we will need to do that as well.

src/main/java/io/vertx/core/net/impl/TCPServerBase.java
```diff
@@ -162,6 +162,26 @@ public Future<Boolean> updateSSLOptions(ServerSSLOptions options, boolean force)
     }
   }
 
+  public void updateTrafficShapingOptions(TrafficShapingOptions options) {
+    if (options == null) {
+      throw new IllegalArgumentException("Invalid null value passed for traffic shaping options update");
+    }
+    if (trafficShapingHandler == null) {
+      throw new IllegalArgumentException("Unable to update traffic shaping options because the server was not configured " +
```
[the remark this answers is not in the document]
sarankk: The error we throw is based on the argument passed. Hence I thought `IllegalArgumentException` would be more appropriate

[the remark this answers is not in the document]
frankgh: For the case of the `trafficShapingHandler` the argument is not passed, but rather we have an instance already.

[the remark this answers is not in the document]
sarankk: Since `trafficShapingHandler` is not passed, will switch to `IllegalStateException`

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
sarankk: Since negative values and 0 have the same effect. I thought we can use 0 for stop monitoring and check for disallow negative values to avoid confusion

src/main/java/io/vertx/core/net/impl/TCPServerBase.java
sarankk: Thanks, addressed above

src/main/java/examples/NetExamples.java
```diff
@@ -745,4 +766,26 @@ public void configureTrafficShapingForHttpServer(Vertx vertx) {
...
+                                  .setTrafficShapingOptions(new TrafficShapingOptions()
+                                                              .setInboundGlobalBandwidth(64 * 1024)
+                                                              .setOutboundGlobalBandwidth(128 * 1024));
+    HttpServer server = vertx.createHttpServer(options);
+    TrafficShapingOptions update = new TrafficShapingOptions()
+                                     .setInboundGlobalBandwidth(2 * 64 * 1024) // twice
+                                     .setOutboundGlobalBandwidth(128 * 1024); // unchanged
+    server
+      .listen(1234, "localhost")
+      .onComplete(res -> {
```
vietj, line 784: use `onSuccess(v -> server.updateTrafficShapinOptions(update))` idiom instead

src/main/java/examples/NetExamples.java
```diff
@@ -735,6 +735,27 @@ public void configureTrafficShapingForNetServer(Vertx vertx) {
...
+                                 .setPort(1234)
+                                 .setTrafficShapingOptions(new TrafficShapingOptions()
+                                                             .setInboundGlobalBandwidth(64 * 1024)
+                                                             .setOutboundGlobalBandwidth(128 * 1024));
+    NetServer server = vertx.createNetServer(options);
+    TrafficShapingOptions update = new TrafficShapingOptions()
+                                     .setInboundGlobalBandwidth(2 * 64 * 1024) // twice
+                                     .setOutboundGlobalBandwidth(128 * 1024); // unchanged
+    server
+      .listen(1234, "localhost")
```
vietj, line 750: use `onSuccess(v -> server.updateTrafficShapinOptions(update))` idiom instead

src/main/java/io/vertx/core/net/TrafficShapingOptions.java
frankgh: makes sense. Previously we were disallowing disabling it, and this addresses that issue.

vietj:
Thanks for the contribution, it has been back-ported to 4.x branch https://github.com/eclipse-vertx/vert.x/pull/5096

PR #5094, opened 2024-02-05
ObjectMapper configuration should apply when pretty printing

tsegismont:
Follows-up on https://github.com/eclipse-vertx/vert.x/pull/5060

Since the pretty mapper is deprecated, any customization of the object mapper should be used when pretty printing with `Json.encodePrettily`.

### DIFF
--- a/src/main/java/io/vertx/core/json/jackson/DatabindCodec.java
+++ b/src/main/java/io/vertx/core/json/jackson/DatabindCodec.java
@@ -27,7 +27,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
@@ -35,10 +34,11 @@ import java.util.concurrent.atomic.AtomicReference;
 public class DatabindCodec extends JacksonCodec {
 
   private static final ObjectMapper mapper = new ObjectMapper();
-  private static final AtomicReference<ObjectMapper> prettyMapper = new AtomicReference<>();
+  private static final ObjectMapper prettyMapper = new ObjectMapper();
 
   static {
     initialize(mapper, false);
+    initialize(prettyMapper, true);
   }
 
   private static void initialize(ObjectMapper om, boolean prettyPrint) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/json/jackson/DatabindCodec.java
```diff
@@ -27,29 +27,28 @@
...
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
 public class DatabindCodec extends JacksonCodec {
 
   private static final ObjectMapper mapper = new ObjectMapper();
-  private static final ObjectMapper prettyMapper = new ObjectMapper();
+  private static final AtomicReference<ObjectMapper> prettyMapper = new AtomicReference<>();
```
vietj, line 38: I do not understand why we need atomic here ? what is the purpose of this ?

vietj: what is the issue with not lazy creating the pretty mapper ?

vietj: yes I do prefer, I don't think we need to save anything here I prefer to use atomic to be used for what they are designed for, if we really need some kind of lazy loading we would use the class lazy loading trick instead

tsegismont: Done

magicprinc:
It is better to use a customized DefaultPrettyPrinter:

`
objectMapper.setDefaultPrettyPrinter(new Pretty());// .writerWithDefaultPrettyPrinter()


private static final class Pretty extends DefaultPrettyPrinter {
  public Pretty (DefaultPrettyPrinter base){
    super(base);  
    fix();
  }

  Pretty (){
    fix();
  }

  private void fix (){
    _objectFieldValueSeparatorWithSpaces = ": ";// otherwise: "propertyName" : 1 < extra space before :
    _objectIndenter = new DefaultIndenter("  ", "\n");// 2 spaces (no changes), LF (vs \r\n on Windows)
  }

  @Override
  public DefaultPrettyPrinter createInstance (){
    return new Pretty(this);
  }
}
`

magicprinc:
io.vertx.core.json.jackson.DatabindCodec#mapper must be created with JacksonCodec#factory (with HybridJacksonPool

PR #5098, opened 2024-02-06
Add helper method to determine if request can be upgraded to websocket.

tsegismont:
Vert.x Web And Vert.x HTTP Proxy have similar code. This method covers all cases and can be reused.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpUtils.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpUtils.java
@@ -983,4 +983,24 @@ public final class HttpUtils {
     int len = host.length();
     return HostAndPortImpl.parseHost(host, 0, len) == len;
   }
+
+  public static boolean canUpgradeToWebSocket(HttpServerRequest req) {
+    if (req.version() != io.vertx.core.http.HttpVersion.HTTP_1_1) {
+      return false;
+    }
+    if (req.method() != io.vertx.core.http.HttpMethod.GET) {
+      return false;
+    }
+    MultiMap headers = req.headers();
+    for (String connection : headers.getAll(io.vertx.core.http.HttpHeaders.CONNECTION)) {
+      if (connection.toLowerCase().contains(io.vertx.core.http.HttpHeaders.UPGRADE)) {
+        for (String upgrade : headers.getAll(io.vertx.core.http.HttpHeaders.UPGRADE)) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/http/impl/HttpUtils.java
```diff
@@ -983,4 +983,24 @@ public static boolean isValidHostAuthority(String host) {
...
+    if (req.version() != io.vertx.core.http.HttpVersion.HTTP_1_1) {
+      return false;
+    }
+    if (req.method() != io.vertx.core.http.HttpMethod.GET) {
+      return false;
+    }
+    MultiMap headers = req.headers();
+    for (String connection : headers.getAll(io.vertx.core.http.HttpHeaders.CONNECTION)) {
+      if (connection.toLowerCase().contains(io.vertx.core.http.HttpHeaders.UPGRADE)) {
+        for (String upgrade : headers.getAll(io.vertx.core.http.HttpHeaders.UPGRADE)) {
```
machine424, line 997: shouldn't we check for the exact `Upgrade: websocket`? it's cheaper and we can always address in future special cases.

machine424 (commented):
Thanks Thomas!

src/test/java/io/vertx/core/http/Http1xTest.java
```diff
@@ -5374,4 +5371,61 @@ public void testMissingHostHeader() throws Exception {
...
+    UnaryOperator<RequestOptions> config = ro -> ro.setMethod(HttpMethod.GET)
+      .putHeader(HttpHeaders.CONNECTION, HttpHeaders.UPGRADE)
+      .putHeader(HttpHeaders.UPGRADE, "foo");
+    doTestCanUpgradeToWebSocket(config, false);
+  }
+
+  private void doTestCanUpgradeToWebSocket(UnaryOperator<RequestOptions> config, boolean shouldSucceed) throws Exception {
+    server.requestHandler(req -> {
+      HttpServerResponse serverResponse = req.response();
+      int statusCode = HttpUtils.canUpgradeToWebSocket(req) ? 200 : 500;
```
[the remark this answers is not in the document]
tsegismont: It's just a test, we don't really care.

src/main/java/io/vertx/core/http/impl/HttpUtils.java
tsegismont: No, because servers should accept any case according to the spec

PR #5103, opened 2024-02-07
refactor: remove unused test method

yeikel:
As far as I can tell, this method is not used

As a bonus, we stop using the deprecated `X509Certificate` class

### DIFF
--- a/src/test/java/io/vertx/test/core/TestUtils.java
+++ b/src/test/java/io/vertx/test/core/TestUtils.java
@@ -31,9 +31,6 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 import java.util.zip.GZIPOutputStream;
-
-import javax.security.cert.X509Certificate;
-
 import io.netty.buffer.Unpooled;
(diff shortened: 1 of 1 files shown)


yeikel:
The build failures a bit confusing. What  is this about?

vietj:
this might be used in other project depending on vertx core so a check on those project should be done

yeikel:
> this might be used in other project depending on vertx core so a check on those project should be done

What other projects are you referring to? Where should I check?

vietj:
other projects in vert-x3 or eclipse-vertx, some of them rely on vertx-core test artifact

yeikel:
> other projects in vert-x3 or eclipse-vertx, some of them rely on vertx-core test artifact

I am a bit confused here because this is coming from a test class and test classes are typically not published for others to use

I [searched for this method](https://github.com/search?q=org%3Aeclipse-vertx%20cnOf&type=code) name, and all the usages are from within this same project (and the method with the other signature)

vietj:
that might be confusing but you can look at vertx-web and how it uses vertx-core VertxTest for instance

PR #5104, opened 2024-02-07
refactor: migrate usages of the Provider class

yeikel:
The constructor we are using was deprecated in Java 9

```java

 /**
     * Constructs a provider with the specified name, version number,
     * and information. Calling this constructor is equivalent to call the
     * {@link #Provider(String, String, String)} with {@code name}
     * name, {@code Double.toString(version)}, and {@code info}.
     *
     * @param name the provider name.
     *
     * @param version the provider version number.
     *
     * @param info a description of the provider and its services.
     *
     * @deprecated use {@link #Provider(String, String, String)} instead.
     */

```


yeikel:
> But jdk1.8 only has original version of api. ref: https://docs.oracle.com/javase/8/docs/api/java/security/Provider.html
> And vertx is support `JDK 1.8 or higher`. ref: https://vertx.io/get-started/
> Is it safe to do this?

Vert.x 5 will target Java 11

vietj:
thanks , there are a few occurrences of this in other classes:

- KeyManagerFactoryWrapper
- TrustAllOptions
- VertxTrustManagerFactory

PR #5106, opened 2024-02-07
refactor: remove check

yeikel:


src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -431,7 +431,7 @@ private HttpClientRequest createRequest(
       if (!ABS_URI_START_PATTERN.matcher(u).find()) {
         int defaultPort = 80;
         String addPort = (authority.port() != -1 && authority.port() != defaultPort) ? (":" + authority.port()) : "";
-        u = (useSSL == Boolean.TRUE ? "https://" : "http://") + authority.host() + addPort + requestURI;
+        u = "http://" + authority.host() + addPort + requestURI;
```
yeikel, line 434: This check is not needed because in line 430, we're checking !useSSL

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
```diff
@@ -182,7 +182,7 @@ private EndpointProvider<EndpointKey, SharedClientHttpStreamEndpoint> httpEndpoi
       ProxyOptions proxyOptions = key.proxyOptions;
       if (proxyOptions != null && !key.ssl && proxyOptions.getType() == ProxyType.HTTP) {
         SocketAddress server = SocketAddress.inetSocketAddress(proxyOptions.getPort(), proxyOptions.getHost());
-        key = new EndpointKey(key.ssl, key.sslOptions, proxyOptions, server, key.authority);
+        key = new EndpointKey(false, key.sslOptions, proxyOptions, server, key.authority);
```
yeikel, line 185: It is more explicit than `key.ssl` , and we are negating it in line 183

src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
yeikel: @vietj Do you know what was the intention of this line?

vietj:
can you rebase on latest master there is a conflict

PR #5117, opened 2024-02-13
Explicit configuration of the TCP over TLS client hostname verification

vietj:
Vert.x TCP client does configure the TLS hostname verification algorithm to the empty string which means no checks of the server identity shall be performed beyond checking the validity of the certificate.

This set the default algorithm to null instead of the empty string, any usage of the TCP client must set the verification algorithm when used over TLS, valid values are empty string, HTTPS or LDAPS.

This is a breaking change when using NetClient over TLS.


yeikel (approved):
Thank you!

PR #5118, opened 2024-02-15
Future control flow using predicates

vietj:
`Future` control flow using predicates.

`Future` has already synchronous operations modifying the control flow
- `map` : transforms a mapper failure into a failed future (e.g. `future.map(val -> throw new VertxException())`)
- `otherwise` : transforms a failure into a success

This contribution aims to transform a success into a failure based on the result of a predicate.

Goals

- capture useful predicates that modify the control flow, e.g. a predicate that checks the HTTP client response  (status code, headers, etc...)
- provide inline guards using lambda expressions

Vert.x Web Client has something close to this with HTTP expectations:

```java
    client
      .get(8080, "myserver.mycompany.com", "/some-uri")
      .expect(ResponsePredicate.SC_SUCCESS)
      .expect(ResponsePredicate.JSON)
      .send()
      .onSuccess(res -> {
        // Safely decode the body as a json object
        JsonObject body = res.bodyAsJsonObject();
        System.out.println(
          "Received response with status code" +
            res.statusCode() +
            " with body " +
            body);
      })
      .onFailure(err ->
        System.out.println("Something went wrong " + err.getMessage()));
```

A predicate like interface `Expectation`: a synchronous predicate that checks a boolean value and has a method to create a meaningful exception.

```java
Future<Buffer> body = client.request(options)
  .compose(req -> req
    .send()
    .expecting(resp -> resp.statusCode() == 200)
    .compose(resp -> resp.body()); 
```

The HTTP response status code can be captured so it can be reused.

```java
Future<Buffer> body = client.request(options)
  .compose(req -> req
    .send()
    .expecting(HttpResponseExpectation.SC_OK)
    .compose(resp -> resp.body()); 
```

### DIFF
--- a/src/main/java/io/vertx/core/Expectation.java
+++ b/src/main/java/io/vertx/core/Expectation.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core;
+
+import java.util.Objects;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+/**
+ * An expectation, very much like a predicate with the ability to provide a meaningful description of the failure.
+ * <p/>
+ * Expectation can be used with {@link Future#expecting(Expectation)}.
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ */
+@FunctionalInterface
+public interface Expectation<V> {
+
+  /**
+   * Check the {@code value}, this should be side effect free, the expectation should either succeed returning {@code true} or fail
+   * returning {@code false}.
+   *
+   * @param value the checked value
+   */
+  boolean test(V value);
+
+  /**
+   * Returned an expectation succeeding when this expectation and the {@code other} expectation succeeds.
+   *
+   * @param other the other expectation
+   * @return an expectation that is a logical and of this and {@code other}
+   */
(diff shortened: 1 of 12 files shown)


src/main/java/io/vertx/core/Expectation.java
```diff
@@ -0,0 +1,33 @@
...
+
+@FunctionalInterface
+public interface Expectation<T> extends Assertion<T> {
+
+  boolean test(T t);
+
+  @Override
+  default void check(T t) throws Throwable {
+    if (!test(t)) {
+      throw new AssertionError(toException(t));
```
Ladicek, line 21: When an `Expectation` is created using a lambda expression, which is likely to be very common, `toException` will be `null` and hence we'll `throw new AssertionError(null)`, which is not very friendly.

Wouldn't something like this be better?

```java
default void check(T t) throws Throwable {
  if (!test(t)) {
    Exception exception = toException(t);
    if (exception == null) {
      throw new AssertionError("Expectation failed for " + t);
    } else {
      throw new AssertionError(exception); // or just `throw exception`?
    }
  }
}
```

Ladicek: OK, I see that the `Future` operation that handles an `Expectation` deals with this situation, so please just ignore me here :-)

vietj:
Further work : for now dropping the assertion part and instead try to develop a test construct to handle this through andThen/onComplete

src/main/java/io/vertx/core/http/HttpResponseExpectation.java
```diff
@@ -0,0 +1,394 @@
...
+      }
+    };
+  }
+
+  /**
+   * Creates a predicate validating the response has a {@code content-type} header matching one of the {@code mimeTypes}.
+   *
+   * @param mimeTypes the list of mime types
+   */
+  static HttpResponseExpectation contentType(List<String> mimeTypes) {
```
oliver-brm, line 356:
```suggestion
  static HttpResponseExpectation contentType(String... mimeTypes) {
```
(so we can set 0..n content types)

vietj: maybe we need both, I expect this code to be also used from WebClient expectations.

PR #5120, opened 2024-02-16
Validate content length early during HttpServerResponse#sendFile

frankgh:
In `HttpServerResponse#sendFile`, when the offset is greater than the file length, the calculated content length is negative. Currently, the value is not validated until after the head (response status and other headers) is written. An application might have downstream processing that would act on the `IllegalArgumentException` thrown when validating the content length. However, it is too late for the downstream processor to write a different response status (for example a 400 BAD REQUEST), to signal the client that the request is not valid. The problem is exacerbated when the downstream processor tries to write a status code, because the `IllegalStateException` exception will be thrown because the `Response head is already sent`. This leaves the connection hanging on the server side, while the client is expecting content.

Motivation:

Conformance:


frankgh:
I will provide a patch for `master` shortly.

frankgh:
I found a couple additional issues that I will address shortly. They've been fixed for the PR against `master`

src/test/java/io/vertx/core/http/HttpTest.java
```diff
@@ -2167,6 +2167,64 @@ public void testSendRangeFileFromClasspath() {
     await();
   }
 
+  @Test
+  public void testSendZeroRangeFileFromClasspath() {
+    server.requestHandler(res -> {
+      // hosts_config.txt is 23 bytes
```
vietj, line 2173: can you check this in the tests (hosts_config.txt is 23 bytes), someday we might modify hosts_config.txt and have this test not working correctly

PR #5121, opened 2024-02-16
Validate content length early during HttpServerResponse#sendFile

frankgh:
In `HttpServerResponse#sendFile`, when the offset is greater than the file length, the calculated content length is negative. Currently, the value is not validated until after the head (response status and other headers) is written. An application might have downstream processing that would act on the `IllegalArgumentException` thrown when validating the content length. However, it is too late for the downstream processor to write a different response status (for example a 400 BAD REQUEST), to signal the client that the request is not valid. The problem is exacerbated when the downstream processor tries to write a status code, because the `IllegalStateException` exception will be thrown because the `Response head is already sent`. This leaves the connection hanging on the server side, while the client is expecting content.

Additionally, the `offset` and `length` parameters are not validated, which can produce negative content length.

Motivation:

Conformance:


vietj:
can you rebase on latest master branch @frankgh ?

frankgh:
@vietj  Done, I'm also generating the test files dynamically for each test.

vietj:
merged in #5127

PR #5125, opened 2024-02-19
Speedup Local-Non-Local Context's data operations

franz1981:
This is addressing 3 performance problems:
- local/non-local context data's operations shouldn't use costy `synchronized` on existing maps (now that biased locking is gone)
- local/non-local context data's lookup/remove doesn't really require to create any map, if not already present
- local-non-local duplicated ctx data operations are not piggybacking on deletated ones, forcing using `default` subptimal implement ones, instead

This PR is pushing improvement furthers, but leveraging the facts that:
- default's `ConcurentHasMap` constructor doesn't initialize the internal table, making very cheap to allocate one
- most of the time there's no concurrent data allocations, which means we can save using `synchronized` and just use `compareAndSet` saving another strong memory barrier (in x86, but likely the same in ARM), speeding up common use cases

### DIFF
--- a/src/main/java/io/vertx/core/impl/ContextImpl.java
+++ b/src/main/java/io/vertx/core/impl/ContextImpl.java
@@ -21,6 +21,7 @@ import io.vertx.core.spi.metrics.PoolMetrics;
 import io.vertx.core.spi.tracing.VertxTracer;
 
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 /**
  * A base class for {@link Context} implementations.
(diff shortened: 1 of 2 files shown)


vietj:
before doing anything here, we should first validate and finish this PR https://github.com/eclipse-vertx/vert.x/pull/5048 which has an impact on the overall design, since I would like to back port this PR

franz1981:
> since I would like to back port this PR to 4.x

this PR you mean #5048 ?

franz1981:
@vietj if merged I have to fix https://github.com/quarkusio/quarkus/blob/4e6908f522190127a28af7c96806c06cc0b0a60e/extensions/vertx/runtime/src/main/java/io/quarkus/vertx/core/runtime/VertxLocalsHelper.java#L25 which is wrongly allocating a local map, although not really required.

PR #5130, opened 2024-02-24
CompositeFuture implementation must unregister against its component on completion

vietj:
A `CompositeFuture` should unregister itself against its components when the composite future completes since the composite future is not interested anymore by the completion of the uncompleted futures. Otherwise the composite future might remain forever in the list of a promise pending completion, which could lead in some situation to a memory leak.

The implementation of `CompositeFuture` now does unregister when it completes, the registration against the component new uses the internal listener interface instead of the API `onComplete` method in order to keep a reference to the listener that was registered in order to unregister the same listener. In addition a single listener is now used, reducing the number of listener required to implement the composition behavior to a single listener instead of a listener per component. The creation of implementation performs now a two phase initialisation instead of a single since a component might already be completed and this should not interfere with the listener registration process, e.g. a component might be completed and the remaining components to register a listener for should be unregistered or avoid to be registered; it seems cleaner to first register against all components and have the listener only update the state of the composite future and then perform an initialization phase that will finish the completion if the composite future was completed during the registration phase.

The implementation of `Future` has been modified to allow the unregistration of a listener as require by the composite future change.

### DIFF
--- a/src/main/java/io/vertx/core/impl/future/CompositeFutureImpl.java
+++ b/src/main/java/io/vertx/core/impl/future/CompositeFutureImpl.java
@@ -16,111 +16,151 @@ import io.vertx.core.Future;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Handler;
 
-import java.util.function.Function;
-import java.util.stream.Collectors;
-
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
-public class CompositeFutureImpl extends FutureImpl<CompositeFuture> implements CompositeFuture {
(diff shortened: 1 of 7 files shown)


Fyro-Ing:
Hi

There is a regression from this PR, no ?

from 4.5.4, when using Future.join, result is CompositeFuture with all results & failed
and we can use #transform with final AsyncResult<CompositeFuture>

but now from 4.5.5, #join return the first failed future and doesn't honor contract as CompositeFuture and return FailedFuture Now

vietj:
can you provide a reproducer ?

Fyro-Ing:
https://github.com/Fyro-Ing/vert.x/commit/23488ac5bda1a95f31eb0090ac2655a2febb3634

same, on branche 4.5.4, & it's ok

PR #5137, opened 2024-02-28
fix: Dropped proxy on redirect

### LINKED ISSUE #5136: HttpClient drops original proxy on redirect
### Version

Which version(s) did you encounter this bug ?
`>= 4.x`

### Context

`RequestOptions` allows us to set a proxy per request, but upon redirect the original proxy gets lost

It also affects `WebClient`.

### Do you have a reproducer?

I have added tests for it in PR, just run `HttpTest`

zekronium:
Motivation:

Fixes #5136

This can also be fixed by setting the original proxy on `HttpClientRequest` but this way it works with custom handlers.

`WebClient` can be fixed naturally via this commit OR can have its own fix, since it contains other ways to retrieve the proxy in request chain

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpClientImpl.java
@@ -467,7 +467,10 @@ public class HttpClientImpl extends HttpClientBase implements HttpClientInternal
       request.redirectHandler(resp -> {
         Future<RequestOptions> fut_ = rHandler.apply(resp);
         if (fut_ != null) {
-          return fut_.compose(this::request);
+          return fut_.compose(o -> {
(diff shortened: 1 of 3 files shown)


vietj:
can you update your PR to not delete the generated vertx classes

zekronium:
> can you update your PR to not delete the generated vertx classes

Ok, no problem

zekronium:
Fixed. I also feel, maybe we should reference only the `ProxyOptions` object instead of the entire RequestOptions?

zekronium:
This also requires adjustment to some of the methods in `4.x` to work, since request options are not passed in to `createStream`

vietj:
can you back port @zekronium

zekronium:
> can you back port @zekronium

Separate PR?

PR #5140, opened 2024-02-29
An SPI for interacting with local context storage

vietj:
The existing implementation uses on a concurrent hash map guarded by the context instance.

This new implementation relies on prior knowledge of the set of context locals, so the context allocates upfront the required space to hold the local storage, reducing synchronisation and simplifying lookup that uses allocated index instead of hashed string lookups.

The existing local context data is retrofit as a context local.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -720,6 +720,21 @@
               </additionalClasspathElements>
             </configuration>
           </execution>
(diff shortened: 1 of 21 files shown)


vietj:
@franz1981

PR #5143, opened 2024-03-06
Remove un-necessary Http1xServerConnection internal state synchronised access

vietj:
The `Http1xServerConnection` synchronize itself to guard the internal state of the request/response in progress.

Such synchronization used to be useful before Vert.x 4 and the clear dispatch between connection/request code. Currently read and writes are confined in the event-loop thread and removing this synchronization has no effect on the read/writes of this state.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
@@ -199,10 +199,7 @@ public class Http1xServerConnection extends Http1xConnectionBase<ServerWebSocket
       return;
     }
     Buffer buffer = Buffer.buffer(VertxHandler.safeBuffer(content.content()));
-    Http1xServerRequest request;
-    synchronized (this) {
-      request = requestInProgress;
-    }
+    Http1xServerRequest request = requestInProgress;
     request.context.execute(buffer, request::handleContent);
     //TODO chunk trailers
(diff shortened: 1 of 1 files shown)


franz1981:
FYI:
https://github.com/quarkusio/quarkus/blob/main/independent-projects/resteasy-reactive/server/vertx/src/main/java/org/jboss/resteasy/reactive/server/vertx/ResteasyReactiveOutputStream.java#L53

and check ` Http1xServerRequest` as well, which is still synchronizing over `conn` -> the connection

vietj:
I think it should be fine for this case as we update the state of the request/response in progress and this sync and the request/close status which are not related to back-pressure/drain. WDYT ?

> FYI: https://github.com/quarkusio/quarkus/blob/main/independent-projects/resteasy-reactive/server/vertx/src/main/java/org/jboss/resteasy/reactive/server/vertx/ResteasyReactiveOutputStream.java#L53
> 
> and check ` Http1xServerRequest` as well, which is still synchronizing over `conn` -> the connection

franz1981:
@vietj 

the best way would be to try running some of the tests on Quarkus which relies on it, by using a vertx version relax these properties @geoand @cescoffier how it looks?

vietj:
which others do you mean ?

franz1981:
> which others do you mean ?

Http1xServerRequest, for example

PR #5144, opened 2024-03-06
Fix InboundBuffer::pause & write performance

franz1981:
This patch improve Quarkus techempower plaintext throughput of ~5% in our lab; nothing fancy but seems related to the excessive synchronization which, by using instance field to synchronize with, doesn't allow the JVM to perform the https://shipilev.net/jvm/anatomy-quarks/1-lock-coarsening-for-loops/ (lock coarsening) optimization.

I've applied an optimization to `onEnd` which benefit vertx as well, which often observe non-paused requests 
i.e. reducing the synchronization over the connection while completing the request.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
@@ -567,32 +567,45 @@ public class Http1xServerRequest extends HttpServerRequestInternal implements io
 
   void handleEnd() {
     InboundBuffer<Object> queue;
+    HttpEventHandler handler = null;
     synchronized (conn) {
       ended = true;
       queue = pending;
+      if (queue == null) {
+        handler = onRequestComplete();
+      }
     }
     if (queue != null) {
       queue.write(InboundBuffer.END_SENTINEL);
     } else {
-      onEnd();
+      handleEnd(handler);
     }
   }
 
-  private void onEnd() {
+  private HttpEventHandler onRequestComplete() {
+    assert Thread.holdsLock(conn);
     if (METRICS_ENABLED) {
       reportRequestComplete();
     }
-    HttpEventHandler handler;
-    synchronized (conn) {
-      if (decoder != null) {
-        endDecode();
-      }
-      handler = eventHandler;
+    if (decoder != null) {
+      endDecode();
     }
+    return eventHandler;
+  }
+
+  private void handleEnd(HttpEventHandler handler) {
     // If there have been uploads then we let the last one call the end handler once any fileuploads are complete
     if (handler != null) {
       handler.handleEnd();
     }
+  }
+
+  private void onEnd() {
+    HttpEventHandler handler;
+    synchronized (conn) {
+      handler = onRequestComplete();
+    }
+    handleEnd(handler);
   }
 
   private void reportRequestComplete() {
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
@@ -123,21 +123,6 @@ public class Http1xServerRequest extends HttpServerRequestInternal implements io
     }
   }
 
-  private InboundBuffer<Object> pendingQueue() {
-    if (pending == null) {
-      pending = new InboundBuffer<>(context, 8);
-      pending.drainHandler(v -> conn.doResume());
-      pending.handler(buffer -> {
-        if (buffer == InboundBuffer.END_SENTINEL) {
-          onEnd();
-        } else {
-          onData((Buffer) buffer);
-        }
-      });
-    }
-    return pending;
-  }
-
   void handleContent(Buffer buffer) {
     InboundBuffer<Object> queue;
     synchronized (conn) {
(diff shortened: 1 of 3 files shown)


franz1981:
@vietj This change here is not in conflict with https://github.com/eclipse-vertx/vert.x/pull/5143 given that it works on different synchronization(s) happening in a inefficient way, so, if you agree on the changes, could be safely                                                        merged.

Just for reference, it improves by ~5% few benchmarks using plaintext, so, not a game changer; but given that fixing the presence of pausing in Quarkus is NOT possible nor is possible to improve the inbound buffer to not create any array q, in case we end a request while still paused, this is the less destructive and simpler change to improve things I could think about.

Said that, sadly profilers can consistently lie and, in an experiment, I've tried removing `InboundBuffer::write` code just to check if it improves performance, and it seems not (!); instead its caller `io/vertx/core/http/impl/Http1xServerConnection.handleMessage` has been blame, which will be improved by  https://github.com/eclipse-vertx/vert.x/pull/5143

franz1981:
@vietj ping

franz1981:
@vietj how this look bud?

vietj:
this looks good, I need to review it.

can you have a quick look at https://github.com/eclipse-vertx/vert.x/pull/5164/files ?

vietj:
@franz1981 before proceeding with this, I think we can review and merge https://github.com/eclipse-vertx/vert.x/pull/5143/files which does remove synchronization use in `Http1xServerConnection`

src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
```diff
@@ -123,18 +123,18 @@ void setRequest(HttpRequest request) {
...
-      pending.handler(buffer -> {
-        if (buffer == InboundBuffer.END_SENTINEL) {
-          onEnd();
-        } else {
-          onData((Buffer) buffer);
-        }
-      });
-    }
+  private InboundBuffer createAndSetInboundBuffer() {
+    assert pending == null;
```
vietj, line 127: that change does not seem necessary for this PR

vietj: what is the benefit of changing this ? am I missing something ?

franz1981: Nope it is related, but probably worth more information to be understood..
In the rest of the method you can find

```java
  private InboundBuffer createAndSetInboundBuffer() {
    assert pending == null;
    InboundBuffer<Object> pending = new InboundBuffer<>(context, 8);
    this.pending = pending;
    pending.drainHandler(v -> conn.doResume());
    pending.handler(buffer -> {
      if (buffer == InboundBuffer.END_SENTINEL) {
        onEnd();
      } else {
        onData((Buffer) buffer);
      }
    });
    return pending;
  }
  ```
as you can see `pending` is using a local just created `InboundBuffer`, calling 2 synchronized methods:
- `pending.drainHandler`
- `pending.handler`

I've measured that these 2 synchronized methods would "merge" into a single acquire/release of monitor instead of 2, thanks to using a local variable (which is supposed to be stable, compared to an access field), see https://github.com/franz1981/java-puzzles/blob/8851ce838603a425f38400c2a0dee739f3ba6d58/src/main/java/red/hat/puzzles/concurrent/LockCoarsening.java#L194 which shows that it perform at twice the speed if compared to the existing one.

File: src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java [hunk 2]
```diff
@@ -346,15 +331,37 @@ public HttpServerRequest exceptionHandler(Handler<Throwable> handler) {
...
+      if (pending != null) {
+        pending.pause();
+      } else {
+        pending = InboundBuffer.createPaused(context, 8, pendingDrainHandler(), pendingHandler());
+      }
       return this;
     }
   }
 
+  private Handler<Object> pendingHandler() {
```
franz1981, line 343: These are to make sure just a single lambda type is generated - If I would inline them I'm not sure they'll be the same lambda type.

src/main/java/io/vertx/core/streams/impl/InboundBuffer.java
```diff
@@ -87,6 +86,27 @@ public InboundBuffer(Context context) {
...
+    return new InboundBuffer<>(context, highWaterMark, 0L, drainHandler, handler);
+  }
+
+  public static <E> InboundBuffer<E> createAndFetch(Context context, long highWaterMark, long demand, Handler<Void> drainHandler, Handler<E> handler) {
+    Objects.requireNonNull(drainHandler);
+    Objects.requireNonNull(handler);
+    checkPositiveAmount(demand);
+    InboundBuffer<E> inboundBuffer = new InboundBuffer<>(context, highWaterMark, Long.MAX_VALUE, drainHandler, handler);
+    if (inboundBuffer.emit(demand)) {
+      inboundBuffer.context.runOnContext(v -> inboundBuffer.drain());
```
franz1981, line 104: I forgot to use asyncDrain here .-.

franz1981:
@vietj I've addressed your comments and added some tests for the additional unsynchronized methods.

franz1981:
@vietj this is fine than? So we can bring it in the next quarkus release

franz1981:
Ping @vietj ? This is ok to merge? So I can verify on the next techempower round if helps?

PR #5150, opened 2024-03-12
Add option for whether semicolon is treated as normal character or not

### LINKED ISSUE #4588: Allow semicolon to be treated as normal character in query params
### Questions

Allow semicolon to be treated as normal character in query params

### Version

4.3.7

### Context

1. Client sends a request to an API e.g. http://host/someBasePath/myResource?transactionDetails=shopping;tesco&date=2013-01-01
2. Notice the `;` within 'shopping;tesco'
3. The API internally tries to fetch the query params from the `HttpServerRequest` i.e
   `httpServerRequest.getParam("transactionDetails")`
   Expectation is this would return `shopping;tesco` however it is only returning `shopping` instead

Semicolons since 2014 from W3C recommendations are now illegal parameter separators and only `&` is allowed.
There is a fix merged in netty -> https://github.com/netty/netty/issues/8855 (see QueryStringDecoder class)

However HttpServerRequest doesn't seem to allow a way to get query params with treating semicolons as normal.

It appears this is down to HttpUtils using a constructor of QueryStringDecoder which default to not treating semicolons as normal characters:
https://github.com/eclipse-vertx/vert.x/blob/1807f72a01c70bb9c189ee232ce635d15be0d726/src/main/java/io/vertx/core/http/impl/HttpUtils.java#L498

Should the above HttpUtils now use the new constructor of QueryStringDecoder and set semicolons to be treated as normal characters by default from vertx point of view? Alternatively some way for users of vertx to set this behaviour on/off?

### Extra

* Java 11
* Vertx 4.3.7

PragmaTwice:
#### Motivation:

Close #4588.

This PR adds an option `uriQuerySemicolonIsNormalChar` to `HttpServerOptions` to support treating `;` a normal character instead of query parameter seperator, since it has already been supported in netty `QueryStringDecoder`.

And the default behavior is still to treat `;` as a seperator to ensure compatibility.

References:
- https://github.com/netty/netty/pull/9701
- https://stackoverflow.com/questions/3481664/semicolon-as-url-query-separator

#### Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
@@ -87,6 +87,7 @@ public class Http1xServerRequest extends HttpServerRequestInternal implements io
   // Cache this for performance
   private Charset paramsCharset = StandardCharsets.UTF_8;
   private MultiMap params;
+  private boolean semicolonIsNormalCharInParams;
   private MultiMap headers;
   private String absoluteURI;
 
@@ -301,9 +302,10 @@ public class Http1xServerRequest extends HttpServerRequestInternal implements io
   }
 
   @Override
-  public MultiMap params() {
-    if (params == null) {
-      params = HttpUtils.params(uri(), paramsCharset);
+  public MultiMap params(boolean semicolonIsNormalChar) {
+    if (params == null || semicolonIsNormalChar != semicolonIsNormalCharInParams) {
+      params = HttpUtils.params(uri(), paramsCharset, semicolonIsNormalChar);
+      semicolonIsNormalCharInParams = semicolonIsNormalChar;
     }
     return params;
   }
--- a/src/main/java/io/vertx/core/http/impl/Http2ServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http2ServerRequest.java
@@ -59,6 +59,7 @@ public class Http2ServerRequest extends HttpServerRequestInternal implements Htt
   // Accessed on context thread
   private Charset paramsCharset = StandardCharsets.UTF_8;
   private MultiMap params;
+  private boolean semicolonIsNormalCharInParams;
   private String absoluteURI;
   private MultiMap attributes;
   private HttpEventHandler eventHandler;
@@ -357,10 +358,11 @@ public class Http2ServerRequest extends HttpServerRequestInternal implements Htt
     return paramsCharset.name();
   }
   @Override
-  public MultiMap params() {
+  public MultiMap params(boolean semicolonIsNormalChar) {
     synchronized (stream.conn) {
-      if (params == null) {
-        params = HttpUtils.params(uri(), paramsCharset);
+      if (params == null || semicolonIsNormalChar != semicolonIsNormalCharInParams) {
+        params = HttpUtils.params(uri(), paramsCharset, semicolonIsNormalChar);
+        semicolonIsNormalCharInParams = semicolonIsNormalChar;
       }
       return params;
     }
--- a/src/test/java/io/vertx/core/http/impl/HttpUtilsTest.java
+++ b/src/test/java/io/vertx/core/http/impl/HttpUtilsTest.java
@@ -8,12 +8,14 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
  */
-package io.vertx.core.http;
+package io.vertx.core.http.impl;
 
+import io.vertx.core.MultiMap;
 import io.vertx.core.http.impl.HttpUtils;
 import org.junit.Test;
 
 import java.net.URI;
+import java.nio.charset.Charset;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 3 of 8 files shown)


vietj:
I don't think we are going to make this an option as it is very specific to the API and not related to the behavior of the HTTP server itself

vietj:
Instead it should be a method parameter of `Http1xServerRequest#params()` method

vietj (changes requested):
Use a method parameter instead of an HTTP server option

PragmaTwice:
> Use a method parameter instead of an HTTP server option

Done according to your suggestions : )

Feel free to review again and comment.

vietj (changes requested):
since the created multimap is cached, we should also cache the delimiter and check the delimiter is the same when accessing again the multimap because this might be called multiple times with different values, when the delimiter is not the same then we should compute it again (and perhaps cache it).

src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
```diff
@@ -88,6 +88,7 @@ public class Http1xServerRequest extends HttpServerRequestInternal implements io
   // Cache this for performance
   private Charset paramsCharset = StandardCharsets.UTF_8;
   private MultiMap params;
+  private Boolean semicolonIsNormalCharInParams;
```
vietj, line 91: we can use a primitive `boolean` here since `params != null` implies `semicolonIsNormalCharInParams` has been set

src/main/java/io/vertx/core/http/impl/Http2ServerRequest.java
```diff
@@ -70,6 +70,7 @@ public class Http2ServerRequest extends HttpServerRequestInternal implements Htt
   // Accessed on context thread
   private Charset paramsCharset = StandardCharsets.UTF_8;
   private MultiMap params;
+  private Boolean semicolonIsNormalCharInParams;
```
vietj, line 73: we can use a primitive `boolean` here since `params != null` implies `semicolonIsNormalCharInParams` has been set

PragmaTwice: Done.

src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
PragmaTwice: Done.

PragmaTwice:
Fixed according to your suggestions : )

Feel free to review again and comment.

src/test/java/io/vertx/core/http/Http2ServerTest.java
```diff
@@ -530,6 +530,16 @@ public void testURI() throws Exception {
...
       assertEquals(Arrays.asList("bar_value_1", "bar_value_2"), params.getAll("bar"));
+
+      // test params()
+      MultiMap params1 = req.params(true);
+      MultiMap params2 = req.params(true);
+      assertTrue(params1 == params2); // got the cached value
+      MultiMap params3 = req.params(false);
+      assertTrue(params1 != params3); // cache refreshed
+      MultiMap params4 = req.params(false);
+      assertTrue(params3 == params4); // got the cached value
```
vietj, line 541: assertSame

vietj (changes requested):
This test only the new feature with HTTP/2, a new test should be added in HttpTest so it will be executed for for HTTP/2 and HTTP/1.

src/test/java/io/vertx/core/http/Http2ServerTest.java
PragmaTwice: done

src/test/java/io/vertx/core/http/Http2ServerTest.java
```diff
@@ -530,6 +530,16 @@ public void testURI() throws Exception {
       assertEquals(Collections.singletonList("foo_value"), params.getAll("foo"));
       assertEquals("bar_value_1", params.get("bar"));
       assertEquals(Arrays.asList("bar_value_1", "bar_value_2"), params.getAll("bar"));
+
+      // test params()
+      MultiMap params1 = req.params(true);
+      MultiMap params2 = req.params(true);
+      assertTrue(params1 == params2); // got the cached value
+      MultiMap params3 = req.params(false);
+      assertTrue(params1 != params3); // cache refreshed
```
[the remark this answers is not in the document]
PragmaTwice: done

src/test/java/io/vertx/core/http/Http2ServerTest.java
```diff
@@ -530,6 +530,16 @@ public void testURI() throws Exception {
       assertEquals(Collections.singletonList("foo_value"), params.getAll("foo"));
       assertEquals("bar_value_1", params.get("bar"));
       assertEquals(Arrays.asList("bar_value_1", "bar_value_2"), params.getAll("bar"));
+
+      // test params()
+      MultiMap params1 = req.params(true);
+      MultiMap params2 = req.params(true);
+      assertTrue(params1 == params2); // got the cached value
```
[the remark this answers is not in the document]
PragmaTwice: done

PragmaTwice:
Fixed according to your suggestions : )

Feel free to review again and comment. cc @vietj

src/test/java/io/vertx/core/http/impl/HttpUtilsTest.java
```diff
@@ -229,4 +231,14 @@ private void check(String base, String ref, String expected) throws Exception {
     URI uri = HttpUtils.resolveURIReference(base, ref);
     assertEquals(expected, uri.getPath());
   }
+
+  @Test
+  public void testParams() {
```
vietj, line 236: this should also be tested in `HttpTest`

PragmaTwice: done

PragmaTwice:
Fixed according to your suggestions : )

Feel free to review again and comment. cc @vietj

vietj (changes requested):
Can you squash everything in a single commit to facilitate back-port to 4.x ?

PragmaTwice:
> Can you squash everything in a single commit to facilitate back-port to 4.x ?

Done. Please check.

vietj:
thanks I'll merge that soon and backport

PR #5157, opened 2024-03-19
More load balancing policies

### LINKED ISSUE #5154: More load balancing policies
In addition to the existing (round-robin and least request) policies in `io.vertx.core.loadbalancing.LoadBalancer`, we should have:

- [x] random
- [x] power of two choices

Possible contributor @machine424

machine424:
Fixes #5154

Add Random and Power of 2 choices policies.
Also, adjust the Round Robin policy to handle a special case.

### DIFF
--- a/src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
+++ b/src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
@@ -12,11 +12,12 @@ package io.vertx.core.loadbalancing;
 
 import io.vertx.core.spi.loadbalancing.DefaultEndpointMetrics;
 import io.vertx.core.spi.loadbalancing.Endpoint;
-import io.vertx.core.spi.loadbalancing.EndpointMetrics;
 import io.vertx.core.spi.loadbalancing.EndpointSelector;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * A load balancer.
@@ -34,6 +35,9 @@ public interface LoadBalancer {
   LoadBalancer ROUND_ROBIN = () -> {
     AtomicInteger idx = new AtomicInteger();
     return (EndpointSelector) endpoints -> {
+      if (endpoints.isEmpty()) {
+        return -1;
+      }
       int next = idx.getAndIncrement();
       return next % endpoints.size();
     };
@@ -58,6 +62,46 @@ public interface LoadBalancer {
       }
       return selected;
     }
+    @Override
+    public <E> Endpoint<E> endpointOf(E endpoint) {
+      return new DefaultEndpointMetrics<>(endpoint);
+    }
+  };
+
+  /**
+   * Random load balancer.
+   */
+  LoadBalancer RANDOM = () -> {
+    return (EndpointSelector) endpoints -> {
+      if (endpoints.isEmpty()) {
+        return -1;
+      }
+      return ThreadLocalRandom.current().nextInt(endpoints.size());
+    };
+  };
+
+  /**
+   * Power of two choices load balancer.
+   */
+  LoadBalancer POWER_OF_TWO_CHOICES = () -> new EndpointSelector() {
+    @Override
+    public int selectEndpoint(List<? extends Endpoint<?>> endpoints) {
+      if (endpoints.isEmpty()) {
+        return -1;
+      } else if (endpoints.size() == 1) {
+        return 0;
+      }
+      int i1 = ThreadLocalRandom.current().nextInt(endpoints.size());
+      int i2 = ThreadLocalRandom.current().nextInt(endpoints.size());
+      while (i2 == i1) {
+        i2 = ThreadLocalRandom.current().nextInt(endpoints.size());
+      }
+      if (((DefaultEndpointMetrics<?>) endpoints.get(i1)).numberOfInflightRequests() < ((DefaultEndpointMetrics<?>) endpoints.get(i2)).numberOfInflightRequests()) {
+        return i1;
+      }
+      return i2;
+    }
+
     @Override
     public <E> Endpoint<E> endpointOf(E endpoint) {
       return new DefaultEndpointMetrics<>(endpoint);
--- a/src/test/java/io/vertx/core/loadbalancing/LoadBalancingTest.java
+++ b/src/test/java/io/vertx/core/loadbalancing/LoadBalancingTest.java
@@ -19,6 +19,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 public class LoadBalancingTest {
 
@@ -52,4 +53,41 @@ public class LoadBalancingTest {
     }
     assertEquals(2, selector.selectEndpoint(metrics));
   }
+
+  @Test
+  public void testRandom() throws Exception {
+    EndpointSelector selector = LoadBalancer.RANDOM.selector();
+    List<Endpoint<?>> metrics = new ArrayList<>();
+    int num = 6;
+    for (int i = 0;i < num;i++) {
+      metrics.add(selector.endpointOf(null));
+    }
+    for (int i = 0; i < 1000; i++) {
+      int selectedIndex = selector.selectEndpoint(metrics);
+      assertTrue(selectedIndex >= 0 && selectedIndex < metrics.size());
+    }
+  }
+
+  @Test
+  public void testPowerOfTwoChoices() throws Exception {
+    EndpointSelector selector = LoadBalancer.POWER_OF_TWO_CHOICES.selector();
+    for (int i = 0; i < 1000; i++) {
+      Endpoint<?> e1 = selector.endpointOf(null);
+      Endpoint<?> e2 = selector.endpointOf(null);
+      List<Endpoint<?>> metrics = Arrays.asList(e1, e2);
+      e2.metrics().initiateRequest();
+      assertEquals(0, selector.selectEndpoint(metrics));
+    }
+
+    List<Endpoint<?>> metrics = new ArrayList<>();
+    int num = 6;
+    for (int i = 0;i < num;i++) {
+      metrics.add(selector.endpointOf(null));
+    }
+    for (int i = 0; i < 1000; i++) {
+      int selectedIndex = selector.selectEndpoint(metrics);
+      assertTrue(selectedIndex >= 0 && selectedIndex < metrics.size());
+    }
+  }
 }
+
--- a/src/main/java/io/vertx/core/spi/loadbalancing/EndpointSelector.java
+++ b/src/main/java/io/vertx/core/spi/loadbalancing/EndpointSelector.java
@@ -23,7 +23,7 @@ public interface EndpointSelector {
    * Select an endpoint among the provided list.
    *
    * @param endpoints the endpoint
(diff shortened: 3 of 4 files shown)


src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
```diff
@@ -34,6 +36,9 @@ public interface LoadBalancer {
   LoadBalancer ROUND_ROBIN = () -> {
     AtomicInteger idx = new AtomicInteger();
     return (EndpointSelector) endpoints -> {
+      if (endpoints.isEmpty()) {
```
machine424, line 39: As stated here https://github.com/eclipse-vertx/vert.x/blob/07daa4eefac7a5dc8a2d03c368e2591588e8fe1e/src/main/java/io/vertx/core/spi/loadbalancing/EndpointSelector.java#L26

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
```diff
@@ -64,6 +69,67 @@ public <E> Endpoint<E> endpointOf(E endpoint) {
     }
   };
 
+  /**
+   * Random load balancer.
+   */
+  // LoadBalancer RANDOM = () -> (EndpointSelector) endpoints -> new Random().nextInt(endpoints.size());
+  // ThreadLocalRandom? contention, seed?
```
machine424, line 76: Do you think ThreadLocalRandom should be used?

src/test/java/io/vertx/core/loadbalancing/LoadBalancingTest.java
```diff
@@ -35,6 +36,11 @@ public void testRoundRobin() throws Exception {
     assertEquals(0, selector.selectEndpoint(metrics));
     assertEquals(1, selector.selectEndpoint(metrics));
     assertEquals(2, selector.selectEndpoint(metrics));
+
+    metrics = new ArrayList<>();
```
machine424, line 40: I don't know if there is a way to factorize these special cases checks for all policies.

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
```diff
@@ -64,6 +69,66 @@ public <E> Endpoint<E> endpointOf(E endpoint) {
...
+  };
+
+  /**
+   * Power of two choices load balancer.
+   */
+  // BEST_OF_TWO? P2C?
+  LoadBalancer POWER_OF_TWO_CHOICES = () -> new EndpointSelector() {
+    final Random random = new Random();
+
+    private List<Integer> sample(int length) {
```
machine424, line 93: I implemented Robert Floyd's here, I couldn't find any existing util for this. Did I miss sth?
Should we move it?

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
```diff
@@ -64,6 +69,66 @@ public <E> Endpoint<E> endpointOf(E endpoint) {
...
+      }
+      return random.nextInt(endpoints.size());
+    };
+  };
+
+  /**
+   * Power of two choices load balancer.
+   */
+  // BEST_OF_TWO? P2C?
+  LoadBalancer POWER_OF_TWO_CHOICES = () -> new EndpointSelector() {
```
machine424, line 90: wdyt about the naming?

tsegismont:
Please verify you have used the email associated to your Eclipse account and that you have signed-off your commit

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
tsegismont: Power of two choices is the generally accepted name, chosen after this paper: https://www.nginx.com/blog/nginx-power-of-two-choices-load-balancing-algorithm/

tsegismont:
I've changed the name of the PR because we'll use it to add all policies listed in #5154

src/test/java/io/vertx/core/loadbalancing/LoadBalancingTest.java
tsegismont: Yes, you could use a separate class with parameterized test junit runner

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
tsegismont: Nice catch

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
tsegismont: I think we don't have any guarantee about the thread invoking the selector, so it would be safer to use `ThreadLocalRandom`

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
```diff
@@ -26,6 +27,7 @@
  *
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
+@FunctionalInterface
```
vietj, line 30: I don't think this is what we want. Declaring functionnal interface has impact on the API and its evolution, it's usually good when it is a user API. This is not a user API (user shall not implement it).

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
tsegismont: I think you could simplify this, take a look at the implementation in Stork https://github.com/smallrye/smallrye-stork/blob/09c4b19c14993dc164a799a1c84f1dc5e9cb695c/load-balancer/power-of-two-choices/src/main/java/io/smallrye/stork/loadbalancer/poweroftwochoices/PowerOfTwoChoicesLoadBalancer.java#L34

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
machine424: I think I accepted the IDE's suggestion by mistake, will remove, thanks.

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
machine424: (okk, just in case it was long for you)

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
machine424: I'm afraid they're not implementing the algorithm as is; there is no guarantee that the two picked instances are different. I don't know the consequences of that.
But if you want to keep it "simple", we can do the same here.

src/test/java/io/vertx/core/loadbalancing/LoadBalancingTest.java
machine424: I'll give that a try, thanks.

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
tsegismont: yes, Stork's impl might not inspect two different endpoints every time. But then you could use a `do/while` to select a random int for the second endpoint until it's different than the first one.

machine424: Was my first choice, not as efficient as Floyd's, but it's simpler.

machine424:
> Please verify you have used the email associated to your Eclipse account and that you have signed-off your commit

I re-signed the ECA, maybe it needs some time to sync.

src/main/java/io/vertx/core/loadbalancing/LoadBalancer.java
tsegismont: Perhaps, as a follow-up PR we can examine how one implementation performs compared to the other, using a JMH test.

src/main/java/io/vertx/core/spi/loadbalancing/Endpoint.java
```diff
@@ -29,4 +29,6 @@ public interface Endpoint<E> {
    */
   EndpointMetrics<?> metrics();
 
+
```
machine424, line 32: remove this.

src/test/java/io/vertx/core/loadbalancing/PreferenceAwareLoadBalancingCornerCasesTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public class PreferenceAwareLoadBalancingCornerCasesTest {
```
machine424, line 17: maybe we should merge this with `LoadBalancingCornerCasesTest`?

src/main/java/io/vertx/core/spi/loadbalancing/PreferenceAwareEndpointSelector.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.List;
+
+public interface PreferenceAwareEndpointSelector extends EndpointSelectorCommons {
+
+  /**
+   * Select an endpoint among the provided list taking an attribute into account.
+   *
+   * @param endpoints the endpoint
+   * @param attribute the attribute TODO
```
machine424, line 11: I think we can do better for the naming/types?
also for the `PreferenceAwarexxx`

src/main/java/io/vertx/core/spi/loadbalancing/PreferenceAwareEndpointSelector.java
```diff
@@ -0,0 +1,16 @@
...
+   * @return the selected index or {@code -1} if selection could not be achieved
+   */
+  int selectEndpoint(List<? extends Endpoint<?>> endpoints, String attribute);
```
machine424, line 14: or should we allow passing this attr/preference to all policies? it can be confusing.

src/main/java/io/vertx/core/loadbalancing/PreferenceAwareLoadBalancer.java
```diff
@@ -0,0 +1,24 @@
+package io.vertx.core.loadbalancing;
+
+import io.vertx.core.spi.loadbalancing.PreferenceAwareEndpointSelector;
+
+public interface PreferenceAwareLoadBalancer {
+
+  /**
+   * TOO SIMPLE Hashing load balancer XXX. TODO
```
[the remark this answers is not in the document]
tsegismont: Honestly, I need to put more thought into this. As I see it, the problem is the following:

- we have for input either a value provided by the user or computed based on one or more request properties (e.g. client IP or, in the case of sticky sessions, the value of the session id).
- given an input value, we need to be able to stick to the same endpoint (even if other endpoints join or leave)

I'm not sure how to do this right now

tsegismont (commented):
@machine424 thank you for this contribution

Would you agree to finish this PR with random and power of two choices algorithms only?

I think we all need to put more though into the problem of hashing/sticky session load balancing. I could create a separate issue and one of us would address in a follow-up PR.

machine424:
> Would you agree to finish this PR with random and power of two choices algorithms only?

> I think we all need to put more though into the problem of hashing/sticky session load balancing. I could create a separate issue and one of us would address in a follow-up PR.

Fair enough, I agree with you.`
Removed the last commit, tell me if anything else needs to be added/adjusted.

tsegismont (approved):
Great job @machine424 ! Thank you very much !

PR #5153, opened 2024-03-13
Refactor HttpUtils#absoluteURI

tsegismont:
This is a fix for #2797 and https://github.com/eclipse-vertx/vertx-tracing/issues/74

A proposal was made originally by @aruis but it did not address all possible cases, as described in https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html

This PR covers all cases and adds tests for them.

PR #5158, opened 2024-03-20
Refactor HttpUtils#absoluteURI

### LINKED ISSUE #2797: The uri() method and the absoluteURI() method in HttpServerRequestImpl are inconsistent in judging the legality of uri.
Vert.x 3.6.2

This is a very simple server code
```java
        vertx.createHttpServer().requestHandler( req -> {
            System.out.println( req.uri());
            System.out.println( req.absoluteURI());

            req.response()
                    .putHeader("content-type", "text/plain")
                    .end("Hello from Vert.x!");
        }).listen(8080);
    }
```
When I request the address `http://127.0.0.1:8080/test?a={1}` in the browser, the uri() method will not report an error, but the absoluteURI() print error will return a null value. This inconsistent performance is not reasonable.
The contents of the error are as follows:
```
/test?a={1}
17:26:41.548 [vert.x-eventloop-thread-0] ERROR io.vertx.core.http.impl.HttpServerRequestImpl - Failed to create abs uri
java.net.URISyntaxException: Illegal character in query at index 8: /test?a={1}
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.checkChars(URI.java:3021)
	at java.net.URI$Parser.parseHierarchical(URI.java:3111)
	at java.net.URI$Parser.parse(URI.java:3063)
	at java.net.URI.<init>(URI.java:588)
	at io.vertx.core.http.impl.HttpUtils.absoluteURI(HttpUtils.java:362)
	at io.vertx.core.http.impl.HttpServerRequestImpl.absoluteURI(HttpServerRequestImpl.java:379)
...
...
...
null
```
The first line is the result returned by the uri() method.
The last line is the result returned by the absoluteURI() method.

tsegismont:
Fixes #2797, fixes eclipse-vertx/vertx-tracing#74

It doesn't seem necessary to validate the request uri using the JDK's URI parser. Instead, we can make a few checks on the format and adapt the result.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
@@ -12,8 +12,8 @@
 package io.vertx.core.http.impl;
 
 import io.netty.handler.codec.DecoderResult;
-import io.netty.handler.codec.http.*;
 import io.netty.handler.codec.http.HttpHeaders;
+import io.netty.handler.codec.http.*;
 import io.netty.handler.codec.http.multipart.Attribute;
 import io.netty.handler.codec.http.multipart.HttpPostRequestDecoder;
 import io.netty.handler.codec.http.multipart.InterfaceHttpData;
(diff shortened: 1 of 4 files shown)


tsegismont:
This is a backport of #5153 for `4.x`

PR #5169, opened 2024-03-26
Vertx should be its virtual thread factory at runtime with correct reflection calls

vietj:
A Vertx uses cached reflection to build the virtual thread factory in order to remain compatible with Java 8. When running in a native image Vert.x will assume that the runtime might not have virtual thread incorrectly.

Defer the virtual thread factory availability when a Vertx instance is created, in addition declares that virtual thread factory API should not be cached so the Vertx instance is able to build the virtual thread factory correctly.


vietj:
@cescoffier can you have a quick look ? this solve the runtime virtual thread factory detection problem

cescoffier (commented):
I need to have a closer look, as it will affects how we initialize Vert.x in native image in Quarkus.

PR #5177, opened 2024-04-03
Inbound message composite bytebuf

vietj:


franz1981 (commented):
Composites buffer are not useful for accumulation, usually, but for flushing.
I suggest to accumulate buffers into a queue, which can try hard to reuse what's left of the last appended buffer and than flush it as composite, if needed, instead.

Check what I have done for AppendBuffer.
Sadly the overhead of composite as generic buffer holder is not worthy, but is useful to traverse just one (using a single listener too) the pipeline, on flush, instead.
The good thing is that, if the composite buffer is made of just direct buffers and it is less then the max iovec supported by the transport, it could directly become a single posix writev. And that's why batching js beneficial

PR #5180, opened 2024-04-05
Provide a better event-loop to context message transfer queue

vietj:
The Vert.x event-loop thread to context thread message transfer relies on the `InboundBuffer` implementation.

`InboundBuffer` performs the add/dispatch in the same method call assuming that the same thread actually handles the dispatch of the message, forcing the dispatch to then schedule the message delivery on the context thread.

## Inbound read queue

The `InboundReadQueue` design actually split this operation in two separate operations providing control to the caller of the message dispatch.

- the `add` operation queues a message and let the producer knows whether messages should be drained from the queue (e.g. if a drain operation is already in progress, then there is no need to schedule another drain).
- the `drain` operation let consumer consume messages from the queue until needed

The event-loop to context message dispatch then becomes:

1. add the message to the queue
2. ping the context thread to drain the queue

### Event-loop thread dispatch

In this use case, the `InboundReadQueue` assumes the same thread is producing/consuming messages and therefore no memory visibility is actually required. In practice the event-loop add to the queue and then drain delivers the message to the connection.

### Generic context thread dispatch

In this use case, the queue will be drained by the context thread and an SPSC + volatile WIP is used. This behaviour also can optimise the message delivery since we don't need anymore a message received ⇒ scheduling a task, the event-loop thread and the context thread can work in an SPSC consumer design.

This use case holds for:

- another event-loop thread (pooled HTTP client connection)
- worker thread
- virtual thread

## Back pressure

The context thread message deals controls the back-pressure and cannot control the inbound channel back-pressure without races. Like Like the `OutboundWriteQueue`, the `InboundReadQueue` relies on an internal buffer and the queue acts as an intermediary for back-pressure.

- the producer `add` operation signals when the queue becomes un-writable (e.g. it turns off Netty auto-read)
- the consumer consumes messages, when messages are consumed it is responsible to signal the queue is writable again

## Inbound message queue

The `InboundMessageQueue` is a construct integrating the `InboundReadQueue` with the Vert.x context and a demand counter (`ReadStream`). The consumer deals with `pause`/`resume`/`fetch` to control demand, the producer implements the message flow pause/resume. This queue is the `InboundBuffer` replacement.

## ConnectionBase message flow changes

`doPause`/`doResume` of `ConnectionBase` has been rewritten to be strict concerning the delivery of messages. Previously these operations were turning on/off channel auto read, however this was not controlling the reads in progress. This changes these operations to control a `paused` flag and buffer any messages to be handled by the connection base when paused.

This change has been made to let the `Http1xServerConnection` control precisely the message flow when processing an HTTP pipelined request content. Previously the content was poured in the pending queue of the pipelined request which is complicated to deliver when the pipelined request is processed and requires some hacks with respect to the request demand. After this change, `Http1xServerConnection` can immediately pause the connection after receiving a pipelined request and have the guarantee that no message will be processed until it is resumed. When the pipelined request is processed, the connection is resumed and will deliver the messages with the regular `handleMessage` flow.

 ## Performance

### Plaintext

Screenshot 2024-04-16 at 17 11 17

- master : 4.5.7 baseline
- 5.0.0 : 5.0.0 baseline
- 5.0.0-irq : this PR branch
- 5.0.0-irq-nv : this PR branch using non volatile but synchronised inbound buffer demand (relying on biased locking)

### Microbenchmark

```
> mvn clean package -DskipTests -Pbenchmarks
> java -jar target/vertx-core-5.0.0-SNAPSHOT-benchmarks.jar HttpServerHandlerBenchmark
```

#### InboundMessageQueue (synchronised long)

```
Benchmark                                      Mode  Cnt     Score    Error   Units
HttpServerHandlerBenchmark.netty              thrpt   10  2398.380 ± 19.954  ops/ms
HttpServerHandlerBenchmark.vertx              thrpt   10  1769.388 ±  6.817  ops/ms
HttpServerHandlerBenchmark.vertxOpt           thrpt   10  2241.688 ±  4.130  ops/ms
HttpServerHandlerBenchmark.vertxOptMetricsOn  thrpt   10  1906.300 ± 28.674  ops/ms
```

#### InboundMessageQueue (volatile long)

```
Benchmark                                      Mode  Cnt     Score    Error   Units
HttpServerHandlerBenchmark.netty              thrpt   10  2445.317 ± 17.642  ops/ms
HttpServerHandlerBenchmark.vertx              thrpt   10  1976.744 ±  7.916  ops/ms
HttpServerHandlerBenchmark.vertxOpt           thrpt   10  1951.065 ±  4.861  ops/ms
HttpServerHandlerBenchmark.vertxOptMetricsOn  thrpt   10  2192.769 ± 26.459  ops/ms
```

#### InboundBuffer (master)

```
Benchmark                                      Mode  Cnt     Score    Error   Units
HttpServerHandlerBenchmark.netty              thrpt   10  2423.540 ± 35.876  ops/ms
HttpServerHandlerBenchmark.vertx              thrpt   10  1860.064 ± 16.155  ops/ms
HttpServerHandlerBenchmark.vertxOpt           thrpt   10  2209.176 ± 11.710  ops/ms
HttpServerHandlerBenchmark.vertxOptMetricsOn  thrpt   10  1896.875 ± 14.296  ops/ms
```

### InboundBuffer (4.x)

```
Benchmark                                      Mode  Cnt     Score    Error   Units
HttpServerHandlerBenchmark.netty              thrpt   10  2368.750 ± 30.001  ops/ms
HttpServerHandlerBenchmark.vertx              thrpt   10  1892.064 ± 21.788  ops/ms
HttpServerHandlerBenchmark.vertxOpt           thrpt   10  2097.548 ±  9.652  ops/ms
HttpServerHandlerBenchmark.vertxOptMetricsOn  thrpt   10  2187.718 ± 10.779  ops/ms
```

## Integration with Vert.x Core

This integrates with

- HTTP/1.1 client/server chunk
- NetSocket
- WebSocket

Past attempts to solve this:
- https://github.com/eclipse-vertx/vert.x/pull/5177
- https://github.com/eclipse-vertx/vert.x/pull/5164

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/InboundMessageQueue.java
+++ b/src/main/java/io/vertx/core/net/impl/InboundMessageQueue.java
@@ -1,3 +1,13 @@
+/*
+ * Copyright (c) 2011-2024 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
 package io.vertx.core.net.impl;
 
 import io.vertx.core.impl.ContextInternal;
--- a/src/main/java/io/vertx/core/net/impl/InboundMessageQueue.java
+++ b/src/main/java/io/vertx/core/net/impl/InboundMessageQueue.java
@@ -41,6 +41,18 @@ public class InboundMessageQueue<M> implements Predicate<M>, Runnable {
     this.eventLoop = eventLoop;
   }
 
+  public InboundMessageQueue(EventLoop eventLoop, ContextInternal context, int lowWaterMark, int highWaterMark) {
+    InboundReadQueue.Factory readQueueFactory;
+    if (context.threadingModel() == ThreadingModel.EVENT_LOOP && context.nettyEventLoop() == eventLoop) {
+      readQueueFactory = InboundReadQueue.SINGLE_THREADED;
+    } else {
+      readQueueFactory = InboundReadQueue.SPSC;
+    }
+    this.readQueue = readQueueFactory.create(this, lowWaterMark, highWaterMark);
+    this.context = context;
+    this.eventLoop = eventLoop;
+  }
+
   @Override
   public boolean test(M msg) {
     while (true) {
@@ -81,7 +93,7 @@ public class InboundMessageQueue<M> implements Predicate<M>, Runnable {
    * @param msg the message
    * @return {@code true} when a {@link #drain()} should be called.
    */
-  public boolean add(M msg) {
+  public final boolean add(M msg) {
     assert eventLoop.inEventLoop();
     int res = readQueue.add(msg);
     if ((res & InboundReadQueue.QUEUE_UNWRITABLE_MASK) != 0) {
@@ -93,10 +105,26 @@ public class InboundMessageQueue<M> implements Predicate<M>, Runnable {
   /**
    * {@link #add(Object)} + {@link #drain()}.
    *
+   * @param messages the messages
+   */
+  public final void write(Iterable<M> messages) {
+    boolean f = false;
+    for (M msg : messages) {
+      f |= add(msg);
+    }
+    if (f) {
+      drain();
+    }
+  }
+
+  /**
+   * {@link #add(Object)} + {@link #drain()}.
+   *
    * @param msg the message
    */
-  public void write(M msg) {
-    if (add(msg)) {
+  public final void write(M msg) {
+    boolean f = add(msg);
+    if (f) {
       drain();
     }
   }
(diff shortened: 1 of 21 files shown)


vietj:
@franz1981 @jponge

vietj:
This is now ready for review @franz1981 @jponge @cescoffier and anyone else that wants to review this

jponge:
(still on my todo)

vietj:
@jponge ok

src/main/java/io/vertx/core/net/impl/InboundMessageQueue.java
```diff
@@ -0,0 +1,206 @@
...
+    DEMAND_UPDATER.set(this, 0L);
+  }
+
+  /**
+   * Add {@code amount} to the current demand.
+   *
+   * @param amount the number of message to consume
+   */
+  public final void fetch(long amount) {
+    if (amount < 0L) {
```
jponge, line 187: You might want to handle `0` as well and reject it.

vietj: I think it's fine to tolerate 0 is a no-op, any reactive-streams implementation that uses it will never pass 0

Issue #5185, opened 2024-04-12
JsonObject: The equals method fails when comparing integers with floats.


ProgMaq:
### Version

4.5.7

### Context

Due to a modification of the equals method made 5 years ago, given an integer (2) and a float (2.3), this condition will not be satisfied "if (n1.doubleValue() == n2.doubleValue())...", so it will advance to the next condition which will truncate the value "if (n1.longValue() == n2.longValue())...", returning an erratic "true".

In other libraries like GSON, this same case returns "false".

### Do you have a reproducer?

```
  public static void main(String[] args) {
    final JsonObject f = new JsonObject("{\"number\": 2.3}");
    final JsonObject i = new JsonObject("{\"number\": 2}");
    System.out.println(f.equals(i)); // print true
  }
```

### Extra

```

      // special case for numbers
      if (thisValue instanceof Number && otherValue instanceof Number && thisValue.getClass() != otherValue.getClass()) {
        Number n1 = (Number) thisValue;
        Number n2 = (Number) otherValue;
        // floating point values
        if (thisValue instanceof Float || thisValue instanceof Double || otherValue instanceof Float || otherValue instanceof Double) {
          // compare as floating point double
          if (n1.doubleValue() == n2.doubleValue()) {
            // same value check the next entry
            continue;
          }
        }
        if (thisValue instanceof Integer || thisValue instanceof Long || otherValue instanceof Integer || otherValue instanceof Long) {
          // compare as integer long
          if (n1.longValue() == n2.longValue()) {
            // same value check the next entry
            continue;
          }
        }
      }

```

tsegismont:
Thanks for the report, we'll look into it

PR #5192, opened 2024-04-27
fix-JsonObject#equals-method-fails-when-comparing-integers-with-floats

sai-manohar-veerubhotla:
Motivation:
The motivation is to fix the bug: https://github.com/eclipse-vertx/vert.x/issues/5185

Context:
The current implementation of JsonObject#equals fails to equate numeric types correctly, leading to false comparisons:
1. `2f` is not considered equal to `2`.
2. `2D` is not considered equal to `2`.

Changes:
This PR updates JsonObject#equals to correctly handle comparisons between different numeric types (float, double, intege, long), ensuring logical equivalence where expected. Additionally, new assertions have been added to JsonObjectTest#testNumberEquality to validate these changes.

This PR tries to fix exactly the same by modifying the JsonObject#equals. Adds new assertions to JsonObjectTest#testNumberEquality.

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/src/main/java/io/vertx/core/json/JsonObject.java
@@ -584,7 +584,7 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
 
   /**
    * Get the binary value with the specified key.
-   *
+   * <p>
    * JSON itself has no notion of a binary, this extension complies to the RFC-7493, so this method assumes there is a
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -584,7 +584,7 @@ public JsonArray getJsonArray(String key) {
 
   /**
    * Get the binary value with the specified key.
-   *
```
vietj, line 587: no need for this change

Issue #5204, opened 2024-05-16
Vert.x does not respond appropriately for HTTP/1.1 request containing no Host header


wmorrison365fr:
### Version
4.5.7

### Context

An error occurred while processing an HTTP/1.1 request with no Host header. The error occurred because we didn't expect the `authority()` to be non-null (and hadn't properly checked):
```
java.lang.NullPointerException: Cannot invoke "io.vertx.rxjava3.core.net.HostAndPort.host()" because the return value of "io.vertx.rxjava3.core.http.HttpServerRequest.authority()" is null
```

We've fixed this in our application, but were surprised that Vert.x doesn't handle validation of an HTTP/1.1 request with no Host header, especially as the spec [RFC9112](https://datatracker.ietf.org/doc/html/rfc9112#section-3.2) indicates that servers should respond with a 400 BadRequest response:

> A client MUST send a Host header field ([Section 7.2](https://www.rfc-editor.org/rfc/rfc9110#section-7.2) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]) in all HTTP/1.1 request messages. If the target URI includes an authority component, then a client MUST send a field value for Host that is identical to that authority component, excluding any userinfo subcomponent and its "@" delimiter ([Section 4.2](https://www.rfc-editor.org/rfc/rfc9110#section-4.2) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]). If the authority component is missing or undefined for the target URI, then a client MUST send a Host header field with an empty field value.
>
> A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that lacks a Host header field and to any request message that contains more than one Host header field line or a Host header field with an invalid field value.

### Do you have a reproducer?

Not at the moment but can do so if 

### Steps to reproduce

We used telnet to fire a request with no Host. Note though that our application reads the `authority()` (trying to align HTTP/1.1 and HTTP/2 requests):
```
$ telnet ig.example.com 8083
Trying ::1...
Connected to localhost.
Escape character is '^]'.
GET /hello HTTP/1.1

HTTP/1.1 500 Internal Server Error
Content-Length: 0
```

### Extra

[...]

PR #5205, opened 2024-05-17
5204 reject missing host

NilsRenaud:
An HTTP/1.1 message MUST be rejected with an HTTP 400 error code when no Host header is present. See https://datatracker.ietf.org/doc/html/rfc9112#section-3.2

See https://github.com/eclipse-vertx/vert.x/issues/5204 for details.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
@@ -48,6 +48,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Objects;
 import java.util.Set;
 
+import static io.netty.handler.codec.DecoderResult.failure;
 import static io.vertx.core.spi.metrics.Metrics.METRICS_ENABLED;
 
 /**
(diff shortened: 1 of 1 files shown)


NilsRenaud:
I've not added any tests since I'm not really sure it's the right place to have this check.
Maybe we could add a new place such as `verifyRequest()` to handle such things.
Any thoughts ?

vietj:
https://github.com/vert-x3/vertx-web/pull/2612

NilsRenaud:
Is there a reason why it's handled in vert.x web rather than Vert.x core ?

vietj:
a couple of reasons:

- most users are using vertx-web, vertx-web already does similar checks, e.g. sending an error on a request with a path not starting with '/' (https://github.com/vert-x3/vertx-web/commit/a330de8d205e2f2cab7752d464d5b317620cce7c#diff-e57da13e074e54a5eef739d72b499b54edf338858495d434f2a991dc229babb2R83)
- in vertx-core we want to provide the ability to have a more generic HTTP server than can handle such things (e.g.  it can be handy for testing)

We might revisit that later I think.

NilsRenaud:
Ok, though I would love to have a flag like `.permitInvalidInput(boolean)` to have 2 versions of Vert.x core:
- `true`: for testing purpose
- `false`: in production (default)

vietj:
I think instead we should try in vertx-web to move this code to a validation handler that would be executed first to make things more clear

PR #5215, opened 2024-05-30
Support duplicated context duplication

vietj:
Duplicating a duplicated context is supported but the semantic of the actual locals is not defined.

This update the duplicated context duplication by doing a copy of each local in the duplicated duplicate. This introduce a duplicator for each local that is responsible for copying the object when it is not null.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/ContextImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/ContextImpl.java
@@ -38,7 +38,7 @@ public final class ContextImpl extends ContextBase implements ContextInternal {
 
   static final boolean DISABLE_TIMINGS = SysProps.DISABLE_CONTEXT_TIMINGS.getBoolean();
 
-  private final VertxInternal owner;
+  private final VertxImpl owner;
   private final JsonObject config;
   private final DeploymentContext deployment;
   private final CloseFuture closeFuture;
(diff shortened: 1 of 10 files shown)


vietj:
@franz1981 @cescoffier  please have a look

franz1981:
This can be related (and better used at) https://github.com/quarkusio/quarkus/pull/40725

can you checkout my branch there and show how the changes there will translate with the new API, bud? I'm travelling, but I would like, before it got merged, to make sure the semantic and behaviours won't change too much and instead we save some code by leveraging this new feature (which is super cool!)

vietj:
for now this is a tentative PR, we can discuss next week @franz1981 I think

franz1981:
Yep, let's sync once I'm back, thanks @vietj :pray:

src/main/java/io/vertx/core/impl/ContextLocalImpl.java
```diff
@@ -12,18 +12,31 @@
...
+    return LocalSeq.add(idx -> new ContextLocalImpl<>(idx, type, duplicator));
   }
 
-  public ContextLocalImpl() {
-    this.index = LocalSeq.next();
+  public static final UnaryOperator<?> IDENTITY = UnaryOperator.identity();
+
+  final int index;
+  final Class<T> type;
+  final  Function<T, ? extends T> duplicator;
```
franz1981, line 35: `valueDuplicator`

vietj:
See https://github.com/eclipse-vertx/vert.x/issues/5437

PR #5242, opened 2024-06-23
Feature/dns over https

imz87:
**Motivation**
The increasing demand for enhanced privacy and security on the internet has led to the adoption of DNS-over-HTTPS (DoH) as a standard protocol for DNS resolution. DoH ensures that DNS queries are encrypted and protected from eavesdropping and tampering, providing a significant improvement over traditional DNS methods.

Currently, Vert.x lacks support for DoH, which limits its ability to provide secure DNS resolution for applications. By adding DoH support to Vert.x, we can offer developers the ability to leverage encrypted DNS queries, aligning with modern security practices and enhancing the overall privacy of applications built on Vert.x.

**Conformance**
I have signed the Eclipse Contributor Agreement as required. I have also ensured that all code changes adhere to the Vert.x code style guidelines, as outlined in the Vert.x Code Style Guidelines.


vietj:
thank you @imz87 can you make a PR for the master branch of Vert.x ? the implementation of the DNS client has changed and I would like to be sure we can continue to support this feature in master before merging this branch.

zekronium:
I see that the `DnsClient`, just like in `4.x`, is not used for internal DNS resolution in `5.x` too, where `AddressResolverGroup` is used. Is there a plan to change that? If not, we would like to integrate with `AddressResolver`, so that `DoH` would be used for `HttpClient` and other resolutions as well. 

This is all preparation for Encrypted Client Hello support, so any technical considerations or tips from your side are welcome

vietj:
it is normal that the DNS client is not used for resolution, since we rely on a resolver which does more than a client, so DoH should be contributed to Netty instead to benefit of this feature so we can have it in vertx DNS resolution.

zekronium:
> it is normal that the DNS client is not used for resolution, since we rely on a resolver which does more than a client, so DoH should be contributed to Netty instead to benefit of this feature so we can have it in vertx DNS resolution.

Yeah, that is known that its not being used. My justification was to extend the AddressResolver on the vertx side, since it handles the quite complex and cumbersome API of HTTP Client of Netty and simply how much easier that would be atleast as a short term solution.

Would you in any case accept such an implementation on vertx side atleast momentarily or we should only do Netty

vietj:
we cannot accept it as is, unless there is a guarantee that it will be provided in Netty (to avoid a feature regression).

perhaps the best would be to develop in Netty then use in vertx and we could backport the DNS client changes from master to 4.x to keep things in par

zekronium:
> we cannot accept it as is, unless there is a guarantee that it will be provided in Netty (to avoid a feature regression).
> 
> perhaps the best would be to develop in Netty then use in vertx and we could backport the DNS client changes from master to 4.x to keep things in par

We are waiting for this to come through, then we can resume
https://github.com/netty/netty/pull/14160

vietj:
this looks like an awesome contribution to vertx and netty @imz87 @zekronium

PR #5262, opened 2024-07-30
Reuse existing allocators, enable experimental adaptive one and enable SSL heap pooling (Fixes #5168)

### LINKED ISSUE #5168: Vertx pooled allocator should be the same as Netty
Currently, in vertx https://github.com/eclipse-vertx/vert.x/blob/adbe97600cc6215f15ce2fac629c89f93618ca8f/src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java#L25 is NOT reusing the Netty's `PooledByteBufAllocator.DEFAULT`, causing creation of more thread-local direct buffers and arenas, enlarging the RSS footprint of vertx application, for no reason.

What's the reason behind this choice @vietj?

franz1981:
Fixes #5168

### DIFF
--- a/src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
+++ b/src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
@@ -11,22 +11,59 @@
 package io.vertx.core.buffer.impl;
 
 import io.netty.buffer.AbstractByteBufAllocator;
+import io.netty.buffer.AdaptiveByteBufAllocator;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.UnpooledByteBufAllocator;
 import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.SystemPropertyUtil;
 
 public abstract class VertxByteBufAllocator extends AbstractByteBufAllocator {
 
   /**
-   * Vert.x pooled allocator.
+   * Vert.x pooled allocator.<br>
+   * Setting {@code -Dio.netty.allocator.type=adaptive} will use {@link AdaptiveByteBufAllocator} in
+   * {@link ByteBufAllocator#DEFAULT} instead of the default {@link PooledByteBufAllocator#DEFAULT} one.<br>
+   * Setting anything different from {@code adaptive} will still use {@link PooledByteBufAllocator#DEFAULT} as the allocator.
    */
-  public static final ByteBufAllocator POOLED_ALLOCATOR = new PooledByteBufAllocator(true);
+  public static final ByteBufAllocator POOLED_ALLOCATOR = pooledAllocator();
+
+  /**
+   * Tries hard to reuse the existing default Netty allocators instances to avoid creating new instances.
+   */
+  private static ByteBufAllocator pooledAllocator() {
+    String allocType = allocType();
+    if (PooledByteBufAllocator.defaultPreferDirect()) {
+      if ("adaptive".equals(allocType)) {
+        assert ByteBufAllocator.DEFAULT instanceof AdaptiveByteBufAllocator;
+        return ByteBufAllocator.DEFAULT;
+      } else {
+        return PooledByteBufAllocator.DEFAULT;
+      }
+    } else {
+      if ("adaptive".equals(allocType)) {
+        return new AdaptiveByteBufAllocator(true);
+      } else {
+        return new PooledByteBufAllocator(true);
+      }
+    }
+  }
+
+  /**
+   * This is how currently Netty decide which allocator to use
+   * @see <a href="https://github.com/netty/netty/blob/netty-4.1.112.Final/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L80-L98">ByteBufUtil</a>
+   */
+  private static String allocType() {
+    return SystemPropertyUtil.get("io.netty.allocator.type", PlatformDependent.isAndroid() ? "unpooled" : "pooled");
+  }
+
   /**
    * Vert.x shared unpooled allocator.
    */
-  public static final ByteBufAllocator UNPOOLED_ALLOCATOR = new UnpooledByteBufAllocator(false);
+  public static final ByteBufAllocator UNPOOLED_ALLOCATOR = !PooledByteBufAllocator.defaultPreferDirect() ?
+    UnpooledByteBufAllocator.DEFAULT :
+    new UnpooledByteBufAllocator(false);
 
   private static final VertxByteBufAllocator UNSAFE_IMPL = new VertxByteBufAllocator() {
     @Override
--- a/src/main/generated/io/vertx/core/net/OpenSSLEngineOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/net/OpenSSLEngineOptionsConverter.java
@@ -1,11 +1,6 @@
 package io.vertx.core.net;
 
 import io.vertx.core.json.JsonObject;
-import io.vertx.core.json.JsonArray;
-import io.vertx.core.json.impl.JsonUtil;
-import java.time.Instant;
-import java.time.format.DateTimeFormatter;
-import java.util.Base64;
 
 /**
  * Converter and mapper for {@link io.vertx.core.net.OpenSSLEngineOptions}.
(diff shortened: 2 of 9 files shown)


franz1981:
@He-Pin do you think this could be useful to use to try the adaptive pool with vertx?

zekronium:
@franz1981 Why is the heap buffer not pooled, shouldn't it be a almost free lunch in performance?

franz1981:
@zekronium 

> Why is the heap buffer not pooled

IDK I would love to know it by @vietj 

probably there's some historical reason for that but I would like to know it.

franz1981:
@vietj this is adding the option to enable heap pooling when SSL is enable, which can be a saviour with JDK SSL, see https://github.com/quarkusio/quarkus/issues/41880#issuecomment-2258734835

src/main/java/io/vertx/core/net/TCPSSLOptions.java
```diff
@@ -872,4 +881,22 @@ public TCPSSLOptions setReusePort(boolean reusePort) {
     return (TCPSSLOptions) super.setReusePort(reusePort);
   }
 
+  /**
+   * Set whether the server should use heap pools while using SSL.
```
vietj, line 885: server -> server/client

src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
```diff
@@ -11,22 +11,52 @@
...
    * Vert.x pooled allocator.
    */
-  public static final ByteBufAllocator POOLED_ALLOCATOR = new PooledByteBufAllocator(true);
+  public static final ByteBufAllocator POOLED_ALLOCATOR = pooledAllocator();
+
+  /**
+   * Tries hard to reuse the existing default Netty allocators instances to avoid creating new instances.
+   */
+  private static ByteBufAllocator pooledAllocator() {
+    String allocType = allocType();
```
vietj, line 33: I think we should log which one gets finally used

franz1981: That's a good idea...trace?

src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
```diff
@@ -11,22 +11,52 @@
...
+      }
+    }
+  }
+
+  private static String allocType() {
+    return SystemPropertyUtil.get("io.netty.allocator.type", PlatformDependent.isAndroid() ? "unpooled" : "pooled");
+  }
+
   /**
    * Vert.x shared unpooled allocator.
```
vietj, line 55: missing: document how this allocator is chosen

src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
```diff
@@ -11,22 +11,52 @@
...
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.UnpooledByteBufAllocator;
 import io.netty.util.internal.PlatformDependent;
+import io.netty.util.internal.SystemPropertyUtil;
 
 public abstract class VertxByteBufAllocator extends AbstractByteBufAllocator {
 
   /**
    * Vert.x pooled allocator.
```
vietj, line 25: missing: document how this allocator is chosen

src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
vietj: I think debug is OK, it appears only once in the lifetime of the JVM

zekronium: FIY: netty is already logging this, is it really worth it to double that?

franz1981: Good point: found https://github.com/netty/netty/blob/netty-4.1.112.Final/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L80-L98

src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
franz1981: I've added some docs: PTAL

What's "weird" is that I'm reusing the existing `-Dio.netty.allocator.type` property to control which allocator use, but currently anything different from `adaptive` will fallback to `PooledByteBufAllocator.DEFAULT` because that's the sole choice for `VertxByteBufAllocator.POOLED_ALLOCATOR`, which, as the name suggest, must be pooled.

Probably it's worthy to use a separate vertx sys property here, for this, but TBH I would expect if the the Netty one is using `adaptive` to be able to reuse the instance allocated by Netty at `ByteBufAllocator.DEFAULT`.

franz1981:
PTAL @vietj

src/main/java/io/vertx/core/net/SSLEngineOptions.java
```diff
@@ -64,4 +64,11 @@ public SSLEngineOptions setUseWorkerThread(boolean useWorkerThread) {
     this.useWorkerThread = useWorkerThread;
     return this;
   }
+
+  /**
+   * @return whether to use pooled heap buffers
+   */
+  public boolean usePooledHeapBuffers() {
```
vietj, line 71: this should be only on JdkSSLEngineOptions as this is where it makes sense, if that is useful later we will move it to main class

vietj: also for data object it should be getter/setter, so that should be get/set

franz1981: So It means that I have to perform instanceof on the SSL helper?

vietj: yes but it is fine

franz1981:
Not good @vietj 

```
Starting test: ClusteredEventBusWithSSLTest#testSendCharacter 
[vert.x-eventloop-thread-1] WARN io.netty.util.concurrent.AbstractEventExecutor - A task raised an exception. Task: io.vertx.core.impl.future.FutureBase$$Lambda/0x00007f59dc23dcd0@54c5bbbf
java.lang.NullPointerException: Cannot invoke "io.vertx.core.net.SSLEngineOptions.usePooledHeapBuffers()" because "this.sslEngineOptions" is null
	at io.vertx.core.net.impl.SSLHelper.preferredNettyAllocatorWith(SSLHelper.java:141)
	at io.vertx.core.net.impl.TCPServerBase.lambda$listen$4(TCPServerBase.java:243)
	at io.vertx.core.impl.future.FutureImpl$4.onSuccess(FutureImpl.java:176)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:66)
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:259)
	at io.vertx.core.impl.future.Mapping.onSuccess(Mapping.java:40)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:66)
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:259)
	at io.vertx.core.impl.future.Mapping.onSuccess(Mapping.java:40)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:66)
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:259)
	at io.vertx.core.impl.future.Mapping.onSuccess(Mapping.java:40)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:66)
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:259)
	at io.vertx.core.impl.future.PromiseImpl.onSuccess(PromiseImpl.java:49)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:66)
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:259)
	at io.vertx.core.impl.future.Composition$1.onSuccess(Composition.java:62)
	at io.vertx.core.impl.future.FutureBase.lambda$emitSuccess$0(FutureBase.java:60)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:173)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:166)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:469)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:569)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:994)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```
not sure the engine options is the right place for that configuration flag...let's see what I can do here...

franz1981:
I've decided to split in 3 separate PRs this effort, to make it easier to integrate (and test)!

PR #5176, opened 2024-04-01
support zstd compression

sullis:
# Motivation

Netty supports zstd compression.

Google Chrome recently added support for zstd
https://www.phoronix.com/news/Zstd-1.5.6-Zstandard

# Conformance

You should have signed the Eclipse Contributor Agreement as explained in https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md
Please also make sure you adhere to the code style guidelines: https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines

PR #5270, opened 2024-08-02
Support snappy compression

lucamolteni:
Motivation:

Support snappy as a compression mechanism 

Similar to https://github.com/eclipse-vertx/vert.x/pull/5176/files but Snappy support is already inside netty so I don't need to include any other dependencies

Conformance:

### DIFF
--- a/vertx-core/src/main/asciidoc/http.adoc
+++ b/vertx-core/src/main/asciidoc/http.adoc
@@ -391,7 +391,7 @@ options when creating the server.
 
 Snappy is supported without external dependencies.
 
-You need to have Brotli4j on the classpath to decompress Brotli:
+You need to have Brotli4j on the classpath to decompress Brotli, and Zstd-jni for Zstandard:
 
 * Maven (in your `pom.xml`):
 
@@ -402,6 +402,11 @@ You need to have Brotli4j on the classpath to decompress Brotli:
   <artifactId>brotli4j</artifactId>
   <version>${brotli4j.version}</version>
 </dependency>
+<dependency>
+  <groupId>com.github.luben</groupId>
+  <artifactId>zstd-jni</artifactId>
+  <version>${zstd-jini.version}</version>
+</dependency>
 ----
 * Gradle (in your `build.gradle` file):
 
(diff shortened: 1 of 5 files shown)


vertx-core/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
```diff
@@ -216,7 +216,7 @@ private HttpRequest createRequest(
 
   static CharSequence determineCompressionAcceptEncoding() {
     if (isBrotliAvailable() && isZstdAvailable()) {
```
vietj, line 218: don't we need to add `isSnappyAvailable()` too ?

lucamolteni: Compared to `brotli` and `zstd`, [snappy is provided directly by Netty itself](https://github.com/netty/netty/blob/c176ce579106d5b8f5221cbdd8eb4a4ea70ebe1f/codec/src/main/java/io/netty/handler/codec/compression/Snappy.java), so we don't need to have a check that a library is on a class path.

vietj: oh that's good to kow

vietj (changes requested):
Documentation should be updated https://vertx.io/docs/vertx-core/java/#_handling_compressed_body

vietj:
For the docs look in the .adoc file in the project and search for brotli

vertx-core/src/main/asciidoc/http.adoc
```diff
@@ -383,7 +383,7 @@ Here's an example of querying and adding cookies:
 
 ==== Handling compressed body
 
-Vert.x can handle compressed body payloads which are encoded by the client with the _deflate_, _gzip_ or _brotli_
+Vert.x can handle compressed body payloads which are encoded by the client with the _deflate_, _gzip_, _snappy_ or _brotli_
```
vietj, line 386: I think we should update the doc and mention that snappy does not require an external dependency unlike brotli

vietj: that would be great to improve the documentation @lucamolteni because this section is rather slim, please go ahead

lucamolteni: @vietj can I do in another PR though? Let's merge this first

vietj: actually I think it would be better to have this done in the same PR , since we will need to backport to 4.x branches

lucamolteni: @vietj I've updated the two sections accordingly, please take a look

lucamolteni: Any news @vietj ? I'm closing this

vietj:
thank you @lucamolteni

vietj (commented):
LGTM

PR #5273, opened 2024-08-03
HAManager couldn't see quorum if during its initialization cluster manager has all nodes joined already #5272

### LINKED ISSUE #5272: HAManager couldn't see quorum if during its initialization cluster manager has all nodes joined already
### Version

4.5.7

### Context

My app is based on Vert.x  with embedded Hazelcast cluster manager. When I run several instances of the app with HA mode enabled some instances cannot deploy HA verticles, and logs show that they have not attain quorum:  `Quorum not attained. Deployment of verticle will be delayed until there's a quorum.` 

But actually they have.

The problem is that Hazelcast cluster could be in a state when all nodes have already joined at the moment when `HAManager` initialization starts. So `HAManager`s method  `nodeAdded` will never be called. Also `checkQuorum` method called on `init` see all nodes from `clusterManager`, but `clusterMap` do not contain information from all nodes. They will put it in several milliseconds later of course, but `checkQuorum` method will be completed at that time and never be called again.

### Steps to reproduce

It just occurs in some cases.

apanasevich:
Fixes #5272

The key idea is to wait until `clusterMap` get information from other nodes like it done in `nodeAdded` method.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/HAManager.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/HAManager.java
@@ -373,6 +373,8 @@ public class HAManager {
           if (group.equals(this.group)) {
             count++;
           }
+        } else if(!attainedQuorum) {
+          checkQuorumWhenAdded(node, System.currentTimeMillis());
         }
       }
       boolean attained = count >= quorumSize;
--- a/vertx-core/src/main/java/io/vertx/core/impl/HAManager.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/HAManager.java
@@ -373,7 +373,7 @@ public class HAManager {
           if (group.equals(this.group)) {
             count++;
           }
-        } else if(!attainedQuorum) {
+        } else if (!attainedQuorum) {
           checkQuorumWhenAdded(node, System.currentTimeMillis());
         }
       }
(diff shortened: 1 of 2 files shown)


vietj:
@apanasevich can you provide a reproducer for this or add a test demonstrating where this is useful ?

apanasevich:
Hi @vietj 

I cannot reproduce a bug. The problem occures when `clusterManager.getNodes();` contains a node, but `clusterMap.get(node);` do not yet. And this case is covered with timered check in method `checkQuorumWhenAdded`, but not covered on `init`.

```
      List<String> nodes = clusterManager.getNodes();
      int count = 0;
      for (String node : nodes) {
        String json = clusterMap.get(node);

```

I'll try to write a test covering the issue.

apanasevich:
I've written the test with a special `ClusterManager` emulating the issue behaviour. It does not call `nodeAdded` listerens that looks like cluster was created before `Vert.x` initiates `HAManager`. 

Please take a look at the PR again.

vietj:
thanks @apanasevich

apanasevich:
`NetTest.testSocketAddress` timed out. But this test does not use HA feature and even does not deploy a verticle.

vietj:
@apanasevich there are some flaky tests, I reran the CI

PR #5282, opened 2024-08-14
 Fixes updates to traffic shaping options with shared servers

arjunashok:
Motivation:

This fixes the update-path for traffic shaping options to check for the existence of the traffic shaping handler only for the actual server. Currently, updating traffic-shaping options in a shared-server setting results in `IllegalStateException`, as the traffic-shaping handler is not set for the worker (non-main) servers.

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/TCPServerBase.java
+++ b/src/main/java/io/vertx/core/net/impl/TCPServerBase.java
@@ -153,14 +153,16 @@ public abstract class TCPServerBase implements Closeable, MetricsProvider {
     if (options == null) {
       throw new IllegalArgumentException("Invalid null value passed for traffic shaping options update");
     }
-    if (trafficShapingHandler == null) {
-      throw new IllegalStateException("Unable to update traffic shaping options because the server was not configured " +
-                                      "to use traffic shaping during startup");
-    }
     TCPServerBase server = actualServer;
+    // Update the traffic shaping options only for the actual/main server
     if (server != null && server != this) {
-      server.updateTrafficShapingOptions(options);
+       server.updateTrafficShapingOptions(options);
     } else {
+      if (trafficShapingHandler == null) {
+        throw new IllegalStateException("Unable to update traffic shaping options because the server was not configured " +
+                                        "to use traffic shaping during startup");
+      }
+
       long checkIntervalForStatsInMillis = options.getCheckIntervalForStatsTimeUnit().toMillis(options.getCheckIntervalForStats());
       trafficShapingHandler.configure(options.getOutboundGlobalBandwidth(), options.getInboundGlobalBandwidth(), checkIntervalForStatsInMillis);
 
@@ -292,8 +294,10 @@ public abstract class TCPServerBase implements Closeable, MetricsProvider {
       } else {
         // Server already exists with that host/port - we will use that
         actualServer = main;
-        metrics = main.metrics;
-        childHandler =  childHandler(listenContext, localAddress, main.trafficShapingHandler);
+        metrics = actualServer.metrics;
+        // Ensure the workers inherit the actual server's traffic-shaping handler
+        trafficShapingHandler = actualServer.trafficShapingHandler;
+        childHandler =  childHandler(listenContext, localAddress, actualServer.trafficShapingHandler);
         worker = ch -> childHandler.accept(ch, actualServer.sslChannelProvider.result().sslChannelProvider());
         actualServer.servers.add(this);
         actualServer.channelBalancer.addWorker(eventLoop, worker);
--- a/src/main/java/io/vertx/core/net/impl/TCPServerBase.java
+++ b/src/main/java/io/vertx/core/net/impl/TCPServerBase.java
@@ -156,22 +156,29 @@ public abstract class TCPServerBase implements Closeable, MetricsProvider {
     TCPServerBase server = actualServer;
     // Update the traffic shaping options only for the actual/main server
     if (server != null && server != this) {
-       server.updateTrafficShapingOptions(options);
+      server.updateTrafficShapingOptions(options);
     } else {
       if (trafficShapingHandler == null) {
         throw new IllegalStateException("Unable to update traffic shaping options because the server was not configured " +
                                         "to use traffic shaping during startup");
       }
 
-      long checkIntervalForStatsInMillis = options.getCheckIntervalForStatsTimeUnit().toMillis(options.getCheckIntervalForStats());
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/core/net/impl/TCPServerBase.java
```diff
@@ -153,14 +153,16 @@ public void updateTrafficShapingOptions(TrafficShapingOptions options) {
...
     }
-    if (trafficShapingHandler == null) {
-      throw new IllegalStateException("Unable to update traffic shaping options because the server was not configured " +
-                                      "to use traffic shaping during startup");
-    }
     TCPServerBase server = actualServer;
+    // Update the traffic shaping options only for the actual/main server
     if (server != null && server != this) {
-      server.updateTrafficShapingOptions(options);
+       server.updateTrafficShapingOptions(options);
```
sarankk, line 159: Can you update this, with `return` here to avoid repeated calls

arjunashok: That is an interesting question. The repeated calls would happen if the consumer of this API invokes this method for all the shared servers. 

The decision here is based on the contract/interface we would like to expose to clients invoking `updateTrafficShapingOptions` (and similar APIs that affect configuration across shared servers)

1. Is it fair to expect clients to apply/change this setting for all the servers, while having the underlying implementation only update it when the actual server is involved? (which is what I think you are suggesting)

2. Alternatively, should we define the contract such that this setting should be applied/changed to ANY of the servers, and under the hood, vertx will only update it for the actual/main server?  (current implementation)

It seems like we currently have a model like (1). My only concern is that for invocations that are not on a actual/main server, it will do nothing (silently). 
If this is indeed the pattern we would like to follow, then I would prefer to have some signal (even if it were a log line) indicating that this was a worker server, so no updates were made. This might be useful, in the event that someone updates a server’s setting (that is not a main server) to not see the change taking effect.

 I see that `updateSSLOptions` does something similar to (1), although if it were called for all the servers, we would update the SSL options for the actual server N times (where N = no. verticles)

(2) comes with the assumption that this is a shared/global setting that applies to all servers, and hence can be updated on ANY of the servers. Although, one could argue that it is ambiguous for the client as to which server's configuration should be updated, and if they were to apply this to all servers, it would result in several redundant invocations for the actual server.
We could address this concern with redundant invocations by comparing the `TrafficShapingOptions` instances to only update when they differ. I would prefer to do something like this to have no ambiguity.

@vietj Do you have any thoughts or comments on this?

arjunashok: In the latest commit, I made a change to only update the traffic-shaping options when they change. So, we address both concerns by:
1) Updating the actual-server's options regardless of which server is used for the update
2) Only perform the above update when the new options differ from the existing ones

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -204,6 +205,54 @@ public void start(Promise<Void> startPromise) {
     Assert.assertTrue(elapsedMillis > expectedTimeMillis(totalReceivedLength.get(), OUTBOUND_LIMIT)); // because there are simultaneous 2 requests
   }
 
+  @Test
+  public void testDynamicOutboundRateUpdateSharedServers() throws IOException, InterruptedException
+  {
+    int numEventLoops = 5; // We start a shared TCP server with 2 event-loops
+    List<HttpServer> servers = new ArrayList<>();
```
vietj, line 212: the list should be synchronised because it will be updated concurrently by different threads.

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -204,6 +205,54 @@ public void start(Promise<Void> startPromise) {
     Assert.assertTrue(elapsedMillis > expectedTimeMillis(totalReceivedLength.get(), OUTBOUND_LIMIT)); // because there are simultaneous 2 requests
   }
 
+  @Test
+  public void testDynamicOutboundRateUpdateSharedServers() throws IOException, InterruptedException
+  {
+    int numEventLoops = 5; // We start a shared TCP server with 2 event-loops
```
vietj, line 211: 2 -> 5

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -204,6 +205,54 @@ public void start(Promise<Void> startPromise) {
...
+                    waitForResponse.countDown();
+                  }));
+      }
+    }));
+    awaitLatch(waitForResponse);
+    TrafficShapingOptions updatedTrafficOptions = new TrafficShapingOptions()
+                                             .setInboundGlobalBandwidth(INBOUND_LIMIT) // unchanged
+                                             .setOutboundGlobalBandwidth(2 * OUTBOUND_LIMIT);
+
+    for (int i = 0; i < numEventLoops; i++) {
```
vietj, line 248: why are options updated after the test ?

arjunashok: Will fix to update options after server has started

src/main/java/io/vertx/core/net/impl/TCPServerBase.java
vietj: @arjunashok the latest commit seems to do like update of SSL options and I think is the most efficient way to go

src/test/java/io/vertx/core/http/HttpBandwidthLimitingTest.java
```diff
@@ -204,6 +206,62 @@ public void start(Promise<Void> startPromise) {
...
+                    if (res.succeeded()) {
+                      // Apply traffic shaping options after the server has started
+                      TrafficShapingOptions updatedTrafficOptions = new TrafficShapingOptions()
+                                                                      .setInboundGlobalBandwidth(INBOUND_LIMIT)
+                                                                      .setOutboundGlobalBandwidth(2 * OUTBOUND_LIMIT);
+
+                      for (int i = 0; i < numEventLoops; i++) {
+                        servers.forEach(s -> s.updateTrafficShapingOptions(updatedTrafficOptions));
+                      }
+                      startPromise.complete();
```
vietj, line 231: we need to complete the start promise with the result of updateTrafficShapingOptions to ensure proper sequentiality in the test and avoid races, e.g. using a CompositeFuture.all(...) with the future returned by the updates seems the easiest way

arjunashok: yep, good point. updateTrafficShapingOptions is synchronous currently, but added steps to track those actions via promises that chains with the start promise.

PR #5288, opened 2024-08-26
Vertx lifecycle hook for VertxBuilder customization

tsegismont:
Given methods like TracingOptions#setTracer have been deprecated, a new hook can help users configure the Vert.x instance.


tsegismont:
For now, just a draft for discussion.

@vietj it's not possible to use the public builder here without breaking the command launcher API.

If we introduce this new method in Vert.x 4, it's going to change a little bit in Vert.x 5. A little change, but still a breaking one, that we need to be aware of.

If we agree on doing this, I'll add a test to the PR

vietj:
I'm good for a breaking change here given it's an API that fades away in 5

tsegismont:
@vietj PTAL

PR #5292, opened 2024-09-01
Try reuse existing Netty pooled allocator singleton (Fixes #5168)

### LINKED ISSUE #5168: Vertx pooled allocator should be the same as Netty

franz1981:
Fixes #5168 

This is part of https://github.com/eclipse-vertx/vert.x/pull/5262

non-SSL connections configure `PooledByteBufAllocator.DEFAULT` as per 
https://github.com/eclipse-vertx/vert.x/blob/f42b5d22c5c738cae1db925a06029212b198c398/vertx-core/src/main/java/io/vertx/core/net/impl/NetServerImpl.java#L515-L519

In this scenario, hitting these call sites:

https://github.com/eclipse-vertx/vert.x/blob/65403248e1e5c679d5418d8040aafea5bd09d0ef/src/main/java/io/vertx/core/net/impl/VertxHandler.java#L61 

https://github.com/eclipse-vertx/vert.x/blob/f42b5d22c5c738cae1db925a06029212b198c398/vertx-core/src/main/java/io/vertx/core/http/impl/Http2ConnectionBase.java#L59

or

https://github.com/eclipse-vertx/vert.x/blob/f42b5d22c5c738cae1db925a06029212b198c398/vertx-core/src/main/java/io/vertx/core/buffer/impl/BufferImpl.java#L49-L54

referencing `VertxByteBufAllocator.DEFAULT` , it triggers the initialization of `VertxByteBufAllocator.POOLED_ALLOCATOR` as well and by consequence its inner structures - making it a GC Root.

It results into a duplication of allocators i.e. `PooledByteBufAllocator.DEFAULT`  and `VertxByteBufAllocator.POOLED_ALLOCATOR`.

This patch aim to reuse the existing singleton i.e. `PooledByteBufAllocator.DEFAULT` if the user doesn't mess-up with `-Dio.netty.noPreferDirect`, in order to guarantee the original behaviour to be preserved.

### DIFF
--- a/src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
+++ b/src/main/java/io/vertx/core/buffer/impl/VertxByteBufAllocator.java
@@ -19,12 +19,15 @@ import io.netty.util.internal.PlatformDependent;
 
 public abstract class VertxByteBufAllocator extends AbstractByteBufAllocator {
 
+  private static final boolean REUSE_NETTY_ALLOCATOR = Boolean.getBoolean("vertx.reuseNettyAllocators");
+
   /**
-   * Vert.x pooled allocator.
+   * Vert.x pooled allocator. It should prefers direct buffers, unless {@link PlatformDependent#hasUnsafe()} is {@code false}.
    */
-  public static final ByteBufAllocator POOLED_ALLOCATOR = new PooledByteBufAllocator(true);
+  public static final ByteBufAllocator POOLED_ALLOCATOR = (REUSE_NETTY_ALLOCATOR && PooledByteBufAllocator.defaultPreferDirect()) ?
+    PooledByteBufAllocator.DEFAULT : new PooledByteBufAllocator(true);
   /**
-   * Vert.x shared unpooled allocator.
+   * Vert.x shared unpooled allocator. It prefers non-direct buffers.
    */
   public static final ByteBufAllocator UNPOOLED_ALLOCATOR = new UnpooledByteBufAllocator(false);
 
@@ -42,6 +45,10 @@ public abstract class VertxByteBufAllocator extends AbstractByteBufAllocator {
     }
   };
 
+  /**
+   * Vert.x shared unpooled heap buffers allocator.<br>
+   * Differently from {@link #UNPOOLED_ALLOCATOR}, its buffers are not reference-counted and array-backed.
+   */
   public static final VertxByteBufAllocator DEFAULT = PlatformDependent.hasUnsafe() ? UNSAFE_IMPL : IMPL;
 
   @Override
--- a/src/test/java/io/vertx/core/buffer/impl/VertxByteBufAllocatorTest.java
+++ b/src/test/java/io/vertx/core/buffer/impl/VertxByteBufAllocatorTest.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2011-2024 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.core.buffer.impl;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.PooledByteBufAllocator;
+
+public class VertxByteBufAllocatorTest {
+
+
+  @Test
+  public void defaultShouldNotReuseExistingNettyPooledAllocators() {
+    Assert.assertNull(System.getProperty("vertx.reuseNettyAllocators"));
+    Assert.assertNotSame(PooledByteBufAllocator.DEFAULT, VertxByteBufAllocator.POOLED_ALLOCATOR);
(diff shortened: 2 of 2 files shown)


franz1981:
@vietj I cannot reproduce the test failures here on my machine  sadly :/ can you retrigger the CI?

cescoffier:
Hold on before merging this one. We need to discuss versioning schemes.

franz1981:
@cescoffier @vietj decided what could be best to do?

franz1981:
@cescoffier 
For quarkus - which is the expected time frame where vertx 5 can be integrated?
Remember that this change I did here could easily be backported too, given that it brings tangible advantage with zero effort, in quarkus

cescoffier:
For Quarkus, Vert.x 5 integration will start in spring 2025 and end (optimistically) in fall 2025. 

The fact that we don't really know when useDirect is true or false (not the same value in tests, if you use JPMS, if you have access to usafe ... ) and that it can drastically change the behavior makes it hard to accept in a version that is intended to be used in the LTS.

franz1981:
> The fact that we don't really know when useDirect is true or false (not the same value in tests, if you use JPMS, if you have access to usafe ... ) and that it can drastically change the behavior makes it hard to accept in a version that is intended to be used in the LTS.

This is the same that will do Netty under the hood actually; it's not at all related to `Unsafe` but is instead related the sys property `-Dio.netty.noPreferDirect`.

Currently Netty allocates its own singleton(s) reading the same property; the point of checking it upfront here is to preserve the same behaviour the vertx users has benefitted till now.

cescoffier:
It's not what we saw with Julien yesterday, where we got prefer direct true in some cases and false in other cases - without setting any system property.

cescoffier:
I will only accept the risk if it does not change the current default except if explicitly configured (like with a system property) - but as I said - it seems to set the default value differently depending on various things (like unsafe, JVM version, JPMS, NEtty 4.1 vs. 4.2...)

franz1981:
I can see why @cescoffier thanks for sharing, that helped:

It is likely due to the presence of cleaner's API accessible, see https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/PlatformDependent.java#L202

which transitively depends by Unsafe presence, see https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/CleanerJava9.java#L79

that raise some concern regardless this PR, because, right now in Vertx:
1. **without SSL**: if unsafe is not present -> cleaner is noop -> Vertx use Netty's allocator singleton which does prefer heap pooled buffers:
->>>> THIS IS BAD!
see
https://github.com/netty/netty/blob/7893d5f150aefc99bc720387875395afad94d807/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java#L328

causing Heap arenas to be allocated (which cost heap, not off-heap) and additional on-the-flight direct buffers while interacting with NIO sockets i.e. more memory/heap used, more cpu used
2. **with SSL*: if unsafe is not present ->  cleaner is noop -> Vertx uses its own partial allocator which force using direct buffers (without a CLEANER!!!!!! - can OOM a container!) but which doesn't pool heap buffers (and causing SSL JDK a performance issue)

In short: if `Unsafe` is not present, it is a disaster, from a performance prespective.
So, if you have any run which have no `Cleaner` API accessible due to lack on `Unsafe`, we need to address it regardless, because it is bad, performance, but most importantly, stability-wise, for our users.

cescoffier:
Well, and it will likely be different between JVM and native. 
We need more investigation to understand all the consequences, even if today's solution is far from ideal. 

As I said, we cannot change the behavior in an LTS.

franz1981:
> Well, and it will likely be different between JVM and native.

which is kind of surprising (to me - because I didn't profiled native image apps enough - mea culpa :( )

> We need more investigation to understand all the consequences, even if today's solution is far from ideal.

+100
And I'll fill some issue with all the consequences and details: this stuff is NOT documented in Netty and only knowing how it works help - so better doing it once for "all" , if you agree

> As I said, we cannot change the behavior in an LTS.

I'll pick the choice of using a sys property to not impact anyone - but I still need to dug into this enough to make sure what we get/loose and risks

franz1981:
PTAL @cescoffier 

I've added a new property `vertx.reuseNettyAllocators`, which is `false` by default, to preserve the existing behaviour.
When merged in Quarkus we can set this property to `true` saving the doubling of RSS mentioned at https://github.com/quarkusio/quarkus/issues/42224

cescoffier:
That looks more acceptable, but I do not see where it's documented and tested (mainly because we may detect behavior change between Vert.x 4.x and 5.x)

franz1981:
Given that it depends by a static final property, if I set it to try it, is going to impact all the tests after it, randomly (because it depends how junit order test execution). 
I have tested that the default has not changed, instead. We can speak with @vietj on how to make it testable, without running the whole test suite again with this property in.
Consider that it doesn't change the classes used to perform the pooled allocations, but just which instance and currently there are no specific tests on vertx to cover what exist now.
I am opened to suggestions.

cescoffier:
We may have a bit more flexibility to test it in Quarkus (I don't care where the tests are, just that we need some coverage)

franz1981:
@vietj Any other concern for this @cescoffier @vietj ?

PR #5303, opened 2024-09-09
Allow configuring Netty compression content size threshold

wvuong:
Motivation:

Netty allows you to customize the minimum content size threshold for compression via a constructor arg in [HttpContentCompressor](https://netty.io/4.1/api/io/netty/handler/codec/http/HttpContentCompressor.html#HttpContentCompressor-int-io.netty.handler.codec.compression.CompressionOptions...-). Configuring this means that Netty won't compress responses if they are below the size threshold which can be useful for small responses where the CPU vs saved bytes tradeoff is not good. The Netty default is 0 which I carried on here. I also updated the documentation.

Conformance:


vertx-core/src/main/java/io/vertx/core/http/HttpServerOptions.java
```diff
@@ -588,6 +596,15 @@ public HttpServerOptions setCompressionLevel(int compressionLevel) {
     return this;
   }
 
+  public int getCompressionContentSizeThreshold() {
```
vietj, line 599: javadoc

vertx-core/src/main/java/io/vertx/core/http/HttpServerOptions.java
```diff
@@ -588,6 +596,15 @@ public HttpServerOptions setCompressionLevel(int compressionLevel) {
...
+    return compressionContentSizeThreshold;
+  }
+
+  public HttpServerOptions setCompressionContentSizeThreshold(int compressionContentSizeThreshold) {
```
vietj, line 603: guard against invalid values ?

vietj:
thank you

vietj:
@wvuong thank you for the contribution

PR #5307, opened 2024-09-10
Save parsing twice numeric IPv4 address

franz1981:
Using machines which provide `Host` headers in the form "y.x.z.w:port" (ith y, x, z, w numeric) will cause `HostAndPortImpl ::parseHost` to execute an additional `parseRegName` because `parseIPv4Address` returns `-1`, failing to detect the address as a valid numeric address with a maybe valid port.

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
@@ -30,11 +30,10 @@ import io.vertx.core.http.impl.headers.HeadersAdaptor;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.impl.VertxInternal;
 import io.vertx.core.impl.future.PromiseInternal;
-import io.vertx.core.impl.logging.Logger;
-import io.vertx.core.impl.logging.LoggerFactory;
 import io.vertx.core.net.HostAndPort;
 import io.vertx.core.net.NetSocket;
 import io.vertx.core.net.SocketAddress;
+import io.vertx.core.net.impl.HostAndPortImpl;
 import io.vertx.core.spi.metrics.HttpServerMetrics;
 import io.vertx.core.spi.tracing.SpanKind;
 import io.vertx.core.spi.tracing.TagExtractor;
@@ -63,7 +62,7 @@ import static io.vertx.core.spi.metrics.Metrics.METRICS_ENABLED;
  */
 public class Http1xServerRequest extends HttpServerRequestInternal implements io.vertx.core.spi.observability.HttpRequest {
 
-  private static final Logger log = LoggerFactory.getLogger(Http1xServerRequest.class);
+  private static final HostAndPort NULL_HOST_AND_PORT = HostAndPort.create("", -1);
 
   private final Http1xServerConnection conn;
   final ContextInternal context;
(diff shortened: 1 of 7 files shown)


franz1981:
PTAL @vietj 
I'm still adding few others improvement and a benchmark to further improve this, but I need first confirmation of my RFC interpretation

src/main/java/io/vertx/core/net/impl/HostAndPortImpl.java
```diff
@@ -30,35 +30,45 @@ public static int parseIPv4Address(String s, int from, int to) {
...
+      // we need : to start a port
+      return -1;
+    }
+    // we (maybe) have a port - even with a single digit; the ipv4 addr is fineFi
+    return from;
   }
 
   static int parseDecOctet(String s, int from, int to) {
     int val = parseDigit(s, from++, to);
-    switch (val) {
```
franz1981, line 38: we have the exact boundaries here, I don't see the point to use a switch - at the mercy of the JIT to improve it

src/test/java/io/vertx/core/net/impl/HostAndPortTest.java
```diff
@@ -56,12 +57,14 @@ public void testParseHost() {
     assertEquals(7, HostAndPortImpl.parseHost("0.0.0.0", 0, 7));
     assertEquals(8, HostAndPortImpl.parseHost("10.0.0.1.nip.io", 0, 8));
     assertEquals(15, HostAndPortImpl.parseHost("10.0.0.1.nip.io", 0, 15));
+    assertEquals(8, HostAndPortImpl.parseHost("10.0.0.1:8080", 0, 15));
```
franz1981, line 60: @vietj this is fine for you?

franz1981:
These are the numbers before 
```
Benchmark                                        (host)  Mode  Cnt   Score   Error  Units
HostAndPortBenchmark.parseAuthority    192.168.0.1:8080  avgt   20  26.070 ± 1.790  ns/op
HostAndPortBenchmark.parseHost         192.168.0.1:8080  avgt   20  16.042 ± 0.030  ns/op
HostAndPortBenchmark.parseIPv4Address  192.168.0.1:8080  avgt   20  11.542 ± 0.023  ns/op
```
and this PR:
```
Benchmark                                        (host)  Mode  Cnt   Score   Error  Units
HostAndPortBenchmark.isValidAuthority  192.168.0.1:8080  avgt   20  13.785 ± 0.016  ns/op
HostAndPortBenchmark.parseAuthority    192.168.0.1:8080  avgt   20  19.384 ± 1.116  ns/op
HostAndPortBenchmark.parseHost         192.168.0.1:8080  avgt   20  10.988 ± 0.227  ns/op
HostAndPortBenchmark.parseIPv4Address  192.168.0.1:8080  avgt   20  10.627 ± 0.052  ns/op
```

The new `isValidAuthority` method "could" be used in vertx-web at https://github.com/vert-x3/vertx-web/blob/84dfc9a78b246089bf5a8e7282ed119c05e0efa6/vertx-web/src/main/java/io/vertx/ext/web/impl/RoutingContextImpl.java#L77 to save allocating `HostAndPortImpl` given that we just check that's a valid host, wdyt?

src/main/java/io/vertx/core/http/impl/Http1xServerRequest.java
```diff
@@ -63,7 +62,7 @@
  */
 public class Http1xServerRequest extends HttpServerRequestInternal implements io.vertx.core.spi.observability.HttpRequest {
 
-  private static final Logger log = LoggerFactory.getLogger(Http1xServerRequest.class);
+  private static final HostAndPort EMPTY_HOST_AND_PORT = HostAndPort.create("", -1);
```
vietj, line 65: name it rather NULL_HOST_AND_PORT

franz1981: :+1:

PR #5312, opened 2024-09-10
Decouple the cluster manager interface from NodeSelector and make this class an implementation details

vietj:
Motivation:

The `NodeSelector` interface has currently a single implementation and is exposed to the cluster manager. The `NodeSelector` interface combines two API, an API selecting node used by the event bus, an API for broadcasting changes to the node selector implementation used by the cluster manager. The cluster manager exposed API should be reduced to the methods it is intended to use, this also allows to reduce the SPI surface of the cluster manager SPI.

Changes:

Extract the `RegistrationListener` out of `NodeSelector`, the `ClusterManager` interface gets getter/setter for the registration listener, similar to the `NodeListener`.

The `NodeSelector` is moved to the cluster manager SPI impl package and is not exposed anymore.

This is an SPI breaking change.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
+++ b/vertx-core/src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
@@ -52,14 +52,10 @@ public interface ClusterManager extends VertxServiceProvider {
 
   /**
    * Invoked before this cluster node tries to join the cluster.
-   * <p>
-   * Implementations must signal the provided {@code nodeSelector} when messaging handler registrations are added or removed
-   * by sending a {@link RegistrationUpdateEvent} with {@link NodeSelector#registrationsUpdated(RegistrationUpdateEvent)}.
    *
    * @param vertx        the Vert.x instance
-   * @param nodeSelector the {@link NodeSelector} that must receive {@link RegistrationUpdateEvent}.
    */
-  void init(Vertx vertx, NodeSelector nodeSelector);
+  void init(Vertx vertx);
 
   /**
    * Return an {@link AsyncMap} for the given {@code name}.
@@ -129,6 +125,14 @@ public interface ClusterManager extends VertxServiceProvider {
    * @return true if active, false otherwise
    */
   boolean isActive();
+
+  /**
+   * Implementations must signal the provided {@code registrationListener} when messaging handler registrations are added or removed
+   * by sending a {@link RegistrationUpdateEvent} with {@link RegistrationListener#registrationsUpdated(RegistrationUpdateEvent)}.
+   *
+   * @param registrationListener the registration listener
+   */
+  void registrationListener(RegistrationListener registrationListener);
 
   /**
    * Share a new messaging handler registration with other nodes in the cluster.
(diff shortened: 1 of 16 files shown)


vertx-core/src/main/java/io/vertx/core/spi/cluster/RegistrationListener.java
```diff
@@ -0,0 +1,36 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.spi.cluster;
+
+public interface RegistrationListener {
```
vietj, line 13: missing doc

vertx-core/src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
```diff
@@ -52,14 +52,10 @@ default void init(VertxBootstrap builder) {
...
    * Invoked before this cluster node tries to join the cluster.
-   * <p>
-   * Implementations must signal the provided {@code nodeSelector} when messaging handler registrations are added or removed
-   * by sending a {@link RegistrationUpdateEvent} with {@link NodeSelector#registrationsUpdated(RegistrationUpdateEvent)}.
    *
    * @param vertx        the Vert.x instance
-   * @param nodeSelector the {@link NodeSelector} that must receive {@link RegistrationUpdateEvent}.
    */
-  void init(Vertx vertx, NodeSelector nodeSelector);
+  void init(Vertx vertx);
```
tsegismont, line 58: I don't understand why the `RegistrationListener` must be set via a separate method. Why can't it replace the `NodeSelector` as a parameter?

vertx-core/src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
```diff
@@ -130,6 +126,19 @@ default void init(VertxBootstrap builder) {
...
+   * by sending a {@link RegistrationUpdateEvent} with {@link RegistrationListener#registrationsUpdated(RegistrationUpdateEvent)}.
+   *
+   * @param registrationListener the registration listener
+   */
+  void registrationListener(RegistrationListener registrationListener);
+
+  /**
+   * @return the registration listener
+   */
+  RegistrationListener registrationListener();
```
tsegismont, line 140: Is this new method needed?

tsegismont (commented):
Looks good to me, just a question about API design

vertx-core/src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
vietj: nope

vertx-core/src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
vietj: we have `nodeListener(NodeListener listener)`, following the same pattern with `RegistrationListener`

vertx-core/src/main/java/io/vertx/core/spi/cluster/ClusterManager.java
vietj: removed, like nodeListener

PR #5316, opened 2024-09-12
Enable buffer pooling settings with a SSL configurable option

franz1981:
SSL JDK is very memory allocation intensive, see https://github.com/netty/netty/issues/14208: by default Vertx is using unpooled heap buffers of 16 KB for each interaction with SSL/TSL, which can lead to increase significantly the memory footprint.

In order to fix it I've created a new ssl configuration property (which should become the default in vertx 5) to enable pooling of heap buffers using JDK SSL, while keeping the existing behaviour by default.

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/NetClientImpl.java
+++ b/src/main/java/io/vertx/core/net/impl/NetClientImpl.java
@@ -28,7 +28,6 @@ import io.vertx.core.Closeable;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Promise;
-import io.vertx.core.buffer.impl.PartialPooledByteBufAllocator;
 import io.vertx.core.impl.CloseFuture;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.impl.VertxInternal;
@@ -49,7 +48,6 @@ import java.io.FileNotFoundException;
 import java.net.ConnectException;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Predicate;
 
 /**
@@ -297,7 +295,8 @@ public class NetClientImpl implements MetricsProvider, NetClient, Closeable {
       Objects.requireNonNull(connectHandler, "No null connectHandler accepted");
       Bootstrap bootstrap = new Bootstrap();
       bootstrap.group(eventLoop);
-      bootstrap.option(ChannelOption.ALLOCATOR, PartialPooledByteBufAllocator.INSTANCE);
+      bootstrap.option(ChannelOption.ALLOCATOR,
+        sslHelper.clientByteBufAllocator(sslChannelProvider.sslContextProvider()));
 
       vertx.transport().configure(options, remoteAddress.isDomainSocket(), bootstrap);
 
--- a/src/main/java/io/vertx/core/net/impl/SSLHelper.java
+++ b/src/main/java/io/vertx/core/net/impl/SSLHelper.java
@@ -11,12 +11,15 @@
 
 package io.vertx.core.net.impl;
 
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.handler.ssl.OpenSsl;
 import io.netty.handler.ssl.SslProvider;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.VertxException;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.buffer.impl.PartialPooledByteBufAllocator;
 import io.vertx.core.http.ClientAuth;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.net.ClientOptionsBase;
@@ -56,6 +59,27 @@ public class SSLHelper {
     CLIENT_AUTH_MAPPING.put(ClientAuth.REQUEST, io.netty.handler.ssl.ClientAuth.OPTIONAL);
     CLIENT_AUTH_MAPPING.put(ClientAuth.NONE, io.netty.handler.ssl.ClientAuth.NONE);
   }
+
+  ByteBufAllocator clientByteBufAllocator(SslContextProvider ctxProvider) {
+    if (usesJDKSSLWithPooledHeapBuffers(ctxProvider)) {
+      return PooledByteBufAllocator.DEFAULT;
+    }
+    return PartialPooledByteBufAllocator.INSTANCE;
+  }
+
+  ByteBufAllocator serverByteBufAllocator(SslContextProvider ctxProvider) {
+    if (!ssl || usesJDKSSLWithPooledHeapBuffers(ctxProvider)) {
+      return PooledByteBufAllocator.DEFAULT;
+    }
+    return PartialPooledByteBufAllocator.INSTANCE;
+  }
+
+  private boolean usesJDKSSLWithPooledHeapBuffers(SslContextProvider ctxProvider) {
+    return ssl && sslEngineOptions instanceof JdkSSLEngineOptions &&
+      ctxProvider.sslProvider() == SslProvider.JDK &&
+      ((JdkSSLEngineOptions) sslEngineOptions).isPooledHeapBuffers();
+  }
+
 
   /**
    * Resolve the ssl engine options to use for properly running the configured options.
@@ -149,11 +173,14 @@ public class SSLHelper {
 
   private class EngineConfig {
 
+    private final SslProvider sslProvider;
     private final SSLOptions sslOptions;
     private final Supplier<SslContextFactory> supplier;
     private final boolean useWorkerPool;
 
-    public EngineConfig(SSLOptions sslOptions, Supplier<SslContextFactory> supplier, boolean useWorkerPool) {
+    public EngineConfig(SslProvider sslProvider, SSLOptions sslOptions, Supplier<SslContextFactory> supplier,
+                        boolean useWorkerPool) {
+      this.sslProvider = sslProvider;
       this.sslOptions = sslOptions;
       this.supplier = supplier;
       this.useWorkerPool = useWorkerPool;
@@ -161,6 +188,7 @@ public class SSLHelper {
 
     SslContextProvider sslContextProvider() {
       return new SslContextProvider(
+        sslProvider,
         clientAuth,
         endpointIdentificationAlgorithm,
         applicationProtocols,
@@ -291,18 +319,26 @@ public class SSLHelper {
       }).compose(v2 -> ctx.<EngineConfig>executeBlockingInternal(p -> {
         Supplier<SslContextFactory> supplier;
         boolean useWorkerPool;
+        SslProvider sslProvider;
         try {
           SSLEngineOptions resolvedEngineOptions = resolveEngineOptions(sslEngineOptions, useAlpn);
           supplier = resolvedEngineOptions::sslContextFactory;
           useWorkerPool = resolvedEngineOptions.getUseWorkerThread();
+          if (resolvedEngineOptions instanceof JdkSSLEngineOptions) {
+            sslProvider = SslProvider.JDK;
+          } else if (resolvedEngineOptions instanceof OpenSSLEngineOptions) {
+            sslProvider = SslProvider.OPENSSL;
+          } else {
+            sslProvider = SslProvider.JDK;
+          }
         } catch (Exception e) {
           p.fail(e);
           return;
         }
-        p.complete(new EngineConfig(sslOptions, supplier, useWorkerPool));
+        p.complete(new EngineConfig(sslProvider, sslOptions, supplier, useWorkerPool));
       })).onComplete(promise);
     } else {
-      sslContextFactorySupplier = Future.succeededFuture(new EngineConfig(sslOptions, () -> new DefaultSslContextFactory(SslProvider.JDK, false), SSLEngineOptions.DEFAULT_USE_WORKER_POOL));
+      sslContextFactorySupplier = Future.succeededFuture(new EngineConfig(SslProvider.JDK, sslOptions, () -> new DefaultSslContextFactory(SslProvider.JDK, false), SSLEngineOptions.DEFAULT_USE_WORKER_POOL));
     }
     return sslContextFactorySupplier;
   }
--- a/src/main/java/io/vertx/core/net/impl/SSLHelper.java
+++ b/src/main/java/io/vertx/core/net/impl/SSLHelper.java
@@ -76,7 +76,7 @@ public class SSLHelper {
 
   private boolean usesJDKSSLWithPooledHeapBuffers(SslContextProvider ctxProvider) {
     return ssl && sslEngineOptions instanceof JdkSSLEngineOptions &&
-      ctxProvider.sslProvider() == SslProvider.JDK &&
+      ctxProvider.jdkSSLProvider() &&
       ((JdkSSLEngineOptions) sslEngineOptions).isPooledHeapBuffers();
   }
 
@@ -173,14 +173,14 @@ public class SSLHelper {
 
   private class EngineConfig {
 
-    private final SslProvider sslProvider;
+    private final boolean jdkSSLProvider;
     private final SSLOptions sslOptions;
     private final Supplier<SslContextFactory> supplier;
     private final boolean useWorkerPool;
 
-    public EngineConfig(SslProvider sslProvider, SSLOptions sslOptions, Supplier<SslContextFactory> supplier,
+    public EngineConfig(boolean jdkSSLProvider, SSLOptions sslOptions, Supplier<SslContextFactory> supplier,
                         boolean useWorkerPool) {
-      this.sslProvider = sslProvider;
+      this.jdkSSLProvider = jdkSSLProvider;
       this.sslOptions = sslOptions;
       this.supplier = supplier;
       this.useWorkerPool = useWorkerPool;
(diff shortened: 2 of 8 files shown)


franz1981:
@cescoffier and @vietj PTAL

I've added some asserts to make it existing tests to fail and make sure the invariant I'm assuming for SSL on vertx are correct.

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -57,6 +61,40 @@ public class SSLHelper {
     CLIENT_AUTH_MAPPING.put(ClientAuth.NONE, io.netty.handler.ssl.ClientAuth.NONE);
   }
 
+  ByteBufAllocator clientByteBufAllocator(SslChannelProvider sslChannelProvider) {
+    assert sslChannelProvider != null && sslChannelProvider.sslContextProvider() != null &&
```
franz1981, line 65: @vietj these are here to compensate to lack of tests here: these should fail in the CI even if the sys prop is not used, so we know that the assumption, while enabled, is broken

franz1981:
@cescoffier this is key for both https and vertx sql client - cause the latter will rely heavily on it for the reactive sql drivers (both mysql and postgres AFAIK) - 

@tsegismont AFAIK vertx sql driver should rely on `NetClientImpl` e.g. https://github.com/eclipse-vertx/vertx-sql-client/blob/master/vertx-sql-client/src/main/java/io/vertx/sqlclient/impl/ConnectionFactoryBase.java#L45 
hence I would expect that the changes on `NetClientImpl`'s allocator should just do "the right thing", wdyt?

franz1981:
After talking with @cescoffier I'm changing this back to use a specific option to enable this feature.

franz1981:
PTAL @vietj 

I've added some test (I could improve thsi further) - but it is not clear to me yet the role of https://vertx.io/docs/apidocs/io/vertx/core/http/HttpClientOptions.html#setSsl-boolean-

I'm running few tests to better understand what's meant to do - because as far as I can tell the `SSLHelper` is created regardless (see my added tests)

src/test/java/io/vertx/core/net/impl/SSLAllocatorTest.java
```diff
@@ -0,0 +1,145 @@
...
+import io.vertx.core.http.HttpServerOptions;
+import io.vertx.core.impl.ContextInternal;
+import io.vertx.core.net.JdkSSLEngineOptions;
+import io.vertx.core.net.OpenSSLEngineOptions;
+import io.vertx.core.net.SSLOptions;
+import io.vertx.test.core.VertxTestBase;
+import io.vertx.test.tls.Cert;
+import io.vertx.test.tls.Trust;
+
+public class SSLAllocatorTest extends VertxTestBase {
```
franz1981, line 31: @cescoffier not yet perfect but decent enough to make sure the default behaviour is still the same
I need to wait till this is merged to add some CI run which enable both this and the unified allocator changes at https://github.com/eclipse-vertx/vert.x/pull/5292 , which is ready for being reviewed too (by @vietj )

franz1981:
@tsegismont 
I need you here bud: how do I benefit from this change into vert-x sql client? 
How we can consume it for vert-x sql client too into Quarkus?

i.e. Here I need users to set  https://vertx.io/docs/apidocs/io/vertx/core/net/TCPSSLOptions.html#setSsl-boolean- and https://vertx.io/docs/apidocs/io/vertx/core/net/TCPSSLOptions.html#setSslEngineOptions-io.vertx.core.net.SSLEngineOptions- (passing a JDK SSL option which enable heap buffer pooling), in order to made vertx to do the "right" thing and avoid https://github.com/quarkusio/quarkus/issues/41880#issuecomment-2258734835 to happen.

tsegismont:
> I need you here bud: how do I benefit from this change into vert-x sql client?

You can set options with `io.vertx.pgclient.PgConnectOptions#setJdkSslEngineOptions`

> How we can consume it for vert-x sql client too into Quarkus?

In Quarkus, you have to customize pool creation, I think.
See https://quarkus.io/guides/reactive-sql-clients#customizing-pool-creation

franz1981:
Sadly, for Quarkus, "supported" users cannot change the SSL engine, so they will be sticked with JDK SSL and the huge heap allocations it bring :"(

franz1981:
Wdyt @cescoffier ? It is ok enough?

Thanks @tsegismont

vietj:
@cescoffier can you validate this one ?

src/main/java/io/vertx/core/net/impl/NetClientImpl.java
```diff
@@ -297,7 +295,8 @@ private void connectInternal2(ProxyOptions proxyOptions,
       Objects.requireNonNull(connectHandler, "No null connectHandler accepted");
       Bootstrap bootstrap = new Bootstrap();
       bootstrap.group(eventLoop);
-      bootstrap.option(ChannelOption.ALLOCATOR, PartialPooledByteBufAllocator.INSTANCE);
+      bootstrap.option(ChannelOption.ALLOCATOR,
+        sslHelper.clientByteBufAllocator(sslChannelProvider.sslContextProvider()));
```
cescoffier, line 299: How to ensure that this does not change the previous behavior?

cescoffier: Ok, saw the next class.

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -149,18 +173,22 @@ private static class CachedProvider {
...
 
+    private final SslProvider sslProvider;
     private final SSLOptions sslOptions;
     private final Supplier<SslContextFactory> supplier;
     private final boolean useWorkerPool;
 
-    public EngineConfig(SSLOptions sslOptions, Supplier<SslContextFactory> supplier, boolean useWorkerPool) {
+    public EngineConfig(SslProvider sslProvider, SSLOptions sslOptions, Supplier<SslContextFactory> supplier,
+                        boolean useWorkerPool) {
+      this.sslProvider = sslProvider;
```
cescoffier, line 183: We will have to see if this has native complexities.

src/main/java/io/vertx/core/net/impl/SSLHelper.java
```diff
@@ -291,18 +319,26 @@ private Future<EngineConfig> build(SSLOptions sslOptions, ContextInternal ctx) {
       }).compose(v2 -> ctx.<EngineConfig>executeBlockingInternal(p -> {
         Supplier<SslContextFactory> supplier;
         boolean useWorkerPool;
+        SslProvider sslProvider;
         try {
           SSLEngineOptions resolvedEngineOptions = resolveEngineOptions(sslEngineOptions, useAlpn);
           supplier = resolvedEngineOptions::sslContextFactory;
           useWorkerPool = resolvedEngineOptions.getUseWorkerThread();
+          if (resolvedEngineOptions instanceof JdkSSLEngineOptions) {
```
cescoffier, line 327: This would need a substitution in native. Can we move this to a separated, easily substituable method.

cescoffier (changes requested):

src/main/java/io/vertx/core/net/impl/SSLHelper.java
franz1981: I've now performed a single check on an already present option now  i.e. `JdkSSLEngineOptions`
and removed the one which is related OpenSSL, it looks better?
I'm using this additional time to add a further test, as suggested by @vietj

franz1981:
@vietj can you paste the additional test you found in our call, bud?

vietj:
I think you want to look at NetTest with NetSocketInternal

franz1981:
Drafting till fixing the latest tests

franz1981:
many thanks @vietj and @cescoffier for this hard round of review: you were right, there's something terrible hidden on SSL...

In a3393ace29680911e11cb8b6705c2369964dbf98 I've put some comment for some shocking outcome, and I think we need to have a sync before moving this forward - which BTW is not enabled by default eh

franz1981:
While working on this PR I've noticed something odd, exactly 2 things:
1. the original Vertx code which was setting `PartialByteBufAllocator.INSTANCE` as Netty allocator for server + SSL or client (SSL or NOT) hides something fishy which looks just plain wrong: given that `PartialByteBufAllocator.INSTANCE` overrides the `buffer()` method which forces to allocates heap unpooled buffers and such method is used used by https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java#L328 - it means that Netty, while sending it to the wire has to create an additional copy off-heap, because all transport layers requires instead direct buffer
2. Both the original vertx code or this PR are still hitting some slow-path on the SSL JDK engine, which requires the JDK engine to allocates on the flight some `byte[]`. This code path is shown below

```
	at com.sun.crypto.provider.GCTR.update(GCTR.java:212)
	at com.sun.crypto.provider.GCTR.doFinal(GCTR.java:291)
	at com.sun.crypto.provider.GaloisCounterMode$DecryptOp.doFinal(GaloisCounterMode.java:1870)
	at com.sun.crypto.provider.GaloisCounterMode$GCMEngine.doLastBlock(GaloisCounterMode.java:934)
	at com.sun.crypto.provider.GaloisCounterMode$GCMDecrypt.doFinal(GaloisCounterMode.java:1611)
	at com.sun.crypto.provider.GaloisCounterMode.engineDoFinal(GaloisCounterMode.java:458)
	at javax.crypto.Cipher.doFinal(Cipher.java:2543)
	at sun.security.ssl.SSLCipher$T13GcmReadCipherGenerator$GcmReadCipher.decrypt(SSLCipher.java:1908)
	at sun.security.ssl.SSLEngineInputRecord.decodeInputRecord(SSLEngineInputRecord.java:239)
	at sun.security.ssl.SSLEngineInputRecord.decode(SSLEngineInputRecord.java:196)
	at sun.security.ssl.SSLEngineInputRecord.decode(SSLEngineInputRecord.java:159)
	at sun.security.ssl.SSLTransport.decode(SSLTransport.java:111)
	at sun.security.ssl.SSLEngineImpl.decode(SSLEngineImpl.java:736)
	at sun.security.ssl.SSLEngineImpl.readRecord(SSLEngineImpl.java:691)
	at sun.security.ssl.SSLEngineImpl.unwrap(SSLEngineImpl.java:506)
	at sun.security.ssl.SSLEngineImpl.unwrap(SSLEngineImpl.java:482)
	at javax.net.ssl.SSLEngine.unwrap(SSLEngine.java:679)
	at io.netty.handler.ssl.SslHandler$SslEngineType$3.unwrap(SslHandler.java:308)
	at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1443)
	at io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1336)
	at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1385)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:530)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:469)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:290)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1407)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:440)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:918)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:166)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:788)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:724)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:650)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:562)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:994)
	at java.lang.Thread.runWith(Thread.java:1596)
	at java.lang.Thread.run(Thread.java:1583)
```
And it refers to https://github.com/openjdk/jdk21/blob/890adb6410dab4606a4f26a942aed02fb2f55387/src/java.base/share/classes/com/sun/crypto/provider/GCTR.java#L212

which shows, for `GCTR` that the input buffer is not an heap one - maybe is a pretty normal case, but I'll report this here, for further analysis.

cescoffier (approved):
Before merging, I would like to ensure it does not change anything for the native compilation.

zakkak:
> Before merging, I would like to ensure it does not change anything for the native compilation.

@cescoffier could you elaborate on what your concerns are for native compilation?

Looking at the changes I don't expect any drastic changes. Trying the change with Quarkus' `vertx` native IT I am getting

```json
{
  "types": {
    "total": 16096,
    "reflection": 4455,
    "jni": 62,
    "reachable": 14093
  },
  "methods": {
    "foreign_downcalls": -1,
    "total": 125514,
    "reflection": 3555,
    "jni": 55,
    "reachable": 72175
  },
  "classes": {
    "total": 16096,
    "reflection": 4455,
    "jni": 62,
    "reachable": 14093
  },
  "fields": {
    "total": 36087,
    "reflection": 138,
    "jni": 67,
    "reachable": 21524
  }
}
``` 

vs 

```json
{
  "types": {
    "total": 16075,
    "reflection": 4451,
    "jni": 62,
    "reachable": 14072
  },
  "methods": {
    "foreign_downcalls": -1,
    "total": 125400,
    "reflection": 3550,
    "jni": 55,
    "reachable": 72092
  },
  "classes": {
    "total": 16075,
    "reflection": 4451,
    "jni": 62,
    "reachable": 14072
  },
  "fields": {
    "total": 36038,
    "reflection": 138,
    "jni": 67,
    "reachable": 21480
  }
}

```

with 4.5.10

which is reflected to an increase of 36kb in the native executable size.

However, I suspect that this IT test doesn't enable SSL. Any hints how I could/should test this with SSL enabled?

zakkak:
Thank you @cescoffier. I am appending the results from `vertx-http`.

**TLDR:** LGTM.

### vertx-core 4.x (tip of vert.x without this PR)

```json
  "types": {
    "total": 14411,
    "reflection": 4023,
    "jni": 62,
    "reachable": 12484
  },
  "methods": {
    "foreign_downcalls": -1,
    "total": 113204,
    "reflection": 3337,
    "jni": 55,
    "reachable": 64071
  },
  "fields": {
    "total": 33089,
    "reflection": 133,
    "jni": 67,
    "reachable": 19472
  }
```

### this PR

```json
  "types": {
    "total": 14419,
    "reflection": 4023,
    "jni": 62,
    "reachable": 12491
  },
  "methods": {
    "foreign_downcalls": -1,
    "total": 113234,
    "reflection": 3338,
    "jni": 55,
    "reachable": 64081
  },
  "fields": {
    "total": 33097,
    "reflection": 133,
    "jni": 67,
    "reachable": 19479
  }
```

### this PR with pooled heap buffers enabled

```json
  "types": {
    "total": 14410,
    "reflection": 4023,
    "jni": 62,
    "reachable": 12483
  },
  "methods": {
    "foreign_downcalls": -1,
    "total": 113201,
    "reflection": 3337,
    "jni": 55,
    "reachable": 64064
  },
  "fields": {
    "total": 33085,
    "reflection": 133,
    "jni": 67,
    "reachable": 19471
  }
```

To enable the pooled heap buffers I patched Quarkus with:

```diff
diff --git a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/options/HttpServerOptionsUtils.java b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/options/HttpServerOptionsUtils.java
index 36fe46a9d3a..872d79a85cb 100644
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/options/HttpServerOptionsUtils.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/options/HttpServerOptionsUtils.java
@@ -82,6 +82,7 @@ public static HttpServerOptions createSslOptions(HttpBuildTimeConfig buildTimeCo
                 serverOptions.setAlpnVersions(Arrays.asList(HttpVersion.HTTP_2, HttpVersion.HTTP_1_1));
             }
         }
+        serverOptions.setSslEngineOptions(new JdkSSLEngineOptions().setPooledHeapBuffers(true));
         setIdleTimeout(httpConfiguration, serverOptions);
 
         TlsConfiguration bucket = getTlsConfiguration(httpConfiguration.tlsConfigurationName, registry);
```

cescoffier:
Thanks @zakkak ! 

All good for me then!

franz1981:
so @vietj here we are: this can be moved forward and I'll create the relevant issues on the quarkus repo for the missing bits to make it available to use there.

I strongly suggest @vietj that vertx users which relies on JDK SSL will enable this feature as well, TBH, as performance with JDK SSL is terrible without it...

vietj:
thanks for the contrib @franz1981

PR #5338, opened 2024-10-01
Provide access to the default redirect handler

geoand:
_Motivation:_

This is useful for building redirect handlers
for clients that need to delegate to the default behavior. In Quarkus we will leverage this to get rid of the deprecated (and since removed) [calls](https://github.com/quarkusio/quarkus/blob/3.15.1/independent-projects/resteasy-reactive/client/runtime/src/main/java/org/jboss/resteasy/reactive/client/impl/ClientImpl.java#L209-L210) to redirectHandler

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/HttpClientAgent.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/HttpClientAgent.java
@@ -1,9 +1,12 @@
 package io.vertx.core.http;
 
+import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.Future;
+import io.vertx.core.http.impl.DefaultRedirectHandler;
 import io.vertx.core.metrics.Measured;
 import io.vertx.core.net.ClientSSLOptions;
+import java.util.function.Function;
 
 /**
  * An asynchronous HTTP client.
@@ -33,6 +36,13 @@ import io.vertx.core.net.ClientSSLOptions;
  */
 @VertxGen
 public interface HttpClientAgent extends HttpClient, Measured {
+
+  /**
+   * Constant containing the default redirect handler of used by the client.
+   * This is useful for building redirect handlers that need to delegate to the default behavior.
+   */
+  @GenIgnore
+  Function<HttpClientResponse, Future<RequestOptions>> DEFAULT_REDIRECT_HANDLER = new DefaultRedirectHandler();
 
   /**
    * <p>Update the client with new SSL {@code options}, the update happens if the options object is valid and different
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/DefaultRedirectHandler.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/DefaultRedirectHandler.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2011-2024 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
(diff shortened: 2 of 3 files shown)


vertx-core/src/main/java/io/vertx/core/http/HttpClient.java
```diff
@@ -113,4 +118,12 @@ default Future<Void> close() {
    */
   Future<Void> shutdown(long timeout, TimeUnit unit);
 
+  /**
+   * Provides a way for users of the client to get access to the default redirect handler.
+   * This is useful for building redirect handlers that need to delegate to the default behavior.
+   */
+  @GenIgnore
+  default Function<HttpClientResponse, Future<RequestOptions>> defaultRedirectHandler() {
```
vietj, line 126: rather an instance of on `HttpClientAgent` instead, in Vert.x 5, HttpClient is the API for interacting with an HTTP server whereas `HttpClientAgent` is the client.

geoand: Good to know! PR updated

vertx-core/src/main/java/io/vertx/core/http/HttpClientAgent.java
```diff
@@ -72,4 +74,13 @@ default Future<Boolean> updateSSLOptions(ClientSSLOptions options) {
    */
   Future<HttpClientConnection> connect(HttpConnectOptions options);
 
+  /**
+   * Provides a way for users of the client to get access to the default redirect handler.
+   * This is useful for building redirect handlers that need to delegate to the default behavior.
+   */
+  @GenIgnore
+  default Function<HttpClientResponse, Future<RequestOptions>> defaultRedirectHandler() {
```
vietj, line 82: make it instead a constant `DEFAULT_REDIRECT_HANDLER` in this interface

vertx-core/src/main/java/io/vertx/core/http/DefaultRedirectHandler.java
```diff
@@ -0,0 +1,71 @@
+package io.vertx.core.http;
+
+import static io.vertx.core.http.HttpHeaders.CONTENT_LENGTH;
+import io.vertx.core.Future;
+import io.vertx.core.http.impl.HttpUtils;
+import java.net.URI;
+import java.util.function.Function;
+
+
+class DefaultRedirectHandler implements Function<HttpClientResponse, Future<RequestOptions>> {
```
vietj, line 10: I prefer to keep this in impl even though it is public, it is part of implementation and JPMS makes it encapsulated

geoand:
PR updated as per review comments

vertx-core/src/main/java/io/vertx/core/http/HttpClientAgent.java
```diff
@@ -34,6 +37,9 @@
 @VertxGen
 public interface HttpClientAgent extends HttpClient, Measured {
 
+  @GenIgnore
+  Function<HttpClientResponse, Future<RequestOptions>> DEFAULT_REDIRECT_HANDLER = new DefaultRedirectHandler();
```
vietj, line 41: Documentation :-)

vertx-core/src/main/java/io/vertx/core/http/impl/DefaultRedirectHandler.java
```diff
@@ -0,0 +1,72 @@
+package io.vertx.core.http.impl;
```
vietj, line 1: License header

vertx-core/src/main/java/io/vertx/core/http/HttpClientAgent.java
```diff
@@ -72,4 +81,13 @@ default Future<Boolean> updateSSLOptions(ClientSSLOptions options) {
    */
   Future<HttpClientConnection> connect(HttpConnectOptions options);
 
+  /**
+   * Provides a way for users of the client to get access to the default redirect handler.
+   * This is useful for building redirect handlers that need to delegate to the default behavior.
+   */
```
vietj, line 87: I prefer not having this method, users should care about the handler when using the builder

geoand: Removed

vietj:
thanks @geoand

geoand:
🙏🏽

PR #5346, opened 2024-10-08
Add peer host and port info for server SslHandler (4.x)

### LINKED ISSUE #5290: Missing peer host and port info in SSLEngine for server SslHandler
### Version

4.4.9

### Context

We have a customized key manager that extends `X509ExtendedKeyManager` that want to override the `public String chooseEngineServerAlias(String keyType, Principal[] issuers, SSLEngine engine)` method to choose the server alias partly depending on the peer host address.
However, the `engine.getPeerHost()` always returns `null`.

After read related code, I find that netty `SslContext.newHandler` do support passing in an advisory peer information of peer host and port.
However, in vert.x, when creating `SslHandler` for server in `SslChannelProvider`, the peer host and port info is not passed to `SslContext.newHandler`, result in `null` for `engine.getPeerHost()` in `X509ExtendedKeyManager.chooseEngineServerAlias`.
(The `SslChannelProvider` do provide peer host and port info when creating client `SslHandler`)

I tried to pass the peer host and port info from `HttpServerWorker` to `SslChannelProvider.createServerHandler` and find the peer host and port are available in the `SSLEngine` in `X509ExtendedKeyManager.chooseEngineServerAlias`:
```diff
diff --git a/src/main/java/io/vertx/core/http/impl/HttpServerWorker.java b/src/main/java/io/vertx/core/http/impl/HttpServerWorker.java
index cf37c4e8b..82402c858 100644
--- a/src/main/java/io/vertx/core/http/impl/HttpServerWorker.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerWorker.java
@@ -35,6 +35,8 @@ import io.vertx.core.impl.VertxInternal;
import io.vertx.core.net.impl.*;
import io.vertx.core.spi.metrics.HttpServerMetrics;

+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
import java.nio.charset.StandardCharsets;
import java.util.List;
import java.util.function.BiConsumer;
@@ -131,7 +133,12 @@ public class HttpServerWorker implements BiConsumer<Channel, SslChannelProvider>
   private void configurePipeline(Channel ch, SslChannelProvider sslChannelProvider) {
     ChannelPipeline pipeline = ch.pipeline();
     if (options.isSsl()) {
-      pipeline.addLast("ssl", sslChannelProvider.createServerHandler());
+      SocketAddress remoteAddress = ch.remoteAddress();
+      if (remoteAddress instanceof InetSocketAddress) {
+        pipeline.addLast("ssl", sslChannelProvider.createServerHandler(((InetSocketAddress) remoteAddress).getHostString(), ((InetSocketAddress) remoteAddress).getPort()));
+      } else {
+        pipeline.addLast("ssl", sslChannelProvider.createServerHandler());
+      }
       ChannelPromise p = ch.newPromise();
       pipeline.addLast("handshaker", new SslHandshakeCompletionHandler(p));
       p.addListener(future -> {
diff --git a/src/main/java/io/vertx/core/net/impl/SslChannelProvider.java b/src/main/java/io/vertx/core/net/impl/SslChannelProvider.java
index 290bf8c23..cb5aba5d1 100644
--- a/src/main/java/io/vertx/core/net/impl/SslChannelProvider.java
+++ b/src/main/java/io/vertx/core/net/impl/SslChannelProvider.java
@@ -144,17 +144,21 @@ public class SslChannelProvider {
   }

   public ChannelHandler createServerHandler() {
+    return createServerHandler(null, -1);
+  }
+
+  public ChannelHandler createServerHandler(String peerHost, int peerPort) {
     if (sni) {
       return createSniHandler();
     } else {
-      return createServerSslHandler(useAlpn);
+      return createServerSslHandler(useAlpn, peerHost, peerPort);
     }
   }

-  private SslHandler createServerSslHandler(boolean useAlpn) {
+  private SslHandler createServerSslHandler(boolean useAlpn, String peerHost, int peerPort) {
     SslContext sslContext = sslServerContext(useAlpn);
     Executor delegatedTaskExec = useWorkerPool ? workerPool : ImmediateExecutor.INSTANCE;
-    SslHandler sslHandler = sslContext.newHandler(ByteBufAllocator.DEFAULT, delegatedTaskExec);
+    SslHandler sslHandler = sslContext.newHandler(ByteBufAllocator.DEFAULT, peerHost, peerPort, delegatedTaskExec);
     sslHandler.setHandshakeTimeout(sslHandshakeTimeout, sslHandshakeTimeoutUnit);
     return sslHandler;
   }

```

There are a few other places calling `SslChannelProvider.createServerHandler` so although this change works in my use case, a more complete fix may be needed.

ben1222:
Fixes #5290 , on 4.x branch.

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/SslChannelProvider.java
+++ b/src/main/java/io/vertx/core/net/impl/SslChannelProvider.java
@@ -18,6 +18,7 @@ import io.netty.handler.ssl.SslHandler;
 import io.netty.util.AsyncMapping;
 import io.netty.util.concurrent.ImmediateExecutor;
 import io.vertx.core.VertxException;
+import io.vertx.core.net.HostAndPort;
 import io.vertx.core.net.SocketAddress;
 
 import javax.net.ssl.KeyManagerFactory;
@@ -143,25 +144,30 @@ public class SslChannelProvider {
     return sslHandler;
   }
 
-  public ChannelHandler createServerHandler() {
+  public ChannelHandler createServerHandler(HostAndPort remoteAddress) {
     if (sni) {
-      return createSniHandler();
+      return createSniHandler(remoteAddress);
     } else {
-      return createServerSslHandler(useAlpn);
+      return createServerSslHandler(useAlpn, remoteAddress);
     }
   }
 
-  private SslHandler createServerSslHandler(boolean useAlpn) {
+  private SslHandler createServerSslHandler(boolean useAlpn, HostAndPort remoteAddress) {
     SslContext sslContext = sslServerContext(useAlpn);
     Executor delegatedTaskExec = useWorkerPool ? workerPool : ImmediateExecutor.INSTANCE;
-    SslHandler sslHandler = sslContext.newHandler(ByteBufAllocator.DEFAULT, delegatedTaskExec);
+    SslHandler sslHandler;
+    if (remoteAddress != null) {
+      sslHandler = sslContext.newHandler(ByteBufAllocator.DEFAULT, remoteAddress.host(), remoteAddress.port(), delegatedTaskExec);
+    } else {
+      sslHandler = sslContext.newHandler(ByteBufAllocator.DEFAULT, delegatedTaskExec);
+    }
     sslHandler.setHandshakeTimeout(sslHandshakeTimeout, sslHandshakeTimeoutUnit);
     return sslHandler;
   }
 
-  private SniHandler createSniHandler() {
+  private SniHandler createSniHandler(HostAndPort remoteAddress) {
     Executor delegatedTaskExec = useWorkerPool ? workerPool : ImmediateExecutor.INSTANCE;
-    return new VertxSniHandler(serverNameMapping(), sslHandshakeTimeoutUnit.toMillis(sslHandshakeTimeout), delegatedTaskExec);
+    return new VertxSniHandler(serverNameMapping(), sslHandshakeTimeoutUnit.toMillis(sslHandshakeTimeout), delegatedTaskExec, remoteAddress);
   }
 
   private static int idx(boolean useAlpn) {
--- a/src/main/java/io/vertx/core/http/impl/HttpServerWorker.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpServerWorker.java
@@ -129,7 +129,7 @@ public class HttpServerWorker implements BiConsumer<Channel, SslChannelProvider>
   private void configurePipeline(Channel ch, SslChannelProvider sslChannelProvider) {
     ChannelPipeline pipeline = ch.pipeline();
     if (options.isSsl()) {
-      pipeline.addLast("ssl", sslChannelProvider.createServerHandler());
+      pipeline.addLast("ssl", sslChannelProvider.createServerHandler(HttpUtils.socketAddressToHostAndPort(ch.remoteAddress())));
       ChannelPromise p = ch.newPromise();
       pipeline.addLast("handshaker", new SslHandshakeCompletionHandler(p));
       p.addListener(future -> {
--- a/src/main/java/io/vertx/core/http/impl/HttpUtils.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpUtils.java
@@ -40,6 +40,8 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.RandomAccessFile;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
@@ -1077,5 +1079,21 @@ public final class HttpUtils {
       }
     }
     return false;
+  }
+
+  /**
+   * Convert a {@link SocketAddress} to a {@link HostAndPort}.
+   * If the socket address is an {@link InetSocketAddress}, the hostString and port are used.
+   * Otherwise {@code null} is returned.
+   *
+   * @param socketAddress The socket address to convert
+   * @return The converted instance or {@code null} if not applicable.
+   */
+  public static HostAndPort socketAddressToHostAndPort(SocketAddress socketAddress) {
+    if (socketAddress instanceof InetSocketAddress) {
+      InetSocketAddress inetSocketAddress = (InetSocketAddress) socketAddress;
+      return new HostAndPortImpl(inetSocketAddress.getHostString(), inetSocketAddress.getPort());
+    }
+    return null;
   }
 }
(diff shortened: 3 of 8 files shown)


src/main/java/io/vertx/core/net/impl/SslChannelProvider.java
```diff
@@ -143,25 +144,31 @@ public SslHandler createClientSslHandler(SocketAddress remoteAddress, String ser
     return sslHandler;
   }
 
-  public ChannelHandler createServerHandler() {
+  public ChannelHandler createServerHandler(java.net.SocketAddress remoteAddress) {
```
vietj, line 147: rather use Vert.x `HostAndPort` type rather than a socket address

vietj:
looks good but I'd prefer use vertx type `HostAndPort` which is more suited for this internally

src/main/java/io/vertx/core/net/impl/SslChannelProvider.java
ben1222: @vietj Thanks for the comment!
The `HostAndPort` seems currently is used to parse/set host/authority in HTTP header, which looks like a totally different use case...
Here we want to pass the socket remote address info to the `SSLEngine` for `X509ExtendedKeyManager` to use.
Are you sure `HostAndPort` is preferred here?
If use `HostAndPort`, I may need to add a conversion method like `@Nullable static HostAndPort socketHostAndPort(SocketAddress address);` to simplify the conversion, is it OK to put it in `HostAndPort` interface (which will become a new external API)?

vietj: it seems in this example we are using SocketAddress only to carry and host and port info, aren't we ?

ben1222:
@vietj Hmm... I'm not sure if it worth to do this extra conversion... it adds overhead...
What's the benefit? Holding minimal required data? What else?

Anyway, I'll try to use `HostAndPort` and update PR for you to review.

ben1222:
@vietj I updated the PR to use `HostAndPort`, please review again.

ben1222:
Hmm... looks like the vertx gen failed... sorry that I didn't notice it when testing in my IDE... seems it didn't do the vertx gen...
I added `@GenIgnore(GenIgnore.PERMITTED_TYPE)` on `HostAndPort.fromSocketAddress` to fix that problem.
@vietj please review again.

ben1222:
@vietj Could you review? If it looks good to you, I'll update the PR for master branch to also use the `HostAndPort`.

src/main/java/io/vertx/core/net/HostAndPort.java
```diff
@@ -55,6 +60,24 @@ static HostAndPort authority(String host) {
...
+  /**
+   * Convert a {@link SocketAddress} to a {@link HostAndPort}.
+   * If the socket address is an {@link InetSocketAddress}, the hostString and port are used.
+   * Otherwise {@code null} is returned.
+   *
+   * @param socketAddress The socket address to convert
+   * @return The converted instance or {@code null} if not applicable.
+   */
+  @Nullable
+  @GenIgnore(GenIgnore.PERMITTED_TYPE)
```
vietj, line 72: this should use instead directly inet socket address and not assume we accept any socket address and instead return null

ben1222: @vietj This method is added to simplify the caller (no need to add if-else to check if the `SocketAddress` is an `InetSocketAddress` in caller side)
If making the parameter to be `InetSocketAddress` then this method is less useful and I'd rather move this method to somewhere else.
(I'm still thinking passing `InetSocketAddress` directly to `SslChannelProvider` is the simplest and having minimal overhead)
What do you think?

vietj: the issue is that we prefer to avoid methods that perform the null check, the caller should not rely on the method to return null when passing an inapropriate argument, instead it should check the argument and perform the work itself, eventually it can be in an internal util class, but we should not encourage nor promote this on a public API.

ben1222: @vietj You didn't object my previous comment https://github.com/eclipse-vertx/vert.x/pull/5346#discussion_r1794552108 about returning the nullable on public API so I assumed you agree on that:
> If use `HostAndPort`, I may need to add a conversion method like `@Nullable static HostAndPort socketHostAndPort(SocketAddress address);` to simplify the conversion, is it OK to put it in `HostAndPort` interface (which will become a new external API)?

I don't understand your expectation and reason behind it now.
You asked to use `HostAndPort` instead of `SocketAddress`, while not allowing adding a method that returns nullable for public API.
If like you said, the caller should check the argument, then it will become:
```
// in HttpServerWorker.configurePipeline
SocketAddress remoteAddress = ch.remoteAddress();
HostAndPort remoteHostAndPort = (remoteAddress instanceof InetSocketAddress ? HostAndPort.fromSocketAddress(remoteAddress) : null);
pipeline.addLast("ssl", sslChannelProvider.createServerHandler(remoteHostAndPort));

// in NetServerImpl.NetServerWorker.configurePipeline
SocketAddress remoteAddress = ch.remoteAddress();
HostAndPort remoteHostAndPort = (remoteAddress instanceof InetSocketAddress ? HostAndPort.fromSocketAddress(remoteAddress) : null);
ch.pipeline().addLast("ssl", sslChannelProvider.createServerHandler(remoteHostAndPort));

// in NetSocketImpl.upgradeToSsl
SocketAddress remoteAddress = chctx.channel().remoteAddress();
HostAndPort remoteHostAndPort = (remoteAddress instanceof InetSocketAddress ? HostAndPort.fromSocketAddress(remoteAddress) : null);
sslHandler = sslChannelProvider.createServerHandler(remoteHostAndPort);
```
This will result in duplicated code and increase maintenance effort, so I don't think it is a good option.

Another option may be moving this method to an internal API, e.g., move to `SslChannelProvider` or other internal util class.
If moving to `SslChannelProvider`, then why do we need brother changing `public ChannelHandler createServerHandler(java.net.SocketAddress remoteAddress)` to `public ChannelHandler createServerHandler(HostAndPort remoteAddress)`?
If move to another internal util class, could you advice a proper place to put this `SocketAddress` to `HostAndPort` conversion method that may return null?

ben1222: I looked around and find `HttpUtils`, I'll try move this method to there.

ben1222:
@vietj I moved the conversion method to `HttpUtils` and updated PR, please review.

vietj:
@ben1222 thank you, can you port the changes to master ?

ben1222:
@vietj Thank you, I updated the PR for master branch (#5347), please review.
By the way, I'm wondering if 4.5.11 will be available before the end of November?

vietj:
@ben1222 it sounds reasonnable to do early november a 4.5.11 release given that there is a fair amount of issues which have been fixed

PR #5348, opened 2024-10-10
Save recursive locks on websocket writeFrame

franz1981:
Vertx is using a nested series of synchronized over the connection field which costs keep on piling-up; this cost is particularly evident while executing writeFrame outside of I/O threads.

https://github.com/franz1981/java-puzzles/blob/583d468a58a6ecaa5e7c7c300895392638f688dd/src/main/java/red/hat/puzzles/concurrent/LockCoarsening.java#L76-L85

contains some microbenchmarks results which show how bad the performance is even in the uncontended case, and it can just become worse under contention.

This is addressing https://github.com/quarkusio/quarkus/issues/39148#issuecomment-2396314537 
by removing the nested serie of synchronized over connection, reducing them to the bare minimum i.e. just 2 in the best case scenario

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/ServerWebSocketImpl.java
+++ b/src/main/java/io/vertx/core/http/impl/ServerWebSocketImpl.java
@@ -21,13 +21,14 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Promise;
-import io.vertx.codegen.annotations.Nullable;
-import io.vertx.core.*;
 import io.vertx.core.http.ServerWebSocket;
 import io.vertx.core.http.WebSocketFrame;
+import io.vertx.core.http.impl.ws.WebSocketFrameImpl;
 import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.net.HostAndPort;
 import io.vertx.core.spi.metrics.HttpServerMetrics;
+
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 import static io.netty.handler.codec.http.HttpResponseStatus.*;
 import static io.vertx.core.http.impl.HttpUtils.*;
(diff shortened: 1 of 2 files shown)


vietj:
are we trying to optimize WebSocket handhsake here ? or fix bug ?

vietj:
maybe that is a smell and we should not have a reject method for a WebSocket, i.e any WebSocket processed should be already accepted

vietj:
still not clear what it changes , can you clearly describe what is changed and what is the motivation (like in netty) without referring to a quarkus comment ?

mkouba:
So my understanding of the new `ServerWebSocketImpl#tryHandshake()` impl is that it attempts to read the `volatile` field first and if not `null`  then no `synchronized` block is needed. This is a common pattern and it should not break anything ;-).

I'm not so sure about the `AtomicReferenceFieldUpdater#lazySet()` used to update the status. @franz1981 what's the diff compared to `AtomicReferenceFieldUpdater#set()`?

`WebSocketImplBase#unsafeWriteFrame()` and `WebSocketImplBase#unsafeIsClosed()` are meant to avoid one unnecessary nested `synchronized` block, right?

> maybe that is a smell and we should not have a reject method for a WebSocket, i.e any WebSocket processed should be already accepted

Well, maybe but you can't just remove `ServerWebSocket#reject()` in 4.x, right?

> and what is the motivation (like in netty) without referring to a quarkus comment ?

I believe that the motivation is to speedup writing messages to a server WebSocket from a non-event loop thread (e.g. from a worker thread) because as stated in the javadoc of `io.vertx.core.http.impl.ServerWebSocketImpl`: _"This class is optimised for performance when used on the same event loop. However it can be used safely from other threads."_.

franz1981:
> This is a common pattern and it should not break anything

Yep; for the same purpose of the `unsafe` methods: to save nested `synchronized` calls - which are costly (!).

> I'm not so sure about the AtomicReferenceFieldUpdater#lazySet() used to update the status. @franz1981 what's the diff compared to AtomicReferenceFieldUpdater#set()?

Here: https://psy-lob-saw.blogspot.com/2012/12/atomiclazyset-is-performance-win-for.html
This is still valuable right now as well: on JMM 9 it has been replaced with a more meaningful name i.e. `VarHandle::setRelease` ; and it still have happens-before guarantees, just with a cheaper cost (and loosing sequential consistency, which we don't care here).

> WebSocketImplBase#unsafeWriteFrame() and WebSocketImplBase#unsafeIsClosed() are meant to avoid one unnecessary nested synchronized block, right?

Yep, that's the main purpose: the `assert` there are indeed to shield developers from misuses and make clear the expected invariants: I usually name those `unguardedSomething` but is not a much better name TBH.

@vietj PTAL

PR #5353, opened 2024-10-12
Support reuse current eventLoop for Deployment

dreamlike-ocean:
add new ThreadingModel to support reuse current eventLoop for deploying verticle

Motivation:

Sometimes we expect to deploy child verticles using the current EventLoop. This can ensure the affinity between multiple child verticles and the parent verticle.

Moreover, when deploying a large number of verticles, if the current EventLoop cannot be reused, it will generate a large amount of system call overhead caused by selector::wakeup.

directly deploy verticle:
directly deploy verticle

reuse eventLoop for deploying verticle
reuse eventLoop for deploying verticle

So I provided a new ThreadingModel enum and made some compatibility. When creating an eventLoopContext, the EventLoop corresponding to the current Context will be used as the EventLoop of the new eventLoopContext.

Conformance:
I have signed  Eclipse Contributor Agreement.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -187,7 +187,9 @@ public class Http1xClientConnection extends Http1xConnection implements HttpClie
       }
     }
     if (!headers.contains(HOST)) {
-      request.headers().set(HOST, authority.toString(ssl));
+      if (authority != null) {
+        request.headers().set(HOST, authority.toString(ssl));
+      }
     } else {
       headers.remove(TRANSFER_ENCODING);
     }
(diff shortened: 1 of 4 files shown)


vertx-core/src/main/java/io/vertx/core/ThreadingModel.java
```diff
@@ -22,6 +22,11 @@ public enum ThreadingModel {
    */
   EVENT_LOOP,
 
+  /**
+   * Tasks are scheduled on the current event-loop thread.
+   */
+  CURRENT_EVENT_LOOP,
+
```
vietj, line 29: current event loop is not a threading model

vietj:
I think we should find another way to provide this, what is your actual use case ?

dreamlike-ocean:
> I think we should find another way to provide this, e.g. a deployment option boolean that says to reuse the current event loop or something else

I think this is a good idea.let me try

dreamlike-ocean:
I have reimplemented it according to your idea. Please take a look again.

vietj:
@dreamlike-ocean thanks I am still wondering if that should not be done diffently using a deployment builder like structure that would allow to pass  a context explicitely instead

dreamlike-ocean:
> @dreamlike-ocean thanks I am still wondering if that should not be done diffently using a deployment builder like structure that would allow to pass a context explicitely instead

What if it is implemented using a special verticle type? For example, we can provide an “abstract class ResuseEventLoopVerticle extends AbstractVerticle” type?

PR #5356, opened 2024-10-13
Set request context before recording metrics

### LINKED ISSUE #5355: Http1xServerConnection doesn't set the Vert.x context when recording metrics
### Version

4.5.10 (relevant code seems to be unchanged in main)

### Context

As part of Keycloak 26, I'm running Quarkus 3.15.1 which include Vert.x 4.5.10. When I enable Quarkus Metrics and OpenTelemetry, I expect Exemplars to be recorded. Unfortunately that doesn't happen, as `Http1xServerConnection` doesn't set the context, and `OpentelemetryExemplarSamplerProvider` expects it to be set.

See https://github.com/quarkusio/quarkus/issues/43845

### Do you have a reproducer?

https://github.com/ahus1/quarkus-otel-missing-examplar-example

### Steps to reproduce

See https://github.com/quarkusio/quarkus/issues/43845

1. Start Keycloak 26.0.0 and enable Metrics and OpenTelementry
2. Login to issue some HTTP requests
3. Retrieve metrics from the metrics endpoint using CURL

### Extra

Running on JDK 17 on Windows. Seems be independent of JDK version and Windows. 

In a locally patched version, I was able to fix it by updating 

https://github.com/eclipse-vertx/vert.x/blob/1e2fdf820be9628af2576c0f32828802111855b1/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java#L265-L270

to wrap the calls for metrics in ` request.context.dispatch()`. 

I'll prepare a draft PR to discuss this. As this is named `Http1x`, I suspect there might be other code for HTTP/2.0

ahus1:
Closes #5355

Motivation:

This sets the context so that the logic in Quarkus will know that this call is traced via OpenTracing, and then records and exemplar for the metrics.

I've been monkey-patching Keycloak 26 with a changed `Http1xServerConnection#reportResponseComplete()` and it worked for me in in a manual test. 

I've been looking for other places of `metrics.requestReset()` and `metrics.responseEnd()` and fixing them the same way.

~Draft PR for now, as I didn't add any tests, and would probably need help writing some if they would be required. Also waiting for a results that the build returns.~

Conformance:

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xServerConnection.java
@@ -262,12 +262,14 @@ public class Http1xServerConnection extends Http1xConnection implements HttpServ
   private void reportResponseComplete() {
     Http1xServerRequest request = responseInProgress;
     if (metrics != null) {
-      flushBytesWritten();
-      if (request.reportMetricsFailed) {
-        metrics.requestReset(request.metric());
-      } else {
-        metrics.responseEnd(request.metric(), request.response(), request.response().bytesWritten());
-      }
+      request.context.dispatch(event -> {
+        flushBytesWritten();
+        if (request.reportMetricsFailed) {
+          metrics.requestReset(request.metric());
+        } else {
+          metrics.responseEnd(request.metric(), request.response(), request.response().bytesWritten());
+        }
+      });
     }
     VertxTracer tracer = context.tracer();
     Object trace = request.trace();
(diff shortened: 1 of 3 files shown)


ahus1:
I've added a reproducer in https://github.com/ahus1/quarkus-otel-missing-examplar-example

Not sure how to best add a test case for that. I'd love to hear a hint on how to do that.

tsegismont:
@ahus1 see my comments in https://github.com/quarkusio/quarkus/issues/43845 and #5355 (rescheduling the task on the context bound to the request is not a good option).

ahus1:
Changing this to draft, and will probably close it as it is likely to be handled on the Quarkus side

ahus1:
Closing as it will be handled on the Quarkus side. Sorry for the noise!

Issue #5371, opened 2024-10-24
Misalignment in content encoding between Http1.x and Http2


bjornhusberg:
Pull request: https://github.com/eclipse-vertx/vert.x/pull/5372

The Http2 server-implementation handles content encoding differently compared to the Http1.x server-implementation, which leads to issues.

Http1.x:
The HttpContentCompressor determines the encoding based on the Accept-Encoding header right before writing the headers. If the request handler has already set a Content-Encoding header there will be no further encoding and if the Content-Encoding header is set to "identity" the header is removed by HttpChunkContentCompressor and the response is returned as-is.

Http2:
The encoding is determined already when setting up the stream in Http2ServerConnection and is then passed down to Http2ServerRequest and Http2ServerResponse where it is set using the Content-Encoding-header. The header content is later picked up by the CompressorHttp2ConnectionEncoder which compresses the response accordingly.

While the Http1.x-implementation works fine the Http2-implementation leads to two problems:

1. If the response is already encoded by the request handler and the Content-Encoding header has been set accordingly, the CompressorHttp2ConnectionEncoder picks up the encoding and encodes the response again, resulting in garbage returned to the client.

2. If the Content-Encoding is explicitly set to "identity", the response is not compressed by the CompressorHttp2ConnectionEncoder but the Content-Encoding holding "identity" leaks out to the client. This is unnecessary and leads to confusion for some clients.

PR #5372, opened 2024-10-24
Fixes Vert.x issue 5371

bjornhusberg:
### Motivation

The Http2 server-implementation handles content encoding differently compared to the Http1.x server-implementation, which leads to issues.

Http1.x:
The HttpContentCompressor determines the encoding based on the Accept-Encoding header right before writing the headers. If the request handler has already set a Content-Encoding header there will be no further encoding and if the Content-Encoding header is set to "identity" the header is removed by HttpChunkContentCompressor and the response is returned as-is.

Http2:
The encoding is determined already when setting up the stream in Http2ServerConnection and is then passed down to Http2ServerRequest and Http2ServerResponse where it is set using the Content-Encoding-header. The header content is later picked up by the CompressorHttp2ConnectionEncoder which compresses the response accordingly.

While the Http1.x-implementation works fine the Http2-implementation leads to two problems:

1. If the response is already encoded by the request handler and the Content-Encoding header has been set accordingly, the CompressorHttp2ConnectionEncoder picks up the encoding and encodes the response again, resulting in garbage returned to the client.

2. If the Content-Encoding is explicitly set to "identity", the response is not compressed by the CompressorHttp2ConnectionEncoder but the Content-Encoding holding "identity" leaks out to the client. This is unnecessary and leads to confusion for some clients.

### Changes

**Removes the setting of the Content-Encoding header in Http2ServerResponse.**
- _Change motivation:_ Setting the header in one place and performing the compression in another makes it difficult to handle scenarios where the header is set by other code. For instance a request handler. Setting the encoding in the HttpServerResponse is also not the way it's done in the Http1-implementation.
- _Actual changes:_ Removed the setting of the header from Http2ServerResponse. After that, the contentEncoding variable could be removed from the Http2ServerResponse, the Http2ServerRequest and the Push class and no longer needed to be determined in the Http2ServerConnection.initStream or Http2ServerConnection.doSendPush.

**Extended the CompressorHttp2ConnectionEncoder with a new VertxCompressorHttp2ConnectionEncoder.**
- _Change motivation:_ The new encoder determines the content-encoding right before writing the headers, similar to how it's done in HttpContentCompressor. This makes it possible to handle the case where the payload has already been encoded by a request handler or explicitly disabled using Content-Encoding: identity.
- _Actual changes:_ Added a VertxCompressorHttp2ConnectionEncoder that determines if the payload should be compressed just before writing the headers. If content-encoding has already been set or could not be determined the compressor encoder is skipped. If the Content-Encoding header is set to "identity" it is removed.

**Adds two new tests in Http2ServerTest that verifies the issues described in the GitHub issue.**

### Outcome

The two issues described are solved and tests are green. :)

https://github.com/eclipse-vertx/vert.x/issues/5371

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/VertxCompressorHttp2ConnectionEncoder.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/VertxCompressorHttp2ConnectionEncoder.java
@@ -1,0 +1,174 @@
+package io.vertx.core.http.impl;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+import io.netty.handler.codec.compression.CompressionOptions;
+import io.netty.handler.codec.http2.CompressorHttp2ConnectionEncoder;
+import io.netty.handler.codec.http2.Http2Connection;
+import io.netty.handler.codec.http2.Http2ConnectionEncoder;
+import io.netty.handler.codec.http2.Http2Exception;
+import io.netty.handler.codec.http2.Http2Flags;
+import io.netty.handler.codec.http2.Http2FrameWriter;
+import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.handler.codec.http2.Http2LifecycleManager;
+import io.netty.handler.codec.http2.Http2RemoteFlowController;
+import io.netty.handler.codec.http2.Http2Settings;
+import io.netty.handler.codec.http2.Http2SettingsReceivedConsumer;
+import java.util.function.Function;
+
+import static io.vertx.core.http.HttpHeaders.CONTENT_ENCODING;
+import static io.vertx.core.http.HttpHeaders.IDENTITY;
+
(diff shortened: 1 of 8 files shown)


vertx-core/src/main/java/io/vertx/core/http/impl/VertxCompressorHttp2ConnectionEncoder.java
```diff
@@ -0,0 +1,174 @@
+package io.vertx.core.http.impl;
```
vietj, line 1: headers

bjornhusberg: Added!

vietj:
the content of the issue was great and very explanatory, it deserves to be used as comment following the pattern outline above

vietj:
@bjornhusberg tests don't see to pass, can you have a look ?

bjornhusberg:
@vietj Tests are updated

bjornhusberg:
Rerun?

vietj:
@bjornhusberg did that indeed, flaky test

PR #5379, opened 2024-10-29
Use Netty default allocator whenever it is pooled otherwise fallback to adaptive allocator

vietj:
Motivation:

Vert.x should use Netty's default allocator whenever possible in order to minimize the resources for pooled allocation (thread-local direct buffers, arenas).

Changes:

`VertxByteBufAllocator.POOLED_ALLOCATOR` reuses `ByteBufAllocator.DEFAULT` when it is pooled otherwise uses `AdaptiveByteBufAllocator.DEFAULT`.

TCP server/client should use `VertxByteBufAllocator.POOLED_ALLOCATOR` instead of `PooledByteBufAllocator.DEFAULT`.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/net/impl/NetServerImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/impl/NetServerImpl.java
@@ -11,7 +11,6 @@
 package io.vertx.core.net.impl;
 
 import io.netty.bootstrap.ServerBootstrap;
-import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.channel.*;
 import io.netty.channel.group.ChannelGroup;
 import io.netty.channel.group.ChannelGroupFuture;
@@ -31,6 +30,7 @@ import io.vertx.core.impl.HostnameResolver;
 import io.vertx.core.internal.ContextInternal;
 import io.vertx.core.internal.PromiseInternal;
 import io.vertx.core.internal.VertxInternal;
+import io.vertx.core.impl.buffer.VertxByteBufAllocator;
 import io.vertx.core.internal.logging.Logger;
 import io.vertx.core.internal.logging.LoggerFactory;
 import io.vertx.core.internal.tls.SslContextManager;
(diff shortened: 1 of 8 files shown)


vietj:
Still missing integration test with `io.netty.allocator.type`

vertx-core/src/main/java/io/vertx/core/net/impl/NetServerImpl.java
```diff
@@ -508,7 +508,7 @@ private void bind(
     if (options.isSsl()) {
       bootstrap.childOption(ChannelOption.ALLOCATOR, PartialPooledByteBufAllocator.INSTANCE);
```
franz1981, line 509: We shouldn't have VertxByteBufAllocator.UNPOOLED_ALLOCATOR, VertxByteBufAllocator.POOLED_ALLOCATOR and  `PartialPooledByteBufAllocator` and we shouldn't set any ALLOCATOR in our code; Netty will pick directly `ByteBufAllocator.DEFAULT`, if the user doesn't customize its own.

franz1981 (changes requested):

vertx-core/src/main/java/io/vertx/core/net/impl/NetServerImpl.java
vietj: yes but for the moment I think it is safe to behave like that, as it has been made impl and not internal anymore, this seems safe to keep like this for a while

franz1981: To me it's fine - but I still don't see the reason to use a partial one - it will hit again https://github.com/eclipse-vertx/vert.x/pull/5316 if we kept on using the partial allocator.
And if we remove the partial one, there's no reason anymore to have the unpooled singleton as well, since it was used just there...

vietj: yes that will be done in a subsequent commit that cares about SSL

PR #5386, opened 2024-11-05
Drop usage of partial byte buf allocator.

vietj:
Motivation:

Vert.x partial byte buf allocator implies the usage of unpooled buffers which can lead to increase significantly the memory footprint for SSL.

Changes:

Delete the partial byte buf allocator and use instead the same allocator than the non SSL case.


vertx-core/src/main/java/io/vertx/core/net/impl/NetServerImpl.java
```diff
@@ -507,13 +506,8 @@ private void bind(
...
     ServerBootstrap bootstrap = new ServerBootstrap();
     bootstrap.group(vertx.getAcceptorEventLoopGroup(), channelBalancer.workers());
-    if (options.isSsl()) {
-      bootstrap.childOption(ChannelOption.ALLOCATOR, PartialPooledByteBufAllocator.INSTANCE);
-    } else {
-      bootstrap.childOption(ChannelOption.ALLOCATOR, VertxByteBufAllocator.POOLED_ALLOCATOR);
-    }
-
     bootstrap.childHandler(channelBalancer);
+    bootstrap.childOption(ChannelOption.ALLOCATOR, VertxByteBufAllocator.POOLED_ALLOCATOR);
```
franz1981, line 510: VertxByteBufAllocator.UNPOOLED_ALLOCATOR is no longer used and could be dropped as well.

franz1981 (changes requested):

vertx-core/src/main/java/io/vertx/core/net/impl/NetServerImpl.java
vietj: done, thanks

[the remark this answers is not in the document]
vietj: ok

PR #5392, opened 2024-11-16
Improve HTTP validation for inlining and specialized types

franz1981:
This change is mostly to improve native-image performance on validation due to closed-world assumption which lead to virtual calls while reading the chars to validate and no loop unrolling during validation.

Additionally, it takes care of moving out of the hot path unlikely validation scenarios for http header values e.g. presence of CRLF or non-printable control chars: this should help JVM mode as well.
Another thing which can help, despite the duplicated code, is the specialized latin head name validation because doesn't require to turn `byte` into `char`s for `AsciiString`.
Without this change the validation of ASCII heard names was slower than normal strings!
FYI @galderz

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/HttpUtils.java
+++ b/src/main/java/io/vertx/core/http/impl/HttpUtils.java
@@ -863,67 +863,181 @@ public final class HttpUtils {
     });
   }
 
-  public static void validateHeaderValue(CharSequence seq) {
+  public static void validateHeaderValue(CharSequence value) {
+    if (value instanceof AsciiString) {
+      validateAsciiHeaderValue((AsciiString) value);
+    } else if (value instanceof String) {
+      validateStringHeaderValue((String) value);
+    } else {
+      validateSequenceHeaderValue(value);
+    }
+  }
 
-    int state = 0;
-    // Start looping through each of the character
-    for (int index = 0; index < seq.length(); index++) {
-      state = validateValueChar(seq, state, seq.charAt(index));
+  private static void validateAsciiHeaderValue(AsciiString value) {
+    final int length = value.length();
+    if (length == 0) {
+      return;
+    }
+    byte[] asciiChars = value.array();
+    int off = value.arrayOffset();
+    if (off == 0 && length == asciiChars.length) {
+      for (int index = 0; index < asciiChars.length; index++) {
+        int latinChar = asciiChars[index] & 0xFF;
+        if (latinChar == 0x7F) {
+          throw new IllegalArgumentException("a header value contains a prohibited character '127': " + value);
+        }
+        // non-printable chars are rare so let's make it a fall-back method, whilst still accepting HTAB
+        if (latinChar < 32 && latinChar != 0x09) {
+          validateSequenceHeaderValue(value, index - off);
+          break;
+        }
+      }
+    } else {
+      validateAsciiRangeHeaderValue(value, off, length, asciiChars);
+    }
+  }
+
+  /**
+   * This method is the slow-path generic version of {@link #validateAsciiHeaderValue(AsciiString)} which
+   * is optimized for {@link AsciiString} instances which are backed by a 0-offset full-blown byte array.
+   */
+  private static void validateAsciiRangeHeaderValue(AsciiString value, int off, int length, byte[] asciiChars) {
+    int end = off + length;
+    for (int index = off; index < end; index++) {
+      int latinChar = asciiChars[index] & 0xFF;
+      if (latinChar == 0x7F) {
+        throw new IllegalArgumentException("a header value contains a prohibited character '127': " + value);
+      }
+      // non-printable chars are rare so let's make it a fall-back method, whilst still accepting HTAB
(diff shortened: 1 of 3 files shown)


franz1981:
I've had some fun with assembly and reading the RFC and I'm going to push a major improvement on this, turning this in draft first

src/main/java/io/vertx/core/http/impl/HttpUtils.java
```diff
@@ -925,23 +928,32 @@ private static int validateLatinValue(CharSequence seq, int state, byte latinCha
...
-        case '\r':
-          return CR_STATE;
-        case '\n':
-          return LF_STATE;
+    if ((latinChar & HIGHEST_INVALID_VALUE_CHAR_MASK) == 0) {
+      // this is a rare scenario
+      validateNonPrintableCtrlLatin(seq, latinChar);
+      // this can include LF and CR as they are non-printable characters
+      if (state == NO_CR_LF_STATE) {
+        // Check the CRLF (HT | SP) pattern
```
vietj: if that is rare, shouldn't we encapsulate the if block in a method to reduce method size ?

franz1981:
TLDR

1. I've streamlined the header name common path (according to the RFC, which is way stricter!)
2. I've retained the original behaviour, with some improvement without duplicating any logic as it is fine to be slower there

I think that I could have further exploded the original state machine, and list let the hotspot JVM to do it right (if able!), but sadly native image isn't that smart and I prefer to keep it like this as it looks more like what Netty do, but with a less strict fallback.

Performance on Hotspot look great, compare to what it was (I will post something home back to my desk), while I can wait @galderz to valide this (I have added a benchmark on purpose, which could be reused) on native image - 🤞

franz1981:
I have pushed https://github.com/eclipse-vertx/vert.x/pull/5392/commits/e4f43420078593445bd8ebe93026b1a1bf82d6f7 and verified it holds from jdk 17 to main.

It solves a problem of "untrusted" types like AsciiString: since the JVM cannot trust offset and end vs arrays boundaries, it would perform additional computations to perform bound checks. 
Giving the JIT an hints that offset is 0 (JIT can profile branches) allow propagating that information, reducing some of these computations in the compiled unrolled version of the loop.
I didn't verify if the same problem happens on end too, since I couldn't spot any weird bookkeeping after this change, although I am sure there is much more to shave.
I would like to produce a code which is less magic 🪄 and still do the right thing.

This explain both why Strings are better performers and why Netty value validation was faster: due to a bug of the Netty version (I will send a pr to fix, since it looks like a potential security issue) the loop variables trigger a better JVM optimization.

I didn't verified

franz1981:
Any chance you looked into the code @vietj ?

vietj:
early next week @franz1981

vietj:
scheduled for 4.5.12 @franz1981

galderz:
These changes have a positive impact on native.

Before PR:
```
Benchmark                                      (ascii)   Mode  Cnt      Score     Error   Units
HeadersValidationBenchmark.validateNameNetty      true  thrpt    5  13564.945 ± 230.330  ops/ms
HeadersValidationBenchmark.validateNameNetty     false  thrpt    5   4406.927 ±   5.628  ops/ms
HeadersValidationBenchmark.validateNameVertx      true  thrpt    5  17456.963 ±  35.617  ops/ms
HeadersValidationBenchmark.validateNameVertx     false  thrpt    5   3678.795 ±  41.234  ops/ms
HeadersValidationBenchmark.validateValueNetty     true  thrpt    5  16350.225 ± 148.984  ops/ms
HeadersValidationBenchmark.validateValueNetty    false  thrpt    5   3750.169 ±  19.454  ops/ms
HeadersValidationBenchmark.validateValueVertx     true  thrpt    5   1889.591 ±   4.780  ops/ms
HeadersValidationBenchmark.validateValueVertx    false  thrpt    5   2189.690 ± 296.273  ops/ms
```

After PR:
```
Benchmark                                      (ascii)   Mode  Cnt      Score     Error   Units
HeadersValidationBenchmark.validateNameNetty      true  thrpt    5  13696.625 ± 111.232  ops/ms
HeadersValidationBenchmark.validateNameNetty     false  thrpt    5   4234.122 ±  49.438  ops/ms
HeadersValidationBenchmark.validateNameVertx      true  thrpt    5  17824.119 ±  52.503  ops/ms
HeadersValidationBenchmark.validateNameVertx     false  thrpt    5   4028.792 ±  23.738  ops/ms
HeadersValidationBenchmark.validateValueNetty     true  thrpt    5  14507.255 ± 455.145  ops/ms
HeadersValidationBenchmark.validateValueNetty    false  thrpt    5   3515.713 ±  30.857  ops/ms
HeadersValidationBenchmark.validateValueVertx     true  thrpt    5  18613.015 ± 227.081  ops/ms
HeadersValidationBenchmark.validateValueVertx    false  thrpt    5   3505.120 ±  10.858  ops/ms
```

The big winner is `HeadersValidationBenchmark.validateValueVertx` when ascii=true. The reason the performance improves so much is because it produces assembly with no call instructions for this particular benchmark. Some smaller improvements seen in other cases. No regressions.

Issue #5399, opened 2024-11-21
Future.all depends on futures implementing FutureInternal, which is surprising


doctorpangloss:
### Version

4.5.11

### Context

```
class com.hiddenswitch.framework.tests.impl.FrameworkTestBase$Checkpoint cannot be cast to class io.vertx.core.impl.future.FutureInternal (com.hiddenswitch.framework.tests.impl.FrameworkTestBase$Checkpoint and io.vertx.core.impl.future.FutureInternal are in unnamed module of loader 'app')
java.lang.ClassCastException: class com.hiddenswitch.framework.tests.impl.FrameworkTestBase$Checkpoint cannot be cast to class io.vertx.core.impl.future.FutureInternal (com.hiddenswitch.framework.tests.impl.FrameworkTestBase$Checkpoint and io.vertx.core.impl.future.FutureInternal are in unnamed module of loader 'app')
	at io.vertx.core.impl.future.CompositeFutureImpl.init(CompositeFutureImpl.java:67)
	at io.vertx.core.impl.future.CompositeFutureImpl.create(CompositeFutureImpl.java:45)
	at io.vertx.core.impl.future.CompositeFutureImpl.all(CompositeFutureImpl.java:29)
	at io.vertx.core.CompositeFuture.all(CompositeFuture.java:88)
```
A `Checkpoint` implements `Future<Void>` which should be allowed. But this breaks in 4.5 when an instance of `Checkpoint` is passed to `Future.all`.

### Do you have a reproducer?

```java
Future.all(new Future<Void> { ... });
```

This will fail for any valid user-authored implementation of future.

This works on 4.4.

* Link to github project/gist

### Steps to reproduce

1. Implement `Future` in a user code class.
2. Observe that `Future.all` will raise an exception trying to cast Future to FutureInternal. This is unexpected

tsegismont:
Fixed by 0b41047d5

PR #5401, opened 2024-11-21
CompositeFuture should support custom future implementations

tsegismont:
See #5399

`CompositeFuture` expected every future to extend `FutureBase`. But sometimes users may come with their own implementation.


vietj:
I'm not sure we should open the door for this

PR #5406, opened 2024-11-22
Use 3 nodes for clustered all tests

frant-hartm:
Motivation:

`ClusteredSharedCounterTest` uses only 2 nodes. When testing with Hazelcast CP subsystem, we need 3 nodes, using 3 nodes makes it easier for us to re-use the test.

I have also unified the common parts in the clustered tests. There is some code duplication, but we can't simply create a common parent to move the code there, because they already extend other classes. Maybe a junit rule would make sense here. Let me know if you want to do something with the duplication, it was already there before this PR.


frant-hartm:
@tsegismont @vietj Could you guys have a quick look at this?

tsegismont (approved):
Thank you @frant-hartm , LGTM

PR #5431, opened 2025-01-03
Discuss how onComplete or onSuccess/onFailure can be used

julianladisch:
Motivation:

Simplify the code by replacing onComplete + if-ar.succeeded-else with
either onComplete(res, e) or with other processing.

Avoiding onComplete + if-ar.succeeded-else results in more concise and understandable
code because it avoids the additional nesting of the if-else clause.

Extend the `exampleFuture*` code to discuss all possibilities how onComplete and
onSuccess/onFailure might be used.

Conformance:

* [x] You should have signed the Eclipse Contributor Agreement as explained in https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md
* [x] Please also make sure you adhere to the code style guidelines: https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines

### DIFF
--- a/vertx-core/src/main/asciidoc/futures.adoc
+++ b/vertx-core/src/main/asciidoc/futures.adoc
@@ -12,6 +12,27 @@ You cannot interact directly with the result of a future, instead you need to se
 {@link examples.CoreExamples#exampleFuture1}
 ----
 
+Instead of using `onComplete` with `if (ar.succeeded()) { … } else { … }` you can use the `onSuccess` and `onFailure` handlers:
+
+[source,$lang]
+----
+{@link examples.CoreExamples#exampleFuture2}
+----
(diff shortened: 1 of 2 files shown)


tsegismont (commented):
Thanks @julianladisch for bringing this up.

I have reservations about this one. Invoking `onSuccess` and `onFailure` at the same place implies creating two future listeners where a single one would do the trick.

If the goal is to avoid writing the `if` block, then there is an `onComplete` overload that takes two arguments, a success and a failure handler.

We should promote the usage of this method, imo.

That doesn't mean `onSucess` and `onFailure` don't have any use case (sometimes they're convenient, like when a method returning a future needs to perform a side action when the future is completed).

PR #5432, opened 2025-01-04
Fix wrong inherited javadoc of AsyncFile#exceptionHandler

julianladisch:
Motivation:

See
https://vertx.io/docs/5.0.0.CR3/apidocs/io/vertx/core/file/AsyncFile.html#exceptionHandler(io.vertx.core.Handler) The inherited javadoc say that it sets the exception handler on the read stream only.

Fix: Mention that is also sets it on the write stream.

Conformance:

* [x] You should have signed the Eclipse Contributor Agreement as explained in https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md
* [x] Please also make sure you adhere to the code style guidelines: https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines


tsegismont (approved):
LGTM, thank you @julianladisch

Issue #5429, opened 2024-12-30
`HostnameResolver` class initialization is pretty heavy


geoand:
This class is loaded as part of the initialization of `AddressResolverOptions`, but the way it tries to extract the `ndots` and `rotate` configuration is pretty heavy (and shows up in flamegrapahs for startup of a Quarkus application).

I believe it could be made to not use regular expressions and also pull all the options in a single pass, no?

tsegismont:
That's possible, if you'd like to contribute such an enhancement, please go ahead and send a PR (ideally with a benchmark)

geoand:
I'll keep it on my radar

geoand:
#5439 is what I have in mind

PR #5439, opened 2025-01-10
Use simple parser instead of regex for parsing resolv.conf

geoand:
- Closes: #5429

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/HostnameResolver.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/HostnameResolver.java
@@ -45,35 +45,25 @@ public class HostnameResolver implements AddressResolver {
 
   private static final Logger log = LoggerFactory.getLogger(HostnameResolver.class);
 
-  private static Pattern resolvOption(String regex) {
-    return Pattern.compile("^[ \\t\\f]*options[^\n]+" + regex + "(?=$|\\s)", Pattern.MULTILINE);
-  }
-
-  private static final Pattern NDOTS_OPTIONS_PATTERN = resolvOption("ndots:[ \\t\\f]*(\\d)+");
-  private static final Pattern ROTATE_OPTIONS_PATTERN = resolvOption("rotate");
+  private static final String NDOTS_LABEL = "ndots:";
+  private static final String ROTATE_LABEL = "rotate";
+  private static final String OPTIONS_ROW_LABEL = "options";
   public static final int DEFAULT_NDOTS_RESOLV_OPTION;
   public static final boolean DEFAULT_ROTATE_RESOLV_OPTION;
 
+
+  private static final int DEFAULT_NDOTS = 1;
+  private static final boolean DEFAULT_ROTATE = false;
+
   static {
-    int ndots = 1;
-    boolean rotate = false;
     if (isLinux()) {
-      File f = new File("/etc/resolv.conf");
-      try {
-        if (f.exists() && f.isFile()) {
-          String conf = Files.readString(f.toPath());
-          int ndotsOption = parseNdotsOptionFromResolvConf(conf);
-          if (ndotsOption != -1) {
-            ndots = ndotsOption;
-          }
-          rotate = parseRotateOptionFromResolvConf(conf);
-        }
-      } catch (Throwable t) {
-        log.debug("Failed to load options from /etc/resolv/.conf", t);
-      }
+      ResolverOptions options = parseLinux(new File("/etc/resolv.conf"));
+      DEFAULT_NDOTS_RESOLV_OPTION = options.effectiveNdots();
+      DEFAULT_ROTATE_RESOLV_OPTION = options.isRotate();
+    } else {
+      DEFAULT_NDOTS_RESOLV_OPTION = DEFAULT_NDOTS;
+      DEFAULT_ROTATE_RESOLV_OPTION = DEFAULT_ROTATE;
     }
-    DEFAULT_NDOTS_RESOLV_OPTION = ndots;
-    DEFAULT_ROTATE_RESOLV_OPTION = rotate;
   }
 
   private final Vertx vertx;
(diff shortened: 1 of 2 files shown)


tsegismont (commented):
Thank you @geoand 

Have you got a chance to write a micro-benchmark and compare results?

geoand:
Not yet

geoand:
Actually I am not sure what a microbenchmark would look like in this case as I've removed the old methods and replaced them with a single one that extracts the necessary data in one pass.

@franz1981 how do you usually handle cases like this? Moreover, do you think a microbenchmark is even necessary in this case?

franz1981:
That's how I do it: https://github.com/franz1981/java-puzzles/blob/74d60bbb8639ccbc1158046ce8d5575720bca96c/src/main/java/red/hat/puzzles/cstatic/StaticInt.java

You can use -prof jfr:dir=profile-results\;configName=jfr-profile.jfc

And use JMH to read the class loading events and compilation events using both -prof cl -prof comp.

These are very tricky benchmarks since:
- class loading can be dominated by I/O files so having enough forks is necessary to make it happen on the OS page cache - but it means it will be less representative of a real use case since the OS page cache would appear like a super fast disk (whilst is not what happen on containers with rate limited disks i.e. AWS)
- beware whatever is loaded by JMH infra which can affect measurements: e.g. if what you want to benchmark is leading an expensive Java  util class A, but the JMH infra does it too, it will make the code under test to appear faster, If it was the dominant factor, but maybe they same class won't be loaded in the real use case

Which means, make sure the effect you want to measure is big enough that observing it, reproduces it with fidelity and is not absorbed by the infrastructure cost

geoand:
Here is what a CPU time flamegraph looks like currently:

orig

And with this patched backported to 4.5.11 it looks like this:

patched

geoand:
It does seem however that this is causing more allocations so I'll see what I can do

geoand:
Ideally I would want this in Vert.x 4 as well :)

tsegismont:
> Actually I am not sure what a microbenchmark would look like in this case as I've removed the old methods and replaced them with a single one that extracts the necessary data in one pass.

It would have two methods comparing the code inside `if (isLinux()) { ... }` before and after the change.



> I pushed an update and now things look rather nice.
> 
> Here is the CPU time flamegraph:

I'm not an expert but it seems both diagrams use `AddressResolver` init as the wide base. So the conclusion that performance has been improved is drawn from the number of samples ? (i.e 60 instead of 118)

geoand:
> So the conclusion that performance has been improved is drawn from the number of samples ? (i.e 60 instead of 118)

Correct.

> I'm not an expert but it seems both diagrams use AddressResolver init as the wide base

I could have also used a base lower down the call stack that would show that `AddressResolver#<clinit>` is now less wide

geoand:
Anything more you would like from this PR?

tsegismont:
> > Actually I am not sure what a microbenchmark would look like in this case as I've removed the old methods and replaced them with a single one that extracts the necessary data in one pass.
> 
> It would have two methods comparing the code inside `if (isLinux()) { ... }` before and after the change.

Have you been able to write a JMH test comparing the two implementations?

geoand:
I have not even tried it. Do you really think it's necessary given the evidence above?

vertx-core/src/main/java/io/vertx/core/impl/HostnameResolver.java
```diff
@@ -45,35 +45,26 @@ public class HostnameResolver implements AddressResolver {
 
   private static final Logger log = LoggerFactory.getLogger(HostnameResolver.class);
 
-  private static Pattern resolvOption(String regex) {
-    return Pattern.compile("^[ \\t\\f]*options[^\n]+" + regex + "(?=$|\\s)", Pattern.MULTILINE);
-  }
-
-  private static final Pattern NDOTS_OPTIONS_PATTERN = resolvOption("ndots:[ \\t\\f]*(\\d)+");
-  private static final Pattern ROTATE_OPTIONS_PATTERN = resolvOption("rotate");
+  private static final Pattern WHITESPACE_PATTERN = Pattern.compile("\\s+");
```
gsmet, line 48: I'm curious if the following would avoid the apparently heavy initialization of what I suspect is the usage of `\s`
```suggestion
  private static final Pattern WHITESPACE_PATTERN = Pattern.compile("[ \\t]+");
```

Asking because I suspect `resolv.conf` would support only traditional spaces and tabs.

Now maybe it's not worth it.

geoand: It's certainly worth a try. I'll be out for a few days, so I can try it when I get back if you haven't already

gsmet: I have no idea which app you used to generate this so I think it can wait for a few days.

geoand: I just used the rest-json Quarkus quickstart on a Linux machine

gsmet: Actually this regexp is not used anymore, AFAICS.

vertx-core/src/main/java/io/vertx/core/impl/HostnameResolver.java
```diff
@@ -129,18 +120,65 @@ public Future<Void> close() {
...
+          int rotateIndex = afterOptions.indexOf(ROTATE_LABEL);
+          if (rotateIndex != -1) {
+            if (!containsNewLine(afterOptions.substring(0, rotateIndex))) {
+              rotate = true;
+            }
+          }
+          int ndotsIndex = afterOptions.indexOf(NDOTS_LABEL);
+          if (ndotsIndex != -1) {
+            if (!containsNewLine(afterOptions.substring(0, ndotsIndex))) {
+              Matcher matcher = Pattern.compile("ndots:[ \\t\\f]*(\\d)+(?=$|\\s)").matcher(afterOptions.substring(ndotsIndex));
```
gsmet, line 159: I see you're not compiling this anymore.

I suppose you have good reasons (like it being used only once?).

geoand: It should probably be moved into a holder instead

vertx-core/src/main/java/io/vertx/core/impl/HostnameResolver.java
geoand: Updated

vertx-core/src/main/java/io/vertx/core/impl/HostnameResolver.java
geoand: Done

gsmet:
OK, what's cool now is that we should be entirely regexp-free in the simple case so I would expect the flamegraph to look even better than the one you posted.

geoand:
Indeed!

vertx-core/src/main/java/io/vertx/core/impl/HostnameResolver.java
```diff
@@ -190,4 +227,31 @@ public void dispose(L data) {
...
+    public ResolverOptions(int ndots, boolean rotate) {
+      this.ndots = ndots;
+      this.rotate = rotate;
+    }
+
+    public int getNdots() {
+      return ndots;
+    }
+
+    public int getEffectiveNdots() {
```
vietj, line 244: rename to `effectiveNdots`

vertx-core/src/main/java/io/vertx/core/impl/HostnameResolver.java
```diff
@@ -190,4 +227,31 @@ public void dispose(L data) {
...
+      return ndots != -1 ? ndots : 1;
+    }
+
+    public boolean isRotate() {
+      return rotate;
+    }
+  }
+
+  private static class Holder {
```
vietj, line 253: is this static class needed because of lazy loading purpose ?

vertx-core/src/main/java/io/vertx/core/DeploymentOptions.java
```diff
@@ -42,6 +43,7 @@ public class DeploymentOptions {
   private int workerPoolSize;
   private long maxWorkerExecuteTime;
   private TimeUnit maxWorkerExecuteTimeUnit;
+  private String exactDeploymentId;
```
vietj, line 46: this should be removed from this PR

geoand: Good point! done

tsegismont (approved):
LGTM, thank you @geoand

geoand:
🙏🏽

PR #5441, opened 2025-01-10
Support duplicate context duplication.

vietj:
Motivation:

Duplicating a duplicated context is supported but the duplication semantic is not defined.

Changes:

This update the duplicated context duplication by doing a copy of each local in the duplicated duplicate. This introduce a duplicator for each local that is responsible for copying the object when it is not null.


vietj:
@franz1981 please have a look

franz1981:
@vietj sorry for being late: how do I query the existing locals?

PR #5442, opened 2025-01-13
Fixes updates to traffic shaping options with shared servers

vietj:
This fixes the update-path for traffic shaping options to check for the existence of the traffic shaping handler only for the actual server. Currently, updating traffic-shaping options in a shared-server setting results in IllegalStateException, as the traffic-shaping handler is not set for the worker (non-main) servers.


vietj:
port from 4.x

PR #5450, opened 2025-01-16
Avoid UUID.randomUUID() in startup code

geoand:
Motivation:

This is done because bootstrapping the plumbing
needed by the JDK to produce a UUID value
is expensive, it thus doesn't make sense to
pay this cost when the property isn't actually
needed

We are making an [effort](https://github.com/quarkusio/quarkus/issues/45436) in Quarkus to improve startup time even further by eliminating various bottlenecks across the board.
The first call to `UUID.randomUUID()` is definitely heavy (as shown in the following flamegraph) and if we can avoid it a startup code (as we have in the development branch of Quarkus), it would be nice.

uuid


vertx-core/src/main/java/io/vertx/core/file/impl/FileCache.java
```diff
@@ -45,7 +45,7 @@ static File setupCacheDir(String fileCacheDir) {
 
     // the cacheDir will be suffixed a unique id to avoid eavesdropping from other processes/users
     // also this ensures that if process A deletes cacheDir, it won't affect process B
-    String cacheDirName = fileCacheDir + "-" + UUID.randomUUID();
+    String cacheDirName = fileCacheDir + "-" + System.nanoTime();
```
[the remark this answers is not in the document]
geoand: > nanoTime is not absolute - it's relative to the process

Correct. But we do think it can be problematic, I'm happy to use `Random.getRandom()` or `System.currentTimeMillis()`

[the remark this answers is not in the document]
franz1981: if we use `Math.random()` it get better - but is still not granted to be unique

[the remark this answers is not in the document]
geoand: I am pretty sure we are not looking for **that** such strong of a guarantee here, but I'll let the maintainers be the judge of that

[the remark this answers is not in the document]
tsegismont: How about an optimistic attempt? Something like (simplifying):

```java
for(;;) {
  try {
    String cacheDirName = fileCacheDir + "-" + System.nanoTime();
    Files.createDirectories(cacheDirName)
  } catch(FileAlreadyExistException ignore) {
  }
}
```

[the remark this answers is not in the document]
tsegismont: In fact, you could use a random instead of `System.nanoTime`, I think it would be faster

vertx-core/src/main/java/io/vertx/core/impl/deployment/DefaultDeploymentManager.java
```diff
@@ -36,7 +36,7 @@ public DefaultDeploymentManager(VertxImpl vertx) {
   }
 
   private String generateDeploymentID() {
-    return UUID.randomUUID().toString();
+    return Long.valueOf(System.nanoTime()).toString();
```
tsegismont, line 39: This needs to be globally unique when running in clustered mode with HA enabled

So perhaps something like:

```java
    if (vertx.isClustered() && vertx.haManager()!=null) {
      return UUID.randomUUID().toString();
    }
    // Use a counter?
```

geoand: 👌

vertx-core/src/main/java/io/vertx/core/file/impl/FileCache.java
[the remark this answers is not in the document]
geoand: @franz1981 is `Random.nextLong()` faster than `System.nanoTime()`?

geoand:
> what seems to take time is the initialization of SecureRandom.getDfaultPrng due to loading providers, I think w ecould generate a faster UUID by using a given provider

But those are not public APIs, no?

vertx-core/src/main/java/io/vertx/core/impl/deployment/DefaultDeploymentManager.java
geoand: Updated

vertx-core/src/main/java/io/vertx/core/file/impl/FileCache.java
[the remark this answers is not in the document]
geoand: Updated

vietj:
I think we should have a way to specify the exact cache dir (e.g. `FileSystemOptions#exactFileCacheDir`), when none is provided then it uses UUID. Quarkus would specify it in VertxOptions. This would easily be back-ported

vertx-core/src/main/java/io/vertx/core/impl/deployment/DefaultDeploymentManager.java
tsegismont: It's pretty common to deploy verticles concurrently. Even when Vert.x is not clustered, the returned value should be unique.

geoand: ?

geoand:
Sure, that would make sense for us too

vertx-core/src/main/java/io/vertx/core/file/impl/FileCache.java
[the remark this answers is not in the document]
franz1981: nope - or better - usually nanoTime (if not on the cloud with unreliable time sources) uses a thing called `rdts` which is as cheap as reading a memory area

vertx-core/src/main/java/io/vertx/core/impl/deployment/DefaultDeploymentManager.java
geoand: I have updated it to use Random, is that what you meant?

geoand:
I have updated the PR per suggestions

geoand:
Is there anything else you would like me to do for this one?

vertx-core/src/main/java/io/vertx/core/impl/deployment/DefaultDeploymentManager.java
tsegismont: I meant incrementing an `AtomicLong` counter instead of using a random value (uniqueness is guaranteed and it shouldn't change the perf results you got)

tsegismont:
> I think we should have a way to specify the exact cache dir (e.g. `FileSystemOptions#exactFileCacheDir`), when none is provided then it uses UUID. Quarkus would specify it in VertxOptions. This would easily be back-ported

For usability, it seems to me adding a boolean to the options would be enough (it's what's computed in the end to determine if a UUID should be added to the path).

But it's a matter of taste so I'm fine with keeping an extra dir option if you choose so @vietj

geoand:
Is there anything more you want me to do with this one?

tsegismont:
@vietj PTAL

geoand:
🙏🏽

vertx-core/src/main/java/io/vertx/core/file/FileSystemOptions.java
```diff
@@ -48,6 +48,7 @@ public class FileSystemOptions {
   private boolean classPathResolvingEnabled = DEFAULT_CLASS_PATH_RESOLVING_ENABLED;
   private boolean fileCachingEnabled = DEFAULT_FILE_CACHING_ENABLED;
   private String fileCacheDir = DEFAULT_FILE_CACHING_DIR;
+  private String exactFileCacheDir;
```
vietj, line 51: should get a constant for this one

vertx-core/src/main/java/io/vertx/core/impl/deployment/DefaultDeploymentManager.java
```diff
@@ -27,6 +28,8 @@ public class DefaultDeploymentManager implements DeploymentManager {
 
   public static final Logger log = LoggerFactory.getLogger(DefaultDeploymentManager.class);
 
+  private static final AtomicLong nextId = new AtomicLong();
```
vietj, line 31: the idea of cache dir is to avoid that, and keep the same behaviour we have, so please no.

geoand: So what do you propose?

tsegismont: I believe you confused two changes @vietj  : the cache dir that used a random UUID, and the verticle id generator

vietj: sorry, can we have the verticle id generator in another PR then ?

vietj: I'd like to keep distinct PR for the changelog

vietj: PR

geoand: https://github.com/eclipse-vertx/vert.x/pull/5469

geoand: PR updated

vietj (changes requested):
can you add a test with a real vertx instance and check those cases

- a missing dir is created
- an existing dir is reused
- an error is thrown when a non dir file exist already

geoand:
Sure, I'll do that when I'm back from JFokus

geoand:
Aren't those cases already covered by the tests for `FileResolver`?

vietj:
> Aren't those cases already covered by the tests for `FileResolver`?

good question, I don't know :-)

vertx-core/src/test/java/io/vertx/tests/file/FileResolverTestBase.java
```diff
@@ -486,4 +488,16 @@ public void testGetTheCacheDirWithoutHacks() {
       }
     }
   }
+
+  @Test
+  public void testGetTheExactCacheDirWithoutHacks() {
+    String cacheDir = new FileResolverImpl(new FileSystemOptions().setExactFileCacheDir(cacheBaseDir + "-exact")).cacheDir();
+    if (cacheDir != null) {
+      System.out.println(cacheDir);
```
vietj, line 496: println

vertx-core/src/test/java/io/vertx/tests/file/FileResolverTestBase.java
```diff
@@ -486,4 +488,16 @@ public void testGetTheCacheDirWithoutHacks() {
       }
     }
   }
+
+  @Test
+  public void testGetTheExactCacheDirWithoutHacks() {
```
vietj, line 493: this test should be moved to `FileCacheTest` instead, `FileResolverTestBase` tests the behaviour of resolver implementations

vietj (changes requested):
We need tests that assesses the behaviour of creating a vertx instance when

1. the cache dir already exists and is a directory (I guess it reuses the directory)
2. the cache dir does not exist (it should create the missing directory)
3. the cache dir string is not a valid value
4. the cache dir does not exists and cannot be created, e.g. the parent path points to a file
5. the cache dir exists but is not a directory, e.g. it is a file

geoand:
I added all but `3` as I didn't find a way to add a ubvalid name

vertx-core/src/main/java/io/vertx/core/file/impl/FileCache.java
```diff
@@ -37,16 +36,15 @@ static FileCache setupCache(String fileCacheDir) {
...
     // ensure that the argument doesn't end with separator
     if (fileCacheDir.endsWith(File.separator)) {
       fileCacheDir = fileCacheDir.substring(0, fileCacheDir.length() - File.separator.length());
     }
 
     // the cacheDir will be suffixed a unique id to avoid eavesdropping from other processes/users
     // also this ensures that if process A deletes cacheDir, it won't affect process B
-    String cacheDirName = fileCacheDir + "-" + UUID.randomUUID();
-    File cacheDir = new File(cacheDirName);
+    File cacheDir = isEffectiveValue ? new File(fileCacheDir) : new File(fileCacheDir + "-" + UUID.randomUUID());
```
pmlopes, line 47: Doesn't this change, break the comment above? when `isEffectiveValue` is true, 2 vert.x instances will interfere with each other's cache. While this is _probably_ ok for the same application, if 2 applications differ, then it could cause invalid states.

One example is (regardless if the 2 applications are the same or not) The moment the 1st terminates, it deletes the cache and would also mean it was deleted for the second, causing inconsistencies and errors.

geoand:
Is there any chance we can move this forward? @gsmet also independently ran into it

geoand:
I would really appreciate some input on this

geoand:
If this is approved, could it be backported to 4?

vietj:
@geoand I would jsut like before to change the configuration option, instead of passing a different path `exactFileCacheDir`, I would prefer a boolean that configurations whether to use the existing path fully or as a directory.

geoand:
@vietj Thanks. Can you elaborate a little more on what exactly you would like to see, because I don't fully underastand your proposal

vietj:
@geoand instead of providing two paths, I prefer to keep the existing path value, but instead introduce a boolean that configures how it should be interpreted, either as a parent directory with a generated random child name or as an exact path directory to use, perhaps we could also improve that and when the path directory already exists, append a suffix "-2", "-3" to guarantee unicity whatsoever.

geoand:
Thanks @vietj. 

I believe the latest version of the PR addresses your concerns

vietj:
lgtm @geoand

geoand:
🙏🏽

vietj:
@geoand I think we also need to rename this contribution to what it actually is : provide a way to use an exact path for a vertx instance instead of generating one based on random

vietj:
@geoand there is a faiilure on windows

geoand:
I will have a look

geoand:
> @geoand I think we also need to rename this contribution to what it actually is : provide a way to use an exact path for a vertx instance instead of generating one based on random

I updated the commit, but I can't update the PR title

vietj:
@geoand can you backport this to 4.x and 5.0, we intent to do patch release this week of those branches and that would make it available

geoand:
Sure thing, I'll do that tomorrow

geoand:
https://github.com/eclipse-vertx/vert.x/pull/5958, https://github.com/eclipse-vertx/vert.x/pull/5959

Issue #5453, opened 2025-01-20
JsonObject and ConfigRetriever: add further options on how to mergeIn JsonObject


rzorzorzo:
#### Describe the feature

currently config files are merged by overwriting null values and by overwriting array values.
in my use case however, array values need to be merged and null values should not override existing values.
example:

>	JsonObject o1 = JsonObject.of("m", null, "a", JsonArray.of(1));
	JsonObject o2 = JsonObject.of("m", JsonObject.of(), "a", JsonArray.of(2));
	JsonObject o3 = o2.mergeIn(o1, true, true, false);
	System.out.println(o3);

should return:
>{"m":{},"a":[1,2]}

#### Use cases

rational: configuration files are structured by module. a module may have a property which is not included in another, therefore null value should not override a non null value.

the values of arrays may need to be "collected" over multiple module configurations.

#### Contribution

JsonObject.java:

>  public JsonObject mergeIn(JsonObject other, int depth, boolean mergeArrays, boolean overwriteNull) {
    if (depth < 1) {
      return this;
    }
    if (depth == 1 && !mergeArrays && overwriteNull) {
      map.putAll(other.map);
      return this;
    }
    for (Map.Entry<String, Object> e : other.map.entrySet()) {
      if (e.getValue() == null && overwriteNull) {
         map.put(e.getKey(), null);
      } else if (e.getValue() != null){
        map.merge(e.getKey(), e.getValue(), (oldVal, newVal) -> {
          if (oldVal instanceof Map) {
            oldVal = new JsonObject((Map) oldVal);
          }
          if (newVal instanceof Map) {
            newVal = new JsonObject((Map) newVal);
          }
          if (mergeArrays && oldVal instanceof JsonArray && newVal instanceof JsonArray)
        	  return ((JsonArray)newVal).addAll((JsonArray)oldVal);
          if (oldVal instanceof JsonObject && newVal instanceof JsonObject) {
            return ((JsonObject) oldVal).mergeIn((JsonObject) newVal, depth - 1, mergeArrays, overwriteNull);
          }
          return newVal;
        });
      }
    }
    return this;
  }

tsegismont:
@rzorzorzo I sent https://github.com/eclipse-vertx/vert.x/pull/5454 

A merge function should be generic enough to handle your case (ignore nulls and merge arrays) as well as others.

tsegismont:
@vietj can you please take a look at #5454 ?

PR #5454, opened 2025-01-21
JsonObject: custom merge with MergeFunction

tsegismont:
In some cases, users may want more flexibility when merging values of JSON objects.

In this case, they can provide an implementation of `MergeFunction`. Because we can't tell if a `null` result means `null` or skip, implementation can choose to return `MergeFunction.SKIP`.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/vertx-core/src/main/java/io/vertx/core/json/JsonObject.java
@@ -16,6 +16,7 @@ import io.vertx.core.shareddata.Shareable;
 
 import java.time.Instant;
 import java.util.*;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


tsegismont:
Any comment @vietj ?

vertx-core/src/main/java/io/vertx/core/json/JsonObject.java
```diff
@@ -995,34 +999,81 @@ public JsonObject mergeIn(JsonObject other, boolean deep) {
...
+    return this;
+  }
+
+  /**
+   * A function used to merge {@link JsonObject} values.
+   *
+   * @see #mergeIn(JsonObject, int, MergeFunction)
+   */
+  @FunctionalInterface
+  public interface MergeFunction extends BiFunction<Object, Object, Object> {
```
vietj, line 1040: I would prefer introducing a new contract only for that purpose, BiFunction is already a functionnal interface

vietj: if we need to skip some entries then an extra predicate should be provided

vietj: maybe a `BiFunction<Object, Object, Optional<Object>>` would be more expressive here

tsegismont: > maybe a `BiFunction<Object, Object, Optional<Object>>` would be more expressive here

That was my initial thought. And then, considering in most cases we won't need to skip, I tried to avoid `Optional` by using a signal `Object`.

I created `MergeFunction` to encapsulate that signal object.

I would be fine with a `BiFunction` without `Optional`.

vietj:
I think we should not rush this one and see if we can provide something different that would work better

tsegismont:
> where are the requirement coming from

https://github.com/eclipse-vertx/vert.x/issues/5453

tsegismont:
@vietj what can I do to move this forward?

PR #5469, opened 2025-01-31
Avoid UUID.randomUUID() in Verticle deployment startup code

geoand:
This is done because bootstrapping the plumbing
needed by the JDK to produce a UUID value
is expensive, it thus doesn't make sense to
pay this cost when the property isn't actually
needed

As requested in https://github.com/eclipse-vertx/vert.x/pull/5450#discussion_r1936894987

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/deployment/DefaultDeploymentManager.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/deployment/DefaultDeploymentManager.java
@@ -19,6 +19,7 @@ import io.vertx.core.internal.logging.LoggerFactory;
 
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * @author <a href="http://tfox.org">Tim Fox</a>
@@ -26,6 +27,8 @@ import java.util.concurrent.ConcurrentHashMap;
 public class DefaultDeploymentManager implements DeploymentManager {
 
   public static final Logger log = LoggerFactory.getLogger(DefaultDeploymentManager.class);
+
+  private static final AtomicLong nextId = new AtomicLong();
 
   private final VertxImpl vertx;
   private final Map<String, DeploymentContext> deploying = new HashMap<>();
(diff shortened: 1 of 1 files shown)


vietj:
I think for this one we might have a deploymend ID in the deployment options to force a specific ID like we have done for file system path

geoand:
Makes sense. I'll update the PR next week

geoand:
Now that I think of it, wouldn't it make more sense for `DeploymentOptions` to use a `Supplier<String>` which would be called by the `DeploymentManager` for each verticle?

vietj:
> Now that I think of it, wouldn't it make more sense for `DeploymentOptions` to use a `Supplier<String>` which would be called by the `DeploymentManager` for each verticle?

actually no, with vertx 5 we want to that Options object are purely data (json) if we need something with a supplier we would go with a builder instead

geoand:
What builder are you talking about?

tsegismont:
@geoand I believe @vietj is talking about the effort we have made in Vert.x 5 to have data objects, well, pure data objects and use builders when non data config is required (see for example https://github.com/vert-x3/vertx-micrometer-metrics/pull/229)

Anyway, we don't need anything from `DeploymentOptions`, it's really the `DeploymentManager that has to assign a unique ID to each deployment.

tsegismont (approved):
LGTM, thank you

vietj:
@geoand sorry for the lack of context :-), indeed @tsegismont gave the context

vietj:
thank you @geoand

geoand:
🙏🏽

gsmet:
@geoand wondering if you wanted this one backported to 4.x? I was in the process of removing the UUID initialization from the pre-init steps and stumbled upon this one.

geoand:
Yes. I was hoping for a backport so we can get rid of the UUID initialization altogether

gsmet:
OK, probably a good idea to open a backport PR then :).

FYI, I'm preparing a PR to include what came in 4.5.25.

geoand:
https://github.com/eclipse-vertx/vert.x/pull/5974

PR #5471, opened 2025-02-04
Message consumer message delivery reimplementation

vietj:
Motivation:

Reimplement `MessageConsumer` message delivery using `InboundMessageChannel`.

Changes:

`MessageConsumer` uses now an instance of `InboundMessageChannel` to handle message delivery. The dynamic upper bound of buffered messages becomes static to facilitate the implementation. This is a breaking change.

`MessageConsumerOptions` have been introduced to facilitate the configuration of a consumer, this will be back-ported to 4.5.x, in addition the dynamic buffered message upper bound will be deprecated in that branch.


vietj:
Some commits for inbound message channel improvements will be moved to a separate PR.

PR #5490, opened 2025-02-26
Vert.x Jackson json factory read constraint configuration should be overridable

vietj:
Motivation:

Read constraints were added since Jackson 2.15 to provide upper bounds for input such as the maximum number of a name. These constraints can be configured per json factory, Vert.x static json factory should provide configurability for it.

Changes:

Add system properties that configure the Vert.x static Jackson json factory.


vietj:
@pmlopes please have a look

PR #5507, opened 2025-03-07
Mitigate usage of NoStackTraceThrowable

vietj:
Motivation:

`NoStackTraceThrowable` currently extends `Throwable` and has been moved to the implementation package not exported in `module-info`.

There are several concerns we need to solve:

1. Virtual threads and Kotlin coroutines need to catch a throwable which is not a recommended practice, e.g. an `Exception` catch declaration let the `NoStackTraceThrowable` go through.
2. Users relying on `NoStackTraceThrowable` developing a modular application have to catch `Throwable` instead since the type is not exported.

We should provide a transitory path to remove `NoStackTraceThrowable` in Vert.x 6.

Changes:

`NoStackTraceThrowable` extends `VertxException` instead of `Throwable`.

Users catching `NoStackTraceThrowable` (which is deprecated) can instead catch `VertxException` and achieve the same effect.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/dns/DnsException.java
+++ b/vertx-core/src/main/java/io/vertx/core/dns/DnsException.java
@@ -11,8 +11,7 @@
 
 package io.vertx.core.dns;
 
-import io.vertx.core.impl.NoStackTraceThrowable;
-import java.util.Objects;
+import io.vertx.core.VertxException;
 
 /**
  * Exception which is used to notify the {@link io.vertx.core.AsyncResult}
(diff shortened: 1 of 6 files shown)


kolstae:
I'm afraid this may cause some unexpected behavior in certain cases. A thinkable scenario is the following
```kotlin
try {
   // vertx code
} catch (e: Exception) {
   // handle
} catch (e: NoStackTraceThrowable) {
   // no longer reachable
}
```
However I don't think it's very likely and I would say it's worth the risk when moving to 5.x with a proper release note for it.

vietj:
this will be in the migration doc, virtual thread remains in preview in 4.x

tsegismont (approved):
LGTM, thanks @vietj

tsegismont:
Good idea to add a note in the migration doc. This small problem will impact users creating modular applications, so at least it won't be a majority.

PR #5514, opened 2025-03-10
Provide a coherent implementation of hashCode for JsonObject/JsonArray with respect to equals.

vietj:
Motivation:

`JsonArray`/`JsonObject` hashCode implementations rely on the wrapped data structure hashCode implementation and on the generic Object `hashCode` function.

1. The hash value might differ from data structure implementations.

2. Number coercion is performed in equals which allows case were we do have equality (after coercion) but hash differs (e.g. 4L and 4D)

Changes:

`JsonObject`/`JsonArray` implement `hashCode` methods.

Number's `hashCode` is computed against the `hashCode` of their double value.

### DIFF
--- a/src/main/java/io/vertx/core/json/JsonArray.java
+++ b/src/main/java/io/vertx/core/json/JsonArray.java
@@ -12,6 +12,7 @@
 package io.vertx.core.json;
 
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.json.impl.JsonUtil;
 import io.vertx.core.shareddata.ClusterSerializable;
 import io.vertx.core.shareddata.Shareable;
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/core/json/JsonArray.java
```diff
@@ -688,7 +694,11 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return list.hashCode();
+    int h = 0;
+    for (Object value : this) {
+      h += JsonUtil.hashCode(value);
```
tsegismont, line 699: How about using a multiplier when combining hashcodes in order to reduce collisions? Same comment for Json objects.

vietj: I copied what is in JDK

vietj: good idea

PR #5516, opened 2025-03-11
Make sure ContextBuilder defaults to EL threading model

tsegismont:
The Javadoc implies that it is the default, but the implementation leaves it to null, forcing callers to set the EL model everywhere.

With this change, EL threading model is the default and invoking the withThreadingModel method with a null parameter is forbidden.


tsegismont:
@vietj this should solve build failures in Vert.x SQL Client, Vert.x Auth, Vert.x Tracing and Vert.x JDBC Client

vietj (approved):
LGTM

vietj:
thanks for taking care of this :-)

PR #5525, opened 2025-03-13
make sure after resume we just deliver one message at one moment

hadadil:
Motivation:

I had a [change](https://github.com/eclipse-vertx/vert.x/pull/5457) recently in 4.5.13 (not a lucky number)
After upgrading to this version we have noticed after resume we do receive more than one message at one moment.
First I extended the test to reproduce the case... and after that I done the prod code change.

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/eventbus/impl/MessageConsumerImpl.java
+++ b/src/main/java/io/vertx/core/eventbus/impl/MessageConsumerImpl.java
@@ -133,36 +133,24 @@ public class MessageConsumerImpl<T> extends HandlerRegistration<T> implements Me
   }
 
   protected boolean doReceive(Message<T> message) {
-    Handler<Message<T>> theHandler;
     synchronized (this) {
       if (handler == null) {
         return false;
       }
-      if (demand == 0L) {
-        if (pending.size() < maxBufferedMessages) {
-          pending.add(message);
-          return true;
-        } else {
-          discard(message);
-          if (discardHandler != null) {
-            discardHandler.handle(message);
-          } else {
-            log.warn("Discarding message as more than " + maxBufferedMessages + " buffered in paused consumer. address: " + address);
-          }
-        }
+      if (pending.size() < maxBufferedMessages) {
+        pending.add(message);
+        checkNextTick();
         return true;
       } else {
-        if (pending.size() > 0) {
-          pending.add(message);
-          message = pending.poll();
+        discard(message);
+        if (discardHandler != null) {
+          discardHandler.handle(message);
+        } else {
+          String pause = demand == 0 ? "paused" : "NOT paused";
+          log.warn("Discarding message as more than " + maxBufferedMessages + " buffered in " + pause + " consumer. address: " + address);
         }
-        if (demand != Long.MAX_VALUE) {
-          demand--;
-        }
-        theHandler = handler;
       }
     }
-    deliver(theHandler, message);
     return true;
   }
 
@@ -171,7 +159,7 @@ public class MessageConsumerImpl<T> extends HandlerRegistration<T> implements Me
     if (handler == null) {
       throw new NullPointerException();
     }
-    context.emit(msg, handler);
+    context.dispatch(msg, handler);
   }
 
   private void deliver(Handler<Message<T>> theHandler, Message<T> message) {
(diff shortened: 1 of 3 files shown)


hadadil:
to be fair I tested this change with our app... this is also not a proper solution...

src/main/java/io/vertx/core/eventbus/impl/MessageConsumerImpl.java
```diff
@@ -171,7 +158,7 @@ protected void dispatch(Message<T> msg, ContextInternal context, Handler<Message
     if (handler == null) {
       throw new NullPointerException();
     }
-    context.emit(msg, handler);
+    context.dispatch(msg, handler);
```
hadadil, line 161: this was reverted as it had a side effect:
after resume we were received many messages even we paused again during the first message processing.

src/main/java/io/vertx/core/eventbus/impl/MessageConsumerImpl.java
```diff
@@ -183,8 +170,8 @@ private void deliver(Handler<Message<T>> theHandler, Message<T> message) {
 
   private synchronized void checkNextTick() {
     // Check if there are more pending messages in the queue that can be processed next time around
-    if (!pending.isEmpty() && demand > 0L) {
-      context.nettyEventLoop().execute(() -> {
+    if (demand > 0L && !pending.isEmpty()) {
```
hadadil, line 173: I flip here so the demand evaulated first

src/main/java/io/vertx/core/eventbus/impl/MessageConsumerImpl.java
```diff
@@ -183,8 +170,8 @@ private void deliver(Handler<Message<T>> theHandler, Message<T> message) {
...
+      context.executor().execute(() -> {
```
hadadil, line 174: here we should schedult the task for the next tick on the context thread (so we have no 2 thread dispatching a message at the same moment).
In the past we had a scenario like:
* received
* pause
* resume
* received + paused
* received ?! 
so after we resume we had sometimes got two messages even in the 1st on we said already we don't want to have more right now.

src/main/java/io/vertx/core/eventbus/impl/MessageConsumerImpl.java
```diff
@@ -133,36 +133,23 @@ public synchronized Future<Void> unregister() {
...
-          discard(message);
-          if (discardHandler != null) {
-            discardHandler.handle(message);
-          } else {
-            log.warn("Discarding message as more than " + maxBufferedMessages + " buffered in paused consumer. address: " + address);
-          }
-        }
+      if (pending.size() < maxBufferedMessages) {
+        pending.add(message);
+        checkNextTick();
```
hadadil, line 142: this method simplified a lot and here we always just put it in hte pending (if the buffer not full yet of course) and send a tick (that will decied if we need to do anything right now or not)
as the tick evaulation happen on the contex thread we are safe (single threaded)

src/main/java/io/vertx/core/eventbus/impl/MessageConsumerImpl.java
```diff
@@ -183,8 +181,8 @@ private void deliver(Handler<Message<T>> theHandler, Message<T> message) {
 
   private synchronized void checkNextTick() {
     // Check if there are more pending messages in the queue that can be processed next time around
-    if (!pending.isEmpty() && demand > 0L) {
-      context.nettyEventLoop().execute(() -> {
+    if (demand > 0L && !pending.isEmpty()) {
+      context.execute(__ -> {
```
hadadil, line 185: here we do execute the resume/fetch from the context thread so we do not have situations when we call resume and the same time we do receive a msg and we get 2msgs (on on the context thread and one on eventloop thread)

src/main/java/io/vertx/core/eventbus/impl/MessageConsumerImpl.java
```diff
@@ -138,28 +138,26 @@ protected boolean doReceive(Message<T> message) {
...
+      } else {
+        discard(message);
+        if (discardHandler != null) {
+          discardHandler.handle(message);
+        } else {
+          String pause = demand == 0 ? "paused" : "NOT paused";
+          log.warn("Discarding message as more than " + maxBufferedMessages + " buffered in " + pause + " consumer. address: " + address);
         }
-        theHandler = handler;
+        return false;
```
hadadil, line 160: this was also not 100% right, when we do discard msg we should return with false to have that in the metrics

PR #5531, opened 2025-03-25
Jackson v3 support

vietj:
- [x] multi release jar
- [x] jackson core
- [x] jackson databind
- [x] hybrid pool
- [x] documentation
- [x] jackson constraint tests


cazacugmihai:
When will Vert.x switch to Jackson 3.x?

vietj:
we will switch I think it vertx 6.0, but provide a replacement for users that want to use Jackson 3.x in addition of 2.x

hantsy:
If possible creating multiple versioned jar to support the new features. 

Jackson v3 requires Java 17, but Vertx also support Java 21 Virtual thread(I know Quarkus has it, and the Quarkus kernel is based on Vertx).

vietj:
@hantsy that's what we might do

jponge (approved):
It looks good to me, nice work!

I've checked Jackson 3 and the baseline is Java 17, so if you want to lower the MR jar to 17 and 21 I think it works as well.

vietj:
@jponge since java 17 is EOL and 21 brings Virtual threads, I'm keen sticking to 21 so we can handle VT specific code in the same Java21 src tree

PR #5546, opened 2025-04-08
Feature/parameterized tests

imz87:
Motivation:

Vert.x uses JUnit4 for testing, but JUnit4's native support for parameterized tests has limitations. Specifically, JUnit4’s `org.junit.runners.Parameterized` allows running the same test with different sets of data, but it runs all test cases in the same class multiple times for each set of parameters, and the configuration applies to the whole test suite. This approach can make it difficult to manage and configure tests, especially when only specific test methods need to be parameterized or when more granular control over test execution is desired.

The goal of this PR is to extend Vert.x's testing capabilities by adding support for parameterized tests in a way that is compatible with JUnit4. Specifically, we are introducing a `ParameterizedTest` feature that allows each test method to be configured with its own set of parameters, and test data can be specified as parameters. This allows for more flexibility, where individual test methods can run with different data sets without affecting other tests in the same class.

This enhancement will improve testing efficiency, reduce code duplication, and provide a cleaner, more maintainable approach to writing tests that require different inputs or data sets.

Samples: 

```java

  //sample1
  @ParameterizedTest.Parameter("profile")
  private Profile profile;

  @ParameterizedTest.Parameters("profiles")
  public List<Profile> getProfiles() {
    return Arrays.asList(new Profile("f1", "L1"), new Profile("f2", "L2"), new Profile("f3", "L3"));
  }

  @ParameterizedTest.Parameterized(parameters = "profiles", parameter = "profile")
  @Test
  public void testProfile() {
    System.out.println("Profile: " + profile);
  }



  //sample2
  @ParameterizedTest.Parameter("profileNew")
  private Profile profileNew;

  @ParameterizedTest.Parameters("profilesNew")
  public List<Profile> getProfilesNew() {
    return Arrays.asList(new Profile("f4", "L4"), new Profile("f5", "L5"));
  }

  @ParameterizedTest.Parameterized(parameters = "profilesNew", parameter = "profileNew")
  @Test
  public void testProfileNew() {
    System.out.println("ProfileNew: " + profileNew);
  }





  private static class Profile {
    private final String firstname;
    private final String lastname;

    public Profile(String firstname, String lastname) {
      this.firstname = firstname;
      this.lastname = lastname;
    }

    @Override
    public String toString() {
      return firstname + " " + lastname;
    }
  }


```


imz87:
@vietj , could you please check this PR?

Issue #4493, opened 2022-09-23
Support DNS over TCP with retry


jacobhampel:
This is a follow up on  #4309

I think the problem was not completely resolved.

When the DNS server answers with the flag truncated but does not deliver any results the DNS client should retry using TCP.

When I'm looking at the current test it only tests that the DNS client can handle a DNS response with the truncated flag set and one result.

Would it be possible to add this additional functionality.

That would be very helpful.

I'm not a Netty expert but it seams to be fixed already there.

The desired behavior would be
1. DNS client tries to resolve the hostname via UDP (default)
2. DNS server answers with no results but the truncated flag is set
3. DNS client retries to resolve the hostname via TCP
4. DNS server answers with results

Unfortunately, I'm not an expert in netty and therefore unable to fix it myself.

If any additional feedback is required I'm more than happy to help here.

Vertx version: 4.3.2

vietj:
can you provide a reproducer project for this ?

vietj:
it might be indeed a Netty effect

jacobhampel:
I do not have a reproducer project for this, yet.

I don't know how Vertx is using Netty for the DNS resolution, but this might have been solved for Netty.
https://github.com/netty/netty/pull/9139

PR #5555, opened 2025-04-17
Implement support for DNS over TCP fallback in Netty.

azotcsit:
This PR resolves https://github.com/eclipse-vertx/vert.x/issues/4493. The high-level idea is to support Netty feature to retry failed UDP DNS queries via TCP - https://github.com/netty/netty/issues/9127.

I'm not sure about _master_ branch, do I need to raise a separate PR for that?


azotcsit:
Hi @vietj,

I'd be very grateful if you could take a look at this PR and guide me on the next steps. I'm new to Vert.x contribution process, so any suggestions would be very welcomed.

Thanks!

vietj:
can you test this new option ?

azotcsit:
@vietj 
Thanks for the prompt reply. Are you talking about manual testing or about writing unit tests?

I think I can test it manually, however it will take some time as the problem is not easily reproducible in our env. WRT unit testing, I'm not sure how to validate that TCP fallback happened. Could you please give me some references / suggestions on the best way to implement that?

vietj:
we could need to mock the TCP server and respond to the same traffic that a server does

PR #5558, opened 2025-04-28
Add the support to FileChannel to sendFile in HttpServerResponse

olivierayache:
Motivation:

As I am using Vert.x to create a REST web server to retrieve files I was really intersted by the [sendFile](https://vertx.io/docs/apidocs/io/vertx/core/http/HttpServerResponse.html#sendFile-java.lang.String-long-io.vertx.core.Handler-) feature of HttpServerResponse. After some performance testing and profiling I discovered that the current implementation of sendFile is not optimal for my use case: high throughput with the same files often accessed. 
The repetitive call to open method on RandomAccessFile had a big impact on the performance of my code. That is why I added the possibility to use sendFile with a FileChannel and let the developper to handle the lifecycle of that channel. 
By doing that I managed to get a huge improve of the performance of my web server (roughly x2 in throughput)


olivierayache:
Hello @vietj 
This is my first PR here. Do you have any feedbacks on that ?

olivierayache:
I can write on the HTTP/2 part too
I was refactoring my code for HTTP/1.1 to reuse code between sendFile with FileChannel and the one with String

olivierayache:
Hello @vietj 
I now have an implementation for HTTP/1 and HTTP/2

vertx-core/src/main/java/io/vertx/core/http/FileSender.java
```diff
@@ -0,0 +1,51 @@
+package io.vertx.core.http;
+
+import io.vertx.codegen.annotations.GenIgnore;
+import io.vertx.core.Future;
+
+import java.nio.channels.Channel;
+
+public interface FileSender<T extends Channel> {
```
vietj, line 8: why do we need T here ?

vietj, line 8: overall I don't get why we need to introduce this new interface instead of having methods on HttpServerResponse to send files like we do have for others

vietj:
I am wondering here is the API should not be to have the ability to send AsyncFile from an HttpServerResponse and have a way to convert a channel into a Vert.x AsyncFile

vertx-core/src/main/java/io/vertx/core/TriFunction.java
```diff
@@ -0,0 +1,8 @@
+package io.vertx.core;
+
+@FunctionalInterface
+public interface TriFunction<U, V, W, T> {
```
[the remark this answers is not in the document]
olivierayache: I understand but do we already have a "TriFunction" in vertx or a library used by vertx?
Because it allows code mutualisation in Http1xServerResponse as you can see here

https://github.com/eclipse-vertx/vert.x/blob/c52cb15b37adb435c1c7c7afb48f4b06c15de635/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xServerResponse.java#L512

olivierayache:
I am available for a talk about the best way to implement that feature

vertx-core/src/main/java/io/vertx/core/http/FileSender.java
olivierayache: T was available to avoid unnecessary cast in FileSender implementations. By doing that I can easily force Http1xServerResponse to use a FileChannel and Http2ServerRepsonse an AsynchronousFileChannel

https://github.com/eclipse-vertx/vert.x/blob/c52cb15b37adb435c1c7c7afb48f4b06c15de635/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xServerResponse.java#L457-L470

https://github.com/eclipse-vertx/vert.x/blob/c52cb15b37adb435c1c7c7afb48f4b06c15de635/vertx-core/src/main/java/io/vertx/core/http/impl/Http2ServerResponse.java#L574

vietj:
@olivierayache I am available for discussion on our Discord server btw

vietj:
@olivierayache any progress on this ?

olivierayache:
I am working on what we ve discussed but it is not as easy as we thank to create AsyncFile from file descriptor

olivierayache:
Hello @vietj I removed the TriFunction interface and just pushed an update to that PR. The idea we got to retrieve the file descriptor from FileChannel is not working for now that is why I push that today.

olivierayache:
Hello @vietj 
Do you have some feedbacks about my last commits?

vietj:
@olivierayache I will have a look soonish

vertx-core/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java
```diff
@@ -74,32 +75,15 @@ public class AsyncFileImpl implements AsyncFile {
   private long readPos;
   private long readLength = Long.MAX_VALUE;
 
-  AsyncFileImpl(VertxInternal vertx, String path, OpenOptions options, ContextInternal context) {
-    if (!options.isRead() && !options.isWrite()) {
-      throw new FileSystemException("Cannot open file for neither reading nor writing");
-    }
+  AsyncFileImpl(VertxInternal vertx, OpenOptions options, Function<OpenOptions, AsynchronousFileChannel> channelFunction, ContextInternal context) {
```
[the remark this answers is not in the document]
vietj: yes please

vertx-core/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java
```diff
@@ -512,7 +512,7 @@ private void checkClosed() {
   private void checkContext() {
     if (!vertx.getContext().equals(context)) {
       throw new IllegalStateException("AsyncFile must only be used in the context that created it, expected: "
-          + context + " actual " + vertx.getContext());
```
vietj, line 515: avoid reformating

vertx-core/src/main/java/io/vertx/core/http/impl/Http1xServerResponse.java
```diff
@@ -1,798 +1,840 @@
...
+/**
+ * @author <a href="http://tfox.org">Tim Fox</a>
+ */
+public class Http1xServerResponse implements HttpServerResponse, HttpResponse {
+
+  private static final Buffer EMPTY_BUFFER = BufferInternal.buffer(Unpooled.EMPTY_BUFFER);
+  private static final String RESPONSE_WRITTEN = "Response has already been written";
+
+  private final VertxInternal vertx;
+  private final HttpRequest request;
```
vietj, line 58: this file seems to have been reformated, please keep only the necessary changes

vertx-core/src/test/java/io/vertx/tests/http/Http1xTest.java
```diff
@@ -42,6 +42,7 @@
 import org.junit.Test;
 
 import java.io.File;
+import java.nio.channels.FileChannel;
```
vietj, line 45: not necessary : please remove

olivierayache: fixed

vertx-core/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java
olivierayache: fixed

PR #5559, opened 2025-05-05
Introduce an option for HTTP/2 upgrade max content length

geoand:
Relates to: https://github.com/quarkusio/quarkus/issues/47649

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
@@ -122,6 +122,11 @@ public class HttpClientOptions extends ClientOptionsBase {
    */
   public static final boolean DEFAULT_HTTP2_CLEAR_TEXT_UPGRADE_WITH_PREFLIGHT_REQUEST = false;
 
+  /**
+   * Default maximum length of the aggregated content in bytes
+   */
+  public static final int DEFAULT_HTTP2_UPGRADE_MAX_CONTENT_LENGTH = 65536;
+
   /*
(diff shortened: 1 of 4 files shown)


vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -530,6 +538,24 @@ public HttpClientOptions setHttp2KeepAliveTimeout(int keepAliveTimeout) {
...
 
+  /**
+   * @return the HTTP/2 upgrade maximum length of the aggregated content in bytes
+   */
+  public int getHttp2UpgradeMaxContentLength() {
+    return http2UpgradeMaxContentLength;
+  }
+
+  /**
+   * Set the HTTP/2 upgrade maximum length of the aggregated content in bytes.
```
vietj, line 549: I think we need to add that this is only effective when `http2ClearTextUpgradeWithPreflightRequest` is set to false (which is the default).

When http2ClearTextUpgradeWithPreflightRequest is true, then the client makes a preflight OPTIONS request and the upgrade will not send a body, voiding the requirements.

geoand: Good point!

vietj:
thank you @geoand

geoand:
🙏🏽

geoand:
@vietj WDYT about backporting to Vert.x 4?

vietj:
I think it's a good idea @geoand

geoand:
https://github.com/eclipse-vertx/vert.x/pull/5560

Issue #5589, opened 2025-06-05
HTTP server started on a duplicated context shares contextual data


tsegismont:
Originally reported by @chrispatmore in https://github.com/eclipse-vertx/vert.x/issues/5505#issuecomment-2903779644

> We stumbled upon this because:
> 
>     We were using the ContextualData class for storing trace information, and for logging that information using log4j
>     ContextualData puts a map into the localContextData on reads (empty)
>     Duplicating a DuplicatedContext with the map in keeps the same map
>     So updating the data also updates the data in other existing copies of the duplicated context, meaning the data can be present for new requests or changed in other requests
>     You can get a DuplicatedContext as the context that is duplicated for each HttpRequest, if the server was started (listen() called) when already in a DuplicatedContext.
> 
> e.g.
> 
>     have a verticle that has a server instance defined.
>     Start the server on an event bus call (duplicated context now), which also calls ContextualData.getAll() (simulate logger calling get) first.
>     send API calls to server, and see that the contextual data is shared between requests now.
> 

If a server is started on duplicated context, the listening context should not be the current context but the unwrapped context.

tsegismont:
Fixed by https://github.com/eclipse-vertx/vert.x/pull/5606

PR #5590, opened 2025-06-05
HTTP server started on a duplicated context shares contextual data

tsegismont:
See #5589

If the server is started from a duplicated context, the listen context should be the underlying context (unwrapped).

And then we should duplicate the unwrapped context.

This is to make sure the creating context and the request context are not sharing local data.


tsegismont:
Superseded by https://github.com/eclipse-vertx/vert.x/pull/5606

Issue #5600, opened 2025-06-11
HttpServerRequest#toNetSocket make the server report its WebSocket capabilities


tsegismont:
### Version

4.5.15

### Context

Initial report in https://github.com/eclipse-vertx/vertx-http-proxy/issues/125

`HttpServerRequest#toNetSocket` can be used in a proxy to manually upgrade a server connection to WebSocket.

When this method is called, Netty's `WebSocketServerExtensionHandler` adds to response headers all the capabilities of the server (e.g. per message compression). This is fine when the origin (backend) has the same configuration. Otherwise, this breaks the WebSocket handshake protocol.

tsegismont:
Fixed by https://github.com/eclipse-vertx/vert.x/pull/5608

PR #5601, opened 2025-06-12
HttpServerRequest#toNetSocket make the server report its WebSocket capabilities

tsegismont:
See #5600

When the connection is upgraded manually, it's up to the user to manage the handshake, including extensions.


tsegismont:
Thanks @vietj for taking a look

PR #5618, opened 2025-06-18
Add the support to FileChannel to sendFile in HttpServerResponse (5.0.x branch)

olivierayache:
Motivation:

As I am using Vert.x to create a REST web server to retrieve files I was really intersted by the [sendFile](https://vertx.io/docs/apidocs/io/vertx/core/http/HttpServerResponse.html#sendFile-java.lang.String-long-io.vertx.core.Handler-) feature of HttpServerResponse. After some performance testing and profiling I discovered that the current implementation of sendFile is not optimal for my use case: high throughput with the same files often accessed.
The repetitive call to open method on RandomAccessFile had a big impact on the performance of my code. That is why I added the possibility to use sendFile with a FileChannel and let the developper to handle the lifecycle of that channel.
By doing that I managed to get a huge improve of the performance of my web server (roughly x2 in throughput)

This PR is the 5.0.x version of https://github.com/eclipse-vertx/vert.x/pull/5558


olivierayache:
Hello @vietj here is the PR for the branch 5.0.x to add the support to FileChannel to sendFile in HttpServerResponse

vietj:
@olivierayache it looks good, I'll merge it soon

vietj:
@olivierayache actually there is one change I would like to see : the methods use the file extension to setup the response content-type. I would like this to be removed and instead have the developer to set himself the content-type header on the response before sending the file.

olivierayache:
Ok I will work on that. I was doing that to mimic what was done with file url

olivierayache:
Hello @vietj 
Just improved by excluding extension

vietj:
great contribution, thanks @olivierayache

vietj:
@olivierayache can you provide a PR for master as well ?

olivierayache:
I will do today or tomorrow

PR #5628, opened 2025-06-27
A couple of HTTP send file fixes

vietj:
Motivation:

- Avoid idle timeout prematurely closing connection when sending files over TLS
- Sending a `FileChannel`/`RandomAccessFile` should not close the descriptor over TLS

Changes:

- Override `ChunkedNioFile` to prevent the file descriptor to be closed
- When configuring a `NetSocket`, add the idle state handler after the chunked write handler


vietj:
wrong branch

PR #5645, opened 2025-07-10
Optimize Buffer's bound checks (Fixes #5643)

### LINKED ISSUE #5643: Buffer's bound check are not well optimized
### Version

5.x

### Context

Same of https://github.com/netty/netty/pull/15462

https://github.com/eclipse-vertx/vert.x/blob/fc4640cb56043f451eb59929aba94d06782561db/vertx-core/src/main/java/io/vertx/core/buffer/impl/BufferImpl.java#L196

is performing math operations (chain of `or`s) to outsmart the JIT compiler, but such can just confuse it, instead, due to lack of normal comparisons - which would help it to understand what's the checks are for (in the Netty pr there's more info about it).

What I don't understand is why such checks exists, too: why not relying on Netty's bound checks instead?
If users want to read past what's written - is their choice - since they have the duty to:
- check how much readable bytes are there already
- store a read index to remember where they are

### Steps to reproduce

checkout https://github.com/franz1981/vert.x/commit/2bda8aeb1771e4d63a2ab445d8c4f18cca427b1d and run the new benchmark

franz1981:
Motivation:

It fixes #5643

on `JDK 17`:

The old results vs Netty were:
```
Benchmark                                        (batchSize)  Mode  Cnt   Score   Error  Units
BufferReadBenchmark.getByteBatchNetty                     64  avgt   20  57.338 ± 0.176  ns/op
BufferReadBenchmark.getByteBatchVertx                     64  avgt   20  81.085 ± 0.109  ns/op
BufferReadBenchmark.getBytesConstantOffsetNetty           64  avgt   20   3.764 ± 0.040  ns/op
BufferReadBenchmark.getBytesConstantOffsetVertx           64  avgt   20   6.233 ± 0.196  ns/op
BufferReadBenchmark.getBytesNetty                         64  avgt   20   6.225 ± 0.022  ns/op
BufferReadBenchmark.getBytesVertx                         64  avgt   20   6.124 ± 0.038  ns/op
```
With these changes:
```
Benchmark                                        (batchSize)  Mode  Cnt   Score   Error  Units
BufferReadBenchmark.getByteBatchNetty                     64  avgt   20  57.338 ± 0.176  ns/op
BufferReadBenchmark.getByteBatchVertx                     64  avgt   20  61.578 ± 0.029  ns/op
BufferReadBenchmark.getBytesConstantOffsetNetty           64  avgt   20   3.764 ± 0.040  ns/op
BufferReadBenchmark.getBytesConstantOffsetVertx           64  avgt   20   5.218 ± 0.147  ns/op
BufferReadBenchmark.getBytesNetty                         64  avgt   20   6.225 ± 0.022  ns/op
BufferReadBenchmark.getBytesVertx                         64  avgt   20   5.160 ± 0.009  ns/op
```

while comparing just vertx with itself:
```
                                                        Before         |         After    
Benchmark                                         Score   Error  Units |  Score   Error  Units 
BufferReadBenchmark.getByteBatchVertx            81.085 ± 0.109  ns/op | 61.578 ± 0.029  ns/op 
BufferReadBenchmark.getBytesConstantOffsetVertx   6.233 ± 0.196  ns/op |  5.218 ± 0.147  ns/op
BufferReadBenchmark.getBytesVertx                 6.124 ± 0.038  ns/op |  5.160 ± 0.009  ns/op
```
There's an interesting "plot-twist" on `BufferReadBenchmark.getBytes` since, thank to this patch it appears 
that vertx is faster than Netty, although performing an additional bound check; it's likely that expressing it proved the subsequent one to be redundant/simplified - making some JIT strenght reduction to kicks-in.

Indeed, on my machine (the short LE read in Netty was
```asm
test   %edi,%edi
jl     0x00007f35381f0650
vmovq  %rax,%xmm0
mov    %r10,%rax
lea    -0x7(%rsi),%ebx
mov    %r9d,%edi
or     $0x4,%edi
or     %ecx,%edi
lea    0x7(%r11),%r10d
or     %r10d,%edi
or     %ebx,%edi
movslq %ebp,%r8
movswl 0x10(%rdx,%r8,1),%ebx
```
whilst the vertx one (with this PR in):
```asm
cmp    $0x3,%r9d
jl     0x00007f2ce01ededc
mov    $0x3,%eax
or     %ebx,%eax
lea    -0x3(%rbx),%esi
or     $0x3,%eax
or     %esi,%eax         
test   %eax,%eax
jl     0x00007f2ce01edf10
movswl 0x11(%r12,%rdi,8),%esi
``` 
which contains less math (e.g. `lea` and `or` and an additional `jl` - although I don't see the value of `cmp    $0x3,%r9d` (maybe a JIT bug?) since `r9d` never change (it is written once before all the reads happen).

FYI @vietj it is likely Vertx 4.x to be affected as well, I haven't checked if we have custom bound checks there too

> NOTE: 
the current Netty 4.2 branch is not including an important improvement re bound checks, see https://github.com/netty/netty/pull/15452

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/buffer/impl/BufferImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/buffer/impl/BufferImpl.java
@@ -195,7 +195,7 @@ public class BufferImpl implements BufferInternal {
 
   private void checkUpperBound(int index, int size) {
     int length = buffer.writerIndex();
-    if ((index | length - (index + size)) < 0) {
+    if (index < 0 || index + size < 0 || index + size > length) {
       throw new IndexOutOfBoundsException(index + " + " + size + " > " + length);
     }
   }
--- a/vertx-core/src/test/java/io/vertx/benchmarks/BufferReadBenchmark.java
+++ b/vertx-core/src/test/java/io/vertx/benchmarks/BufferReadBenchmark.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2011-2025 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+
+package io.vertx.benchmarks;
+
+
+import io.netty.buffer.ByteBuf;
+import io.netty.util.ResourceLeakDetector;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.impl.buffer.VertxByteBufAllocator;
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Warmup(iterations = 10, time = 1)
+@Measurement(iterations = 10, time = 1)
+@Fork(2)
+public class BufferReadBenchmark {
+
+  static {
+    ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.DISABLED);
+  }
+
+  private ByteBuf nettyByteBuf;
+  private Buffer vertxBuffer;
+  @Param({"64"})
+  private int batchSize;
+
+  @Setup
+  public void setup() {
+      nettyByteBuf = VertxByteBufAllocator.DEFAULT.heapBuffer(batchSize);
+      nettyByteBuf.ensureWritable(batchSize);
+      vertxBuffer = Buffer.buffer(batchSize);
+      for (int i = 0; i < batchSize; i++) {
+          nettyByteBuf.setByte(i, (byte) i);
+          vertxBuffer.setByte(i, (byte) i);
+      }
+  }
+
+  @Benchmark
+  public void getByteBatchNetty(Blackhole bh) {
+    ByteBuf buffer = this.nettyByteBuf;
+    for (int i = 0, size = batchSize; i < size; i++) {
+      bh.consume(buffer.getByte(i));
+    }
+  }
+
+  @Benchmark
+  public void getByteBatchVertx(Blackhole bh) {
+    Buffer buffer = this.vertxBuffer;
+    for (int i = 0, size = batchSize; i < size; i++) {
+      bh.consume(buffer.getByte(i));
+    }
+  }
+
+  @Benchmark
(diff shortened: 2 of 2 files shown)


franz1981:
@vietj refreshed my agreement - it should be fine now

franz1981:
It looks like Java 9 got an intrinsics Java 9 method we could leverage i.e. https://docs.oracle.com/javase/9/docs/api/java/util/Objects.html#checkIndex-int-int-

franz1981:
It looks like using Java 9+ intrinsics deliver a better perf too:
```
BufferReadBenchmark.getByteBatchVertx                         64  avgt   20  59.914 ± 0.478  ns/op
BufferReadBenchmark.getBytesConstantOffsetVertx               64  avgt   20   5.020 ± 0.011  ns/op
BufferReadBenchmark.getBytesVertx                             64  avgt   20   5.028 ± 0.024  ns/op
```
marginally, but costantly better.

And still I won't use it - it doesn't fit well - if possible would be better to just drop the check wdyt @vietj ?

franz1981:
@cescoffier too much fun today!

so...
if we run the benchmark `$ java -jar vertx-core/target/vertx-core-5.1.0-SNAPSHOT-benchmarks.jar BufferReadBenchmark --jvmArgs="--sun-misc-unsafe-memory-access=allow"
`

with JDK 24 we won't have big impacts vs JDK 21, but without, any buffer access (netty or vetx) will get some severe penalty;  it's described at https://github.com/netty/netty/wiki/Java-24-and-sun.misc.Unsafe

I have to check why it's biting us with 4.2 as well

franz1981:
https://github.com/eclipse-vertx/vert.x/blob/a952b9218c82f6a2b310e45d5d8a394a040d49c4/vertx-core/src/main/java/io/vertx/core/impl/buffer/VertxByteBufAllocator.java#L47 this can be the reason why JDK 24 has a perf hit, it requires investigation before it hits Quarkus

franz1981:
I have found the culprit and is very very tricky:
- as said in https://github.com/eclipse-vertx/vert.x/pull/5645#issuecomment-3062749503 we now use `Unsafe`-free Heap ByteBufs - which is not great for multi-bytes primitive operations (let's talk about it in another issue..)
- since `Unsafe` is missing the reference count scheme of Netty cannot perform non-volatile read vs the ref cnt - because it is declared as volatile (with VarHandle or other classes maybe we could? let's see)
- because of this, any tight loop/sequence calling buffer's read/write has (because of the Java Memory Model) to perform a ref cnt volatile load which force re-reading the array in the buffer, the array's length, performing a bound check each time, limiting unrolling...etc etc TLDR a sh*t-show, literally :"(

I'll send soon a patch in Netty, and yes, the workaround to this is https://github.com/eclipse-vertx/vert.x/pull/5645#issuecomment-3062488651

vietj:
@franz1981 ECA check failing

franz1981:
Strange, i have updated it

franz1981:
@vietj it doesn't fail anymore?

In the meantime I'm fixing the Unsafe lack in JDK 24 for Netty :P , FYI https://github.com/netty/netty/pull/15486
This will make the Vertx buffers to **fly** in Quarkus @geoand :P

PR #5651, opened 2025-07-17
Verticle instance leak when using shared servers

### LINKED ISSUE #5621: Verticle instance leak when using shared servers
### Version

4.5.11, probably 5.x too

### Context

We've encountered some memory leaks related to quite frequently deploying and undeploying verticles, and where verticles are starting Http Servers. The started http servers all share the same port number. Rough timeline:

1. First verticle deployed and starts a http server
2. Verticles are deployed and undeployed quite frequently - they start http servers using the same port and address as the first server

Now, the started http servers use the first server under the hood since it was the first one to start listening - and they correctly register and unregister themselves as workers on the first server, however:

1. There's a `Set<TCPServerBase> servers` in TCPServerBase (vert.x 4) or `Set<NetServerImpl> servers` in NetServerImpl (vert.x 5)
2. All subsequent http servers seem to append themselves to this field on the first server
3. But they don't seem to remove themselves when the server is closed
4. Thus, the first server keeps a strong reference to all closed http servers
5. Thus, all closed servers keep a strong reference to the (closed) context
6. Thus, keeping a reference to the deployment
7. Thus, keeping a reference to undeployed verticles etc.

As I understand this, all these objects will be kept alive until all the server for the "address & port combination" have been closed for at least Vert.x 4, but probably Vert.x 5 too.

For reference, retention path for a dead verticle from a memory dump:

Image

My interpretation is that in TCPServerBase.close, in addition to calling balancer.removeWorker, actualServer.servers.remove(this) should also be called (Vert.x 4)

I think that the same would be done in NetServerImpl.doShutdown for Vert.x 5..?

### Steps to reproduce

1. Deploy verticle with one http server on port X
2. Deploy and undeploy a verticle continuosly. Let the verticle start a http server on the same port X.
3. Observe the undeployed verticles in step 2 being leaked

### Do you have a reproducer?

_No response_

tsegismont:
Fixes #5621

Get rid of the servers set that was introduced in 86ca0cf but became redundant when close implementation was reworked using close futures in #4352

### DIFF
--- a/src/main/java/io/vertx/core/net/impl/TCPServerBase.java
+++ b/src/main/java/io/vertx/core/net/impl/TCPServerBase.java
@@ -11,7 +11,6 @@
 package io.vertx.core.net.impl;
 
 import io.netty.bootstrap.ServerBootstrap;
-import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
@@ -25,10 +24,9 @@ import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Promise;
-import io.vertx.core.buffer.impl.PartialPooledByteBufAllocator;
 import io.vertx.core.impl.ContextInternal;
-import io.vertx.core.impl.future.PromiseInternal;
 import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.impl.future.PromiseInternal;
 import io.vertx.core.impl.logging.Logger;
 import io.vertx.core.impl.logging.LoggerFactory;
 import io.vertx.core.net.NetServerOptions;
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/core/net/impl/TCPServerBase.java
```diff
@@ -11,24 +11,13 @@
...
 
 import io.netty.bootstrap.ServerBootstrap;
-import io.netty.buffer.PooledByteBufAllocator;
-import io.netty.channel.Channel;
-import io.netty.channel.ChannelFuture;
-import io.netty.channel.ChannelFutureListener;
-import io.netty.channel.ChannelOption;
-import io.netty.channel.EventLoop;
-import io.netty.channel.EventLoopGroup;
+import io.netty.channel.*;
```
vietj, line 14: pls no star imports :-)

tsegismont: I'll update the PR.

This is a good example of the limitation of using "the IntelliJ defaults" as stated in https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines

Issue #5666, opened 2025-08-04
Extend the DNS hosts refresh period, the current limit is < 3 seconds.


sagaxu:
### Describe the feature

According to the following document
[https://vertx.io/docs/apidocs/io/vertx/core/dns/AddressResolverOptions.html#setHostsRefreshPeriod(int)](https://vertx.io/docs/apidocs/io/vertx/core/dns/AddressResolverOptions.html#setHostsRefreshPeriod(int))

The max int nanos is 2147483648，only 2.147 seconds. 

The underlying implementation actually uses the long type
[https://github.com/eclipse-vertx/vert.x/blob/8ce1eb1d982953e18fbd4a249b6d4142b603488f/vertx-core/src/main/java/io/vertx/core/dns/impl/DnsAddressResolverProvider.java#L55](https://github.com/eclipse-vertx/vert.x/blob/8ce1eb1d982953e18fbd4a249b6d4142b603488f/vertx-core/src/main/java/io/vertx/core/dns/impl/DnsAddressResolverProvider.java#L55)

Maybe we could change the hostsRefreshPeriod below to long.

[https://github.com/eclipse-vertx/vert.x/blob/8ce1eb1d982953e18fbd4a249b6d4142b603488f/vertx-core/src/main/java/io/vertx/core/dns/AddressResolverOptions.java#L115](https://github.com/eclipse-vertx/vert.x/blob/8ce1eb1d982953e18fbd4a249b6d4142b603488f/vertx-core/src/main/java/io/vertx/core/dns/AddressResolverOptions.java#L115)

and then use setHostsRefreshPeriod(long) for longer  period.

tsegismont:
Closed by https://github.com/eclipse-vertx/vert.x/pull/5669

PR #5668, opened 2025-08-04
Use long to set the DNS hosts file refresh period instead of int

sagaxu:
Motivation:
  int can only express nanoseconds of less than 3 seconds, which is probably not enough

See #5666


vietj:
@sagaxu thanks, instead can you introduce a TimeUnit that is by default nano seconds but could be changed to another value instead of changing int to long. You can take TCPSSLOptions which has idleTimeout / idleTimeoutUnit as reference

sagaxu:
> @sagaxu thanks, instead can you introduce a TimeUnit that is by default nano seconds but could be changed to another value instead of changing int to long. You can take TCPSSLOptions which has idleTimeout / idleTimeoutUnit as reference

thanks, please see [https://github.com/eclipse-vertx/vert.x/pull/5669](https://github.com/eclipse-vertx/vert.x/pull/5669)

Issue #5487, opened 2025-02-19
Memory leak bug in Http1xClientConnectionClass


serez:
### Version

Which version(s) did you encounter this bug ? 4.2.6

### Context

In http1xClientConnection.handleHttpMessage() the code throws a "Received HTTP message with no request in progress" exception when a response is received without a request in progress. In this case, the content of a DefaultHttpContent is a byteBuf and hold data.  Per netty's reference counted objects documentation, the code should release the byteBuf buffer before throwing the exception, as nothing will access it anymore.
_The general rule of thumb is that the party that accesses a reference-counted object last is also responsible for the destruction of that reference-counted object. Specifically:

    If a [sending] component is supposed to pass a reference-counted object to another [receiving] component, the sending component usually does not need to destroy it but defers that decision to the receiving component.
    If a component consumes a reference-counted object and knows nothing else will access it anymore (i.e., does not pass along a reference to yet another component), the component should destroy it._
  

### Do you have a reproducer?
The code was ran in Http1xTest class in the vertx code.
Updating requestNum to big enough of a value will create an out of memory exception. Current code will show the increase in used Direct Memory 

```
@Test
  public void testMemoryLeak() {

    String longParam = TestUtils.randomAlphaString(5000);
    String content = TestUtils.randomAlphaString(1000000);
    NetServer server = vertx.createNetServer();
    AtomicBoolean isConnected = new AtomicBoolean(false);
    final int requestNum = 10000;
    AtomicInteger requestCounter = new AtomicInteger(requestNum);
    ByteBufAllocatorMetricProvider unpooledMetricProvider =
      (ByteBufAllocatorMetricProvider) VertxByteBufAllocator.POOLED_ALLOCATOR;
    PooledByteBufAllocatorMetric metric = (PooledByteBufAllocatorMetric) unpooledMetricProvider.metric();

    long prevUsedDirectMemory = metric.usedDirectMemory();
    long firstUsedDirectMemoryAllocation = 0;


    server.connectHandler(so -> {
      System.out.println("server.connectHandler");
      for (int i = 0; i < 2; i++) {
        so.write("HTTP/1.1 200 OK\r\n" +
          "Content-Type: text/plain\r\n" +
          "Content-Length: " + content.length() + "\r\n" +
          "\r\n" +
          content);

      }
    });


    try {
      server.listen(testAddress, onSuccess(v -> {
        isConnected.set(true);

        }));
      while(!isConnected.get()) {
          try {
[...]

PR #5670, opened 2025-08-05
Exclude connections with outstanding bytes to read from being used via connection pooling. Fixes #5663

### LINKED ISSUE #5663: Bytes window state reset during connection reusing from connection pool
### Version

4.2.6

### Context

We are encountering an issue when resetting a Http1xClientConnection. The byteswindow variable does not get reset to 0 under certain conditions and when the connection is reused by a new request, since the byteswindow is not reset, it is causing the socket reads to pause for the new request. This happens when the byteswindow goes higher than the higher water mark. Here is the flow of events:

1. first request - reads 56KiB + 1 from socket. auto read -> true. bytes window: 0
2. first request client stream is paused. bytes window: 56KiB + 1. auto read -> true
3. first request - complete because socket read was complete. connection is back to pool. bytes window: 56KiB + 1, auto read -> true
4. second request - read some amount of data < total content length from socket
5. second request - splits into 8KiB chunks, passes one to Vert.x. bytes window: 64KiB + 1. auto read -> false
6. second request - not paused, so it will ack 8KiB + 1. bytes window: 56KiB. auto read -> false
7. second request - will read remaining content that was read from socket. But since auto read is false, 
8. it will never read any additional data from the socket and will hang

We noticed that the variable byteswindow has been removed in release 5.0.1. We are currently on version 4.2.6. Since the upgrade is a big jump for us, we wanted to check the possibility of getting the change back ported to version 4.2.6. 
 

### Steps to reproduce

**Test for 4.2.6. This was tested in Http1xClientConnectionTest.java**

```
@Test
  public void testStuckGet() throws TimeoutException {
    HttpServer server = vertx.createHttpServer();

    Future<HttpServer> listen = server.requestHandler(req -> {
      int length = 65536 + 1;
      req.response().putHeader("Content-Length", length + "");
      req.response().write(Buffer.buffer(new byte[length]));
      req.response().end();
    }).listen(9080);
    listen.result();

    CompletableFuture<Void> firstRequestEnded = new CompletableFuture<>();
    AtomicReference<String> chosenPort = new AtomicReference<>();
    AtomicReference<String> pausedRequestId = new AtomicReference<>();

    client.request(HttpMethod.GET, 9080, "localhost", "/", response -> {
      AtomicInteger buffer = new AtomicInteger();

      HttpClientRequest result = response.result();
      String requestId = UUID.randomUUID().toString();
      pausedRequestId.set(requestId);
      log.error("The problematic request id is: " + requestId);
      if (response.result().connection() instanceof Http1xClientConnection) {
        Http1xClientConnection conn = (Http1xClientConnection) response.result().connection();
//        conn.setRequestId(requestId);
      }

      netty reads 64 at a time
        passes 8 to vertx at a time
        read all 56+1 from socket
        + 1 is because we read 8 kib at a time -> if we set at 56


      chosenPort.set(response.result().connection().localAddress().toString());
      ports.putIfAbsent(chosenPort.get(), 0);
      ports.put(chosenPort.get(), ports.get(chosenPort.get()) + 1);
      result.response(response1 -> {
        HttpClientResponse result1 = response1.result();
        result1.pause();
        result1.handler(event -> buffer.getAndAdd(event.length()));
        result1.endHandler(event -> firstRequestEnded.complete(null));
      });
      result.end();
    });

    while (true) {
      CompletableFuture<Void> future = new CompletableFuture<>();
      String requestId = UUID.randomUUID().toString();
      client.request(HttpMethod.GET, 9080, "localhost", "/", response -> {
        if (response.result().connection() instanceof Http1xClientConnection) {
          Http1xClientConnection conn = (Http1xClientConnection) response.result().connection();
//          conn.setRequestId(requestId);
        }

        AtomicInteger buffer = new AtomicInteger();
        HttpClientRequest result = response.result();
        ports.putIfAbsent(response.result().connection().localAddress().toString(), 0);
        ports.put(response.result().connection().localAddress().toString(), ports.get(response.result()
          .connection().localAddress().toString()) + 1);
        if (response.result().connection().localAddress().toString().equals(chosenPort.get())) {
          log.error(requestId + " should be a stuck get on port " + chosenPort.get() + " which is caused by " +
            pausedRequestId.get());
        }
        log.error(ports);

        result.response(response1 -> {
          HttpClientResponse result1 = response1.result();
          if (result1 == null) {
            log.error(requestId + " had a null result, oops");
            future.completeExceptionally(new NullPointerException());
          } else {
            result1.handler(event -> {
              buffer.getAndAdd(event.length());
              log.error(requestId + ": buffer up to " + buffer.get());
            });
            result1.endHandler(event -> future.complete(null));
            result1.exceptionHandler(t -> {
              log.error(t.getMessage(), t);
              future.completeExceptionally(t);
            });
          }
        });
        result.end();
      });
      try {
        log.error("Waiting on: " + requestId);
        future.get(30000, TimeUnit.MILLISECONDS);
      } catch (InterruptedException | ExecutionException e) {
        throw new RuntimeException(e);
      }
    }
  }
```

**Test for 5.0.1. This was tested in Http1xClientConnectionTest.java**
```
@Test
  /*
   * This test reproduces an issue involving connection pooling and parallel reads to the same server. When
   * you have a client which pauses its reading of data AND that remaining data is able to be read into InboundBuffer,
   * AND then you have another client to the same server on the same port (via pooling) which attempts to read all
   * of the data. Then the second client will "hang" and eventually fail due to a timeout (if there is one).
   *
   * In order to avoid this as-is, connection pooling has to be disabled which may not be practical.
   */
  public void testStuckGet() throws TimeoutException {
    int portToUse = 9080;
    // This magic number is based on the high water mark of Http1xClientConnection, which is based on Netty.
    // The HTTP client uses Netty's HTTP object decoder, which uses a chunk size of 8KiB. This means that
    // a socket read of 64KiB (the write queue limit of the channel) will result in up to 8 messages which are
    // passed down to the Vert.x handler. When the original request is paused after reading one buffer, it
    // leaves 56KiB in the bytesWindow (not enough to hit the HWM and disable autoRead).
    // When the next request comes in, it will read 64KiB from the socket. Then 8KiB into the InboundBuffer. Since
    // this will be > the HWM (64KiB + 1), autoRead will be disabled. The second request is not paused, so it will
    // then read the 8KiB which was buffered. However, since autoRead is disabled now, no further bytes will be read
    int contentLength = 57344 + 1;
    // This is the max number of requests to make before considering the test a success
    int maxRequests = 5000;
    int idleTimeout = 30000;
    Map<String, Integer> ports = new HashMap<>();
    HttpServer server = vertx.createHttpServer();

    // Set up a generic server which responds with data that will exceed the HWM of Http1xClientConnection (64KiB)
    // However, the data can't be too large because it needs to all be buffered by InboundBuffer, otherwise
    // the original request which is paused won't be complete
    Future<HttpServer> listen = server.requestHandler(req -> {
      req.response().putHeader("Content-Length", contentLength + "");
      req.response().write(Buffer.buffer(new byte[contentLength]));
      req.response().end();
    }).listen(portToUse);
    listen.result();

    CompletableFuture<Void> firstRequestEnded = new CompletableFuture<>();
    AtomicReference<String> chosenPort = new AtomicReference<>();
    AtomicReference<String> pausedRequestId = new AtomicReference<>();
    AtomicInteger pausedRequestBuffer = new AtomicInteger();

    Future<HttpClientRequest> originalRequest = client.request(HttpMethod.GET, portToUse, "localhost", "/");
    originalRequest.onComplete(response -> {
      HttpClientRequest result = response.result();
      String requestId = UUID.randomUUID().toString();
      pausedRequestId.set(requestId);
      log.error("The problematic request id is: " + requestId);
      if (response.result().connection() instanceof Http1xClientConnection) {
        Http1xClientConnection conn = (Http1xClientConnection) response.result().connection();
//        conn.setRequestId(requestId);
      }
      chosenPort.set(response.result().connection().localAddress().toString());
      ports.putIfAbsent(chosenPort.get(), 0);
      ports.put(chosenPort.get(), ports.get(chosenPort.get()) + 1);

      result.response().onComplete(response1 -> {
        HttpClientResponse result1 = response1.result();
        // Even though the response stream is paused, the remaining data will fit into InboundBuffer's pending
        // queue. This means its own end handler will be triggered, and the connection is available for reuse
        result1.handler(event -> pausedRequestBuffer.getAndAdd(event.length()));
        result1.pause();
        result1.endHandler(event -> firstRequestEnded.complete(null));
      });
      result.end();
    });

    // Now loop forever until we can get a request which shares the same port as the original request which is paused
    // This means the connection was reused
    int requestCount = 0;
    while (requestCount < maxRequests) {
      requestCount++;
      // We create requests sequentially and block on the future completing. The future is complete when all data
      // is retrieved. There is no need to do parallel requests
      CompletableFuture<Void> future = new CompletableFuture<>();
      String requestId = UUID.randomUUID().toString();
      AtomicInteger buffer = new AtomicInteger();

      client.request(HttpMethod.GET, portToUse, "localhost", "/").onComplete(response -> {
        if (response.result().connection() instanceof Http1xClientConnection) {
          Http1xClientConnection conn = (Http1xClientConnection) response.result().connection();
//          conn.setRequestId(requestId);
        }

        HttpClientRequest result = response.result();
        ports.putIfAbsent(response.result().connection().localAddress().toString(), 0);
        ports.put(response.result().connection().localAddress().toString(), ports.get(response.result()
          .connection().localAddress().toString()) + 1);
        log.error("Ports in use are below. The problematic request was on port " + chosenPort.get() + " so there must " +
          "be > 1 request on that port to reproduce this issue");
        log.error(ports);

        if (response.result().connection().localAddress().toString().equals(chosenPort.get())) {
          // We can detect that this request should hang because it's reusing the paused request's port.
          log.error(requestId + " should be a stuck get on port " + chosenPort.get() + " which is caused by " +
            pausedRequestId.get());
        }

        // This block of code just processes the results from the server. It completes the request future once
        // the end handler runs. If the end handler never runs, it means we have successfully hit a "stuck get"
        result.response().onComplete(response1 -> {
          HttpClientResponse result1 = response1.result();
          if (result1 == null) {
            log.error(requestId + " had a null result, oops");
            future.completeExceptionally(new NullPointerException());
          } else {
            result1.handler(event -> {
              buffer.getAndAdd(event.length());
              log.error(requestId + ": The client's buffer has accumulated up to " + buffer.get() + " bytes");
            });
            result1.endHandler(event -> {
              if (buffer.get() != contentLength) {
                future.completeExceptionally(new RuntimeException(String.format("content length doesn't match: Was %s" +
                  " but expected %s", buffer.get(), contentLength)));
              } else {
                future.complete(null);
              }
            });
            result1.exceptionHandler(t -> {
              log.error(t.getMessage(), t);
              future.completeExceptionally(t);
            });
          }
        });
        result.end();
      });
      try {
        log.error("Waiting on: " + requestId);
        future.get(idleTimeout, TimeUnit.MILLISECONDS);
      } catch (InterruptedException | ExecutionException e) {
        throw new RuntimeException(e);
      } catch (TimeoutException e) {
        log.error("Timed out waiting on: " + requestId + ". Received " + buffer
          .get() + " out of " + contentLength + " bytes");
        log.error(ports);
        throw e;
      }
    }
  }
```

### Do you have a reproducer?

_No response_

cfolgere:
Applies to 4.x

Motivation:

This is a proposed fix for https://github.com/eclipse-vertx/vert.x/issues/5663. This issue can cause a request using Vert.x client to "hang" if it involves connection reuse which has a previous request on it. The previous request must have already ended (as in all data was retrieved) but the data was not consumed (as in, the client response stream was paused). If this occurs, then the subsequent request can hang because the readWindow value, which is used to determine if the Channel auto read setting is enabled, is left over from the previous request. Under the right conditions, it can be impossible for the second request to make progress.

In this fix, connections are not reused from the connection pool if there is an outstanding value for readWindow by adding an isValid check before a connection is checked out (This would be correct anyway in case the connection has since expired but the PoolChecker has not cleaned it up yet). This means the outstanding request will need to complete before a second request can use the same connection. However, in the scenario above the second request would never succeed, so there is no loss in throughput there.

I ran the same test described in https://github.com/eclipse-vertx/vert.x/issues/5663. I ran it on the 4.x branch and it still occurs, but with this fix applied it no longer occurs. 

Before fix, where the request using the same port hangs indefinitely
```
Ports in use are below. The problematic request was on port 127.0.0.1:63902 so there must be > 1 request on that port to reproduce this issue 
{127.0.0.1:63902=3, 127.0.0.1:63903=1098} 
cc483753-b354-4352-9fef-0da3df6c24bb should be a stuck get on port 127.0.0.1:63902 which is caused by 2f39ec98-75c0-4833-8d4b-622e586daa6e
```

After fix, since the connection is available for reuse, a new connection is opened (which then itself is reused many times).
```
Ports in use are below. The problematic request was on port 127.0.0.1:63917 so there must be > 1 request on that port to reproduce this issue 
{127.0.0.1:63917=1, 127.0.0.1:63918=3943, 127.0.0.1:63919=1057} 
```

Conformance:

### DIFF
--- a/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
+++ b/src/main/java/io/vertx/core/http/impl/Http1xClientConnection.java
@@ -1311,7 +1311,12 @@ public class Http1xClientConnection extends Http1xConnectionBase<WebSocketImpl> 
 
   @Override
   public boolean isValid() {
-    return expirationTimestamp == 0 || System.currentTimeMillis() <= expirationTimestamp;
+    // The readWindow check is here because it's possible for a connection to released back to the pool
+    // if a request ends, but the client has not consumed the response. In this case, readWindow will be
+    // high and can lead to the channel autoRead being disabled if a new request comes in, causing this new
+    // request to hang until the first client consumes the response of the first request.
+    // So instead, exclude such connections from being reused until the client has finished consuming the response.
+    return (expirationTimestamp == 0 || System.currentTimeMillis() <= expirationTimestamp) && readWindow == 0;
   }
 
   private synchronized void shutdownNow() {
--- a/src/main/java/io/vertx/core/http/impl/SharedClientHttpStreamEndpoint.java
+++ b/src/main/java/io/vertx/core/http/impl/SharedClientHttpStreamEndpoint.java
@@ -45,9 +45,11 @@ class SharedClientHttpStreamEndpoint extends ClientHttpEndpointBase<Lease<HttpCl
       if (pooled.available() > 0) {
         HttpClientConnection conn = pooled.get();
         if (selected == null) {
-          selected = pooled;
+          if (pooled.get().isValid()) {
+            selected = pooled;
+          }
         } else {
-          if (conn.lastResponseReceivedTimestamp() > last) {
+          if (conn.lastResponseReceivedTimestamp() > last && conn.isValid()) {
             selected = pooled;
           }
         }
--- a/src/test/java/io/vertx/core/http/Http1xClientConnectionTest.java
+++ b/src/test/java/io/vertx/core/http/Http1xClientConnectionTest.java
@@ -11,15 +11,28 @@
 package io.vertx.core.http;
 
 import io.netty.buffer.Unpooled;
+import io.vertx.core.Future;
 import io.vertx.core.MultiMap;
 import io.vertx.core.Promise;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.impl.HttpRequestHead;
 import io.vertx.core.impl.ContextInternal;
+import io.vertx.core.impl.logging.Logger;
+import io.vertx.core.impl.logging.LoggerFactory;
 import org.junit.Test;
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class Http1xClientConnectionTest extends HttpClientConnectionTest {
+  private static final Logger log = LoggerFactory.getLogger(Http1xClientConnectionTest.class);;
 
   @Test
   public void testResetStreamBeforeSend() throws Exception {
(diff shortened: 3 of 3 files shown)


vietj:
@cfolgere can you write a test for this ?

cfolgere:
@vietj I have added a test which is based on the original test from the bug report. 

before fix:

```
[ERROR] Tests run: 7, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 31.37 s <<< FAILURE! -- in io.vertx.core.http.Http1xClientConnectionTest
[ERROR] io.vertx.core.http.Http1xClientConnectionTest.testConnectionReuseLeadsToStuckGetRequest -- Time elapsed: 30.67 s <<< ERROR!
java.util.concurrent.TimeoutException
        at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1960)
        at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2095)
        at io.vertx.core.http.Http1xClientConnectionTest.testConnectionReuseLeadsToStuckGetRequest(Http1xClientConnectionTest.java:243)
```

after fix:
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running io.vertx.core.http.Http1xClientConnectionTest
Starting test: Http1xClientConnectionTest#testResetStreamRequestSent 
Starting test: Http1xClientConnectionTest#testResetStreamBeforeSend 
Starting test: Http1xClientConnectionTest#testConnectionReuseLeadsToStuckGetRequest 
The problematic request id is: 96a65dac-74ff-4d58-a377-257c6d5fff8d 
Starting test: Http1xClientConnectionTest#testServerConnectionClose 
Starting test: Http1xClientConnectionTest#testGet 
Connection evicted 
Starting test: Http1xClientConnectionTest#testStreamGet 
Connection evicted 
Starting test: Http1xClientConnectionTest#testConnectionClose 
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.619 s -- in io.vertx.core.http.Http1xClientConnectionTest        
```

vietj:
thanks I am not sure for the fix yet

I am thinking it should be done differently, e.g. returning the connection to the pool after the request has been fully sent instead that does makes sense to me

cfolgere:
@vietj Thanks for looking, I wasn't confident I could submit a change that would guarantee the connection is returned to the pool. I saw in debugging that when this occurs, requests and responses are both empty since all data has been retrieved (but not ack'd). So it seemed like such a change would be more complex. Since this is only affecting 4.x (same test passes in 5.x) I went with this specific change instead because it guarantees the connection is closed when it's detected as not valid. It will cause an additional connection to be opened for a second request, but the second request will succeed this time.

vietj:
@cfolgere the original issue mentions the request is reset (which interacts with the pool) as normally a connection goes back to the pool when both request and response have ended, which means having received the last response message from netty and having written the last request message.

The connection should go back to the pool when the stream when `handleClose` is called. I can indeed see that it is called by `HttpClientStream#reset` and perhaps that is causing the issue.

I think the fix should be that handleClose is called at the appropriate time instead, so perhaps defer the call to close handler if the connections is not in a suitable state for going back to the pool.

However the reproducer does not use reset the stream. So it is not yet clear to me how the connection can go back to the pool if the request is not reset.

cfolgere:
@vietj Sorry, there is no connection reset being triggered, we were referring to bytesWindow not being reset between requests. The only condition required for this to happen is that a client makes a request, the complete data is buffered in InboundBuffer, and the client pauses the response stream. Then the connection is eligible for reuse so that when the second request comes in, the readWindow will hit the HWM and prevent further data from being read. The reason this happens when the first response is this in HttpClientImpl.java:
```java
            stream.closeHandler(v -> {
              lease.recycle();
            });
```
The stream referenced above is a StreamImpl, and the close handler is called in Http1xClientConnection (this is where the last bit of data is received, but the client doesn't need to read it
```java
      if (!isConnect && chunk instanceof LastHttpContent) {
        handleResponseEnd(stream, (LastHttpContent) chunk);
      }
```
which leads to
```java
    lastResponseReceivedTimestamp = System.currentTimeMillis();
    stream.context.execute(trailer, stream::handleEnd);
    if (stream.requestEnded) {
      stream.context.execute(null, stream::handleClosed);
    }
```

So the response is ended (buffered in InboundBuffer), the close handler runs, connection back in pool, but at the same time the client has paused the stream. This is what the reproducer is doing which leads to a timeout.

Here is a screenshot that shows the bytesWindow (renamed readWindow in 4.x branch) > 0 when a connection would be released back to the pool (you can see stack trace is from handleResponseEnd above)
image

vietj:
thansk for the explanations.

the reproducer looks complicated I think it might possible to make a simpler one.

I think we might have to do something about the read window instead which is one of:

- change the back pressure system to instead use the boolean returned by InboundBuffer#write (that means stop writing to the queue), e.g. when `queue.write` returns `false` we pause the connection and we set a `drainHandler` on the queue to resume when it is called. When the response is fully received we bypass this and resume the connection if it was paused before calling the close handler
- release the connection to the pool when the last bytes of the stream has been acked instead (which might be less performant but respects more the heap usage)

vietj:
does this happen behaviour with master ? I'm thinking that master should be modified to resume before the close handler is called

vietj:
@cfolgere please see https://github.com/eclipse-vertx/vert.x/pull/5671 which implements the first option

cfolgere:
@vietj Thanks for the investigation, when I apply this fix (https://github.com/eclipse-vertx/vert.x/pull/5671) on 4.x I do see the reproducer posted does now pass. I also tried the test you added and it fails without the fix, and passes with the fix so it looks like it does reproduce the problem and is much simpler. Neither test fails on master. Conceptually I think the change makes sense (adding a per-stream readWindow and ack'ing the remaining bytes on the connection if the response ends)

vietj:
> @vietj Thanks for the investigation, when I apply this fix (#5671) on 4.x I do see the reproducer posted does now pass. I also tried the test you added and it fails without the fix, and passes with the fix so it looks like it does reproduce the problem and is much simpler. Neither test fails on master. Conceptually I think the change makes sense (adding a per-stream readWindow and ack'ing the remaining bytes on the connection if the response ends)

thanks for your time spent on this and helping us figure out what the issue is.

surabhisarnot:
Hi @vietj Thank you for merging the PR in 4.5.18. We are currently on version 4.2.6. We are looking for two fixes:

1) https://github.com/eclipse-vertx/vert.x/issues/5487
2) https://github.com/eclipse-vertx/vert.x/pull/5671/commits/c4343ad87f43cc26496db638e3241d6c752f66bd

Will it be possible to get these two fixes in 4.2.6, it would be easier for us to upgrade.

vietj:
@surabhisarnot you should move to 4.5, we are only supporting the latest minor release

Issue #5665, opened 2025-08-03
Reading files create useless copies


franz1981:
### Version

4.x and 5.x

### Context

https://github.com/eclipse-vertx/vert.x/blob/8ce1eb1d982953e18fbd4a249b6d4142b603488f/vertx-core/src/main/java/io/vertx/core/file/impl/FileSystemImpl.java#L879 shows that we perform a double copy of the byte[] since Buffer perform another copy of it again

doxlik:
Hello dear @franz1981 and Vert.x community, I really love this project and want to help it as much as possible. I don't have super deep expertise in all Java / Netty file optimizations (while I actively dive into it) so I just ask ChatGPT for possible variants to solve this copy issue and one of its suggestions seems for me quite good. I know anyone of us can just go and ask ChatGPT but I reviewed its code and for me it seems as quite strong choice and I am interested in your opinion.

```
// in io.vertx.core.file.impl.FileSystemImpl

import io.netty.buffer.ByteBuf;
import io.netty.buffer.Unpooled;
import io.vertx.core.buffer.Buffer;
import io.vertx.core.buffer.impl.BufferImpl;

import java.io.IOException;
import java.nio.MappedByteBuffer;
import java.nio.channels.FileChannel;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.StandardOpenOption;

private BlockingAction<Buffer> readFileInternal(String path) {
  Objects.requireNonNull(path);
  return new BlockingAction<Buffer>() {
    @Override
    public Buffer perform() {
      final Path target;
      try {
        target = resolveFile(path).toPath();
      } catch (Throwable t) {
        throw new FileSystemException(getFileAccessErrorMessage("read", path), t);
      }

      try {
        final long size = Files.size(target);

        // If the file fits into a single ByteBuffer (<= 2 GiB), prefer mmap to avoid heap allocation + copy.
        if (size <= Integer.MAX_VALUE) {
          try (FileChannel ch = FileChannel.open(target, StandardOpenOption.READ)) {
            // Map the whole file read-only. This yields a direct MappedByteBuffer; no heap array allocation.
            MappedByteBuffer mbb = ch.map(FileChannel.MapMode.READ_ONLY, 0, size);
            // Wrap without copying into a Netty ByteBuf, then into a Vert.x Buffer.
            ByteBuf bb = Unpooled.wrappedBuffer(mbb);    // no copy
            return new BufferImpl(bb);                   // no copy
[...]

PR #5677, opened 2025-08-13
Avoiding extra byte copy inside FileSystemImpl.readFileInternal

doxlik:
Motivation:

In this short PR I did optimization that previously was discussed with @franz1981 in this issue: #5665 regarding avoiding of extra bytes copy in FileSystemImpl.readFileInternal method.

Additional notes:

1. I could not either "mvn package" or "mvn test" as there multiple Http2 related files inside master is not compiling now.
2. It is my totally first contribution, sorry if I missed something.

Conformance:

Everything is signed, my eclipse account is identical to github one.


vertx-core/src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
```diff
@@ -876,7 +853,8 @@ public Buffer perform() {
         try {
           Path target = resolveFile(path).toPath();
           byte[] bytes = Files.readAllBytes(target);
-          return Buffer.buffer(bytes);
+          ByteBuf bb = Unpooled.wrappedBuffer(bytes);
```
franz1981, line 856: @vietj : this is ok? Or you prefer to add a new constructor to `BufferImpl' which still uses the existing Vertx heap types eg VertxUnsafeHeapByteBuf/VertxHeapByteBuf ?
The reason I'm asking is:
- `Unpooled::wrappedBuffer` still have Netty reference count checks under the hood, implemented - which have a cost
- IIRC Vertx itself contains checks to make sure the allocator of a vertx buffer is "known" to vertx e.g. https://github.com/eclipse-vertx/vert.x/blob/ff78924cff931f275238bac3dfc64bacb27991d1/vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java#L48-L67

One pain point is that AFAIK Netty's heap buffers (one of the two variant) doesn't expose in public to directly set the wrapped buffer - and that's why Netty itself is using only non-Unsafe variants in `Unpooled::wrappedBuffer(byte[])`: if we want to do this right in Vertx we could leverage a single implementation (i.e.non-unsafe) as well or modify Netty to expose it for Unsafe variants (I can do it if needed)

doxlik:
@franz1981 thank you for your comments, I removed unrelated changes. Actually they were made by Intelij IDEA code autoformatting and I thought that maybe they are okay, but I agree that they create unnecessary noise, so next time I will try to avoid any of them.

franz1981:
Fyi @vietj

franz1981:
@tsegismont 

It looks like 

https://github.com/netty/netty/blob/f922892f44743411ce89e8259d705ed9c9d3763e/buffer/src/main/java/io/netty/buffer/UnpooledHeapByteBuf.java#L292

that's going to allocate the `internal` NIO buffer which usually is not used in the Vertx heap ones (the internal NIO buffer is something used by Netty while sending data over the wire while communicating w NIO JDK layers...).
But It's still a good suggestion and still an improvement vs what we got - while retaining the same behaviour as we have now

@doxlik wdyt?

doxlik:
@franz1981 Regarding the suggestion, what if file really big and bigger than 2 gb? Then here "            res.getByteBuf().writeBytes(fc, 0, (int) length); // check actual size before the cast" even if we check the size, we still need to fallback to something that can handle it, either wrapping or have some "while" loop until fully read the file to the buffer.

Regarding generally low level ByteBuffs stuff, it is super interesting but I have almost no solid knowledge of all its internals, so in terms of buffers / allocators optimization I fully trust to you.

tsegismont:
> @franz1981 Regarding the suggestion, what if file really big and bigger than 2 gb? Then here " res.getByteBuf().writeBytes(fc, 0, (int) length); // check actual size before the cast" even if we check the size, we still need to fallback to something that can handle it, either wrapping or have some "while" loop until fully read the file to the buffer.

I think it's safe to stick to the behavior of `java.nio.file.Files#readAllBytes`

If the size is bigger than `java.nio.file.Files#MAX_BUFFER_SIZE`, an `OutOfMemoryError` error is thrown

doxlik:
@tsegismont then if we want stick to behavior of java.nio.file.Files#readAllBytes, then should we move to the channel approach as you suggested or we can simply leave the current code that already use readAllBytes?

doxlik:
@franz1981 @vietj I also wanted to ask, what will be further process of the review / PR? I mean is it kinda in direct queue of pending PR's for Julien's review / approval or we are waiting for Vert.x 5.1.0 to proceed with this PR?

tsegismont:
> @tsegismont then if we want stick to behavior of java.nio.file.Files#readAllBytes, then should we move to the channel approach as you suggested or we can simply leave the current code that already use readAllBytes?

We should move to the channel approach but, for the error cases, we can reproduce the behavior of `java.nio.file.Files#readAllBytes`

doxlik:
@tsegismont I published solution where tried to go FileChannel option but within error logic of readAllBytes. Check when you have time, please.

tsegismont:
Thank you for the updates @doxlik 

I've created https://github.com/eclipse-vertx/vert.x/pull/5730 , that supersedes this one.

PR #5683, opened 2025-08-25
Ack the HTTP client response window from the correct event-loop thread.

vietj:
Motivation:

The recent change of the HTTP client response flow control to avoid releasing a connection to the pool not fully received introduced a bug creating a data race when acking the response window bytes. This happens when the stream event-loop is different from the connection event-loop.

Changes:

Ack the response window from the connection thread instead of the stream thread.


ppalaga:
@vietj I think I am facing this issue in https://github.com/quarkiverse/quarkus-cxf/issues/1893 
Is there any plan to release 4.5.19 any time soon?

vietj:
@ppalaga release is expected tomorrow

Issue #5676, opened 2025-08-13
Enable DNS Resolving via TCP in case of Timeout


cgm-aw:
### Describe the feature

Hi everyone,

I noticed that Vertx enables Netty DNS Resolving via TCP (see https://github.com/eclipse-vertx/vert.x/blob/4.5.17/src/main/java/io/vertx/core/impl/resolver/DnsResolverProvider.java#L108):
`builder.socketChannelFactory(() -> (SocketChannel) vertx.transport().channelFactory(false).newChannel());`

However, this sets "retryOnTimeout" to "false". It would be nice if Vertx would enable TCP resolving on timeouts - maybe just with some kind of environment variable for toggling, like this:

`builder.socketChannelFactory(() -> (SocketChannel) vertx.transport().channelFactory(false).newChannel(), Boolean.getBoolean("vertx.dns.retryOnTimeout"));`

I'm happy to provide a PR if you approve of this.

Best regards

cgm-aw:
By the way, I think https://github.com/eclipse-vertx/vert.x/issues/5114 requests the same thing.

vietj:
@cgm-aw this should be configured using vertx options or address resolver options, contribution is welcome

cgm-aw:
Done: https://github.com/eclipse-vertx/vert.x/pull/5692
I just noticed this is flagged for 5.x, I based the fix on 4.x. Not sure what your policy on backports is - should I base the fix on 5.x?

vietj:
we can backport very small features

vietj:
@cgm-aw can you port this contribution to master ?

cgm-aw:
Sure, here is the PR for master:
https://github.com/eclipse-vertx/vert.x/pull/6116

codingkiddo:
Hi @cgm-aw / maintainers, I’d like to work on this if it’s still available.

I checked current master and the equivalent call now appears in:

`vertx-core/src/main/java/io/vertx/core/dns/impl/DnsAddressResolverProvider.java`

My proposed approach:

- add a Vert.x system property, for example `vertx.dns.retryOnTimeout`
- wire it into Netty’s `DnsNameResolverBuilder#socketChannelFactory(ChannelFactory, boolean)`
- keep the default behavior unchanged by using `false`
- add focused test coverage where feasible, making sure the system property is restored after the test

Does this API shape sound acceptable, or would you prefer this as an `AddressResolverOptions` property instead of a system property?

[...]

PR #5692, opened 2025-09-01
Issue 5676: Introduced the option to fall back to TCP when DNS resolv…

cgm-aw:
See https://github.com/eclipse-vertx/vert.x/issues/5676

### DIFF
--- a/src/main/generated/io/vertx/core/dns/AddressResolverOptionsConverter.java
+++ b/src/main/generated/io/vertx/core/dns/AddressResolverOptionsConverter.java
@@ -105,6 +105,11 @@ public class AddressResolverOptionsConverter {
             obj.setServers(list);
           }
           break;
+        case "useTcpForFallbackDnsResolving":
+          if (member.getValue() instanceof Boolean) {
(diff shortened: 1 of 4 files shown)


vietj:
I think we are missing a test for this

cgm-aw:
I thought about testing this, but all Vertx does here is pass the parameter to the Netty resolver. So I just tested this new functionality. Writing a test which checks whether the fallback actually works is out of scope here from my point of view, and I think it would require a lot of mocking and/or dependencies to Netty internals, which would make the test hard to maintain.
Let me know if you still want a test to test the actual fallback in Netty and I will give it a shot.

vietj:
@cgm-aw can you sign the Eclipse Contributor Agreement

cgm-aw:
Sure, you can find me under "cgmaw"

vietj:
vertx has a mock dns resolver we can use

cgm-aw:
New PR for merge into master: https://github.com/eclipse-vertx/vert.x/pull/6116

PR #5712, opened 2025-09-11
Replace with logger.error.

kunaljani1100:
Motivation:

Currently, line 290 in the `private URI replaceFragment(URI oldURI, String fragment)` method of the `JsonPointerImpl` class has the functionality to replace an old fragment in the JSON URL with a new fragment in the URL. In case this operation fails, the `e.printStackTrace()` method that can cause various problems such as the inability to distinguish log levels and being hard to filter logs in the production environment. Therefore `e.printStackTrace()` is being replaced with the  `logger.error()` method.

Conformance:

I have signed the eclipse contributor agreement.

I am adhering to the code style guidelines.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/json/pointer/impl/JsonPointerImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/json/pointer/impl/JsonPointerImpl.java
@@ -282,14 +282,13 @@ public class JsonPointerImpl implements JsonPointer {
     return replaceFragment(oldURI, null);
   }
 
-  private URI replaceFragment(URI oldURI, String fragment) {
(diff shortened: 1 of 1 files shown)


vietj:
Actually I am not sure it is correct to swallow the exception  and instead we should rethrow the exception wrapped with a VertxException, can you take care of this and add a test for it ?

kunaljani1100:
Adding a test for this method will not work, since generating in the `JsonPointerImpl` class, the URI check is performed during the object creation itself, and therefore it is not possible for us to reach the condition where a `URISyntaxException` is caught and the statements after line 294 is executed. 

The reason why `URISyntaxException` is being caught is because the constructor for the URI class requires us to catch the exception.

PR #5715, opened 2025-09-15
Ensure HttpServerResponse#endHandler and HttpServerResponse#bodyEndHandler are always executed before VertxTracer#sendResponse

vietj:
Motivation:

`HttpServerResponse#endHandler` and `HttpServerResponse#bodyEndHandler` are after `VertxTracer#sendResponse` for HTTP/1.x and before for HTTP/2.

We should have a consistent behaviour on ordering in order to guarantee predictability for users of this API and ensure that those handlers are invoked before the span end is signalled.


brunobat:
Tested it on Quarkus app and it seems to fix this problem: https://github.com/quarkusio/quarkus/issues/43570

PR #5730, opened 2025-09-23
Avoid extra byte copy inside FileSystemImpl.readFileInternal

tsegismont:
See #5665

The improvement consists in creating the buffer upfront and using Netty API to read the file channel.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2026 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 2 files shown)


tsegismont:
@franz1981 please take a look

tsegismont:
@vietj PTAL

I added a  test with JMH. Here are the results on my machine.

## Latency

| File size | `fileChannel` (ns/op) | `readAllBytes` (ns/op) | Speedup |
|-----------|-----------------------|------------------------|---------|
| 128 B     | 1,384 +/- 25          | 1,554 +/- 23           | ~11%    |
| 4 KB      | 1,887 +/- 20          | 2,495 +/- 25           | ~24%    |
| 64 KB     | 10,911 +/- 136        | 18,657 +/- 455         | ~42%    |
| 1 MB      | 220,739 +/- 8,878     | 420,946 +/- 38,985     | ~48%    |

The advantage grows with file size. At 1 MB, `fileChannel` is nearly 2x faster.

## Allocations per operation

| File size | `fileChannel` (B/op) | `readAllBytes` (B/op) | Saved |
|-----------|----------------------|-----------------------|-------|
| 128 B     | 784                  | 837                   | ~6%   |
| 4 KB      | 4,752                | 8,776                 | ~46%  |
| 64 KB     | 66,192               | 131,656               | ~50%  |
| 1 MB      | 1,049,238            | 2,097,728             | ~50%  |

From 4 KB up, `readAllBytes` allocates almost exactly double: the intermediate `byte[]` plus the ByteBuf copy. `fileChannel` eliminates the `byte[]` entirely.

At 128 B the gap is smaller, probably because of the fixed overhead (FileChannel/Path objects, ByteBuf header) dominates over the data copy.

vertx-core/src/main/java/io/vertx/core/file/impl/FileSystemImpl.java
```diff
@@ -876,8 +852,16 @@ private BlockingAction<Buffer> readFileInternal(String path) {
       public Buffer perform() {
         try {
           Path target = resolveFile(path).toPath();
-          byte[] bytes = Files.readAllBytes(target);
-          return Buffer.buffer(bytes);
+          try (FileChannel fc = FileChannel.open(target, StandardOpenOption.READ)) {
+            long size = fc.size();
+            if (size > (long) Integer.MAX_VALUE) {
+              throw new OutOfMemoryError("File is too big");
```
vietj, line 859: add a comment justifying that

tsegismont: Done

PR #5736, opened 2025-10-01
Add of() overloading with variable map entries.

kunaljani1100:
Motivation:

Currently the static method `JsonObject.of()` method has 10 different definitions to accept key value pairs as parameters to create a new JSON object. The new implementation of this method intends to ensure that we do not need 10 different definitions based on the number of parameters to create a JSON object. Instead, the new method implements an array of map entries (tuples) that consists of a key and a value corresponding to the key. It is possible to have multiple such map entries, so instead of defining the `of()` method 10 times, this method should be declared only once with the ability to process a variable number of arguments.

Conformance:

I am adhering to the code style guidelines.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/json/JsonObject.java
+++ b/vertx-core/src/main/java/io/vertx/core/json/JsonObject.java
@@ -396,6 +396,19 @@ public class JsonObject implements Iterable<Map.Entry<String, Object>>, ClusterS
   }
 
   /**
+   * Create a JsonObject from the given map entries.
+   * @param entries All the map entries to add the JSON object.
+   * @return a JsonObject containing the specified mappings.
+   */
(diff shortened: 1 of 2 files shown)


vietj:
what is the purpose of this ?

kunaljani1100:
We do not want to have 10 separate declarations of methods when it is possible to declare a single method that can accept a variable number of arguments.

vietj:
I mean what is the purpose of using map.entry while vertx already have the same method with key/value ?

kunaljani1100:
When we want to create a JSON object with a large number of entries, such as 12 or more entries, the functions that have been already defined would not be able to accommodate these changes, so we will need to use a map.entry object when there are a large number of key value pairs that we want to add to our JSON object.

vietj:
it seems very specific to your case and does not add much, I think you can make your own static method for this

PR #5737, opened 2025-10-02
Clarify how to terminate an http response in the javadoc

FroMage:
It was not clear to us how to signal an error during response processing in Vert.x. One that occurs _after_ we've sent the headers and some chunks of data. Turns out we have to use `reset()` but we did not know about this.

Hopefully this is now clearer.

Note that I am not sure if `sendFile` is also a terminal operation, or if you should call `end()` after it?

See https://github.com/quarkusio/quarkus/pull/50361#discussion_r2397871691

CC @geoand @tsegismont


FroMage:
CI failure doesn't seem related :-/

tsegismont:
> Note that I am not sure if `sendFile` is also a terminal operation, or if you should call `end()` after it?

Yes it is, no need to call `end()` after it.

> CI failure doesn't seem related :-/

No worries

tsegismont (commented):
Thank you @FroMage ! Would you mind adding the comments for `sendFile`?

FroMage:
Sure.

vietj:
@FroMage can you rebase on latest master, I just fixed a couple of tests failing on Java 25 CI

FroMage:
Done.

Issue #5691, opened 2025-09-01
isDIGIT http1.xserver ArrayOtOfboundException


jr981008:
### Version

4.5.13

### Context
2025-08-31 21:52:34.578 [ERROR] [Edge.java-57] [vert.x-eventloop-thread-14]-vertx unexpected fail.
java.lang.ArrayIndexOutOfBoundsException: no message
at io.vertx.core.net.impl.HostAndPortImpl.isDIGIT(HostAndPortImpl.java:129)
at io.vertx.core.net.impl.HostAndPortImpl.isUnreserved(HostAndPortImpl.java:120)
at io.vertx.core.net.impl.HostAndPortImpl.parseRegName(HostAndPortImpl.java:108)
at io.vertx.core.net.impl.HostAndPortImpl.parseHost(HostAndPortImpl.java:25)
at io.vertx.core.net.impl.HostAndPortImpl.isValidAuthority(HostAndPortImpl.java:151)
at io.vertx.core.http.impl.Http1xServerRequest.isValidAuthority(Http1xServerRequest.java:251)
at io.vertx.core.http.impl.HttpServerRequestWrapper.isValidAuthority(HttpServerRequestWrapper.java:123)
at io.vertx.ext.web.impl.RoutingContextImpl.route(RoutingContextImpl.java:81)
at io.vertx.ext.web.impl.RouterImpl.handle(RouterImpl.java:69)
at io.vertx.ext.web.impl.RouterImpl.handle(RouterImpl.java:37)
at io.vertx.core.http.impl.Http1xServerRequestHandler.handle(Http1xServerRequestHandler.java:150)
at io.vertx.core.http.impl.Http1xServerRequestHandler.handle(Http1xServerRequestHandler.java:42)
at io.vertx.core.impl.ContextImpl.emit(ContextImpl.java:342)
at io.vertx.core.impl.DuplicatedContext.emit(DuplicatedContext.java:163)
at io.vertx.core.http.impl.Http1xServerConnection.handleMessage(Http1xServerConnection.java:174)
at io.vertx.core.net.impl.ConnectionBase.read(ConnectionBase.java:159)
at io.vertx.core.net.impl.VertxHandler.channelRead(VertxHandler.java:153)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:442)
[...]

PR #5752, opened 2025-10-15
HostAndPortImpl#isDIGIT throws ArrayOtOfboundException

tsegismont:
See #5691

Happens when non-ascii characters are used in domain names


tsegismont:
cc @franz1981

PR #5754, opened 2025-10-16
Reorg some of the classes of the http.impl package: group HTTP/1.x related as http.impl.http1x and WebSocket as http.impl.ws

vietj:


vietj:
will be done later after HTTP/3 has been implemented

PR #5756, opened 2025-10-20
HTTP/3 API and implementation

vietj:
Implementation of HTTP API with HTTP/3.

This implementation makes use of Vert.x Quic stream implementation.

## Todo

- [ ] implementation
  - [x] outer context integration and testing
  - [ ] metrics
  - [ ] tracing
  - [ ] send file
  - [ ] push
    - [ ] client
    - [ ] server
  - [ ] network activity logging
- [ ] API
  - [ ] breaking changes
    - [ ] identify
    - [ ] decide
  - [ ] dev
    - [ ] options
    - [ ] connection close payload
  - [ ] request cancellation superseding request reset 
- [ ] Improve Netty Quic reset
- [ ] go away strategies ?
- [ ] documentation
- [ ] ...

## Breaking changes

Implementing HTTP/3 implies a few potential breaking changes, the strategy to handle them will be on a case by case basis and attempt to minimize the amount of breaking changes to delay them until Vert.x 6 is implemented, here is a non exhaustive list:

- stream id declares long type instead of int
- settings
- ...

### Shutdown / GoAway

- H2: `GoAway` carries stream id + error code + debug data
- H3: `GoAway` only carries stream id, but quic can close with an error code + reason data, error code seems similar but reason is not the same semantic than debug.

### Reset semantic

- H2: reset means immediate termination of a stream
- H3: reset only applies to the stream outbound and means no more data will be set

Introduce stream cancellation that performs the most adapted operation according to the protocol:

- H2 : reset with `0x8` (CANCEL) code
- H3 : reset or abort reading with REQUEST_CANCELLED code

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/Vertx.java
+++ b/vertx-core/src/main/java/io/vertx/core/Vertx.java
@@ -230,6 +230,24 @@ public interface Vertx extends Measured {
   HttpServer createHttpServer(HttpServerOptions options);
 
   /**
+   * Create an HTTP3 client using the specified options
+   *
+   * @param options  the options to use
+   * @return the server
+   */
+  default HttpClientAgent createHttpClient(Http3ClientOptions options) {
+    return httpClientBuilder().with(options).build();
+  }
+
+  /**
+   * Create an HTTP3 server using the specified options
+   *
+   * @param options  the options to use
+   * @return the server
+   */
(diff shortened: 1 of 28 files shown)


vietj:
See https://github.com/eclipse-vertx/vert.x/pull/5802

PR #5761, opened 2025-10-25
Handle stale outbound connections after connection failures in clustered event bus

chipnesh:
Motivation:

Fixes a bug where a failed attempt to establish a clustered event bus connection leaves a stale OutboundConnection instance in the outboundConnections map. 
This prevents subsequent reconnect attempts and can cause writes to hang/fail repeatedly. The fix ensures the failed connection is cleaned up consistently, enabling a fresh connection on the next send. 
A regression test is added.

This problem was introduced during refactoring in commit ffb5d7fb. 
That cleanup path was dropped, causing a regression where stale entries could persist and block reconnection.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
+++ b/vertx-core/src/main/java/io/vertx/core/eventbus/impl/clustered/ClusteredEventBus.java
@@ -348,11 +348,11 @@ public final class ClusteredEventBus extends EventBusImpl {
   }
 
   private void sendRemote(String remoteNodeId, MessageImpl<?, ?> message, Promise<Void> writePromise) {
-    OutboundConnection outboundConnection = getOutboundConnection(remoteNodeId);
+    OutboundConnection outboundConnection = getOutboundConnection(remoteNodeId, writePromise);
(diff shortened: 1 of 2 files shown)


chipnesh:
@vietj please take a look

vietj:
@chipnesh good catch and thanks for your contribution, I can spot a couple of issues here

1. indeed on a lookup failure, the connection slot should be cleaned up along with the connection state cleanup (handleClose), the PR removes the slot cleanup when the connection closes which will prevent connection failures to clean the map.

2. the current message write triggering the lookup tries to piggy back on the connection failure. I think instead we should let it continue and be added to the connection. However the issue we have is that the connection once closed will continue pending message writes to happen. Instead we should record the closed state and when a pending write is added to the connection, we should fail the promise.

Here is an attempt to implement this https://github.com/eclipse-vertx/vert.x/pull/5763 , can you have a look ?

chipnesh:
good point. looks good to me.

chipnesh:
Can we get your pr merged asap so we can fix this?
Its blocking me from using the new version of vertx in production.

vietj:
https://github.com/eclipse-vertx/vert.x/pull/5767

PR #5782, opened 2025-11-13
Streams doc improvements

tsegismont:
Currently, the doc uses the same `NetServer` socket as a `ReadStream` and as `WriteStream` for examples.

This can cause some confusion: snippets effectively show the `WriteStream` being paused when the write queue is full...

To avoid that sort of confusion, we can use different types for streams: `NetSocket` for `ReadStream`, `AsyncFile` for `WriteStream`.

In fact, the documentation talks about files already, which makes me think we might have used these types originally and then changed for some reason.


tsegismont:
cc @AdrianVasiliu

PR #5800, opened 2025-11-18
Decouple HttpServerMetrics and HttpClientMetrics from TCPMetrics.

vietj:
Motivation:

HttpServerMetrics and HttpClientMetrics should extend the new TransportMetrics instead of TCPMetrics.


tsegismont:
This is necessary to prepare for HTTP/3 so :+1: 

Nevertheless, I think the PR should have a milestone set so that it appears in the release notes because it's an important change. Also, the breaking changes page must be updated (perhaps it is already, I haven't checked).

Also, it seems some changes will be required in Micrometer and Dropwizard metrics implementation of the SPI

tsegismont:
> Also, it seems some changes will be required in Micrometer and Dropwizard metrics implementation of the SPI

At least we should add a new tag to the meters that indicates the `type`

tsegismont:
I filed https://github.com/vert-x3/issues/issues/663

PR #5823, opened 2025-11-27
NetSocket.exceptionHandler reports UnsupportedOperationException on close

tsegismont:
Follows-up on https://github.com/eclipse-vertx/vert.x/pull/5783

If the user closes the connection and then the client is closed, UnsupportedOperationException is reported to the socket exception handler.

Instead of throwing UOE, the redundant shutdown event coming is ignored.

Also, make sure to complete channel promises when the channel is closed.
Otherwise, the connection close future won't be completed.


tsegismont:
@vietj this is ready for review. Updated it as per discussion offline (ignore the shutdown event coming from client close if the connection is being shutdown already).

PR #5824, opened 2025-11-28
Clarify raw vs normalized path usage in Vert.x Core documentation

mathias82:
### Motivation

Developers often assume that `HttpServerRequest#path()` returns a normalized
version of the request path. However, this value reflects the raw path sent
by the client and may contain repeated separators or path traversal markers
(e.g. `//` or `..`). This misunderstanding can lead to incorrect or unsafe
path-based logic, especially when working with Vert.x Web routing.

### What this PR does

This PR updates the Vert.x Core HTTP documentation to:

- state clearly that `HttpServerRequest#path()` returns the raw path
- highlight the presence of non-normalized values such as `//` or `..`
- provide guidance that security-sensitive checks and route comparisons
  should use `RoutingContext#normalizedPath()` when running under Vert.x Web
- ensure consistency with the updated Vert.x Web documentation

### Why this matters

Vert.x Web performs canonicalization before matching routes. Using the raw
path for authorization, auditing, or routing decisions may produce unexpected
results or introduce subtle bugs.

By documenting this distinction in Vert.x Core, this PR aligns expectations
and prevents misuse of `HttpServerRequest#path()` in scenarios where a
normalized path is required.

### Related work

- Vert.x Web documentation update: https://github.com/vert-x3/vertx-web/pull/2832
- Issue discussion: https://github.com/vert-x3/vertx-web/issues/2831

---

This PR contains documentation-only changes.

### DIFF
--- a/vertx-core/src/test/java/io/vertx/tests/timer/TimerTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/timer/TimerTest.java
@@ -17,6 +17,7 @@ import io.vertx.core.impl.VertxImpl;
 import io.vertx.core.internal.VertxInternal;
 import io.vertx.test.core.Repeat;
 import io.vertx.test.core.VertxTestBase;
+import org.hamcrest.Matchers;
 import org.junit.Test;
 
 import java.util.concurrent.CancellationException;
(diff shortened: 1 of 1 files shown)


sberyozkin:
CC @vietj @cescoffier

vietj (commented):
I think we should also add such information in the javadoc itself of HttpServerRequest

mathias82:
> I think we should also add such information in the javadoc itself of HttpServerRequest

@vietj Added a Javadoc note to path() explaining that it returns the raw request path (not normalized) and should not be used directly for security-sensitive logic. No functional changes.

PR #5829, opened 2025-12-03
Add support for hybrid key exchange protocol x25519mlkem768.

anavarr:
Motivation:
The rise of quantum computers threatens traditional asymmetric key exchange protocol due to their ability to break private keys.
Post-quantum cryptography must use problems that quantum computers can't solve as quickly. Module-lattice-based problems have been found to resist quantum computers. ML-KEM, for module-lattice-based key encapsulation mechanism, is an instance of a key exchange protocol resistant to quantum computers. Due to its relative short existence, it has been recommended to use it alongside traditional Diffie-Hellman with elliptic curve.
Thus, the new hybrid key exchange protocol x25519mlkem768 uses both Diffie-Hellman with elliptic curve 25519 and ML-KEM. Its has been integrated in OpenSsl starting with version 3.5.

Changes:
This features relies on the netty-tcnative-openssl-dynamic bound to version 3.6 of openssl at runtime, and netty-tcnative-classes at build-time.

- Add boolean useHybrid field to io.vertx.core.net.SSLOptions and create getters/setters in io.vertx.core.net.SSLOptions as well as every implementation of io.vertx.core.net.TCPSSLOptions.
- If this value is set to true (default false), the ssl handler will be set to use x25519mlkem768 instead of x25519.
- If key exchange protocol is set to x25519mlkem768 but JdkSsl is used instead of OpenSsl, or the version of openssl used at runtime does not suppot x25519mlkem768, the user is informed that hybrid key exchange is impossible, then channel is closed.

### DIFF
--- a/.github/workflows/ci-4.x.yml
+++ b/.github/workflows/ci-4.x.yml
@@ -20,6 +20,7 @@ jobs:
         jdk: [8]
         include:
           - os: ubuntu-latest
+            container: 'ubuntu:26.04'
             jdk: 25
           - os: ubuntu-latest
             jdk: 8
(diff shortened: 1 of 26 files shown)


anavarr:
added tests and checked the latest build of netty-tcnative

anavarr:
I removed the dependency on `netty-tcnative-boringssl-static` because it doesn't behave properly in tests alongside smallrye-openssl. Smallrye-openssl provides openssl 3.x but I think we can make them coexist by excluding boringssl from hybrid key exchange tests in maven config if you want to keep `netty-tcnative-boringssl-static`

PR #5836, opened 2025-12-05
Add defaultServer to HttpClientOptions

### LINKED ISSUE #5102: Add setDefaultServer to HttpClientOptions
In #4874 , HTTP Client `RequestOptions` has been improved so that users can set a server `Address`  that works well with all service resolvers.

In HTTP Client options, it is possible to set a default host and port, but not a default server.

For usability, it would be nice to be able to set a default server too.

tsegismont:
Closes #5102

`HttpClientOptions` allows to set a default port and a default host.

It can also be convenient to set a default address (in particular, when using a domain socket address or a service address from the Vert.x service resolver).


vietj:
I am not in favor of this change because an `Address` cannot translate to JSON back and forth as it is.

In practice most users will use a default port and a default host, so it does not add much usability in practice and we should not do it only because we can do it.

tsegismont:
Fair enough

PR #5850, opened 2025-12-22
Vert.x CodeGen will always use LF line endings

### LINKED ISSUE #5849: Generated Java files are generated on Windows with LF, showing up in IDEs as modified
### Describe the feature

When developing on Windows with its CRLF line endings, I noticed the following: 

* Git checks out the previously generated files with CRLF
* The next generator run re-creates them with LF
* Now the IDE shows a lot of changes

I'm suggesting to add a `.gitattributes` file to ensure that Git checks out the files with LF on Windows, and I will provide a PR

### Contribution

@ahus1

ahus1:
Closes #5849

Motivation:

This aligns the Git settings with the behavior of CodeGen.

An alternative approach would be to teach CodeGen to use the platform specific line endings. Let me know if that is preferred instead, or any other solution.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Those files are auto-generated, and will always have LF line endings when re-generated locally
+**/generated/**/*.java text=auto eol=lf


vietj:
as I understand, that will have an effect on repository commits since in our projects we do checkin the generated code ?

ahus1:
Actually it will have an effect on all git checkouts. AFAIK Git already has a handling to normalize CRLR / LR on all text files before committing them. And when checking them out, it will apply the platform specific line handling to all text files. 

This now ensures that whenever a generated file is checkout out from Git, it will always have LF line endings the way the Vert.x code generator uses. This way an IDE will no longer show the files as changed. 

Alternatives: (1) do not check in generated code, or (2) teach the generator to generate files with the line endings of the OS it runs on. 

See for a longer and official explanation: https://git-scm.com/docs/gitattributes#_eol

ahus1:
As discussed in the Vert.x maintainer hangout, this should rather be tackled in the generator.

PR #5858, opened 2025-12-25
Ensure setup is only called once in VirtualThreadEventBusTest, VirtualThreadContextTest and NetBandwidthLimitingTest

### LINKED ISSUE #5857: Avoid calling setup multiple times in several test classes
### Version

main

### Context

Follow-up to #5851

Addressing a similar problem in VirtualThreadEventBusTest, VirtualThreadContextTest and NetBandwidthLimitingTest

### Steps to reproduce

Look at the CI logs, the start test log lines are duplicated

### Do you have a reproducer?

_No response_

ahus1:
Closes #5857

Motivation:

Ensure setup is only called once. Otherwise there are multiple methods annotated with `@Before`.


ahus1:
@vietj - This applies the same pattern to avoid calling `setUp()` twice as in #5852. Please review when you have the time. Thanks!

PR #5866, opened 2025-12-27
Fix initialization and re-computation of expirationTimestamp in Http2ClientConnectionImpl

### LINKED ISSUE #5865: Unstable test Http2Test.testKeepAliveTimeout
### Version

main

### Context

In https://github.com/eclipse-vertx/vert.x/actions/runs/20499367576/job/58904378867 the test failed, and it is an unstable test. 

When analyzing the test, I found the following: 

* `Http2ClientConnectionImpl.onStreamClosed()` calls `super.onStreamClosed(s)` which will return the connection to the pool
* When doing that, the `expirationTimestamp` is not set yet, and due to that the `isValid()` will return false and evict the `SharedHttpClientConnectionGroup.checkExpired()` will have a short time window to close the connection before the `Http2ClientConnectionImpl.onStreamClosed()` set the `expirationTimestamp` and make the connection valid again.

### Steps to reproduce

I was able to reproduce it locally with several breakpoints and some delays in the test. 

I'll prepare a pull request.

### Do you have a reproducer?

_No response_

ahus1:
Closes #5865

Motivation:

Setting `expirationTimestamp` is required before the connection is recycled to ensure that the background timer executed correctly.

In a [previous attempt](https://github.com/eclipse-vertx/vert.x/compare/e0df732d1d402e7986dafb9b8cca6299bd7069d2..1eb0c76d21d1fdbed66cee40fb254030002f1433) I tried to set the initial value to `Long.MAX_VALUE`, but then `Http2Test.testClientKeepAliveTimeoutNoStreams` started failing. So I'm no setting it a value using `recycle()`.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/http2/codec/Http2ClientConnectionImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/http2/codec/Http2ClientConnectionImpl.java
@@ -62,6 +62,8 @@ public class Http2ClientConnectionImpl extends Http2ConnectionImpl implements Ht
     this.authority = authority;
     this.lifetimeEvictionTimestamp = maxLifetime > 0 ? System.currentTimeMillis() + maxLifetime : Long.MAX_VALUE;
     this.handler = connHandler;
+    // Ensure that the expirationTimestamp is initialized
+    recycle();
   }
 
   @Override
@@ -199,8 +201,9 @@ public class Http2ClientConnectionImpl extends Http2ConnectionImpl implements Ht
 
   @Override
   void onStreamClosed(Http2Stream s) {
-    super.onStreamClosed(s);
+    // Prepare this connection for returning as onStreamClosed() will return it to the pool.
     recycle();
+    super.onStreamClosed(s);
   }
 
   @Override


vietj:
@ahus1 is that possible to have a test for this ?

ahus1:
@vietj - this is fixing an unstable test that I referenced in the issue #5865. As it is about a timing issue, I wouldn't know how to implement a test case for this. 

Can you please elaborate on what kind if test case you are looking for?

jnbdz:
I gave the missing test a try. The `Http2Test` case below reproduces the first half of the problem deterministically: with a pooled connection that never carries a stream (server `maxConcurrentStreams=0`, same setup as `testClientKeepAliveTimeoutNoStreams`), `setHttp2KeepAliveTimeout(3)` and `PoolOptions#setCleanerPeriod(1)`, the connection must be closed after ~3s. On current master it fails with

```
Expected connection to be closed after the keep-alive timeout, was closed after 1ms
```

because `expirationTimestamp` is still `0` when the cleaner first runs; with the `recycle()` call in the constructor from this PR it passes, and the full `Http2Test` is green with the change.

```java
@Test
public void testClientKeepAliveTimeoutNoStreamsNotExpiredEarly(Checkpoint checkpoint) throws Exception {
  // A pooled connection that never carried a stream must be considered valid until the keep-alive timeout
  // elapses: the pool cleaner must not evict it right after it is created
  server = vertx.createHttpServer(Http2TestBase.createHttp2ServerOptions().setInitialSettings(new Http2Settings().setMaxConcurrentStreams(0)));
    req.response().end();
  });
  startServer();
  int keepAliveTimeoutSeconds = 3;
  AtomicBoolean closed = new AtomicBoolean();
  client = vertx
    .httpClientBuilder()
    .withConnectHandler(conn -> {
      long now = System.currentTimeMillis();
      conn.closeHandler(v -> {
        if (closed.compareAndSet(false, true)) {
          long elapsed = System.currentTimeMillis() - now;
          long low = keepAliveTimeoutSeconds * 1000 - 500;
          long high = keepAliveTimeoutSeconds * 1000 + 500;
          assertTrue("Expected connection to be closed after the keep-alive timeout, was closed after " + elapsed + "ms", low < elapsed);
          assertTrue("Expected connection to be closed after the keep-alive timeout, was closed after " + elapsed + "ms", elapsed < high);
          client.close().onComplete(v2 -> {
            checkpoint.succeed();
          });
        }
      });
    })
    .with(Http2TestBase.createHttp2ClientOptions().setHttp2KeepAliveTimeout(keepAliveTimeoutSeconds))
    .with(new PoolOptions().setCleanerPeriod(1))
    .build();
  client.request(requestOptions);
}
```

The reordering in `onStreamClosed` closes a window between two statements executed on the event loop while the cleaner runs on another thread, so I don't see a way to make that part deterministic without a hook; the test above at least pins the initialisation, which is what makes `testKeepAliveTimeout` fail in CI.

Note the diff no longer applies cleanly on current master (`lifetimeEvictionTimestamp` is gone), the change itself rebases trivially. Happy to open it as a PR on top of yours if that helps, @ahus1.

PR #5868, opened 2026-01-07
Add option for setting max query params

geoand:
Relates to: https://github.com/quarkusio/quarkus/issues/47431

### DIFF
--- a/vertx-core/src/test/java/io/vertx/tests/http/Http1xTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/http/Http1xTest.java
@@ -423,6 +423,9 @@ public class Http1xTest extends HttpTest {
     assertEquals(256, options.getDecoderInitialBufferSize());
     assertIllegalArgumentException(() -> options.setDecoderInitialBufferSize(-1));
 
+    assertEquals(HttpServerOptions.DEFAULT_MAX_QUERY_PARAMS, options.getMaxQueryParams());
+    assertEquals(options, options.setMaxQueryParams(1025));
+    assertEquals(1025, options.getMaxQueryParams());
   }
 
   @Test
@@ -787,6 +790,7 @@ public class Http1xTest extends HttpTest {
     boolean decompressionSupported = rand.nextBoolean();
     boolean acceptUnmaskedFrames = rand.nextBoolean();
     int decoderInitialBufferSize = TestUtils.randomPositiveInt();
+    int maxQueryParams = TestUtils.randomPositiveInt();
 
     options.setSendBufferSize(sendBufferSize);
     options.setReceiveBufferSize(receiverBufferSize);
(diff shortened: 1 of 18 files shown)


vertx-core/src/test/java/io/vertx/tests/http/Http1xTest.java
```diff
@@ -2619,6 +2627,48 @@ private void testServerMaxHeaderSize(int maxHeaderSize) {
     await();
   }
 
+  @Test
+  public void testMaxQueryParams() {
```
vietj: this test should be moved to the parent class `HttpTest` so it can be verified for HTTP/2 as well

vietj:
thanks @geoand I am actually wondering why we have such limit, this is probably due to the decoder we use, but in practice I think this should be unbounded because I don't see this limit preventing any kind o DOS kind of style attack

geoand:
>  I am actually wondering why we have such limit,

When develoiping this patch, I was wondering the same :)

vietj:
> > I am actually wondering why we have such limit,
> 
> When develoiping this patch, I was wondering the same :)

maybe we should simply configure the parser with the highest possible limit and spare us supporting an un-necessary configuration knob ?

geoand:
If you believe that is safe, I'm fine with that

vietj:
it is safe because the request URI we use is already limited by the `maxInitialLineLength`

keep the `params(String uri, Charset charset, int maxParams, boolean semicolonIsNormalChar)` overload for compatibility and pass it the highest possible value

geoand:
The Netty code mentions `Maximum number of query parameters allowed, to mitigate HashDOS`. Are you sure we want to remove it?

vietj:
@geoand let me have a look

vietj:
@geoand let be safe and proceed as originally intended by this PR

geoand:
FWIW, I agree

geoand:
@vietj is there anything more you want me to do about this?

vietj:
the Eclipse Contributor Agreement validation seems to fail, do you have a valid eclipse account ?

geoand:
I thought I did as I have done this in the past...

vietj:
can you double check that ? or what account you have used ?

geoand:
Yes, I will check later on today

geoand:
I have no idea why the Ecplise thing is complaining as:

Screenshot from 2026-01-26 12-37-39


Screenshot from 2026-01-26 12-38-10

vietj:
@geoand thanks for updating this to the latest changes in HTTP configuration layer

geoand:
🙏

geoand:
I rebased onto the latest `master`

vietj:
@geoand Http3ServerFileUploadTest don't seem to pass with this new feature yet

geoand:
I see. I'll have a look next week

vietj:
fixed by https://github.com/eclipse-vertx/vert.x/pull/6124

PR #5881, opened 2026-01-09
Move the max lifetime connection pooling implementation to the pool.

vietj:
Motivation:

The implementation of max lifetime connection pooling is implemented in the HttpClientConnection#isValid method of each implementation we have.

This feature can actually be implemented by the pool instead of being implemented in HttpClientConnection#isValid as long as the pool can determine the creation timestamp of the connection.

Changes:

- Stop checking the max lifetime validity in each implementation of HttpClientConnection#isValid
- Move the max lifetime validity check in the caller of HttpClientConnection#isValid using the creation timestamp of the connection

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xOrH2ChannelConnector.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/Http1xOrH2ChannelConnector.java
@@ -22,7 +22,6 @@ import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.timeout.IdleStateHandler;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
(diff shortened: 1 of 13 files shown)


tsegismont:
Thanks @vietj . Have you planned to apply this change to the SQL Client pool as well?

tsegismont:
Forget about it, I misunderstood the goal of the change. It does not move the check to the base pool, it simplifies it for HTTP alone

vietj:
> Forget about it, I misunderstood the goal of the change. It does not move the check to the base pool, it simplifies it for HTTP alone

right, it might move to the pool itself some day but I'd like to keep the pool currently as stable as possible.

the win here is that the max lifetime parameter avoids to be propagated to the connection which simplifies a bit things

PR #5892, opened 2026-01-13
Update the Vert.x logging implementation to log better human readable message in absence of a log message.

vietj:
Motivation:

Some logging event might arrive with a null message and trigger the logging of the NULL string which is very confusing for human analyzing the a log tail.

We should improve this and provide a better message.

Changes:

Use the <missing-log-message> when no log message is provided instead of NULL.

### DIFF
--- a/src/main/java/io/vertx/core/logging/JULLogDelegate.java
+++ b/src/main/java/io/vertx/core/logging/JULLogDelegate.java
@@ -22,6 +22,39 @@ import java.util.logging.LogRecord;
  * @author <a href="kenny.macleod@kizoom.com">Kenny MacLeod</a>
  */
 public class JULLogDelegate implements LogDelegate {
+
+  private static final String MISSING_LOG_MESSAGE = "<missing-log-message>";
+
+  /**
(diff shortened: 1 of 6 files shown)


rsvoboda:
The fix doesn't fix native binary, some details in https://github.com/quarkus-qe/quarkus-test-suite/pull/3019#issuecomment-4591318327 where we tried to have additional coverage for the QUARKUS-6806 issue

vietj:
> The fix doesn't fix native binary, some details in [quarkus-qe/quarkus-test-suite#3019 (comment)](https://github.com/quarkus-qe/quarkus-test-suite/pull/3019#issuecomment-4591318327) where we tried to have additional coverage for the QUARKUS-6806 issue

is there a reason explaining this ? we are not testing native in vertx, so there is no way to speculate about what works and what does not work in native

rsvoboda:
Hi Julien, we can have the native thing tracked in https://github.com/quarkusio/quarkus/issues/40918 comment.

I wanted to add heads up that the trouble is not fully fixed. It works fine for JVM mode.

PR #5900, opened 2026-01-15
Replace apacheds-protocol-dns with mock Netty dns server in tests [Cherry-Pick]

pendula95:


vietj:
that looks great, thanks @pendula95

PR #5902, opened 2026-01-19
Enable the io_uring sendfile implementations.

### LINKED ISSUE #5743: Zero copy file transferring (via FileRegion) is disabled for IoUring transport.
Hello dear Vert.x team. 

As I see, in the IoUringTransport the support of file region is disabled by 
```
public class IoUringTransport implements Transport {
...
  @Override
  public boolean supportFileRegion() {
    return false;
  }
...
```
which mean that **Http1xServerResponse.sendFile** function will always fallback to read / write model instead of zero copy. 

And as Netty have direct support for file regions over io_uring (via IoUringFileRegion) I think it is good idea to enable this support for Vert.x as well, what do you think?

P.S. I tried to "hack" transport just to make this getter return true (via Unsafe) and IoUringFileRegion start to work BUT it has constant **fd** leak, actually they were never closed and I thought that probably it is bug of IoUringFileRegion but I check in reactor-netty lib and it utilizing IoUringFileRegion don't have leak which mean my hack was not enough and we also need somewhere handle FileRegion closure / releasing and as you have super huge experience with Netty and DefaultFileRegion I think handling IoUringFileRegion will not cause big problems for you. 

Thank you for bringing such perfect software as Vert.x to JVM community)

dreamlike-ocean:
Motivation:

Enable the io_uring UDS and sendfile implementations

Current Netty io_uring transport already fully provides sendfile and Unix domain socket functionality. This PR enables Vert.x support for these features.
	1.	`io.vertx.core.impl.transports.IoUringTransport#supportsDomainSockets` will always return true, and  `io.vertx.core.impl.transports.IoUringTransport#configure(io.vertx.core.net.TcpOptions, boolean, io.netty.bootstrap.ServerBootstrap)` and `io.vertx.core.impl.transports.IoUringTransport#configure(io.vertx.core.net.TcpOptions, boolean, io.netty.bootstrap.Bootstrap)` are adjusted to support Unix domain sockets.
	2.	`io.vertx.core.impl.transports.IoUringTransport#supportFileRegion`: if the current Linux kernel supports IORING_OP_SPLICE and it is not disabled by system parameters, sendfile support is enabled.
	3.	Due to performance issues with io_uring splice (see https://github.com/netty/netty/issues/15747), support is added to forcibly disable io_uring sendfile via a system parameter.
	4.	Introduce UncloseableFileRegion to replace the current DefaultFileRegion in Vert.x code, preventing resource leaks caused by non-zero reference counts. fix #5743
	
Conformance:

I have signed the Eclipse Contributor Agreement

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
@@ -20,6 +20,7 @@ import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.channel.unix.UnixChannelOption;
 import io.netty.channel.uring.*;
 import io.vertx.core.datagram.DatagramSocketOptions;
+import io.vertx.core.impl.SysProps;
 import io.vertx.core.net.TcpConfig;
 import io.vertx.core.net.TcpOption;
 import io.vertx.core.net.impl.SocketAddressImpl;
@@ -34,6 +35,8 @@ import static io.vertx.core.impl.transports.NioTransport.configOption;
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
 public class IoUringTransport implements Transport {
+
+  private static final boolean DISABLE_SENDFILE = SysProps.DISABLE_IO_URING_SENDFILE.getBoolean();
 
   private static volatile int pendingFastOpenRequestsThreshold = 256;
 
@@ -68,7 +71,7 @@ public class IoUringTransport implements Transport {
 
   @Override
   public boolean supportFileRegion() {
-    return IoUring.isSpliceSupported();
+    return IoUring.isSpliceSupported() && !DISABLE_SENDFILE;
   }
 
   @Override
--- a/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
@@ -71,7 +71,7 @@ public class IoUringTransport implements Transport {
 
   @Override
   public boolean supportFileRegion() {
-    return IoUring.isSpliceSupported() && !DISABLE_SENDFILE;
+    return !DISABLE_SENDFILE && IoUring.isSpliceSupported();
   }
 
   @Override
--- a/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
@@ -20,7 +20,6 @@ import io.netty.channel.unix.DomainSocketAddress;
 import io.netty.channel.unix.UnixChannelOption;
 import io.netty.channel.uring.*;
 import io.vertx.core.datagram.DatagramSocketOptions;
-import io.vertx.core.impl.SysProps;
 import io.vertx.core.net.TcpConfig;
 import io.vertx.core.net.TcpOption;
 import io.vertx.core.net.impl.SocketAddressImpl;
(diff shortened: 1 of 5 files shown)


dreamlike-ocean:
Additionally, regarding advanced features supported by Netty io_uring such as BufferGroup read and sendzc, I believe users can implement them manually, for example, in the following way.
https://gist.github.com/dreamlike-ocean/38d8e62c043018f1ab4dd872414a0b33

vietj:
@dreamlike-ocean can you rebase on latest master ?

dreamlike-ocean:
> @dreamlike-ocean can you rebase on latest master ?

done

tsegismont:
@dreamlike-ocean can you please rebase your PR? Please note that UDS is now present in master branch

dreamlike-ocean:
> @dreamlike-ocean can you please rebase your PR? Please note that UDS is now present in master branch

done! please review it

vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
```diff
@@ -68,7 +71,7 @@ public boolean supportsDomainSockets() {
 
   @Override
   public boolean supportFileRegion() {
-    return false;
+    return IoUring.isSpliceSupported() && !DISABLE_SENDFILE;
```
doxlik: @dreamlike-ocean Thank you!

Maybe we should order this as:

return !DISABLE_SENDFILE && IoUring.isSpliceSupported();

This lets the explicit sendfile-disable switch short-circuit before touching IoUring splice support. The runtime difference is probably only cold-path / initialization, because isSpliceSupported() returns a cached static boolean after IoUring is initialized, but it seems cleaner to respect the disable flag first.

dreamlike-ocean: good idea!

dreamlike-ocean:
Due to the Netty pull request #16571 “Add generic FileRegion support in io_uring stream channel”, it is no longer necessary to check whether the splice operation is available.

vertx-core/src/main/java/io/vertx/core/net/impl/UncloseableFileRegion.java
```diff
@@ -0,0 +1,37 @@
...
+ * @author <a href="mailto:dreamlike.vertx@gmail.com">MengYang Li</a>
+ */
+public class UncloseableFileRegion extends DefaultFileRegion {
+  public UncloseableFileRegion(FileChannel fileChannel, long position, long count) {
+    super(fileChannel, position, count);
+  }
+
+  public UncloseableFileRegion(File file, long position, long count) {
+    super(file, position, count);
+  }
```
tsegismont: I think this constructor is not required

vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
```diff
@@ -68,7 +71,7 @@ public boolean supportsDomainSockets() {
 
   @Override
   public boolean supportFileRegion() {
-    return false;
+    return !DISABLE_SENDFILE;
```
tsegismont: How is this going to help? As far as I understand, if send file is disabled with this new sysprop, the fallback mechanism is to use a `ChunkedNioFile`, not epoll's sendfile.

If my understanding is correct, I believe we should not have this sysprop. If sending files with io_uring is sometimes slower than with epoll, then I think we can admit the limitation until https://github.com/netty/netty/issues/15747 is resolved.

tsegismont (commented):
Thank you @dreamlike-ocean, a couple comments/concerns.

vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
dreamlike-ocean: I agree with your point. I’m going to return true directly, given the Netty pull request #16571, “Add generic FileRegion support in io_uring stream channel.”
In the future, we could potentially implement a higher-performance file-oriented ChunkedWriteHandle based on io_uring. Introducing this system property would make more sense at that stage.

tsegismont (approved):
LGTM, thank you @dreamlike-ocean !

vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
tsegismont: Sounds good to me

PR #1058, opened 2015-06-04
Add OSGi metadata to the vert.x core jar

cescoffier:
Make vert.x core a valid OSGi bundle.

* Generate OSGi metadata
* Fix ServiceFactory classloading
* Add documentation

This commit fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=469363

The resulting jar is still a valid jar.

PR #5903, opened 2026-01-20
Avoid calling ServiceLoader twice with same class loader

gsmet:
Motivation:

The ServiceLoader goes through all the jars to find its files and it seems like a good idea to avoid doing it twice for the same class loader.

This is somehow related to the work we are doing in Quarkus for Leyden.

Conformance:

I think I have signed the ECA already to work on Bean Validation.

### DIFF
--- a/vertx-core/src/test/java/io/vertx/it/servicehelper/ServiceHelperTest.java
+++ b/vertx-core/src/test/java/io/vertx/it/servicehelper/ServiceHelperTest.java
@@ -16,12 +16,16 @@ import io.vertx.test.core.TestUtils;
 import org.junit.Test;
 
 import java.io.File;
+import java.io.IOException;
 import java.lang.reflect.Method;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Collection;
+import java.util.Enumeration;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
 
 /**
  * Check the service helper behavior.
@@ -125,5 +129,44 @@ public class ServiceHelperTest {
     Method method = serviceHelperClass.getMethod("loadFactories", Class.class);
     Collection collection = (Collection) method.invoke(null, someFactoryClass);
     assertThat(collection).hasSize(1);
+  }
+
+  @Test
+  public void shouldInvokeGetResourcesOnce() throws Exception {
+    // We want to make sure we don't try to get resource from CL twice
+    // When the TCCL is the same as the ServiceHelper's classloader
+
+    AtomicInteger invocations = new AtomicInteger();
+
+    // Load the ServiceHelper class from a custom classloader.
+    ClassLoader custom = new URLClassLoader(new URL[]{
+      new File(TestUtils.MAVEN_TARGET_DIR, "classes").toURI().toURL(),
+      new File(TestUtils.MAVEN_TARGET_DIR, "test-classes").toURI().toURL(),
+      serviceHelperFile.toURI().toURL(),
+    }, null) {
+
+      @Override
+      public Enumeration<URL> getResources(String name) throws IOException {
+        invocations.incrementAndGet();
+        return super.getResources(name);
+      }
+    };
+
+    Class serviceHelperClass = custom.loadClass(ServiceHelper.class.getName());
+    Class notImplementedSPIClass = custom.loadClass(NotImplementedSPI.class.getName());
+    assertThat(serviceHelperClass.getClassLoader()).isEqualTo(custom);
+    assertThat(notImplementedSPIClass.getClassLoader()).isEqualTo(custom);
+
+    final ClassLoader originalTCCL = Thread.currentThread().getContextClassLoader();
+    try {
+      // Set TCCL to the CL that loaded the ServiceHelper class
+      Thread.currentThread().setContextClassLoader(custom);
+      Method method = serviceHelperClass.getMethod("loadFactories", Class.class);
+      Collection collection = (Collection) method.invoke(null, notImplementedSPIClass);
+      assertThat(collection).hasSize(0);
+      assertEquals(1, invocations.get());
+    } finally {
+      Thread.currentThread().setContextClassLoader(originalTCCL);
+    }
   }
 }


gsmet:
FYI @geoand

gsmet:
I'm interested in having this backported to 4.x for current Quarkus.

gsmet:
BTW, just to clarify: in my case, it was for a service file that didn't exist so we were going through all the calls as the list was empty.

tsegismont:
> I'm interested in having this backported to 4.x for current Quarkus.

@gsmet fine with me. When this one is merged, can you send PRs to the `4.x` and `5.0` branches please? Thanks

Please note that, in Vert.x 5, this class has moved to an impl package (not visible in a modular Java app)

gsmet:
So it all depends on what you want to do. My initial PR didn't change the logic.

Your commit changes the logic right? If you enforce a CL, we don't try to fallback to the `ServiceHelper` CL anymore? Also didn't you drop the condition that is the very reason of my PR in the TCCL case? Avoid interrogating the same CL twice?

tsegismont:
> So it all depends on what you want to do. My initial PR didn't change the logic.

Right

> Your commit changes the logic right? If you enforce a CL, we don't try to fallback to the `ServiceHelper` CL anymore? Also didn't you drop the condition that is the very reason of my PR in the TCCL case? Avoid interrogating the same CL twice?

Yes, it does. I think we messed-up somewhere, and started to try loading factories twice even when the user provided a specific classloader.

As the comment at the bottom suggests, we should try again only when the first attempt was with the TCCL.

The original change is https://github.com/eclipse-vertx/vert.x/pull/1058

gsmet:
But the TCCL might be what loaded `ServiceHelper` so I'm not sure why you dropped my optimization in the TCCL case.

tsegismont:
If you invoke `io.vertx.core.impl.ServiceHelper#loadFactories(java.lang.Class<T>, java.lang.ClassLoader)` by providing a classloader instead of `null`, there will be a single attempt.

gsmet:
Yeah, I understand that. But why make the TCCL case worse than with my change? I don't understand the rationale?

gsmet:
I'm specifically talking about having ` && TCCL != ServiceHelper.class.getClassLoader()` in the condition before falling back.

vietj:
is that possible to get a test for this ?

tsegismont:
As discussed off-list, the code in `master` right now tries loading factories twice, even when a classloader is provided, hence the confusion.

The original intent of this PR is to avoid loading factories twice when the TCCL happens to be the same as the `ServiceHelper`'s classloader.

tsegismont:
> is that possible to get a test for this ?

If we had a mocking lib in the test scope, we could verify `io.vertx.core.impl.ServiceHelper#loadFactories(java.util.ServiceLoader<T>)` is invoked exactly once when:

- `io.vertx.core.impl.ServiceHelper#loadFactories(java.lang.Class<T>, java.lang.ClassLoader)` is invoked with classloader set to `null`
- the TCCL is set to `ServiceHelper.class.getClassLoader()`

Otherwise, I don't know how we can check what our code does with the classloader

vietj:
I think we already have structure for this to verify loading with a classloader

tsegismont:
> I think we already have structure for this to verify loading with a classloader

I see that in `ServiceHelperTest`, thanks.

@gsmet can you share a stacktrace or flamegraph that shows how the classloader was searched several times so that we can intercept the right methods and verify assumptions?

gsmet:
Typically, you have this case:

Screenshot From 2026-01-26 14-05-28

gsmet:
Note that in this particular run, the TCCL was different from the `ServiceHelper` class loader and in this case, we will still try to load files twice, unfortunately.

With fast-jar, the optimization should work, with what we have to use for AOT, it won't as we have to load the classes from JDK class loader for them to be cached.

Let me clarify a bit:

- fast-jar: TCCL is also the CL loading most of the application classes
- aot-jar: TCCL is just a thin layer to load resources (including service files) and the loading of classes themselves has to be delegated to the JDK class loader for the AOT cache class loading to work

I'm wondering if having all frameworks to pay the price for OSGi might be a bit counter productive but these issues are hard to solve :/.

tsegismont:
Thanks for the detailed answer @gsmet 

@vietj @gsmet review the last commit that provides a test

gsmet:
I was thinking a bit more about it and I was wondering if we shouldn’t have the option to pass a CL to this various ServiceLoader calls when initializing the VertxBuilder. 

Because in our case, we would like to enforce the CL and always avoid this two calls. 

Doesn’t prevent us to merge this PR but I wonder if we should push it a bit further so that we are sure we never do the dual calls with Quarkus.

tsegismont:
Perhaps we could have an option in Vert.x Bootstrap that configures the classloader to use when searching for services.

@vietj thoughts?

vietj:
@gsmet I think perhaps you could instead use the VertxBootstrap and avoid the service loader part by setting all the services you want on the VertxBootstrap, we could modify the bootstrap to avoid discovering things and thus provide you full control over things, it seems better for you.

gsmet:
But that would prevent people from tweaking things, right? I'm not sure we want to be as opinionated as that.

gsmet:
Here is the dirty patch I applied to achieve what I wanted (the work in this PR is still needed).

It gives an idea of what I would like to achieve.

It allows us to define the class loader without it being an issue if people want to extend Vert.x.

Now, what I would like to know is if it would be acceptable, if things should be included in VertxOptions instead... Maybe we could have a short call about it?

```diff
diff --git a/src/main/java/io/vertx/core/impl/VerticleManager.java b/src/main/java/io/vertx/core/impl/VerticleManager.java
index dc4adaaa9..95f88ee69 100644
--- a/src/main/java/io/vertx/core/impl/VerticleManager.java
+++ b/src/main/java/io/vertx/core/impl/VerticleManager.java
@@ -45,7 +45,7 @@ public class VerticleManager {
   }
 
   private void loadVerticleFactories() {
-    Collection<VerticleFactory> factories = ServiceHelper.loadFactories(VerticleFactory.class);
+    Collection<VerticleFactory> factories = ServiceHelper.loadFactories(VerticleFactory.class, vertx.classLoader());
     factories.forEach(this::registerVerticleFactory);
     VerticleFactory defaultFactory = new JavaVerticleFactory();
     defaultFactory.init(vertx);
diff --git a/src/main/java/io/vertx/core/impl/VertxBuilder.java b/src/main/java/io/vertx/core/impl/VertxBuilder.java
index 48d963552..7c31cfba4 100644
--- a/src/main/java/io/vertx/core/impl/VertxBuilder.java
+++ b/src/main/java/io/vertx/core/impl/VertxBuilder.java
@@ -49,6 +49,7 @@ public class VertxBuilder {
 
   private VertxOptions options;
   private JsonObject config;
+  private ClassLoader classLoader;
   private Transport transport;
   private ClusterManager clusterManager;
   private NodeSelector clusterNodeSelector;
@@ -85,6 +86,11 @@ public class VertxBuilder {
     return config;
   }
 
+  public VertxBuilder classLoader(ClassLoader classLoader) {
+    this.classLoader = classLoader;
+    return this;
+  }
+
   /**
    * @return the transport to use
    */
@@ -235,7 +241,8 @@ public class VertxBuilder {
       transport,
       fileResolver,
       threadFactory,
-      executorServiceFactory);
+      executorServiceFactory,
+      classLoader);
     vertx.init();
     return vertx;
   }
@@ -257,7 +264,8 @@ public class VertxBuilder {
       transport,
       fileResolver,
       threadFactory,
-      executorServiceFactory);
+      executorServiceFactory,
+      classLoader);
     vertx.initClustered(options, handler);
   }
 
@@ -271,7 +279,7 @@ public class VertxBuilder {
     initMetrics(options, providers);
     initTracing(options, providers);
     initClusterManager(options, providers);
-    providers.addAll(ServiceHelper.loadFactories(VertxServiceProvider.class));
+    providers.addAll(ServiceHelper.loadFactories(VertxServiceProvider.class, classLoader));
     initProviders(providers);
     initThreadFactory();
     initExecutorServiceFactory();
diff --git a/src/main/java/io/vertx/core/impl/VertxImpl.java b/src/main/java/io/vertx/core/impl/VertxImpl.java
index 3b4916429..b33977780 100644
--- a/src/main/java/io/vertx/core/impl/VertxImpl.java
+++ b/src/main/java/io/vertx/core/impl/VertxImpl.java
@@ -165,15 +165,24 @@ public class VertxImpl implements VertxInternal, MetricsProvider {
   private final ThreadLocal<WeakReference<ContextInternal>> stickyContext = new ThreadLocal<>();
   private final boolean disableTCCL;
   private final Boolean useDaemonThread;
+  private final ClassLoader classLoader;
 
   VertxImpl(VertxOptions options, ClusterManager clusterManager, NodeSelector nodeSelector, VertxMetrics metrics,
             VertxTracer<?, ?> tracer, Transport transport, FileResolver fileResolver, VertxThreadFactory threadFactory,
             ExecutorServiceFactory executorServiceFactory) {
+    this(options, clusterManager, nodeSelector, metrics, tracer, transport, fileResolver, threadFactory, executorServiceFactory, null);
+  }
+
+  VertxImpl(VertxOptions options, ClusterManager clusterManager, NodeSelector nodeSelector, VertxMetrics metrics,
+            VertxTracer<?, ?> tracer, Transport transport, FileResolver fileResolver, VertxThreadFactory threadFactory,
+            ExecutorServiceFactory executorServiceFactory, ClassLoader classLoader) {
     // Sanity check
     if (Vertx.currentContext() != null) {
       log.warn("You're already on a Vert.x context, are you sure you want to create a new Vertx instance?");
     }
 
+    this.classLoader = classLoader;
+
     Boolean useDaemonThread = options.getUseDaemonThread();
     int workerPoolSize = options.getWorkerPoolSize();
     int internalBlockingPoolSize = options.getInternalBlockingPoolSize();
@@ -352,6 +361,11 @@ public class VertxImpl implements VertxInternal, MetricsProvider {
     return transport;
   }
 
+  @Override
+  public ClassLoader classLoader() {
+    return classLoader;
+  }
+
   @Override
   public boolean isNativeTransportEnabled() {
     return !(transport instanceof JDKTransport);
diff --git a/src/main/java/io/vertx/core/impl/VertxInternal.java b/src/main/java/io/vertx/core/impl/VertxInternal.java
index 77bc250f8..83e7a4a3e 100644
--- a/src/main/java/io/vertx/core/impl/VertxInternal.java
+++ b/src/main/java/io/vertx/core/impl/VertxInternal.java
@@ -88,6 +88,8 @@ public interface VertxInternal extends Vertx {
 
   Transport transport();
 
+  ClassLoader classLoader();
+
   /**
    * Create a WebSocket client using the specified options and close future
    *
diff --git a/src/main/java/io/vertx/core/impl/VertxWrapper.java b/src/main/java/io/vertx/core/impl/VertxWrapper.java
index 520b12e6c..982c2b40b 100644
--- a/src/main/java/io/vertx/core/impl/VertxWrapper.java
+++ b/src/main/java/io/vertx/core/impl/VertxWrapper.java
@@ -348,6 +348,11 @@ public abstract class VertxWrapper implements VertxInternal {
     return delegate.transport();
   }
 
+  @Override
+  public ClassLoader classLoader() {
+    return delegate.classLoader();
+  }
+
   @Override
   public WebSocketClient createWebSocketClient(WebSocketClientOptions options) {
     return delegate.createWebSocketClient(options);
```

vietj:
In vertx 5, options should only be data, never some kind of live object, such objects should be in builder.

vietj:
@gsmet we can discuss the classloader in another place or issue, I strongly believe that actually the VertxBootstrap should not perform any kind of loading and the builder instead should. The bootstrap should be more or less like a factory that initialize and build a Vert.x instance. This would allow bootstrap users to load services themselves using whatever mechanism they like and have full and precise control over what gets loaded and what is not.

tsegismont:
@vietj as discussed off-list, I believe we can merge this PR and backport it. Then, we can continue the conversation about improving the builder here, or maybe better on a separate issue.

gsmet:
Yes so the thing is that without the CL passed, we end up in the very same situation, even with this patch. We need both for the issue to get fixed on our side.

tsegismont:
> Yes so the thing is that without the CL passed, we end up in the very same situation, even with this patch. We need both for the issue to get fixed on our side.

In a previous comment you said this would do the trick, and then we discuss the option to improve the builder so that there is greater control of the service loading. What do I miss?

gsmet:
What changed is that we now have a TCCL that is different from the class loader loading the classes, also in the AOT case (at the beginning we were experimenting with only one CL).

So if we don't pass the CL, we end up being in the second case and we do two lookups because the TCCL and the CL loading the class are different. But what you did is an improvement in the case of `fast-jar` and also a bug fix so +1 to merge it.

We had a discussion today with @vietj about this class loader thing and he's going to prototype something but he wanted this one merged and backported so I think we can go ahead.

Let me know if things are still unclear.

tsegismont:
OK, I didn't know your requirements had changed. I'll go ahead and merge this

vietj:
@gsmet here is an initial draft for Vert.x 5.1 https://github.com/eclipse-vertx/vert.x/pull/5946

gsmet:
Thanks @tsegismont , let me know if you want me to do the backport to 4.x.

PR #5938, opened 2026-02-06
Fix unstable TimerTest

### LINKED ISSUE #5937: Unstable Test TimerTest
### Version

5.0

### Context

When running job https://github.com/eclipse-vertx/vert.x/actions/runs/21745456833/job/62730103189 the following test failed:

```
Error:  Failures: 
Error:    TimerTest.access$600:33->AsyncTestBase.assertTrue:378
```

### Steps to reproduce

This is a random failure in CI.

Looking at the code, it relies on a monotonic clock, wich might have not worked in that execution on a Windows system. 

https://github.com/eclipse-vertx/vert.x/blob/f1a7da10dc12e871b1e2d34a49ea6e3d45f68b0f/vertx-core/src/test/java/io/vertx/tests/timer/TimerTest.java#L126

It would be more reliable to use System.nanoTime() which is guaranteed to be monotonic as a first attempt.

### Do you have a reproducer?

_No response_

ahus1:
Closes #5937

(cherry picked from commit 01e00da0e9d2a2cf360cc0a1e2001f231b725b09)

Motivation: System.nanoTime() should be monotonic. Also more descriptive assertion with a hamcrest matcher so we can dive deeper if it fails again.

### DIFF
--- a/vertx-core/src/test/java/io/vertx/tests/timer/TimerTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/timer/TimerTest.java
@@ -17,6 +17,7 @@ import io.vertx.core.impl.VertxImpl;
 import io.vertx.core.internal.VertxInternal;
 import io.vertx.test.core.Repeat;
 import io.vertx.test.core.VertxTestBase;
+import org.hamcrest.Matchers;
 import org.junit.Test;
 
 import java.util.concurrent.CancellationException;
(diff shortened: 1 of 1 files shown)


ahus1:
@vietj - One more attempt to stabilize the CI by taking care of this test. I'll provide a backport once it is merged to main. Thanks!

PR #5942, opened 2026-02-06
Fix unstable QuicServerTest.testRebind

### LINKED ISSUE #5941: Unstable test QuicServerTest.testRebind
### Version

master

### Context

This failed randomly in CI, see https://github.com/eclipse-vertx/vert.x/actions/runs/21747092960/job/62735543905 for an example

```
Error:  Errors: 
Error:    QuicServerTest.testRebind » Bind Address already in use
Error:  Tests run: 6639, Failures: 0, Errors: 1, Skipped: 100
```

### Steps to reproduce

This might be due to not calling an explicit close on the QuicServer. I'll prepare a PR

### Do you have a reproducer?

_No response_

ahus1:
Closes #5941

Motivation: Closing the server explicitly when the verticle is undeployed to avoid the "Bind Address already in use" error. 

Let me know if this is a valid fix, or if the test was trying to test something else.

### DIFF
--- a/vertx-core/src/test/java/io/vertx/tests/net/quic/QuicServerTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/net/quic/QuicServerTest.java
@@ -900,8 +900,15 @@ public class QuicServerTest extends VertxTestBase {
     };
     int num = 4;
     for (int i = 0;i < num;i++) {
-      String id = vertx.deployVerticle(deployable).await();
-      vertx.undeploy(id).await();
+      try {
+        String id = vertx.deployVerticle(deployable).await();
+        // Undeploying the verticle is expected to close the QuicServer as it is bound to that context, see CleanableQuicServer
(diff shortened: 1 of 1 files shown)


vietj:
@ahus1 normally the HTTP server bound by this verticle should be automatically undeployed as it is registered against the verticle for being undeployed via a hook

ahus1:
> normally the HTTP server bound by this verticle should be automatically undeployed as it is registered against the verticle for being undeployed via a hook

OK, I wasn't aware of that automatic handling - it came to me as a surprise. I see that this is mentioned in the docs. 

As I wasn't able to reproduce the problem, I first enhanced the code to provide some additional information when this happens the next time. Please let me know if this is ok. 

BTW, the GitHub CI failed due to another problem, and I opened yet another unstable test case issue: #5952

ahus1:
This has run stale. Closing.

PR #5971, opened 2026-02-13
Make sure that Context.unwrap always returns a raw context

tsegismont:
Even when the instance was created by duplicating a duplicated context.


tsegismont:
@vietj this follows-up on our conversation during the community meeting.

In fact, the implementation already ensures that a context created by duplicating a duplicated context simply wraps the raw context :sunglasses: :smile:

Issue #5452, opened 2025-01-19
Access TLVs of Proxy Protocol v2 header from NetSocket


trnguyencflt:
#### Describe the feature

Ability to read TLVs of Proxy Protocol header from NetSocket. To be more clear, in HAProxyMessageCompletionHandler#decode function, we only use `HAProxyMessage msg` only for overriding remote address and local address, however, in some use case, there are information in the proxy protocol header that are store in `List<HAProxyTLV> tlvs` of `msg` that are needed. Currently there is not a way to access that data when enabling proxy protocol in NetServerOptions of NetServer

vietj:
I see, you could like to access those info from the NetSocket ?

trnguyencflt:
> I see, you could like to access those info from the NetSocket ?

Yes, NetSocket is created from NetServer (in #connectHandler), when NetServer is init with proxy protocol enabled then there should be a way to access proxy protocol data, the current way of getting only remote/local address isn't sufficient for all use cases. Thanks

vietj:
so in practice a map would be sufficient

trnguyencflt:
> so in practice a map would be sufficient

Not sure if I understand this correctly, TLVs are represented as a list in HAProxyMessage (of netty), could you explain more how the map works here? Thanks.

vietj:
I meant a Java map to expose this in API

trnguyencflt:
> I meant a Java map to expose this in API

That'd work, thanks 👍

PR #5972, opened 2026-02-13
Add support for HA Proxy Protocol TLV's

damianorenfer:
Motivation:

TLV's are supported in netty, but currently not accessible to vert.x users.
Here the TLV's are exposed in `NetSocket#tlvs`

This also answers to #5452 

Conformance:
ECA signed

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/HttpConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/HttpConnection.java
@@ -23,6 +23,7 @@ import java.security.cert.Certificate;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -303,5 +304,15 @@ public interface HttpConnection {
    * @return the indicated server name
    */
   String indicatedServerName();
+
+  /**
+   * @return the type-length-values present in the TCP header as a list of map entries
+   * where the key contains the TLV type and the value contains the TLV value.
+   * This is mainly used for HA Proxy Protocol v2
+   */
+  @GenIgnore()
+  default List<Map.Entry<Buffer, Buffer>> proxyProtocolV2HeaderTLVs() {
+    return List.of();
+  }
 
 }
--- a/vertx-core/src/main/java/io/vertx/core/net/NetSocket.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/NetSocket.java
@@ -18,7 +18,6 @@ import io.vertx.codegen.annotations.Nullable;
 import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
-import io.vertx.core.VertxException;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.streams.ReadStream;
 import io.vertx.core.streams.WriteStream;
@@ -29,6 +28,7 @@ import java.security.cert.Certificate;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Represents a socket-like interface to a TCP connection on either the
@@ -313,6 +313,14 @@ public interface NetSocket extends Socket {
    * @return the application-level protocol negotiated during the TLS handshake
    */
   String applicationLayerProtocol();
+
+  /**
+   * @return the type-length-values present in the TCP header as a list of map entries
+   * where the key contains the TLV type and the value contains the TLV value.
+   * This is mainly used for HA Proxy Protocol v2
+   */
+  @GenIgnore()
+  List<Map.Entry<Buffer, Buffer>> proxyProtocolV2HeaderTLVs();
 
 }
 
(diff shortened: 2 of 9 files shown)


vietj:
I'd rather expose that has a Map<Buffer, Buffer> instead (that map can be ordered if needed)

damianorenfer:
Sure, I just updated the code accordingly

vertx-core/src/main/java/io/vertx/core/net/NetSocket.java
```diff
@@ -316,5 +316,12 @@ default List<Certificate> peerCertificates() throws SSLPeerUnverifiedException {
    */
   String applicationLayerProtocol();
 
+  /**
+   * @return the type-length-values present in the TCP header as map where the key contains the TLV type and
+   * the value contains the TLV value, mainly used for HA Proxy Protocol v2
+   */
+  @GenIgnore()
+  Map<Buffer, Buffer> tlvs();
```
vietj: why do we need that ? Quic does not yet support proxying at the moment and we are not sure it makes sense

vietj: that seems at the moment to be a TCP concern only

tmonney:
The protocol technically allows for multiple occurrences of TLVs with the same type. It's probably not that common, but I am wondering if it makes sense to add unnecessary constraints. The underlying Netty decoder also uses a List, presumably for that reason.

vertx-core/src/main/java/io/vertx/core/net/NetSocket.java
damianorenfer: Do you see a better place to put this one ? The goal is to access TLV's in `NetServer#connectHandler`

damianorenfer:
> The protocol technically allows for multiple occurrences of TLVs with the same type. It's probably not that common, but I am wondering if it makes sense to add unnecessary constraints. The underlying Netty decoder also uses a List, presumably for that reason.

@vietj I was also not able to find anything about TLV uniqueness based on type [here](https://www.haproxy.org/download/3.1/doc/proxy-protocol.txt).
What do you think ?

vertx-core/src/main/java/io/vertx/core/http/HttpConnection.java
```diff
@@ -298,4 +299,11 @@ default List<Certificate> peerCertificates() throws SSLPeerUnverifiedException {
    */
   String indicatedServerName();
 
+  /**
+   * @return the type-length-values present in the TCP header as map where the key contains the TLV type and
+   * the value contains the TLV value, mainly used for HA Proxy Protocol v2
+   */
+  @GenIgnore()
+  Map<Buffer, Buffer> tlvs();
```
vietj:
```suggestion
  default Map<Buffer, Buffer> tlvs() {
    return Map.of();
  }
```

vertx-core/src/main/java/io/vertx/core/net/QuicConnection.java
```diff
@@ -176,4 +177,11 @@ default Future<Void> shutdown(Duration timeout) {
    */
   SocketAddress localAddress();
 
+  /**
+   * @return the type-length-values present in the TCP header as map where the key contains the TLV type and
```
vietj:
```suggestion
```

damianorenfer: I just wanted four feedback on this one https://github.com/eclipse-vertx/vert.x/pull/5972#issuecomment-3907793562

Before doing further modifications

vietj:
> > The protocol technically allows for multiple occurrences of TLVs with the same type. It's probably not that common, but I am wondering if it makes sense to add unnecessary constraints. The underlying Netty decoder also uses a List, presumably for that reason.
> 
> @vietj I was also not able to find anything about TLV uniqueness based on type [here](https://www.haproxy.org/download/3.1/doc/proxy-protocol.txt). What do you think ?

I think we can use an Iterable<Map.Entry<Buffer, Buffer>> instead

damianorenfer:
> > > The protocol technically allows for multiple occurrences of TLVs with the same type. It's probably not that common, but I am wondering if it makes sense to add unnecessary constraints. The underlying Netty decoder also uses a List, presumably for that reason.
> > 
> > 
> > @vietj I was also not able to find anything about TLV uniqueness based on type [here](https://www.haproxy.org/download/3.1/doc/proxy-protocol.txt). What do you think ?
> 
> I think we can use an Iterable<Map.Entry<Buffer, Buffer>> instead

Sure, done!

vertx-core/src/main/java/io/vertx/core/http/HttpConnection.java
```diff
@@ -304,4 +305,14 @@ default List<Certificate> peerCertificates() throws SSLPeerUnverifiedException {
    */
   String indicatedServerName();
 
+  /**
+   * @return the type-length-values present in the TCP header as an iterable of map entries
+   * where the key contains the TLV type and the value contains the TLV value.
+   * This is mainly used for HA Proxy Protocol v2
+   */
+  @GenIgnore()
+  default Iterable<Map.Entry<Buffer, Buffer>> tlvs() {
```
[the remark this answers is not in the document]
damianorenfer: Sure, should we rename it everywhere ? also in NetSocket etc ?

vietj:
thanks @damianorenfer

drenferwl:
@vietj Do you think we can backport this to 4.x ?
Would be really appreciated to have it in Quarkus 3

PR #5975, opened 2026-02-17
TCP socket client/server API

vietj:
Provide a TCP socket interface as a replacement of the `NetSocket`/`NetServer`/`NetClient` API, the configuration layer uses the new TCP client/server configuration classes.

While the API is very close, this makes explicit reference to being TCP oriented and makes also some cleanup on those interfaces.


vietj:
on hold for 5.1

PR #5983, opened 2026-02-18
WIP - delay locals capture from Vertx to contexts

jponge:
I am wondering if we'd benefit from shifting the context local captures from the `Vertx` instance to the actual contexts creation. This involves a few allocations on duplication, but I don't think it'd be that much of a problem.

One possible improvement (but I have not verified) is that it could allow new `ContextLocal` instances to be registered after a `Vertx` context has been initialized.

WDYT?

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/LocalSeq.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/LocalSeq.java
@@ -15,7 +15,6 @@ import io.vertx.core.spi.context.storage.ContextLocal;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
 
 /**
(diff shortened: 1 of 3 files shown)


vietj:
what is the goal ?

jponge:
> One possible improvement (but I have not verified) is that it could allow new ContextLocal instances to be registered after a Vertx context has been initialized.

jponge:
It's not necessarily meant to be merged, I'm really wondering if we need to set in stone the locals array this early in the lifetime. I've had a test in the IDE + debugger where running would fail because the `ContextLocal` arrived later so the array index was wrong.

vietj:
the main reason the context local list is captured when creating a vertx instance, is that context are created assuming the list does not change and therefore there is an exact mapping of `ContextBase#locals` and `VertxImpl#contextLocals`.

If `contextLocals` changes, then a context created before the list is updated will have an incorrect local storage.

PR #5990, opened 2026-02-21
improve performance of Vertx timer

### LINKED ISSUE #4971: Memory optimization
There is a memory optimization that can be used in Vert.x in several places.
For example, here: 
`io.vertx.core.impl.VertxImpl.InternalTimerHandler`
`private final AtomicBoolean disposed = new AtomicBoolean();`
`AtomicBoolean` is quite a heavy object (many times bigger than volatile boolean)
It can be replaced with `VarHandle`. The `VarHandle` has all the same operations, but it is a static field member of a class and in object there is only a volatile byte or int field. You can see this technic in JDK 
`java.util.concurrent.FutureTask#set`

Replacing `Atomic`* with `VarHandle` and `volatile` in libraries makes sense.
It dramatically reduces the memory footprint without disadvantages.

I could make a PR for InternalTimerHandler.

re-thc:
Motivation: vertx timer handling and cancellation accesses a volatile field, this could be improved using an updater. Fixes #4971.

Changes: replace direct access to volatile field in timer with updater

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/VertxImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/VertxImpl.java
@@ -80,7 +80,6 @@ import java.net.InetSocketAddress;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


vietj:
Unless you spotted this on a hot path, I don't think I want to trade readability for performance in a case that seems to be narrow.

zekronium:
Timer is quite the hot path for a lot of operations especially if using VT

vietj:
@zekronium yes timers is certainly worth it

re-thc:
So yes, this was a hot path. I didn't see any regressions. Some memory improvements.

vietj:
@re-thc can you only contribute this change and discard the rest ?

re-thc:
@vietj done

vietj:
can you update the title of the issue as well as the description and what it changes

vietj:
please merge in single commit and follow the commit format: motivation/changes/results

re-thc:
@vietj done

vietj:
@re-thc can you update the motivation / PR title to specifically that you are improving the performance of vertx timer, that's what I meant. Currently as is we see "optimize vertx memory" or "Replace Atomic with Var Handle / Volatile to reduce memory usage." which is very generic and is not explanatory instead we should know that it impacts timer, e.g.


Title : improve performance of Vertx timer

Motivation: vertx timer handling and cancellation accesses a volatile field, this could be improved using an updater

Changes: replace direct access to volatile field in timer with updater

re-thc:
@vietj PR updated

vietj:
thanks @re-thc

PR #6008, opened 2026-03-04
Fix broken Jackson integration test.

vietj:
Motivation:

Jackson bump broke the integration test that was not incorrect due to a fix in max document length enforcement.

Changes:

Increase the max document length so the integration test is correct.

Add a test enforcing max document length is respected.


tsegismont:
Thanks for taking care of this @vietj

PR #6026, opened 2026-03-18
Support Unix Domain Sockets with io_uring transport

tsegismont:
In Netty 4.2, Unix Domain Sockets with io_uring are supported.

But they were not integrated by Vert.x yet.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2024 Contributors to the Eclipse Foundation
+ * Copyright (c) 2011-2026 Contributors to the Eclipse Foundation
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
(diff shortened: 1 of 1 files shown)


tsegismont:
@vietj please take a look at this draft.

In the long run, do you want to have a job for native io_uring + domain sockets, as we have for native epoll?

Here it's convenient because the job for native io_uring is temporarily disabled, so at least we have some smoke tests for TCP and HTTP client / servers.

vietj:
I'd like to have things tests as they should be

tsegismont:
Can you elaborate?

vietj:
@tsegismont do we have io uring fully passing on CI ?

vietj:
I can see now only a few tests are executed in the new io uring profile

tsegismont:
> @tsegismont do we have io uring fully passing on CI ?
> I can see now only a few tests are executed in the new io uring profile

Not yet. This is why this is a draft PR that we will be ready for review when the `NativeIoUring` CI job can be enabled again.

In the meantime, I was wondering if, eventually, we'd need the `NativeIoUring+DomainSockets` job like we have for epoll

vietj:
I don't know yet, I need to see how io uring testing comes

vietj:
great, but for now I'd like to stick to smoke tests in the io_uring test profile

tsegismont:
Merging as the failures are unrelated

PR #6064, opened 2026-04-15
Fix HttpClientResponse body()/end() race

snazy:
This change fixes the race in `HttpClientResponseImpl` described in #6038: `handleTrailers()`/`handleException()` can run before `body()`/`end()` creates the promise. If that happens, the completion signal is dropped and the future can hang. This is observed from the call site as a HTTP request timeo ut.

This fix is only applied to `HttpClientResponseImpl`:
- keep ended/failure state on `HttpClientResponseImpl`
- in `body()`/`end()`, create the `HttpEventHandler` promise while holding `synchronized(conn)`
- after unlock, replay completion/failure via `handleEnd()`/`handleException()` when needed

Although `HttpEventHandler` looks very similar (like it _could_ be affected by the same/similar race condition), it is not affected. I intentionally did not change `HttpEventHandler`, because it is also used in server request processing. The only case when server request processing could be affected , is when `body()` is called from a _different_ thread while `end()` is being processed - that is very unlikely (feels like a misuse).

Also added two new tests to `Http1xTest` that reliably fail without the change to `HttpClientResponseImpl` and reliably pass with that change. The added regression tests are `testResponseBodyAfterResponseEnd` and `testResponseEndAfterResponseEnd`.

Fixes #6038

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
@@ -17,6 +17,7 @@ import io.vertx.codegen.annotations.Nullable;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.MultiMap;
+import io.vertx.core.Promise;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.*;
 import io.vertx.core.http.impl.headers.HeadersAdaptor;
@@ -24,7 +25,6 @@ import io.vertx.core.internal.buffer.BufferInternal;
 import io.vertx.core.internal.logging.Logger;
 import io.vertx.core.internal.logging.LoggerFactory;
 import io.vertx.core.net.NetSocket;
-import io.vertx.core.net.impl.ConnectionBase;
 import io.vertx.core.streams.WriteStream;
 
 import java.util.ArrayList;
@@ -49,10 +49,42 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
   private Handler<StreamPriority> priorityHandler;
 
   // Cache these for performance
-  private MultiMap headers;
+  private final MultiMap headers;
   private MultiMap trailers;
   private List<String> cookies;
   private NetSocket netSocket;
+
+  /**
+   * This {@code ended} field is used internally to track whether the response has ended.
+   * The {@link #completion} promise is used to expose the response result, completed as failed
+   * or succeeded. The promise's {@link Future} is exposed to external users via {@link #end()}.
+   *
+   * <p>The {@code ended} / {@link #completion} pair decouples the "response ended" state handling
+   * from the potentially expensive work that might be incurred by handlers registered on the promise's
+   * {@link Future}.
+   *
+   * <p>The pattern for both fields in {@link #handleException(Throwable)} and {@link #handleTrailers(MultiMap)}
+   * is:
+   * <ol>
+   *   <li>Acquire lock ({@code synchronized (conn)}).
+   *   <li>Check {@code ended}, if {@code ended != null} return immediately.
+   *   <li>Set {@code ended}.
+   *   <li>Release lock.
+   *   <li>Complete the {@link #completion} promise.
+   * </ol>
+   *
+   * <p>Possible states of {@code ended}:
+   * <ul>
+   *   <li>{@code null} - the response has not ended yet</li>
+   *   <li>{@link #ENDED_SENTINEL ENDED_SENTINEL} - the response has ended successfully</li>
+   *   <li>any other {@link Throwable} - the response has ended with an exception</li>
+   * </ul>
+   *
+   * <p>All accesses to this field must be guarded by {@code conn}.
+   */
+  private Throwable ended;
+  private static final Throwable ENDED_SENTINEL = new Throwable();
+  private final Promise<Void> completion;
 
   HttpClientResponseImpl(HttpClientRequestBase request, HttpVersion version, HttpClientStream stream, int statusCode, String statusMessage, MultiMap headers) {
     this.version = version;
@@ -61,9 +93,11 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
     this.request = request;
     this.stream = stream;
     this.conn = stream.connection();
+    this.completion = request.context.promise();
     this.headers = headers;
   }
 
+  // Must be guarded by a `synchronized (conn)` block.
   private HttpEventHandler eventHandler(boolean create) {
     if (eventHandler == null && create) {
       eventHandler = new HttpEventHandler(request.context);
@@ -158,6 +192,10 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
 
   @Override
   public String getTrailer(String trailerName) {
+    MultiMap trailers;
+    synchronized (conn) {
+      trailers = this.trailers;
+    }
     return trailers != null ? trailers.get(trailerName) : null;
   }
 
@@ -175,9 +213,10 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
     }
   }
 
+  /** Must be called within a {@code synchronized (conn)} block. */
   private void checkEnded() {
-    if (trailers != null) {
-      throw new IllegalStateException();
+    if (ended != null) {
+      throw new IllegalStateException("Response already ended");
     }
   }
 
@@ -271,10 +310,24 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
 
   void handleTrailers(MultiMap trailers) {
     HttpEventHandler handler;
+    Throwable wasEnded;
+    // This synchronized block is used to guarantee that a handler's `handleEnd()` is
+    // called only once and that setting/updating `trailers` does not race with
+    // `trailers()`, where the `trailers` field can escape.
     synchronized (conn) {
-      this.trailers = trailers;
+      wasEnded = this.ended;
+      if (wasEnded != null) {
+        return;
+      }
+      if (this.trailers == null) {
+        this.trailers = trailers;
+      } else if (this.trailers != trailers) {
+        this.trailers.setAll(trailers);
+      }
+      this.ended = ENDED_SENTINEL;
       handler = eventHandler;
     }
+    completion.tryComplete();
     if (handler != null) {
       handler.handleEnd();
     }
@@ -282,12 +335,18 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
 
   void handleException(Throwable e) {
     HttpEventHandler handler;
+    Throwable wasEnded;
     synchronized (conn) {
-      if (trailers != null) {
+      // Only report the first exception, and generally only report when the
+      // response has not yet ended.
+      wasEnded = this.ended;
+      if (wasEnded != null) {
         return;
       }
+      this.ended = e;
       handler = eventHandler;
     }
+    completion.tryFail(e);
     if (handler != null) {
       handler.handleException(e);
     } else {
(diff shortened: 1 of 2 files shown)


snazy:
Backport for 5.0 : https://github.com/snazy/vert.x/pull/new/fix-race-6038-5.0
Backport for 4.x: https://github.com/snazy/vert.x/pull/new/fix-race-6038-4.x

vietj:
to simplify, ended should be a Throwable and we should use a specific throwable constant to signal proper ending, e.g. END_SENTINEL

snazy:
@vietj For the END_SENTINEL, you mean to throw that constant/stackless exception from `checkEnded()`?

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -48,11 +47,25 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
   private Handler<HttpFrame> customFrameHandler;
   private Handler<StreamPriority> priorityHandler;
 
+  static final class RequestEndedException extends IllegalStateException {
```
vietj: no need for a specific class here, we can simply use == comparison instead

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -176,8 +189,8 @@ public List<String> cookies() {
   }
 
   private void checkEnded() {
-    if (trailers != null) {
-      throw new IllegalStateException();
+    if (ended == ENDED_SENTINEL) {
+      throw ENDED_SENTINEL;
```
vietj: we should throw a new IllegalStateException each time to give proper stack information and not try to reuse the sentinel

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -283,9 +300,14 @@ void handleTrailers(MultiMap trailers) {
   void handleException(Throwable e) {
     HttpEventHandler handler;
     synchronized (conn) {
-      if (trailers != null) {
+      if (ended == ENDED_SENTINEL) {
```
vietj: I think here we should ignore any exception if ended is not null and consider that receiving any excxeption will not lead to further event delivery

snazy: Yea, I had that before (and I agree that would be the correct behavior), but some tests like `Http1xTest.testInvalidTrailersInHttpClientResponse()` started to fail (timeout), so I reverted the previous behavior. Mean, I can try to adopt the test. WDYT?

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
snazy: okay, updated

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
snazy: fixed

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -297,13 +309,38 @@ void handleException(Throwable e) {
...
+    deliverEndOrException(ended, eventHandler);
+    return future;
   }
 
   @Override
-  public synchronized Future<Void> end() {
-    checkEnded();
-    return eventHandler(true).end();
+  public Future<Void> end() {
+    HttpEventHandler eventHandler;
```
vietj: I think this part should be reimplemented more cleanly with an actual Promise set of the client response that becomes failed/succeeded by events

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
snazy: Pushed 9a3a0ede841f4316e5eb776d17f254f9b3137ff9 with that change and updates to the affected tests.

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
snazy: Hm, yea, a cleaner state handling would be nice. Let me give that a try.

snazy: Implemented via 0bbc30be1493e6260e0d7892f069f3c1d512f2a4

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -61,9 +62,20 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
     this.request = request;
     this.stream = stream;
     this.conn = stream.connection();
+    this.completion = request.context.promise();
+    if (completion.future() != completion) {
```
vietj: in which case would we need that ?

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -61,9 +62,20 @@ public class HttpClientResponseImpl implements HttpClientResponse  {
...
+      // We rely on the implementation of `Promise.future()` to return 'this', as it's implemented
+      // in `io.vertx.core.internal.ContextInternal.promise()`. This saves a separate field holding
+      // a `Future` set to the same object instance as `completion`.
+      throw new IllegalStateException("Implementation expects Promise.future() to return this");
+    }
     this.headers = headers;
   }
 
+  private Future<Void> completionFuture() {
```
vietj: I don't think this is necessary

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -271,8 +287,20 @@ void handleChunk(Buffer data) {
 
   void handleTrailers(MultiMap trailers) {
     HttpEventHandler handler;
+    // This synchronized block is used to guarantee an atomic check-then-act against
+    // the completion-future.
     synchronized (conn) {
-      this.trailers = trailers;
+      if (completionFuture().isComplete()) {
```
vietj: I think instead we should return when we call tryComplete which returns false, in that case it means we didn't complete the future, some other event did and we should not try to make more progress

vietj: like it has been done in handleException

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
snazy: Not strictly necessary, just added it due to the hard implementation-behavior-assumption that the returned `Future` is always `==` the returned `Promise` (in the c'tor).

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
snazy: Just a defensive check, if the implementation changes.
The alternative would have been to always have a field `Future completionFuture` with the same object instance as the `Promise completion`.

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
snazy: If we return too early here, accesses to an escaped `MultiMap` from `trailers()` could miss the the trailers (after observing the completion).

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
vietj: I don't think this is necessary as this always is the case

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
vietj: I think we can still make the update and then check return after state update

snazy: Hm, that could mutate the `trailers` field/content after the response has been ended (failure). You think that's acceptable? (I guess you're worried about the extra `synchronized` cost of `FutureImpl.isComplete()`?)

vietj: yes that is acceptable

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -271,8 +282,22 @@ void handleChunk(Buffer data) {
...
+      // additional `synchronized` cost in `FutureImpl`.
+      //
+      // 'trailers' may have already been initialized in the 'trailers()' function
+      // and escaped to callers.
+      if (this.trailers == null) {
+        this.trailers = trailers;
+      } else if (this.trailers != trailers) {
+        this.trailers.setAll(trailers);
+      }
+      if (!completion.tryComplete()) {
```
vietj: we should call completion.tryComplete after the synchronized block, because this can call handlers in cascade (side effects), synchronized block should only be used to ensure state visibility.

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
```diff
@@ -283,7 +308,9 @@ void handleTrailers(MultiMap trailers) {
   void handleException(Throwable e) {
     HttpEventHandler handler;
     synchronized (conn) {
-      if (trailers != null) {
+      // Only report the first exception, and generally only report when the
+      // response has not yet ended.
+      if (!completion.tryFail(e)) {
```
vietj: move tryFail before the synchronized block to avoid side effects within a synchronized block

snazy: Similar here, moving `tryFail()` before the lock feels risky for the same ordering reasons.

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientResponseImpl.java
snazy: Hm, I’m a bit worried, because moving it out of the lock could introduce races with `body()`/handler setup (missed or duplicate end delivery).

vietj: it is not possible that the HttpClientStream calls HttpClientResponseImpl concurrently, calls are serialized

snazy: Agree on the `HttpClientStream` part, but a call to `body()` could still see "not completed" 

If we move it out of the `synchronized` block:

| Time | Thread 1 (`handleTrailers()`) | Thread 2 (`body()`) |
|---|---|---|
| t0 | State: `eventHandler == null`, `completion` not complete |  |
| t1 | Enter `synchronized(conn)`, set `this.trailers`, snapshots `handler = eventHandler` (`null`), exits lock |  |
| t2 |  | Enters `synchronized(conn)`, creates `eventHandler`, creates `bodyPromise`, reads `completion.isComplete() == false`, exits lock |
| t3 | Calls `completion.tryComplete()` (now becomes complete) |  |
| t4 | `handler` snapshot was `null`, so no `handler.handleEnd()` call | `completed == false` snapshot, so no replay (`handleEnd()` not called) |
| t5 | Returns | Returns `bodyFuture`, but it never gets completed |

So the "end signal" is list and `body()` can hangs (observed timeout).

Replacing line with `event = eventHandler(true)` mitigates that, but there's then still a tiny window where a `checkEnded()` can see a not-ended response and do the "wrong thing".

We could however use a `boolean ended` field instead as the "ended signal/guard" and set that in the `synchronized` block. Then we can safely move the `completion.tryComplete()` out of the `synchronized` block.

WDYT?

snazy: Added 569e17ba5ddd30fba729e7eeff97043e9439b006 with a new field to move the future-completion out of the synchronize-block.

PR #6067, opened 2026-04-16
Improve the resiliency of Vert.x Logger.

vietj:
Motivation:

Vert.x should improve it's logger resiliency when handling faulty loggers, otherwise special care needs to be taken when logging happens in order to avoid breaking invariants, for instance `TaskQueue` log taks failures and assumes that the logging will not fail.

Changes:

Vert.x should redirect logger failures to a special global handler (to avoid ignoring them) and isolate the caller of such issues as part of the `Logger` contract.


catorthosesis (commented):
all looks good , thanks

PR #6069, opened 2026-04-16
Fix Http2 server - compressed and uncompressed responses over same connection

### LINKED ISSUE #5896: Http2 server - compressed and uncompressed responses over same connection
### Version

5.0.6

### Context

Hi, I've encountered the problem with server compression and http2 connection reuse. The problem is in VertxCompressorHttp2ConnectionEncoder

As soon as you send one response that is uncompressed(for instance content-encoding:identity for some small healthcheck endpoint) the next method will switch the encoder from compressed to plain. It will do so only once, so then when you send compressed body, you'll get content-encoding: gzip/zstd but body will be uncompressed cause the codec was already switched to plain by this line `delegate = plainEncoder;`

 ```private void beforeWritingHeaders(ChannelHandlerContext ctx, int streamId, Http2Headers responseHeaders) {
    String contentEncodingToApply = determineContentEncodingToApply(ctx, streamId, responseHeaders);
    if (contentEncodingToApply == null || contentEncodingToApply.equalsIgnoreCase(IDENTITY.toString())) {
      if (responseHeaders.contains(CONTENT_ENCODING, IDENTITY)) {
        responseHeaders.remove(CONTENT_ENCODING);
      }
      delegate = plainEncoder;
    } else {
      responseHeaders.set(CONTENT_ENCODING, contentEncodingToApply);
    }
  }
```

This leads to a client exception like

```
io.netty.handler.codec.compression.DecompressionException: Input is not in the GZIP format
	at io.netty.handler.codec.compression.JdkZlibDecoder.readGZIPHeader(JdkZlibDecoder.java:341)
```

So the response body does not match content-encoding header. Also user should not be forced to use all compressed or all uncompressed requests/response per connection. Described scenario worked properly with vertx 4.*

### Steps to reproduce

1. Create http server with http2 and server compression.  
2. Implement first endpoint - it should return anything and put response header `content-encoding : identity`. 
3. Implement second endpoint-  return anything without additional header to allow compression
4. Create http client with http2 support. Make to sequential requests to endpoint without compression and endpoint with compression. 
5. On the second response you'll receive `content-encoding: *` but response body is not actually compressed which leads to decompression exception on the client

### Do you have a reproducer?

_No response_

### PR DESCRIPTION
**Pull request by zyclonite**
Motivation:

This PR fixes #5896 by allowing multiple concurrent HTTP/2 streams on the same connection to independently select their encoder (compressed vs. plain) based on their own headers, without interfering with each other.
This problem surfaced when different browsers connect via a HTTP/2 enabled proxy that multiplexes multiple connections into one single HTTP/2 connection. If one client supports compression and another not, the server sends the proper headers but the wrong payload, triggering decompression errors.

Changes:

Refactored `VertxCompressorHttp2ConnectionEncoder` to store the selected encoder (compressed or plain) as a property on each `Http2Stream`, rather than using a mutable connection-wide delegate field.

Added `testResponseCompressionEnabledMixedStreams` in `Http2ServerTest` which verifies that each stream independently receives the correct compression behavior.

Results:

Each HTTP/2 stream now independently selects its encoder at header-write time, and subsequent data frames use that same encoder. Concurrent streams with different compression requirements on the same connection are handled correctly. Stream property lifecycle is managed automatically by Netty when streams are closed.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/http2/codec/VertxCompressorHttp2ConnectionEncoder.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/http2/codec/VertxCompressorHttp2ConnectionEncoder.java
@@ -35,23 +35,26 @@ import static io.vertx.core.http.HttpHeaders.IDENTITY;
 
 public class VertxCompressorHttp2ConnectionEncoder implements Http2FrameWriter, Http2ConnectionEncoder, Http2SettingsReceivedConsumer {
 
-  private Http2ConnectionEncoder delegate;
+  private final Http2ConnectionEncoder delegate;
   private final Http2ConnectionEncoder plainEncoder;
+  private final Http2Connection.PropertyKey encoderKey;
 
   public VertxCompressorHttp2ConnectionEncoder(Http2ConnectionEncoder plainEncoder, CompressionOptions[] compressionOptions) {
     this.delegate = new CompressorHttp2ConnectionEncoder(plainEncoder, compressionOptions);
     this.plainEncoder = plainEncoder;
+    this.encoderKey = plainEncoder.connection().newKey();
   }
 
-  private void beforeWritingHeaders(ChannelHandlerContext ctx, int streamId, Http2Headers responseHeaders) {
+  private Http2ConnectionEncoder selectEncoder(ChannelHandlerContext ctx, int streamId, Http2Headers responseHeaders) {
     String contentEncodingToApply = determineContentEncodingToApply(ctx, streamId, responseHeaders);
     if (contentEncodingToApply == null || contentEncodingToApply.equalsIgnoreCase(IDENTITY.toString())) {
       if (responseHeaders.contains(CONTENT_ENCODING, IDENTITY)) {
         responseHeaders.remove(CONTENT_ENCODING);
       }
-      delegate = plainEncoder;
+      return plainEncoder;
     } else {
       responseHeaders.set(CONTENT_ENCODING, contentEncodingToApply);
+      return delegate;
     }
   }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by vietj** [comment]
rebased since I've done fixes to improve test passing more consistently

**Comment by zyclonite** [comment]
thank you, with a small assertion fix it was successful

**Comment by zyclonite** [comment]
@vietj based on the last commit (https://github.com/eclipse-vertx/vert.x/pull/6069/commits/4736fbfdd20342bd9696f88af2472546e4d44e4f) the fix looks way better now

what do you think?

File: vertx-core/src/main/java/io/vertx/core/http/impl/http2/codec/VertxCompressorHttp2ConnectionEncoder.java [hunk 1]
```diff
@@ -35,23 +35,26 @@
 
 public class VertxCompressorHttp2ConnectionEncoder implements Http2FrameWriter, Http2ConnectionEncoder, Http2SettingsReceivedConsumer {
 
-  private Http2ConnectionEncoder delegate;
+  private final Http2ConnectionEncoder compressingEncoder;
```
vietj: can you avoid renaming for the sake of clarify of the PR ?

vietj:
can you make a single commit and explain what has changed and how it solves the issue ?

basically I'm looking for a commit

Motivation:

Changes:

Results:

that helps reviewing :-)

zyclonite:
fixed the PR based on your input

vietj:
excellent @zyclonite

vietj:
can you backport this PR @zyclonite ?

zyclonite:
sure, onto the 5.0 branch - in 4.x this was not an issue

zyclonite:
@vietj that's the backport PR #6080

PR #6080, opened 2026-04-21
Move http2 compression handling from connection to stream level (5.0 branch)

zyclonite:
backport on 5.0 branch from #6069

Motivation:

This PR fixes https://github.com/eclipse-vertx/vert.x/issues/5896 by allowing multiple concurrent HTTP/2 streams on the same connection to independently select their encoder (compressed vs. plain) based on their own headers, without interfering with each other.
This problem surfaced when different browsers connect via a HTTP/2 enabled proxy that multiplexes multiple connections into one single HTTP/2 connection. If one client supports compression and another not, the server sends the proper headers but the wrong payload, triggering decompression errors.

Changes:

Refactored VertxCompressorHttp2ConnectionEncoder to store the selected encoder (compressed or plain) as a property on each Http2Stream, rather than using a mutable connection-wide delegate field.

Added testResponseCompressionEnabledMixedStreams in Http2ServerTest which verifies that each stream independently receives the correct compression behavior.

Results:

Each HTTP/2 stream now independently selects its encoder at header-write time, and subsequent data frames use that same encoder. Concurrent streams with different compression requirements on the same connection are handled correctly. Stream property lifecycle is managed automatically by Netty when streams are closed.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/VertxCompressorHttp2ConnectionEncoder.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/VertxCompressorHttp2ConnectionEncoder.java
@@ -33,24 +33,45 @@ import static io.vertx.core.http.HttpHeaders.IDENTITY;
 
 public class VertxCompressorHttp2ConnectionEncoder implements Http2FrameWriter, Http2ConnectionEncoder, Http2SettingsReceivedConsumer {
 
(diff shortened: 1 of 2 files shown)


vietj:
can you update comment message with the full detail so it appears directly in the issue ?

zyclonite:
sure, done

vietj:
perfect, thanks @zyclonite

Issue #6021, opened 2026-03-16
EventBus: tracer.sendResponse() called prematurely for fire-and-forget messages dispatched to blocking handlers


zhfeng:
## Problem

In `InboundDeliveryContext.execute()`, for fire-and-forget EventBus messages (send/publish with no `replyAddress`), `tracer.sendResponse()` is called synchronously immediately after `dispatch()` returns:

```java
// InboundDeliveryContext.execute() pseudocode:
message.trace = tracer.receiveRequest(ctx, ...);  // 1. sets up trace context
dispatch(message, ctx, handler);                    // 2. for blocking handlers, submits to worker pool and returns immediately
Object trace = message.trace;
if (replyAddress == null && trace != null) {
    tracer.sendResponse(ctx, null, trace, ...);     // 3. closes trace context BEFORE worker thread runs
}
```

When the handler is a **blocking handler** (e.g., Quarkus `@ConsumeEvent(blocking = true)`), `dispatch()` submits the work to a worker thread pool and returns immediately. The `sendResponse()` call at step 3 then removes the tracing context (e.g., OpenTelemetry span/scope) from the Vert.x context locals **before the worker thread has a chance to execute the handler**.

This causes the trace context to be lost inside the blocking handler — for example, `io.opentelemetry.context.Context.current()` returns the root context instead of the expected trace context, breaking distributed tracing.

## Expected Behavior

For blocking/async dispatch, `tracer.sendResponse()` should be deferred until after the handler has completed execution on the worker thread, so that the trace context remains available throughout the handler's lifecycle.

## Current Workaround

In [quarkusio/quarkus#53056](https://github.com/quarkusio/quarkus/pull/53056), we work around this by using reflection to null out `MessageImpl.trace` before `dispatch()` returns, preventing `InboundDeliveryContext.execute()` from calling `sendResponse()`. We then call `sendResponse()` ourselves in the worker thread's `finally` block after the handler completes.

This works but is fragile since it depends on the internal `MessageImpl.trace` field.

## Proposed Solution

`InboundDeliveryContext.execute()` could be made aware of whether the dispatch is synchronous or asynchronous. For asynchronous dispatch (blocking handlers, virtual threads), `sendResponse()` should be deferred — either via:

1. A callback/completion mechanism from the dispatch (e.g., using the `Future` returned by `executeBlocking`)
2. Making `dispatch()` responsible for calling `sendResponse()` when it knows the handler has completed
[...]

PR #6083, opened 2026-04-23
Fix premature tracer span closure for fire-and-forget and publish

tsegismont:
See #6021

Add `MessageConsumerOptions#setAutoProcessed` and `Message#processed()` so consumers can defer trace span closure until async/blocking handler completion.

### DIFF
--- a/vertx-core/src/main/asciidoc/eventbus.adoc
+++ b/vertx-core/src/main/asciidoc/eventbus.adoc
@@ -240,6 +240,25 @@ persisted in storage, or `false` if not.
 * A message consumer which processes an order might acknowledge with `true` when the order has been successfully processed
 so it can be deleted from the database
 
+==== Manual Ack Mode
+
+By default, Vert.x automatically acks messages when the message handler returns.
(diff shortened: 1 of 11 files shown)


vietj:
I'd rather name processed by ack which is a common terminology, we also used this for acking MQTT messages

tsegismont:
Done

vertx-core/src/main/asciidoc/eventbus.adoc
```diff
@@ -240,6 +240,25 @@ persisted in storage, or `false` if not.
 * A message consumer which processes an order might acknowledge with `true` when the order has been successfully processed
 so it can be deleted from the database
 
+==== Manual Ack Mode
+
+By default, Vert.x automatically acks messages when the message handler returns.
```
vietj: I think we need to explain what means "ack" in this context as it is overloaded.

vietj: we should also mention that it is only relevant to tracing and metrics, and it is not related to producer interaction.

tsegismont: Done

tsegismont:
Superseded by https://github.com/eclipse-vertx/vert.x/pull/6136

PR #6085, opened 2026-04-23
Reusing dispatch method in the emit

### LINKED ISSUE #6086: Probably dispatch method should be reused inside emit
### Describe the feature

Currently **emit** method looks like
```
  @Override
  public final <T> void emit(T argument, Handler<T> task) {
    if (executor().inThread()) {
      ContextInternal prev = beginDispatch();
      try {
        task.handle(argument);
      } catch (Throwable t) {
        reportException(t);
      } finally {
        endDispatch(prev);
      }
    } else {
      executor().execute(() -> emit(argument, task));
    }
  }
```

while dispatch is
```
  default <E> void dispatch(E event, Handler<E> handler) {
    ContextInternal prev = beginDispatch();
    try {
      handler.handle(event);
    } catch (Throwable t) {
      reportException(t);
    } finally {
      endDispatch(prev);
    }
  }

```

so it is effective duplicate logic of **dispatch** in the   _"if (executor().inThread())"_ branch.

### Contribution

here is PR https://github.com/eclipse-vertx/vert.x/pull/6085

doxlik:
Motivation:

Reuse **dispatch** method in the **emit** to consolidate logic.

Closes https://github.com/eclipse-vertx/vert.x/issues/6086

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/ContextBase.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/ContextBase.java
@@ -77,14 +77,7 @@ abstract class ContextBase implements ContextInternal {
   @Override
   public final <T> void emit(T argument, Handler<T> task) {
     if (executor().inThread()) {
-      ContextInternal prev = beginDispatch();
-      try {
-        task.handle(argument);
-      } catch (Throwable t) {
-        reportException(t);
-      } finally {
-        endDispatch(prev);
-      }
+      dispatch(argument, task);
     } else {
       executor().execute(() -> emit(argument, task));
     }


doxlik:
@vietj hi, it is "suggestion" PR to a bit consolidate logic, in case if you prefer to have explicit separated logic of emit and dispatch even if part of emit now seems to be same as dispatch - just feel free to close this issue / PR as not planned.

vietj:
@doxlik I think I did that originally because I had inlining issues when using dispatch directly

vietj:
that would need to be checked again

doxlik:
@vietj very important point, I fully agree. If this kind of refactoring can introduce overhead on a hot path because the JIT does not inline dispatch, then it is definitely better to keep the repeated code as it is now. I really appreciate this kind of performance-minded decision in such hot paths.

vietj:
@doxlik thanks for proposing this change.

PR #6088, opened 2026-04-28
Fix multipart decoder cleanup on abnormal termination

### LINKED ISSUE #6014: Multipart decoder retention on abort/reset/exception paths
### Version

5.0.x

### Context

While looking into multipart abort/reset handling, I noticed that multipart decoders can remain retained on exception/reset/close paths instead of being cleaned up like on normal completion.

I reproduced this by sending incomplete `multipart/form-data` requests and closing the connection before the final boundary. In my PoC, terminal decoders remained retained for more than 30 seconds after traffic stopped.

The previous discussion and PoC details are here:
- https://gitlab.eclipse.org/security/vulnerability-reports/-/issues/370

### Steps to reproduce

1. Run a simple Vert.x HTTP server with a `POST /upload` endpoint that calls `setExpectMultipart(true)`.
2. Send many incomplete `multipart/form-data` requests and close the connection before the final boundary.
3. Observe that decoder instances can remain retained on terminal paths for an extended time.

### Do you have a reproducer?

_No response_

shblue21:
Motivation:

Multipart decoding should release the Netty multipart decoder when a request terminates abnormally. The normal completion path already cleans it up, but abort/reset/close paths could leave decoder state attached longer than intended.

This updates the abnormal termination paths to clean up the multipart decoder for HTTP/1 and the shared HTTP/2/HTTP/3 request implementation. It also adds regression coverage for incomplete multipart uploads aborted by the client.

Fixes #6014

Conformance:

Confirmed.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
@@ -567,8 +567,7 @@ public class Http1ServerRequest extends HttpServerRequestInternal implements io.
         } catch (HttpPostRequestDecoder.ErrorDataDecoderException |
                  HttpPostRequestDecoder.TooLongFormFieldException |
                  HttpPostRequestDecoder.TooManyFormFieldsException e) {
-          decoder.destroy();
-          decoder = null;
+          cleanupDecoder();
           handleException(e);
         }
       }
@@ -628,6 +627,7 @@ public class Http1ServerRequest extends HttpServerRequestInternal implements io.
           } catch (Exception e) {
             // Will never happen, anyway handle it somehow just in case
             handleException(e);
+            return;
           } finally {
             attr.release();
           }
@@ -640,8 +640,7 @@ public class Http1ServerRequest extends HttpServerRequestInternal implements io.
     }  catch (HttpPostRequestDecoder.EndOfDataDecoderException e) {
       // ignore this as it is expected
     } finally {
-      decoder.destroy();
-      decoder = null;
+      cleanupDecoder();
     }
   }
 
@@ -649,11 +648,13 @@ public class Http1ServerRequest extends HttpServerRequestInternal implements io.
     HttpEventHandler handler = null;
     Http1ServerResponse resp = null;
     InterfaceHttpData upload = null;
+    HttpPostRequestDecoder decoderToCleanup = null;
     synchronized (conn) {
       if (!isEnded()) {
         handler = eventHandler;
         if (decoder != null) {
           upload = decoder.currentPartialHttpData();
+          decoderToCleanup = decoder;
         }
       }
       if (!response.ended()) {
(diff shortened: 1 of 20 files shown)


vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
```diff
@@ -628,6 +627,7 @@ private void endDecode() {
           } catch (Exception e) {
             // Will never happen, anyway handle it somehow just in case
             handleException(e);
+            return;
```
vietj: I don't think it is intended, the finally clause will be executed anyway

vertx-core/src/test/java/io/vertx/tests/http/fileupload/HttpServerFileUploadTest.java
```diff
@@ -56,6 +57,58 @@ public void setUp() throws Exception {
     testDir = testFolder.newFolder();
   }
 
+  protected void testClientResetMultipartUploadCleansDecoder() throws Exception {
```
vietj: make it a test and instead use Assume on the HTTP version to make it pass only for the intended HTTP versions

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
```diff
@@ -663,14 +664,37 @@ void handleException(Throwable t) {
...
+  }
+
+  private void cleanupDecoder() {
+    HttpPostRequestDecoder decoderToCleanup;
+    synchronized (conn) {
+      decoderToCleanup = decoder;
     }
-    if (handler != null) {
-      handler.handleException(t);
+    cleanupDecoder(decoderToCleanup);
```
vietj, line 692: just call decoderToCleanup.destroy() and make the decoder field null

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
```diff
@@ -640,20 +640,21 @@ private void endDecode() {
...
     HttpEventHandler handler = null;
     Http1ServerResponse resp = null;
     InterfaceHttpData upload = null;
+    HttpPostRequestDecoder decoderToCleanup = null;
     synchronized (conn) {
       if (!isEnded()) {
         handler = eventHandler;
         if (decoder != null) {
           upload = decoder.currentPartialHttpData();
+          decoderToCleanup = decoder;
```
vietj, line 660: make the decoder null here instead of trying to make it null the cleanupDecoder

vietj:
I think overall for HTTP/1.x we should try to have a single method guaranteed to be called to centralize the cleanup logic of the request before doing this

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
shblue21: Still checking this. The return was meant to avoid continuing `decoder.hasNext()` after `handleException()` may clear the decoder, but I’ll verify the cleanup order.

Thank you.

shblue21:
I kept this focused on the multipart decoder cleanup paths. A broader HTTP/1.x request cleanup refactor can be handled separately.

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
shblue21: Removed the return. handleException is now called after the attribute is released.

vertx-core/src/test/java/io/vertx/tests/http/fileupload/HttpServerFileUploadTest.java
shblue21: The reset cleanup test now lives in the base file upload test and is gated by HTTP version with Assume.

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
shblue21: cleanupDecoder(HttpPostRequestDecoder) now only destroys the captured decoder.

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerRequest.java
shblue21: The decoder is now captured and nulled under the connection lock before callbacks run.

PR #5970, opened 2026-02-12
EPoll: Respect extended TCP Keepalive parameters for clients, not just servers

ivu-mawi:
I want to use a lower TCP Keepalive interval as a client. Currently this does not work. It does work for servers using the EPoll transport in Vert.x.

This PR makes it work for clients using EPoll.

I noticed that even the code for servers seems to be absent in master / 5.x, maybe worth a "forward-port"?

I have not yet signed the Eclipse Contributor Agreement. I can do that if necessary for this small change.

PR #5183, opened 2024-04-10
Feature 5163 configurable socket keepalive interval

aksabg:
Feature according to: [github.com/eclipse-vertx/vert.x/issues/5163](github.com/eclipse-vertx/vert.x/issues/5163)

Issue #5163, opened 2024-03-22
Configurable socket keep alive interval


aksabg:
`NetServer` now supports `tcpKeepAlive` flag which is translated to Netty's `SO_KEEPALIVE` flag. However, the keepalive interval, idle time and count cannot be set through settings, which means it defaults to OS values (usually the idle time is 2 hours).

When using `NetServer` in conjuction with the AWS load balancer, the connections are closed when idle for more than 350 seconds.

The following Netty fields should be supported in `NetServerOptions`:

1. TCP_KEEPIDLE
2. TCP_KEEPINTVL
3. TCP_KEEPCNT

Reference: https://github.com/reactor/reactor-netty/issues/2167

aksabg:
@vietj I can add this feature, just need the permission to branch.

vietj:
@aksabg please contribute it via a pull request

aksabg:
@vietj I don't have permissions to create a remote branch required for a pull request.

jakub-id:
@vietj any chance to get this merged? This is a PITA to configure externally on Kubernetes < 1.29 where configuring these defaults via sysctl is considered unsafe.

vietj:
@jakub-id the issue is that the PR does not have an eclipse agreement signed

jakub-id:
@vietj I saw that but the Eclipse report is weird as her email is mentioned once in the green box and three times in the red box. Unless it's a different email but it doesn't look like it.

aksabg:
@vietj @jakub-id I have signed the agreement (https://accounts.eclipse.org/users/vealrad) 
Is there anything else I need to do?

jakub-id:
@aksabg Great! Can you remove the JavaDoc formatting changes and empty new lines as @vietj asked on the PR?

vietj:
where is the PR ?

aksabg:
@vietj 
https://github.com/eclipse-vertx/vert.x/pull/5183

PR #6089, opened 2026-04-29
Support extended keepalive parameters in epoll and io_uring

ivu-mawi:
EpollTransport: support extended keepalive parameters TCP_KEEPIDLE, TCP_KEEPCNT and TCP_KEEPINTVL

Builds on #5183. This work has not made it into master, so I'm catching up on that.

I hope I've ported it correctly given the refactoring of the TCPSSLOptions / TCPConfig.

For motivation, see #5163 and #5183 and #5970

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
@@ -14,6 +14,7 @@ package io.vertx.core.impl.transports;
 import io.netty.bootstrap.Bootstrap;
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.channel.*;
+import io.netty.channel.epoll.EpollChannelOption;
 import io.netty.channel.socket.DatagramChannel;
 import io.netty.channel.socket.InternetProtocolFamily;
 import io.netty.channel.unix.DomainSocketAddress;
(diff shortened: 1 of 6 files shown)


vietj:
I am wondering for the new config to potentially have an Option object to be more extensible than what we have now.

vietj:
@ivu-mawi I have introduced generic TCP option, can you rebase on master

vietj:
that only affects the new config

ivu-mawi:
Seems good! I'll gladly rebase when I get the chance, which will be next week only I'm afraid.

tsegismont:
No problem @ivu-mawi . Thank you!

vietj:
why is that PR still in draft mode ?

ivu-mawi:
I'm trying to verify locally that it actually works. But I can't seem to get packet capturing to work in WSL (Windows subsystem for linux), and that's the only Linux I have access to at the moment. :(

I'll give up the test. Code is ready, I'll mark it ready.

vietj:
@ivu-mawi an you rebase on latest branch to resolve the git conflict ?

ivu-mawi:
Fixed unit tests by disallowing `0` as value again

ivu-mawi:
One thing I'm not sure about with the new TcpOption is we're setting various parameters to `-1` explicitly now, even though the intention in some of the code seems to have been that `-1` means "don't set the option on the actual socket, leave whatever the OS default is". I couldn't find any docs as to whether setting these options to `-1` explicitly is allowed or breaks something.

ivu-mawi:
Oh, it looks like exactly that is actually causing an error in the pipeline runs.

ivu-mawi:
We can see [here](https://github.com/torvalds/linux/blob/27fa82620cbaa89a7fc11ac3057701d598813e87/net/ipv4/tcp.c#L3776) that Linux does not allow any value < 1.

I have made it so that `0` instead of `-1` is now the special value that means "do not set, leave OS-default". This is technically incompatible to the code from vert.x 4.x, but I'd say that's okay. Probably very few people will explicitly use `-1`, and if they do, they will get a runtime exception after upgrading to vert.x 5, and have to change it to `0`.

I've done that just because it's nicer to have a `isNullOrZero` method than a `isNullOrMinusOne` method ...

vertx-core/src/main/java/io/vertx/core/net/TcpConfig.java
```diff
@@ -210,6 +211,17 @@ public <T> T getOption(TcpOption<T> option) {
     return options != null ? option.type.cast(options.get(option)) : null;
   }
 
+  /**
+   * Returns true if the given option is not set or set to {@code 0}.
+   * @param option a tcp option
+   * @return true if the given option is not set or set to {@code 0}.
+   * @param <T> the value type of the option, must be a {@link Number}.
+   */
+  public <T extends Number> boolean isNullOrZero(TcpOption<T> option) {
```
vietj, line 222: can you move this method in the transport package as a package private as it is not useful on this class

vertx-core/src/main/java/io/vertx/core/impl/transports/IoUringTransport.java
```diff
@@ -14,6 +14,7 @@
 import io.netty.bootstrap.Bootstrap;
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.channel.*;
+import io.netty.channel.epoll.EpollChannelOption;
```
vietj: Remove EpollChannelOption import

ivu-mawi:
Thanks!

Issue #6098, opened 2026-05-04
lookup4 can reuse an IPv6-only resolver after lookup6


shblue21:
### Version

5.x

### Context

`DnsClientImpl` keeps separate resolver caches for `lookup()`, `lookup4()`, and `lookup6()`.

Resolver lookup uses the requested address type, but newly created resolvers are always stored in the IPv4 resolver cache.

As a result, `lookup6()` can create an IPv6-only resolver and store it in the IPv4 cache for the event loop. A later `lookup4()` call on the same event loop may then reuse that IPv6-only resolver instead of an IPv4-only resolver.

`lookup4()` should always use an IPv4-only resolver, even if `lookup6()` was called first.

### Steps to reproduce

1. Configure a `DnsClient` with a DNS server that returns both A and AAAA records for the same name.
2. Call `lookup6(name)`.
3. From the same Vert.x context, call `lookup4(name)`.

`lookup4()` may reuse the resolver created by `lookup6()` and return the IPv6 result instead of the A record.

PR #6099, opened 2026-05-04
Fix DNS resolver cache selection

shblue21:
Motivation:

#6098

`DnsClientImpl` uses separate resolver caches for the default resolver, IPv4-only lookups, and IPv6-only lookups.

Newly created resolvers were always stored in the IPv4 resolver cache. This meant that after `lookup6()` ran on an event loop, a later `lookup4()` on the same event loop could reuse an IPv6-only resolver.

The fix stores each resolver in the cache selected for that lookup. It also adds a regression test for `lookup6()` followed by `lookup4()`.

Conformance:

Confirmed.

### DIFF
--- a/vertx-core/src/test/java/io/vertx/tests/dns/DNSTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/dns/DNSTest.java
@@ -12,6 +12,7 @@
 package io.vertx.tests.dns;
 
 import io.netty.handler.codec.dns.DnsMessage;
+import io.netty.handler.codec.dns.DnsRecordType;
 import io.netty.resolver.dns.DnsNameResolverTimeoutException;
 import io.vertx.core.Vertx;
 import io.vertx.core.VertxException;
@@ -28,6 +29,7 @@ import org.junit.Test;
 
 import java.net.InetSocketAddress;
 import java.util.Collections;
+import java.util.List;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


vertx-core/src/test/java/io/vertx/tests/dns/DNSTest.java
```diff
@@ -356,6 +357,34 @@ public void testLookup6() throws Exception {
...
+      if (questionRecord.type() == DnsRecordType.A) {
+        return Collections.singleton(MockDnsServer.a("vertx.io", 100, ipv4));
+      }
+      return Collections.emptySet();
+    });
+    DnsClient dns = prepareDns();
+
+    dns
+      .lookup6("vertx.io")
+      .onComplete(TestUtils.onSuccess(ipv6 -> {
```
vietj: rather use await on Future twice to make the test more linear

then no final await() is needed

vertx-core/src/test/java/io/vertx/tests/dns/DNSTest.java
```diff
@@ -356,6 +357,27 @@ public void testLookup6() throws Exception {
     await();
   }
 
+  @Test
+  public void testLookup6ThenLookup4() throws Exception {
+    final String ipv4 = "10.0.0.1";
+    mockDnsServer.store(questionRecord -> {
+      if (questionRecord.type() == DnsRecordType.AAAA) {
+        return Collections.singleton(MockDnsServer.aaaa("vertx.io", 100, "::1"));
```
vietj: minor : use List.of instead

vietj:
@shblue21 can you provide a backport for the 5.0 branch ?

Issue #5839, opened 2025-12-09
WebSocket Client Connection Evicted/Shutdown During Handshake in Vert.x 5.0.5


yogurt-c:
### Version

5.0.5

### 🎯 Version

* **Vert.x Version:** 5.0.5
* **Operating System:** macOS (Apple Silicon - M1)
* **Java:** 21

Java Version: OpenJDK 21

### 📝 Description

I'm encountering an issue where the **WebSocket client connection is being unexpectedly shut down (`ShutdownEvent` / `Connection evicted`) immediately after sending the WebSocket handshake request, and before receiving the server's handshake response. This issue occurs **intermittently (occasionally)**, not on every attempt, but when it does, it prevents the WebSocket connection from being established. The client is connecting to a locally developed Vert.x WebSocket server.

### ✅ Expected Behavior

The WebSocket client should wait for the server's handshake response. If the response is valid, the connection should be upgraded to a WebSocket connection and stay active. The connection should not be shut down or evicted during the handshake phase unless explicitly requested or due to a network error/timeout.

### ❌ Actual Behavior

**Intermittently**, the connection is closed/evicted during the handshake phase, specifically after the handshake request is sent and before a response is received.

The log shows a `USER_EVENT: io.vertx.core.net.impl.ShutdownEvent` which leads to a `Connection evicted` warning and subsequent connection closure, as shown below:

```log
... (Handshake Request WRITE/FLUSH occurs before this)
2025-12-05 16:25:21 DEBUG [vert.x-eventloop-thread-2] i.n.handler.logging.LoggingHandler - [id: 0xe82a509f, L:/127.0.0.1:55615 - R:localhost/127.0.0.1:47136] ACTIVE
2025-12-05 16:25:21 DEBUG [vert.x-eventloop-thread-2] i.n.handler.logging.LoggingHandler - [id: 0xbe1d97b4, L:/127.0.0.1:55615 - R:localhost/127.0.0.1:47136] USER_EVENT: io.vertx.core.net.impl.ShutdownEvent@4909845e
2025-12-05 16:25:21 WARN  [vert.x-eventloop-thread-2] i.v.c.h.i.HttpClientConnectionInternal - Connection evicted
2025-12-05 16:25:21 DEBUG [vert.x-eventloop-thread-2] i.n.handler.logging.LoggingHandler - [id: 0xbe1d97b4, L:/127.0.0.1:55615 - R:localhost/127.0.0.1:47136] WRITE: 0B
[...]

PR #6110, opened 2026-05-08
Keep WebSocket client reachable during connect

shblue21:
Motivation:

Addresses #5839 for direct `WebSocketClient.connect(...)`.

When `WebSocketClient.connect(options)` is used directly, the returned future can
outlive the caller's reference to the client. Since `CleanableWebSocketClient`
runs cleanup once it becomes unreachable, GC can close the underlying client
before the connect future completes.

Keep the cleanable client reachable until the connect attempt completes.

Conformance:

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/CleanableWebSocketClient.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/CleanableWebSocketClient.java
@@ -19,6 +19,7 @@ import io.vertx.core.spi.metrics.Metrics;
 import io.vertx.core.spi.metrics.MetricsProvider;
 
 import java.lang.ref.Cleaner;
+import java.lang.ref.Reference;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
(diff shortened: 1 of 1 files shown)


vietj:
does that happen because the no reference is kept on the WebSocket client ?

shblue21:
Yes, that seems to be the case.

Thanks for clarifying. I agree that the `WebSocketClient` should be kept referenced, so I closed this draft PR.

Meghajnct:
I am facing this issue again. We are using the vertx version 5.1.5. and we have client referred. 
this.referencedClient = vertx().createWebSocketClient(clientOptions);
                               Cleaner.create()
                                      .register(referencedClient, () -> log.info("WebSocketClient garbage collected!"));
                               WebSocketConnectOptions connectOptions = new WebSocketConnectOptions()
                                       .setAbsoluteURI(absoluteURI)
                                       .addHeader("Host", "127.0.0.1:9094")
                                       .addHeader("Origin", "http://127.0.0.1:9094");

                               //
                               referencedClient.connect(connectOptions)
                                               .onSuccess(ws -> {
                                                   log.info("websocket connected .. ");
                                                   ws.textMessageHandler(message ->
                                                           receivedMessages.get(smId).add(message));
                                                   clientWebsockets.add(ws);
                                               })
                                               .onFailure(err -> log.error("Got Error ", err))
                                               .toCompletionStage()
                                               .toCompletableFuture()
                                               .get(1, TimeUnit.MINUTES);

shblue21:
@Meghajnct Is this assignment repeated for each smId? If so, the field only keeps the last client. Also, is get() running on a Vert.x event-loop thread?

Issue #6111, opened 2026-05-13
HybridJacksonPool can lose pooled recyclers due to stack node publication race


dreamlike-ocean:
### Version

every version

### Context

`HybridJacksonPool.StripedLockFreePool#releasePooled` publishes a new stack head with CAS before initializing the node's `next` pointer.

current:

  ```java
  if (topStacks.compareAndSet(slot, next, newHead)) {
    newHead.next = next;
    return;
  }
```
A concurrent acquirePooled can observe newHead immediately after the CAS, before newHead.next is assigned. In that window it may pop newHead and replace the stack head with null, detaching the previous stack.

I will submit a pr to fix it

tsegismont:
Fixed by https://github.com/eclipse-vertx/vert.x/pull/6114

tsegismont:
Thank you @dreamlike-ocean !

PR #6112, opened 2026-05-13
Fix HybridJacksonPool stack node publication race

dreamlike-ocean:
Motivation:

Fixes a race in `HybridJacksonPool.StripedLockFreePool#releasePooled`. #6111 

The virtual-thread pool was publishing a newly created stack node with CAS before assigning its `next` pointer. A concurrent acquire could observe and pop that partially initialized node, see `next == null`, and drop the rest of the stack.
This change initializes `newHead.next` before the CAS, so the node is fully linked before it becomes visible to other threads.

Conformance:

I have signed the Eclipse Contributor Agreemen


tsegismont (approved):
LGTM, thank you @dreamlike-ocean !

PR #6116, opened 2026-05-18
Issue 5676: Introduced useTcpForFallbackDnsResolving

cgm-aw:
See https://github.com/eclipse-vertx/vert.x/issues/5676

### DIFF
--- a/vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
+++ b/vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
@@ -11,17 +11,18 @@
 package io.vertx.test.fakedns;
 
 import io.netty.bootstrap.Bootstrap;
+import io.netty.bootstrap.ServerBootstrap;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.*;
 import io.netty.channel.nio.NioIoHandler;
 import io.netty.channel.socket.DatagramChannel;
+import io.netty.channel.socket.SocketChannel;
 import io.netty.channel.socket.nio.NioDatagramChannel;
+import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.handler.codec.dns.*;
 import io.netty.handler.codec.dns.DnsRecord;
 import io.netty.util.internal.PlatformDependent;
-import io.vertx.core.Vertx;
-import io.vertx.core.internal.VertxInternal;
 
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
@@ -42,21 +43,15 @@ public class MockDnsServer {
   private String ipAddress = IP_ADDRESS;
   private int port = PORT;
 
-  private EventLoopGroup eventLoop;
-  private ChannelFactory<? extends Channel> channelFactory;
-  private Channel channel;
+  private final EventLoopGroup eventLoop = new MultiThreadIoEventLoopGroup(1, NioIoHandler.newFactory());
+  private final ChannelFactory<? extends Channel> channelFactory = NioDatagramChannel::new;
+  private Channel udpChannel;
+  private Channel tcpChannel;
   private RecordStore store;
-  private List<Throwable> failures = new CopyOnWriteArrayList<>();
+  private final List<Throwable> failures = new CopyOnWriteArrayList<>();
   private final Deque<DnsMessage> currentMessage = new ArrayDeque<>();
-
-  public MockDnsServer(Vertx vertx) {
-    this.eventLoop = ((VertxInternal) vertx).nettyEventLoopGroup().next();
-    this.channelFactory = ((VertxInternal) vertx).transport().datagramChannelFactory();
-  }
-
-  public MockDnsServer() {
-    this.eventLoop = null;
-  }
+  private boolean enableTcp = false;
+  private boolean enableUdp = true;
 
   public MockDnsServer store(RecordStore store) {
     this.store = store;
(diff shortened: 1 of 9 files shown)


vietj:
we should have a test for this that checks TCP is actually used

cgm-aw:
I added a test - I'm not really happy with it, this is Netty code it's testing, but it should do the job.
By the way: Building master from scratch is not not possible without locally building vertx-codegen and vertx-dependencies to have the snapshots available. If you want I add an update to the BUILDING.md.

vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
```diff
@@ -47,9 +48,12 @@ public class MockDnsServer {
   private EventLoopGroup eventLoop;
   private ChannelFactory<? extends Channel> channelFactory;
   private Channel channel;
+  private Channel tcpChannel;
   private RecordStore store;
   private List<Throwable> failures = new CopyOnWriteArrayList<>();
   private final Deque<DnsMessage> currentMessage = new ArrayDeque<>();
+  private boolean tcpEnabled;
+  private boolean failOnUdp;
```
vietj: I think instead this should be a enableUdp boolean like TCP, because anyway there the protocol is datagram oriented so the test will work without an UDP listener

vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
```diff
@@ -47,9 +48,12 @@ public class MockDnsServer {
   private EventLoopGroup eventLoop;
   private ChannelFactory<? extends Channel> channelFactory;
   private Channel channel;
+  private Channel tcpChannel;
   private RecordStore store;
   private List<Throwable> failures = new CopyOnWriteArrayList<>();
   private final Deque<DnsMessage> currentMessage = new ArrayDeque<>();
+  private boolean tcpEnabled;
```
vietj: rename enableTcp

cgm-aw: Done

vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
cgm-aw: Done

vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
```diff
@@ -99,6 +113,9 @@ public void start() {
       .handler(new ChannelInitializer<DatagramChannel>() {
         @Override
         protected void initChannel(DatagramChannel ch) throws Exception {
+          if (!enableUdp) {
```
vietj: I think here we should refactor to avoid binding the UDP channel instead of checking the init, like you have done for startTcp

cgm-aw: That's a bit more tricky, look into the newest commit for the proposed change

vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
```diff
@@ -156,9 +184,63 @@ protected void channelRead0(ChannelHandlerContext ctx, DatagramDnsQuery msg) thr
...
+    }
+  }
+
   public List<Throwable> stop() throws Exception {
-    Channel channel = this.channel;
-    this.channel = null;
+    Channel tc = this.tcpChannel;
+    this.tcpChannel = null;
+    if (tc != null) {
+      tc.close().sync();
```
vietj: I think this should be treated like the UDP channel close that uses shutdownGracefully

vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
```diff
@@ -169,6 +251,8 @@ public List<Throwable> stop() throws Exception {
             .shutdownGracefully(0, 10, TimeUnit.SECONDS);
         }
       }
+    } else if (tc != null && this.eventLoop == null) {
+      tc.eventLoop().parent().shutdownGracefully(0, 10, TimeUnit.SECONDS);
```
vietj, line 225: to make this simpler, I think we should modify the server to always create an event loop for the test using the NIO transport, which would simplify the close of the server. Can you take care of this ?

vietj: I am thinking about decoupling the mock DNS server from vertx and make it generally available as a test facility.

vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
cgm-aw: done

vertx-core/src/test/java/io/vertx/test/fakedns/MockDnsServer.java
cgm-aw: I removed the dependency to the Vertx event loop, it was pretty straight forward because the MockDnsServer already had a fallback solution

vietj: thanks, that's great.

vietj:
@cgm-aw can you address the confict on MockDnsServer with a rebase ? can you squash your contribution too ?

cgm-aw:
@vietj sure done

PR #6126, opened 2026-05-23
feat: support custom proxy authorization header

### LINKED ISSUE #5129: WebClient through kerberos secured proxy is not working
### Questions

I am trying to access a website using the vert.x WebClient but it seems that the WebClient is not able to make the connection.

Checked the "httpclient" inside org.apache.httpcomponents, and I can go through the proxy without issues, but I am using vert.x so need the WebClient to work.

Looked into the org.apache.httpcomponents code and found this is what its doing:
1. Create a socket connection to the secure proxy.
2. Issue a connect request to the proxy server over the socket.
3. The proxy server challenge the connect request and returns back headers that includes "Negotiate".
4. Generate a kerberos token from the KDC for the proxy server.
5. Send over the socket the connect request again with the header:
"Proxy-Authorization: Negotiate <token>"
6. The connection is then successful.
7. The socket is then upgraded to TLS.

WebClient is not able to do that, I tried to do the bits of obtaining the Token and passing it in the "Proxy-Authorization" header in a connect request, but its not working, Still getting "407 Proxy Authentication Required".

### Version

Same issue on 4.4.3 and 4.5.3

### Context

I encountered an exception which looks suspicious while ...

### Do you have a reproducer?

No

### Steps to reproduce

Just a simple WebClient.getAbs(...) request where ProxyOptions are specified.

Harshal96:
Fixes #5129.

- Add `ProxyOptions#setProxyAuthorization(...)` for prebuilt proxy auth values such as `Negotiate <token>`.
- Send the custom `Proxy-Authorization` value for plain HTTP proxy requests and HTTPS CONNECT tunneling.
- Use Netty `HttpProxyHandler` outbound headers for CONNECT so non-Basic proxy schemes can be supplied by callers.
- Include the proxy authorization value in endpoint keys to avoid sharing pooled connections across different proxy auth values.
- Add proxy option, endpoint key, HTTP proxy, and HTTPS CONNECT coverage.

### DIFF
--- a/vertx-core/src/test/java/io/vertx/tests/http/Http1xProxyTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/http/Http1xProxyTest.java
@@ -205,7 +205,6 @@ public class Http1xProxyTest extends HttpTestBase2 {
   public void testHttpsProxyRequestCustomAuthorization() throws Exception {
     String authorization = "Negotiate token";
 
-    server.close();
     server = vertx.createHttpServer(createBaseServerOptions()
       .setSsl(true)
       .setKeyCertOptions(Cert.SERVER_JKS.get()));
--- a/vertx-core/src/test/java/io/vertx/tests/http/Http1xProxyTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/http/Http1xProxyTest.java
@@ -186,6 +186,53 @@ public class Http1xProxyTest extends HttpTestBase2 {
 
   @WithProxy(kind = ProxyKind.HTTP)
   @Test
+  public void testHttpProxyRequestCustomAuthorization() throws Exception {
+    String authorization = "Negotiate token";
+    client = vertx.createHttpClient(new HttpClientOptions()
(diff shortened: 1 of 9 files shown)


vertx-core/src/test/java/io/vertx/tests/http/Http1xProxyTest.java
```diff
@@ -184,6 +184,53 @@ public void testHttpProxyRequestAuth() throws Exception {
...
+
+    assertEquals(authorization, proxy.lastRequestHeaders().get("Proxy-Authorization"));
+  }
+
+  @WithProxy(kind = ProxyKind.HTTP)
+  @Test
+  public void testHttpsProxyRequestCustomAuthorization() throws Exception {
+    String authorization = "Negotiate token";
+
+    server.close();
```
vietj: no need to close the server

Harshal96: done

Harshal96:
@vietj all the comments are handled, can you give it another review?

PR #6136, opened 2026-05-28
Fix premature tracer span closure for fire-and-forget and publish

tsegismont:
See #6021

Add MessageConsumer#processor so consumers can defer trace span closure until async/blocking handler completion.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/eventbus/MessageConsumer.java
+++ b/vertx-core/src/main/java/io/vertx/core/eventbus/MessageConsumer.java
@@ -52,7 +52,7 @@ public interface MessageConsumer<T> extends ReadStream<Message<T>> {
    * @return this consumer
    */
   @GenIgnore(PERMITTED_TYPE)
-  MessageConsumer<T> processor(Function<Message<T>, Future<Void>> processor);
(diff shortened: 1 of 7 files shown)


vertx-core/src/main/java/io/vertx/core/eventbus/MessageConsumer.java
```diff
@@ -38,6 +43,17 @@ public interface MessageConsumer<T> extends ReadStream<Message<T>> {
...
+  /**
+   * Set a processor for the consumer.
+   * <p>
+   * The returned future completion signals the completion of message processing.
+   *
+   * @param processor the processor
+   * @return this consumer
+   */
+  @GenIgnore(PERMITTED_TYPE)
+  MessageConsumer<T> processor(Function<Message<T>, Future<Void>> processor);
```
vietj: here we should use instead `Function<Message<T>, Future<?>>`

vietj (approved):

vertx-core/src/main/java/io/vertx/core/eventbus/MessageConsumer.java
tsegismont: Good point, done

Issue #6144, opened 2026-05-30
WebSocketBaseImpl closeStatusCode can throw an NPE


cescoffier:
### Version

5.0

### Context

The method is the following:

```
@Override
    public Short closeStatusCode() {
      Short sc;
      boolean isClosed;
      synchronized (this) {
        sc = closeStatusCode;
        isClosed = closed;
      }
      return isClosed ? (sc == null ? 1006 : sc) : sc;
    }
```

Look at the return statement.

The inner ternary `sc == null ? 1006 : sc` has an int literal on one branch and Short on the other.  Since  `1006` is a constant that fits in short, Java resolves the result type to short (primitive), which means `sc` would be unboxed if this branch is taken. So, that makes the outer ternary `isClosed ? (short primitive) : (Short boxed)` . When one branch is primitive short, and the other is boxed Short, the result type is short... Java unboxes `sc` in the false branch too. Unboxing null throws NPE.

The trivial fix is: return isClosed ? (sc == null ? Short.valueOf((short) 1006) : sc) : sc;

cescoffier:
This is clearly a bug. Could we get it into 5.0. x?

vietj:
> This is clearly a bug. Could we get it into 5.0. x?

we haven't planned to release anymore 5.0.x release since Vert.x 5.1 is out

tsegismont:
Fixed by https://github.com/eclipse-vertx/vert.x/pull/6178

PR #6153, opened 2026-06-02
WebSocketBaseImpl closeStatusCode can throw an NPE

tsegismont:
See #6144

A single test is not less clear and avoids the NPE.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/websocket/WebSocketImplBase.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/websocket/WebSocketImplBase.java
@@ -204,7 +204,10 @@ public abstract class WebSocketImplBase<S extends WebSocket> implements WebSocke
       sc = closeStatusCode;
       isClosed = closed;
     }
-    return isClosed ? (sc == null ? 1006 : sc) : sc;
+    if (isClosed && sc == null) {
+      return 1006;
+    }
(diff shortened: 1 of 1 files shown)


vertx-core/src/main/java/io/vertx/core/http/impl/websocket/WebSocketImplBase.java
```diff
@@ -204,7 +204,10 @@ public Short closeStatusCode() {
       sc = closeStatusCode;
       isClosed = closed;
     }
-    return isClosed ? (sc == null ? 1006 : sc) : sc;
+    if (isClosed && sc == null) {
```
vietj, line 207: can we keep a one liner ?

vietj: can this case be tested ?

tsegismont: We can, but I'd argue it doesn't make it clearer:

```
return (isClosed && sc == null) ? Short.valueOf((short) 1006) : sc;
```

tsegismont: What's your concern with the if statement?

Issue #6146, opened 2026-06-01
WebSocketClient ignores setSslEngineOptions


phraktle:
### Version

5.1.0

### Context

Regression in 5.1.0 compared to 5.0.12: a `WebSocketClient` built with `setSslEngineOptions(new OpenSSLEngineOptions())` always uses the JDK SSLEngine, never native OpenSSL — even with tcnative available.

This cause seems to be: `VertxImpl.createWebSocketClientImpl` builds the `NetClient` via `new NetClientBuilder(this, config.getTcpConfig())`, whose `TcpClientConfig` constructor never sets `sslEngineOptions` and the call site never chains `.sslEngineOptions(...)`, so `resolveEngineOptions(null, useAlpn=false)` defaults to JDK.

### Steps to reproduce

1. Have `netty-tcnative-boringssl-static` on the classpath (`OpenSsl.isAvailable()` is true).
2. Connect to any wss:// host with:
```java
vertx.createWebSocketClient(new WebSocketClientOptions()
    .setSsl(true)
    .setSslEngineOptions(new OpenSSLEngineOptions()));
```
3. On connect, check `ws.sslSession().getClass().getName()`

**Expected**: `io.netty.handler.ssl.*` (native OpenSSL).
**Actual**: `sun.security.ssl.SSLSessionImpl` (JDK).

tsegismont:
Fixed by https://github.com/eclipse-vertx/vert.x/pull/6156

PR #6154, opened 2026-06-02
Fix WebSocketClient ignoring setSslEngineOptions

tsegismont:
See #6146

WebSocketClient was ignoring the SslEngineOptions and always using the JDK SSLEngine instead of the configured engine.

### DIFF
--- a/vertx-core/src/test/java/io/vertx/it/tls/SSLEngineTest.java
+++ b/vertx-core/src/test/java/io/vertx/it/tls/SSLEngineTest.java
@@ -15,14 +15,16 @@ import io.netty.handler.ssl.OpenSslSessionContext;
 import io.netty.handler.ssl.SslContext;
 import io.vertx.core.http.*;
 import io.vertx.core.internal.VertxInternal;
+import io.vertx.core.internal.net.NetServerInternal;
 import io.vertx.core.internal.tls.ServerSslContextProvider;
 import io.vertx.core.net.JdkSSLEngineOptions;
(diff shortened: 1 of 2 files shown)


vertx-core/src/test/java/io/vertx/it/tls/SSLEngineTest.java
```diff
@@ -128,4 +128,83 @@ private void doTest(SSLEngineOptions engine,
       .onComplete(onSuccess(v -> testComplete()));
     await();
   }
+
+  @Test
+  public void testWebSocketDefaultEngine() {
```
vietj, line 107: can you move this test rather to the WebSocketTest class ? this class aims to test the SSL engine, not the integration with WebSocket

tsegismont: I added the tests here precisely because the goal is to test whether the SSL engine is picked correctly.

I can't move them to `WebSocketTest`, because this test is executed automatically with surefire, without configuration to include/exclude netty tc native classes and boring ssl. Or did I got this wrong?

vietj: you are right, I made the mistake to think about the other SSLEngineTest that is not IT. Now I'm looking at it, the test looks a lot similar to the HttpServer implementation, we should try to come up with the same code.

tsegismont: Done

vietj (approved):

Issue #5025, opened 2023-12-08
Add an optional  argument to HttpServerResponse.sendFile() to control whether the response is ended


oxsdd:
Add an optional argument to `HttpServerResponse.sendFile()` that indicates whether the response should be ended.

Currently, `sendFile` always ends the response. This prevents sending data from more than one file.

This would be similar in nature to `Pipe.endOnComplete()`.

This option would be also be available on the `sendFile(filename, offset)` and `sendFile(filename, offset, length)` variants as well, so partial files could also be sent.

Use case:  We have an S3 storage service where multipart objects are stored in multiple files. We can send single part S3 objects with `sendFile`. But, when sending multipart objects (ranges), we must compose `Pipe` futures with `endOnComplete(false)`, which does not take advantage of the zero-copy IO features of Linux. 

I would not be able to contribute, but would be able to test the feature.

Thanks!

oxsdd:
I was not sure if I would be able to contribute thinking that it is core internal code but if you would like me to I will make an attempt. Note that I am a first time contributor so I may need some assistance along the way.

I was not considering chunked encoding, since my use case does not need it.  Instead I am looking for  a means of pumping data from *multiple* files to the response that utilizes zero-copy.  Currently the only option seems to be a composition of `Pipe.endOnComplete(false)` that call `HttpServerResponse::write(Buffer)` which is not zero-copy.  Since the application can compute the content length from the files, it's not clear to me that chunked encoding support would be required?

Looking at `Http1xServerResponse` I am now thinking that instead of changing `sendFile`, that it might be better to add *new* methods to `HttpServerResponse` like the following, which have similar behavior to `HttpServerResponse::write(Buffer)`:

```
Future<Void> writeFile(String path);
Future<Void> writeFile(string path, long offset);
Future<Void> writeFile(String path, long offset, long length);
```
The implementation in `Http1xServerResponse` would:

* Ensure the response was not already written, and throw an exception if it is.
* Ensure the response is not chunked, and throw an exception if it is.
* Ensure that `Content-Length` is set, and throw an exception if it is not.
* Open the `RandomAccessFile`
* If headers have not been written, prepare them (call `prepareHeaders(-1)`) and write them to the connection.
* Call `ConnectionBase::sendFile(RandomAccessFile raf, long offset, long length)` to send  the file.
* **Anything else at this point, since we are not ending the request?**

[...]

PR #6159, opened 2026-06-04
Configurable chunk size for chunked send file

### LINKED ISSUE #6160: Make sendFile fallback chunk size configurable
### Describe the feature

`HttpServerResponse.sendFile(...)` can use different internal paths depending on the protocol and connection capabilities.

When the native file-region/sendfile path is available, the file can be transferred efficiently by the OS. However, when this path is not available, for example with TLS, HTTP/2, HTTP/3, traffic shaping, or unsupported transports, Vert.x falls back to streaming the file through userspace.

Today these fallback paths use a fixed 8 KiB chunk/read size, for example in `ChunkedNioFile` and the `AsyncFile`-based fallback path. This value is conservative, but it can be suboptimal for larger files.

In a local benchmark serving a 1 MiB file, increasing the fallback chunk size to 32 KiB, 64 KiB, or 128 KiB significantly reduced overhead compared to the current 8 KiB default.

It would be useful to expose an option for configuring the fallback send-file chunk size, while keeping the current default behavior unchanged. The option should be ignored when the native file-region/sendfile path is used.

### Contribution

(https://github.com/eclipse-vertx/vert.x/pull/6159)

doxlik:
Currently for non-sendfile (non zero copy) send files the chunk size is hardcoded to 8kb which is good default but not always best size, as example in my benchmarks 32 kb showed much better throughput.

Closes https://github.com/eclipse-vertx/vert.x/issues/6160


jnbdz (commented):
Took a look at this while going through the sendFile-related tickets (I have #5025 in progress on the same files, see the last point). The plumbing is correct as far as I can tell (`ChunkedNioFile` / `UncloseableChunkedNioFile` / `AsyncFile.setReadBufferSize` / `VertxConnection.sendFile` all receive the value, existing behaviour is unchanged by default). A few remarks, mostly about API shape:

**API surface** – 9 new overloads on `HttpServerResponse` for a single int. Two possible ways to keep that smaller, for @vietj to weigh in on:
- put `offset` / `length` into `SendFileOptions` too, so only `sendFile(String, SendFileOptions)`, `sendFile(FileChannel, SendFileOptions)` and `sendFile(RandomAccessFile, SendFileOptions)` are added (3 methods instead of 9), or
- make it a server-wide default (`HttpServerOptions.setSendFileChunkSize(int)`), which matches the motivation of the issue ("the fallback default is conservative") without any new response API; a per-call variant could still come later.

**Layering** – `io.vertx.core.net.impl.VertxConnection` now imports `io.vertx.core.http.SendFileOptions` only for `DEFAULT_CHUNK_SIZE`, so `net.impl` depends on `http`. A constant in `net.impl` (with `SendFileOptions.DEFAULT_CHUNK_SIZE` referencing it) avoids that. `UncloseableChunkedNioFile` also has an unused import of `SendFileOptions` left over.

**Default methods drop the options** – the interface defaults `sendFile(..., SendFileOptions)` delegate to the variant without options, so a third-party implementation silently ignores them. Both core implementations override, so no issue in core, but the javadoc could say that.

**Tests** – `testSendFileWithOptions` / `testSendFileRangeWithOptions` in the HTTP/1 plain subclass go through the zero-copy path where the option is ignored by design; in the TLS / HTTP/2 subclasses the fallback is used but nothing observes the chunk size. One assertion that the value actually reaches the pump (e.g. the `chunkSize=4` range test checking that the body arrives in more than one DATA frame / read, or checking `AsyncFile` read buffer size for the non-sendfile path) would make the tests meaningful.

**Docs** – no `http.adoc` paragraph / `HttpExamples` snippet for the new option.

**Minor** – `SendFileOptions` has no `equals`/`hashCode`/`toString`; `NetSocket.sendFile` keeps the hard-coded 8 KiB (parity, possibly out of scope).

**Overlap heads-up** – #5025 (`writeFile`, i.e. sendFile without ending the response) is being implemented on the same methods and uses `UncloseableChunkedNioFile` / `VertxConnection.sendFile` too. If a `SendFileOptions` object lands, `writeFile` should take the same object, which is one more reason to keep the overload count low (offset/length inside the options, or a server-level default). Whichever lands first, I'm fine rebasing the other.

doxlik:
Hi @jnbdz

Thank you for review. I will have detailed view once I have time.
Meanwhile related put offset and length to options, not sure if we can do that, because it will be public api breakage, not sure if it is something that allowed to be done.

vietj:
at the moment I think we should have instead of a new options object, a configuration for all send files on the HttpServerConfig object

doxlik:
@vietj please have a look now

Issue #6169, opened 2026-06-08
Connection pool cancel reports incorrect result for waiter dispatched by recycle


tsegismont:
### Version

5.1.0+

### Context

Reported originally in vert-x3/vertx-web#2911 — under high concurrency with short connect/acquire timeouts, `WebClient` triggers unhandled `IllegalStateException: Promise already completed` exceptions. Reproducible with Vert.x 5.1.0 and 5.1.1.

### Root cause analysis

`SimpleConnectionPool` misses `disposed` flag on waiter dispatch

When a connection is recycled and the pool dispatches it to a queued waiter, `Recycle.execute()` polls the waiter from the queue but does **not** set `waiter.disposed = true`. The same omission exists in `ConnectSuccess.execute()` (secondary waiter loop) and `SetConcurrency.execute()`.

Contrast with `ConnectSuccess.execute()` handling of the **initiating** waiter, which correctly sets `waiter.disposed = true`.

This causes `Cancel.execute()` to enter the `else if (!waiter.disposed)` branch, incorrectly reporting `cancelled = true` for a waiter that has already been dispatched a connection. The cancel callback then attempts to fail the promise even though the connection was already given to the waiter.

#### Race sequence

1. Waiter W is queued in the pool; a timeout timer is set via `onEnqueue`
2. A connection is recycled → `Recycle.execute()` polls W from the queue **without setting `W.disposed`**
3. The timeout timer fires → `Cancel.execute()` finds W not in queue, `!W.disposed` is true → `cancelled = true`
4. The `CombinerExecutor` produces both tasks: a `RecycleTask` (to emit the lease) and a `CancelTask` (to call `promise.fail(timeout)`)
5. `LeaseImpl.emit()` completes `promise` directly (bypassing `Request.complete()`)
6. `CancelTask` calls `promise.fail(timeout)` → **`IllegalStateException: Promise already completed`**

tsegismont:
Fixed by https://github.com/eclipse-vertx/vert.x/pull/6211

PR #6170, opened 2026-06-08
Fix connection pool race causing double-completion of acquire promise

tsegismont:
See #6169
    
A bug in `SimpleConnectionPool` allows a pool waiter's promise to be completed twice under concurrent recycle + cancel with a short connect/acquire timeout.
    
`Recycle`, `ConnectSuccess` (secondary waiter loop), and `SetConcurrency` polled waiters without setting `waiter.disposed=true`.
    
`Cancel` therefore saw `disposed=false`, reported cancellation success, and failed the promise, racing with the recycle path that had already completed it.
    
Some portions of this content were created with the assistance of Claude Code.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/internal/pool/SimpleConnectionPool.java
+++ b/vertx-core/src/main/java/io/vertx/core/internal/pool/SimpleConnectionPool.java
@@ -323,7 +323,9 @@ public class SimpleConnectionPool<C> implements ConnectionPool<C> {
           c += m;
           leases = new LeaseImpl[m];
           for (int i = 0;i < m;i++) {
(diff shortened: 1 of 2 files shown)


vietj:
how is the change in SharedHttpClientGroup tested ?

tsegismont:
It wasn't, indeed. I think it could be handled separately if really required (given this is a critical piece of Vert.x core).

This more focused patch fixes the problem reported by the user so let's stick to it.

tsegismont:
@vietj PTAL

vietj:
> It wasn't, indeed. I think it could be handled separately if really required (given this is a critical piece of Vert.x core).
> 
> This more focused patch fixes the problem reported by the user so let's stick to it.

ok, I think we should if that is a real issue

vietj (approved):

PR #6173, opened 2026-06-09
Improve Http3ContextTest

vietj:
Modernize test a bit.

Fix racy test relying on unconfined future composition.


vietj:
see https://github.com/eclipse-vertx/vert.x/pull/6175

PR #6177, opened 2026-06-10
Flush pending writes before closing the connection on caught exceptions

### LINKED ISSUE #6155: HTTP/1.1 response written from an exception handler is discarded before reaching the client
### Version

5.0.x – 5.2.0-SNAPSHOT

### Context

When an HTTP/1.1 server has `setDecompressionSupported(true)` and receives a body that fails to decompress (Netty's `HttpContentDecompressor` throws `DecoderException`), a response written from `request().exceptionHandler()` never reaches the client — the client sees a connection reset instead of the intended error response.

**Note:** I used AI tooling to help investigate and draft this report and the fix. I've reviewed, tested, and verified everything myself and am happy to discuss any part of it.

### Steps to reproduce

  **Affected:** Vert.x 5.0.x – 5.2.0-SNAPSHOT — HTTP/1.1 server with `setDecompressionSupported(true)`.

  **1. Start a server that answers decode failures from the request exception handler:**

  ```java
  import io.vertx.core.Vertx;
  import io.vertx.core.http.HttpServerOptions;

  public class Repro {
      Vertx vertx = Vertx.vertx();
      vertx.createHttpServer(new HttpServerOptions().setDecompressionSupported(true))
        .requestHandler(req -> {
          req.exceptionHandler(err -> {
            if (!req.response().ended()) {
              req.response().setStatusCode(400).end("bad encoding: " + err.getMessage());
            }
          });
          req.bodyHandler(body -> req.response().end()); // not reached for a malformed body
        })
        .listen(8080)
        .await();
      System.out.println("listening on http://localhost:8080");
    }
  }
  ```

  **2. Send a body declared `Content-Encoding: gzip` whose gzip header is valid but whose payload is corrupted**, so `HttpContentDecompressor` fails mid-decode:

  ```bash
  # valid gzip, then corrupt 10 bytes of the deflate payload (past the 10-byte header)
  printf 'hello world this is a test payload for decompression' | gzip > /tmp/body.gz
  dd if=/dev/urandom of=/tmp/body.gz bs=1 seek=10 count=10 conv=notrunc 2>/dev/null

  curl -sv --data-binary @/tmp/body.gz -H 'Content-Encoding: gzip' http://localhost:8080/
  ```

  **Expected** — the client receives the response written by the exception handler:

  ```
  < HTTP/1.1 400 Bad Request
  ...
  bad encoding: ...
  ```

  **Actual (before the fix)** — the connection closes before the response is flushed; the client gets no response:

  ```
  curl: (52) Empty reply from server      # or: curl: (56) Recv failure: Connection reset by peer
  ```
  ### Proposed fix

  Root cause: `VertxHandler.exceptionCaught` dispatches the throwable to the user handlers and then calls `chctx.close()` **without flushing**, so the response queued by the handler is discarded by Netty's `ChannelOutboundBuffer` when the channel closes.

  Routing the close through the channel makes it traverse the pipeline tail and go through `VertxHandler#close` → `VertxConnection#writeClose` — the existing flush-then-close path already used by graceful shutdown, idle close and `close()`:

  ```java
  if (close) {
    if (connection != null) {
      chctx.channel().close();
    } else {
      chctx.close();
    }
  }
  ```

  I have a working patch (one line in `VertxHandler` + a regression test in `Http1xTest`) verified against `master`; it also applies to `5.1` (clean cherry-pick) and `5.0` (adapted, since `handleException` returns `void` there). Happy to open a PR if this direction looks right.

### Do you have a reproducer?

_No response_

asalajan:
Fixes https://github.com/eclipse-vertx/vert.x/issues/6155

Motivation:

When an exception is caught on a connection, VertxHandler#exceptionCaught dispatches the throwable to the user handlers and then closes the channel with chctx.close(). A response written from the handler -- for instance a 400 written from request().exceptionHandler() when HttpContentDecompressor throws a DecoderException on a malformed compressed body -- is enqueued in the outbound message queue but not flushed, because VertxConnection#write skips the flush while a read is in progress. chctx.close() then causes Netty to discard the unflushed entries from the ChannelOutboundBuffer, so the response never reaches the client: the client observes a connection reset instead of the intended HTTP response.

Changes:

Route the exception-driven close through the channel so it traverses the pipeline tail and invokes VertxHandler#close -> VertxConnection#writeClose, which flushes pending writes (an empty buffer with forceFlush=true) and closes the channel only once the flush has completed. This is the same flush-then-close path already used by graceful shutdown, idle close and the regular close(); the exception path simply was not using it. The bare chctx.close() is kept as a fallback for the case where no connection has been set yet (handlerAdded has not run), since VertxHandler#close dereferences it.

Add a regression test, Http1xTest#testRequestDecompressionInvalidBodyDeliversErrorResponse, that sends a malformed gzip body to a server with decompression enabled and asserts the client receives the response written from the request exception handler.

HTTP/2 is unaffected: its codec replaces VertxHandler in the pipeline, so connection-level exceptions do not go through this path.

I'm happy to open the backport PRs for version 5.0.x and 5.1.x once this one is reviewed/merged

Developed with assistance from an AI coding agent (Claude Code). I reviewed and tested the changes.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java
@@ -151,9 +151,6 @@ public final class VertxHandler<C extends VertxConnection> extends ChannelDuplex
       close = true;
     }
     if (close) {
-      // Close through the channel so it traverses the pipeline tail into VertxHandler#close ->
-      // VertxConnection#writeClose, flushing pending writes before the channel closes -- the same
-      // path used by a regular connection close.
       chctx.channel().close();
     }
   }
--- a/vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java
@@ -151,7 +151,15 @@ public final class VertxHandler<C extends VertxConnection> extends ChannelDuplex
       close = true;
     }
     if (close) {
-      chctx.close();
+      if (connection != null) {
+        // Route the close through the pipeline tail so VertxHandler#close -> VertxConnection#writeClose
+        // flushes any pending writes (e.g. a response written from the user's exception handler) before
+        // closing the channel. A bare chctx.close() would let Netty's ChannelOutboundBuffer discard the
+        // unflushed entries.
+        chctx.channel().close();
+      } else {
+        chctx.close();
+      }
     }
   }
 
--- a/vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java
@@ -151,15 +151,10 @@ public final class VertxHandler<C extends VertxConnection> extends ChannelDuplex
       close = true;
     }
     if (close) {
-      if (connection != null) {
-        // Route the close through the pipeline tail so VertxHandler#close -> VertxConnection#writeClose
-        // flushes any pending writes (e.g. a response written from the user's exception handler) before
-        // closing the channel. A bare chctx.close() would let Netty's ChannelOutboundBuffer discard the
-        // unflushed entries.
-        chctx.channel().close();
-      } else {
-        chctx.close();
-      }
+      // Close through the channel so it traverses the pipeline tail into VertxHandler#close ->
+      // VertxConnection#writeClose, flushing pending writes before the channel closes -- the same
+      // path used by a regular connection close.
+      chctx.channel().close();
     }
   }
 
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/websocket/WebSocketConnectionImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/websocket/WebSocketConnectionImpl.java
@@ -36,7 +36,7 @@ import static io.vertx.core.spi.metrics.Metrics.METRICS_ENABLED;
  */
 public final class WebSocketConnectionImpl extends VertxConnection {
 
-  private final long closingTimeoutMS;
+  private long closingTimeoutMS;
   private ScheduledFuture<?> closingTimeout;
   private final boolean server;
   private final WebSocketMetrics webSocketMetrics;
@@ -137,6 +137,7 @@ public final class WebSocketConnectionImpl extends VertxConnection {
 
   @Override
   public boolean handleException(Throwable t) {
+    closingTimeoutMS = 0L;
     WebSocketImplBase<?> ws = webSocket;
     if (ws != null) {
       ws.context().execute(t, ws::handleException);
--- a/vertx-core/src/test/java/io/vertx/tests/http/Http1xTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/http/Http1xTest.java
@@ -2876,50 +2876,6 @@ public class Http1xTest extends HttpTest {
   }
 
   @Test
-  // Regression test: when an inbound request body fails to decompress, the application's
-  // exception handler must still be able to deliver an error response to the client. Before
-  // the VertxHandler#exceptionCaught flush-before-close fix, the response written from the
-  // exception handler was discarded and the client only saw a connection reset.
-  public void testRequestDecompressionInvalidBodyDeliversErrorResponse() throws Exception {
-    server = vertx.createHttpServer(new HttpServerOptions().setDecompressionSupported(true));
-    // A valid gzip header (the first 10 bytes) followed by a corrupted deflate payload, so that
-    // Netty's HttpContentDecompressor accepts the stream as gzip and then fails while decoding it.
-    byte[] corrupted = TestUtils.compressGzip(TestUtils.randomAlphaString(256));
-    for (int i = 10; i < corrupted.length; i++) {
-      corrupted[i] = (byte) ~corrupted[i];
-    }
-    server.requestHandler(req -> {
-      req.exceptionHandler(err -> {
-        if (!req.response().ended()) {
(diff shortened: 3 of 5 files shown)


vietj:
for this fix we need a test in `VertxConnectionTest` that reproduces the situation as well.

asalajan:
> for this fix we need a test in `VertxConnectionTest` that reproduces the situation as well.

I have added `testFlushPendingWriteOnExceptionClose()` in  `VertxConnectionTest`

asalajan:
Hello @vietj ! 
Is there anything that still needs addressing for this PR ?

vietj:
@asalajan as I understand we are testing the behavior using the NetClient communication to the HTTP server ?

asalajan:
@vietj Yes, for `testRequestDecompressionInvalidBodyDeliversErrorResponse`, raw `NetClient` is used to capture the response bytes directly and assert the 400 reaches the wire before the connection closes which is the exact behaviour the fix restores.
A `HttpClient` version would work too, but `NetClient` gives a direct signal (empty buffer vs. the full `HTTP/1.1 400`) without going through the client HTTP stack.

For the second test, `testFlushPendingWriteOnExceptionClose`, `EmbeddedChannel` is used to isolate the change from the HTTP/decompression specifics

asalajan:
@vietj I've edited my previous response to try to clarify why NetClient was used.

vietj:
I would like actually this test to be in NetClientTest and avoid make it an HTTP test because it is not related to HTTP protocol

asalajan:
> I would like actually this test to be in NetClientTest and avoid make it an HTTP test because it is not related to HTTP protocol

@vietj I moved the test. Please let me know if you see any other issues.

vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java
```diff
@@ -151,7 +151,15 @@ public void exceptionCaught(ChannelHandlerContext chctx, final Throwable t) {
       close = true;
     }
     if (close) {
-      chctx.close();
+      if (connection != null) {
+        // Route the close through the pipeline tail so VertxHandler#close -> VertxConnection#writeClose
+        // flushes any pending writes (e.g. a response written from the user's exception handler) before
+        // closing the channel. A bare chctx.close() would let Netty's ChannelOutboundBuffer discard the
+        // unflushed entries.
+        chctx.channel().close();
```
vietj: I am wondering if we should not actually always do that instead of calling chctx.close, this is what we are currently doing on a regular close of the connection

vietj: and I don't see a good reason to close from the channel handler context perspective

asalajan: I agree chctx.channel().close() should be the default. 
I'm not sure about connection == null  since channel().close() routes through VertxHandler.close() -> conn.handleClose() which may result in a NPE if we don't make VertxHandler.close() null safe. 
Do you think connection can be null at that point ?

vietj: yes the connection could be null, although I am not 100% certain.

for now let's use inconditionally channe close please, I'll think more about it

vertx-core/src/main/java/io/vertx/core/net/impl/VertxHandler.java
```diff
@@ -151,7 +151,10 @@ public void exceptionCaught(ChannelHandlerContext chctx, final Throwable t) {
       close = true;
     }
     if (close) {
-      chctx.close();
+      // Close through the channel so it traverses the pipeline tail into VertxHandler#close ->
```
vietj: this comment is not necessary

asalajan: Removed

vietj: thanks, the point i s that indeed it is true, but there are many others similar implied that are as true as that :-)

asalajan:
I see 3 WebSocket tests are failing because of this change. I missed running them locally:
```
Error:  Errors: 
Error:    WebSocketTest.testReportProtocolViolationOnClient » Timeout Unsatisfied checkpoint
Error:    WebSocketTest.testReportProtocolViolationOnServer » Timeout Unsatisfied checkpoint
Error:    WebSocketTest.testServerWebSocketExceptionHandlerIsCalled » Timeout Unsatisfied checkpoint
Error:  Tests run: 6770, Failures: 0, Errors: 3, Skipped: 142
```
They pass on master and fail with the fix with both the unconditional and the earlier  guarded form - anything that reaches VertxHandler.close.

I am not familiar with WebSockets but I can try to look into it more with the help of AI.

Maybe one option could be to narrow the fix to the HTTP/1 layer by flushing the response in the server connection exception handling before close, and  VertxHandler.exceptionCaught remains as it is: chctx.close(). I don't know if a generic protocol agnostic solution exists.

vietj:
@asalajan I will have a look myself

vietj:
@asalajan after analysis here is my conclusion at the moment after looking more carefully at the issue.

The test is failing because when the client WebSocket handles the exception and close the WebSocket, the connection still attempts to use the closing timeout and in such case we should not.

The `WebSocketConnectionImpl` should be modified to set `closingTimeoutMS` to `0` in `handleException(Throwable)` method. 

The field should be not final anymore, which is fine from visibility perspective as it is initialized in constructor and then only modified from the event-loop.

asalajan:
@vietj Implemented like you suggested. 
Ran tests.http.WebSocketTest locally and they pass.
Thank you for your help!

Issue #6143, opened 2026-05-30
Performance regression in HTTP/2 codec server path in 5.1.0


doxlik:
### Version

5.1.0

### Context

Hi,

I noticed a possible HTTP/2 codec server performance regression after moving from Vert.x `5.0.12` to Vert.x `5.1.0`.

This is only about the **HTTP/2 codec implementation**, not the multiplex implementation.

Environment:

```text
Ubuntu Linux
Java 21
native epoll transport
```

Benchmark shape:

```text
h2load -n 5000000 -c 8 -m 128 -t 8 http://127.0.0.1:8080
```

Minimal h2c server returning a small fixed body.

Rough numbers on my machine:

```text
5.0.12:
  usually around 2.15M req/s
  often up to ~2.20M

5.1.0:
  usually around 2.02M–2.10M
```

So the difference looks roughly like a 5–8% throughput drop in this synthetic HTTP/2 hot-path benchmark, and h2load reported request latency also gets worse.

I collected async-profiler CPU/allocation profiles. Allocation did not look significantly worse overall. CPU looked like a distributed overhead increase around the new HTTP/2 request/response path, especially around:

```text
HttpRequestHeaders validate/sanitize
HttpRequestHead / HttpResponseHead
HttpServerRequestImpl / HttpServerResponseImpl
DefaultHttp2ServerStream.writeHead
request dispatch / stream close handling
```

My current understanding is that this may be related to the HTTP/2 codec path now going through the generic HTTP stream/request/response model, while HTTP/1 still has specialized `Http1ServerRequest` / `Http1ServerResponse` classes.

Question: would it make sense for the HTTP/2 codec server path to also have a specialized request/response fast path, similar to HTTP/1, while keeping the generic stream model for HTTP/2 multiplex and HTTP/3?

I am not asking to revert the 5.1 HTTP stream architecture. I wanted to ask whether this performance tradeoff is intentional/acceptable, or whether a specialized HTTP/2 codec path would be considered if it can be done cleanly.

vietj:
that would mean analysis, we cannot really indeed revert and want to keep the new implementation to avoid proliferation of code to support, I think with the multiplex architecture it might actually have more impact

vietj:
eventually also we will at some point move HTTP/1.1 to the new architecture as well, it might give significantly worse raw performance but on the long run this is what is sustainable and will be capable to provide an overall better architecture with unified middleware pipeline

PR #6179, opened 2026-06-10
Configurable http1 headers map bucket size

### LINKED ISSUE #6180: Make Http1Headers map bucket size configurable
### Describe the feature

Http1xHeaders currently always allocates a 16-bucket MapEntry array. For workloads where HTTP/1.x messages usually have only a few headers, for example 1–3 headers, a smaller bucket array can reduce memory usage and allocation pressure.

### Contribution

(https://github.com/eclipse-vertx/vert.x/pull/6179)

doxlik:
Motivation:

Http1xHeaders currently always allocates a 16-bucket MapEntry array. For workloads where HTTP/1.x messages usually have only a few headers, for example 1–3 headers, a smaller bucket array can reduce memory usage and allocation pressure.

This change keeps the default bucket size as 16, but allows tuning it with a system property for specific high-load workloads.

Closes https://github.com/eclipse-vertx/vert.x/issues/6180

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/headers/Http1xHeaders.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/headers/Http1xHeaders.java
@@ -17,6 +17,7 @@ import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.util.AsciiString;
 import io.netty.util.CharsetUtil;
 import io.vertx.core.MultiMap;
+import io.vertx.core.impl.SysProps;
 import io.vertx.core.http.impl.HttpUtils;
 import io.vertx.core.internal.http.HttpHeadersInternal;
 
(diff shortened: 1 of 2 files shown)


doxlik:
I think I’ll close this PR.

After @vietj’s feedback on Discord, I agree that a global system property is probably not the right approach, because it would affect all Vert.x instances. Making this a proper HTTP client/server configuration option would require a larger change: adding the option to `Http1ClientConfig` / `Http1ServerConfig`, passing it through the HTTP/1 client and server internals, and making `Http1xHeaders` bucket sizing per-instance instead of static/global.

The allocation reduction still looks interesting, especially for workloads with only a few HTTP/1.x headers, but I don’t currently have enough time to do this properly and safely.

Also, as discussed in https://github.com/eclipse-vertx/vert.x/issues/6143, the HTTP/1.1 path may eventually be refactored to use common HTTP components, so it is probably better not to introduce a larger HTTP/1.1-specific refactoring for this.

Thanks for the feedback.

PR #6182, opened 2026-06-11
Fix incorrect Buffer#setBytes(ByteBuffer) length calculation

### LINKED ISSUE #6181: Buffer#setBytes(int, ByteBuffer) expands length using limit instead of remaining
### Version

4.x, 5.x

### Context

Buffer.setBytes(int, ByteBuffer) expands the buffer using ByteBuffer.limit():
```java
ensureLength(pos + b.limit());
buffer.setBytes(pos, b);
```

ByteBuf.setBytes(int, ByteBuffer) only copies b.remaining() bytes. If the source ByteBuffer has a non-zero position, the Vert.x buffer length becomes larger than the copied data.

The buffer should expand by b.remaining().

### Steps to reproduce

```java
ByteBuffer src = ByteBuffer.wrap(new byte[] { 10, 20, 30, 40, 50 });
src.position(2); // remaining == 3, limit == 5

Buffer dst = Buffer.buffer();
dst.setBytes(0, src);

System.out.println(dst.length());
System.out.println(Arrays.toString(dst.getBytes()));
```

Actual output:

```text
5
[30, 40, 50, 0, 0]
```

Expected output:

```text
3
[30, 40, 50]
```

### Do you have a reproducer?

_No response_

shblue21:
Motivation:

Fixes #6181

BufferImpl#setBytes(int, ByteBuffer) expands the buffer using ByteBuffer#limit(), while ByteBuf#setBytes(int, ByteBuffer) copies only remaining() bytes.

When the source buffer has a non-zero position, the Vert.x buffer length can become larger than the bytes actually copied.

Changes:

Use ByteBuffer#remaining() when ensuring the buffer length.

Conformance:

I have signed the ECA

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/buffer/impl/BufferImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/buffer/impl/BufferImpl.java
@@ -502,7 +502,7 @@ public class BufferImpl implements BufferInternal {
   }
 
   public BufferImpl setBytes(int pos, ByteBuffer b) {
-    ensureLength(pos + b.limit());
+    ensureLength(pos + b.remaining());
     buffer.setBytes(pos, b);
     return this;
   }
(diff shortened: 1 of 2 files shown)


vietj:
can you backport to 5.1 and 4.x branches ?

Issue #6168, opened 2026-06-08
5.1.x Jackson 3 make ObjectMapper configurable


bitkid:
### Describe the feature

for jackson 2.x i can access the ObjectMapper used by Vert.x by using DatabindCodec.mapper() and configure it (adding custom modules, enable/disable features etc.)

in jackson 3.x ObjectMapper is immutable and configuration can be done using JsonMapper.builder(). This happens somewhere deep in the Vert.x code and is not accessible from outside.

I would need a hook which either provides a Builder or another way to apply my configurations.

bitkid:
my suggested fix: 

- make static mapper in io.vertx.core.json.jackson.v3.DatabindCodec not final
- add method public static void updateMapper(Function<MapperBuilder<ObjectMapper, ?>, ObjectMapper> f) {
    mapper = f.apply(mapper().rebuild());
  }
- update mapper like this DatabindCodec.updateMapper(builder -> builder.enable(SerializationFeature.APPLY_JSON_INCLUDE_FOR_CONTAINERS).build());

this keeps the original configuration of the mapper (with JacksonCodec.factory and VertxModule) and gives you the builder for your own modifications

bitkid:
https://github.com/eclipse-vertx/vert.x/pull/6183

NavidMitchell:
I created this with Claude as an interim solution. I need to be able to use a modified ObjectMapper with Jackson 3 and Vertx

https://github.com/kinotic-ai/vertx-jackson3-codec

PR #6183, opened 2026-06-11
implement issue #6168 configure jackson 3 mapper

bitkid:
Motivation:

this was not possible before.

### DIFF
--- a/vertx-core/src/main/java21/io/vertx/core/json/jackson/v3/DatabindCodec.java
+++ b/vertx-core/src/main/java21/io/vertx/core/json/jackson/v3/DatabindCodec.java
@@ -17,6 +17,7 @@ import tools.jackson.core.StreamReadFeature;
 import tools.jackson.core.StreamWriteFeature;
 import tools.jackson.core.type.TypeReference;
 import tools.jackson.databind.ObjectMapper;
+import tools.jackson.databind.cfg.MapperBuilder;
 import tools.jackson.databind.json.JsonMapper;
 import io.netty.buffer.ByteBufInputStream;
 import io.vertx.core.buffer.Buffer;
@@ -33,13 +34,14 @@ import java.io.Reader;
 import java.io.Writer;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
 public class DatabindCodec extends JacksonCodec {
 
-  private static final ObjectMapper mapper = JsonMapper
+  private static ObjectMapper mapper = JsonMapper
     .builder(JacksonCodec.factory)
     .addModule(new VertxModule())
     .build();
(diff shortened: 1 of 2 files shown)


phile314:
Any chance that this is merged/released any time soon? It does block Jackson3 adaptation for us.

bitkid:
i think this PR has problems .. because it changes the objectmapper for the rest of the test suite. the clean solution would be to instantiate the mapper correctly, but that wasnt possible for jackson2 as well, so i thought i go for the low hanging fruit.

vietj:
@phile314 I'm currently not happy modifying this on the fly and need to think about a proper way to implement this

what is your use case ?

vietj:
perhaps we could make it available statically like the jackson 2 data bind codec class does

phile314:
We register custom modules with Jackson (special datatype conversions, null handling). As the Object Mapper is immutable with Jackson3, I do not see any way to do that with Vertx and Jackson 3 right now. Maybe I am missing something?

In the end, we probably will just call Jackson directly ourselves without the detour via Vertx to work around this issue.

vietj:
@phile314 makes sense to do like that, however I think that doing that the field then should not be final anymore but should be volatile instead to ensure visibility

vertx-core/src/main/java21/io/vertx/core/json/jackson/v3/DatabindCodec.java
```diff
@@ -28,13 +29,14 @@
...
 import java.util.Map;
+import java.util.function.Function;
 
 /**
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
 public class DatabindCodec extends JacksonCodec {
 
-  private static final ObjectMapper mapper = JsonMapper
+  private static ObjectMapper mapper = JsonMapper
```
vietj: this field should be come volatile

vertx-core/src/main/java21/io/vertx/core/json/jackson/v3/DatabindCodec.java
```diff
@@ -46,6 +48,10 @@ public static ObjectMapper mapper() {
     return mapper;
   }
 
+  public static void updateMapper(Function<MapperBuilder<ObjectMapper, ?>, ObjectMapper> f) {
```
vietj: rather name it _rebuildMapper_ as this is what it does

vietj: add javadoc also explaining what the method does

bitkid:
like with all static things i see problems when running the testsuite in parallel

vietj:
we don't run test suite in parallel at the moment

bitkid:
updated the PR

vietj:
can you make it a single message for this commit and have the message follow the scheme:

Motivation:
XXX
Changes:
XXX

currently we do not have a proper description of this

vietj (changes requested):
Asked for a better description of the changes and impact, other than that we are good.

Also a single commit would help backporting to 5.1 branch

vietj:
@bitkid ping

vietj:
@bitkid ping again

PR #6186, opened 2026-06-11
Correctly handle pending reads during a start tls upgrade.

vietj:
Motivation:

`NetSocket` implementation does pause the socket on a TLS upgrade but relies on the `VertxConnection` pause that still cumulates buffers in its pending queue.

Since obtaining an SSL context is asynchronous, there can be a client hello sent by the client that will be read and sit in the VertxConnection pending list and will be missed by the SSL handler.

Changes:

During a TLS upgrade we want buffers to actually remains in the channel so we should pause channel reading instead of pausing the `VertxConnection` to ensure the pipeline won't see any read during the SSL context resolution.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/net/impl/tcp/NetSocketImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/impl/tcp/NetSocketImpl.java
@@ -127,7 +127,7 @@ public class NetSocketImpl extends StreamChannelBase<NetSocketImpl> implements N
       applicationProtocols = null;
     }
     if (chctx.pipeline().get("ssl") == null) {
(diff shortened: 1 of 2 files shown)


tsegismont:
@vietj this change broke the MySQL Client test

```
Error:  Failures: 
Error:    MySQLTLSTest.lambda$testVerifyIdentityInvalidHostname$22:299 Was expecting <Failed to read any response from the server, the underlying connection may have been lost unexpectedly.> to contain <No name matching localhost found>. Expected true
```

See https://github.com/eclipse-vertx/vertx-sql-client/actions/runs/27401881759/job/80981630732#step:5:918

Can you please take a look?

PR #6191, opened 2026-06-14
Copy followAlternativeServices in HttpClientOptions

shblue21:
Motivation:

The HttpClientOptions copy constructor did not copy followAlternativeServices, so copied options could lose that setting.

Conformance:

I have signed the ECA


vietj:
actually I think that the followAlternativeServices from HttpClientOptions should be removed instead because it does not make much sense

shblue21:
Thanks. Since #6192 removes `followAlternativeServices` from `HttpClientOptions`, this copy constructor fix is no longer needed. I’ll close it.

PR #6199, opened 2026-06-15
fix(dns): support bare IPv6 resolver addresses

### LINKED ISSUE #6194: AddressResolverOptions does not accept bare IPv6 DNS server addresses
### Version

4.x, 5.x

### Context

AddressResolverOptions#setServers says each DNS server entry is the IP address of the DNS server, optionally followed by `:port`. The Vert.x guide also says the default DNS server port is 53 when no port is specified.

Based on that, I expected a bare IPv6 DNS server address such as Google Public DNS IPv6 to be accepted and use the default port.

Currently, resolver initialization fails for this value, apparently because the IPv6 colons are interpreted as part of the host/port syntax.

This is only about the bare IPv6/no-port case.

### Steps to reproduce

```java
Vertx.vertx(new VertxOptions()
  .setAddressResolverOptions(new AddressResolverOptions()
    .addServer("2001:4860:4860::8888"))); // Google Public DNS IPv6
```

### Do you have a reproducer?

_No response_

Harshal96:
Motivation:

Fixes #6194

Fixes parsing of configured DNS resolver addresses so bare IPv6 literals are accepted without being mistaken for `host:port` values.

IPv6 DNS server addresses contain `:` characters, so the previous `lastIndexOf(':')` parsing treated bare IPv6 literals as if they included a port. That could reject valid resolver addresses or parse them incorrectly.

## How

- Detects whether a configured DNS server string actually includes a port.
- Uses Vert.x URI parsing helpers to distinguish IP literals from `host:port` values.
- Preserves support for IPv6 resolver addresses that explicitly include a port.
- Adds regression coverage for bare IPv6 resolver addresses and IPv6 addresses with ports.

## Testing

- Added `NameResolverTest` coverage for bare IPv6 DNS servers.
- Added coverage for IPv6 DNS server entries with explicit ports.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/dns/impl/DnsAddressResolverProvider.java
+++ b/vertx-core/src/main/java/io/vertx/core/dns/impl/DnsAddressResolverProvider.java
@@ -23,6 +23,7 @@ import io.vertx.core.buffer.Buffer;
 import io.vertx.core.dns.AddressResolverOptions;
 import io.vertx.core.internal.ContextInternal;
 import io.vertx.core.internal.VertxInternal;
+import io.vertx.core.net.impl.UriParser;
 import io.vertx.core.spi.dns.AddressResolverProvider;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


shblue21:
Thanks!

PR #6212, opened 2026-06-18
Add RFC 10008 (QUERY Method) Support

desiderantes:
Motivation:

In order to add RFC 10008 support elsewhere in vert.x, these are the basic blocs required to support the QUERY Method.

Conformance:

- [x] You should have signed the Eclipse Contributor Agreement as explained in https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md

- [x] Please also make sure you adhere to the code style guidelines: https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/HttpClientConfig.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/HttpClientConfig.java
@@ -79,6 +79,7 @@ public class HttpClientConfig {
   private String defaultHost;
   private int defaultPort;
   private int maxRedirects;
+  private int maxRedirectBufferSize;
   private ObservabilityConfig observabilityConfig;
   private boolean shared;
   private String name;
@@ -97,6 +98,7 @@ public class HttpClientConfig {
     this.defaultHost = HttpClientOptions.DEFAULT_DEFAULT_HOST;
     this.defaultPort = HttpClientOptions.DEFAULT_DEFAULT_PORT;
     this.maxRedirects = HttpClientOptions.DEFAULT_MAX_REDIRECTS;
+    this.maxRedirectBufferSize = HttpClientOptions.DEFAULT_MAX_REDIRECT_BUFFER_SIZE;
     this.observabilityConfig = null;
     this.shared = HttpClientOptions.DEFAULT_SHARED;
     this.name = HttpClientOptions.DEFAULT_NAME;
@@ -116,6 +118,7 @@ public class HttpClientConfig {
     this.defaultHost = other.defaultHost;
     this.defaultPort = other.defaultPort;
     this.maxRedirects = other.maxRedirects;
+    this.maxRedirectBufferSize = other.maxRedirectBufferSize;
     this.observabilityConfig = other.observabilityConfig != null ? new ObservabilityConfig(other.observabilityConfig) : null;
     this.shared = other.shared;
     this.name = other.name;
@@ -143,6 +146,7 @@ public class HttpClientConfig {
     this.defaultHost = options.getDefaultHost();
     this.defaultPort = options.getDefaultPort();
     this.maxRedirects = options.getMaxRedirects();
+    this.maxRedirectBufferSize = options.getMaxRedirectBufferSize();
     this.observabilityConfig = observabilityConfig;
     this.shared = options.isShared();
     this.name = options.getName();
@@ -448,6 +452,24 @@ public class HttpClientConfig {
    */
   public HttpClientConfig setMaxRedirects(int maxRedirects) {
     this.maxRedirects = maxRedirects;
+    return this;
+  }
+
+  /**
+   * @return the maximum size of the redirect buffer when redirecting QUERY requests
+   */
+  public int getMaxRedirectBufferSize() {
+    return maxRedirectBufferSize;
+  }
+
+  /**
+   * Set the maximum size of the redirect buffer in bytes when redirecting QUERY requests.
+   *
+   * @param maxRedirectBufferSize the maximum buffer size
+   * @return a reference to this, so the API can be used fluently
+   */
+  public HttpClientConfig setMaxRedirectBufferSize(int maxRedirectBufferSize) {
+    this.maxRedirectBufferSize = maxRedirectBufferSize;
     return this;
   }
 
--- a/vertx-core/src/main/java/io/vertx/core/http/HttpClientConfig.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/HttpClientConfig.java
@@ -10,6 +10,7 @@
  */
 package io.vertx.core.http;
 
+import io.vertx.core.impl.Arguments;
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.Unstable;
@@ -456,7 +457,7 @@ public class HttpClientConfig {
   }
 
   /**
-   * @return the maximum size of the redirect buffer when redirecting QUERY requests
+   * @return the maximum size in bytes of the redirect buffer when redirecting QUERY requests
    */
   public int getMaxRedirectBufferSize() {
     return maxRedirectBufferSize;
@@ -469,6 +470,7 @@ public class HttpClientConfig {
    * @return a reference to this, so the API can be used fluently
    */
   public HttpClientConfig setMaxRedirectBufferSize(int maxRedirectBufferSize) {
+    Arguments.require(maxRedirectBufferSize >= 0, "Max redirect buffer size must be >= 0");
     this.maxRedirectBufferSize = maxRedirectBufferSize;
     return this;
   }
--- a/vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
@@ -11,6 +11,7 @@
 
 package io.vertx.core.http;
 
+import io.vertx.core.impl.Arguments;
 import io.netty.handler.logging.ByteBufFormat;
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.codegen.json.annotations.JsonGen;
@@ -134,9 +135,9 @@ public class HttpClientOptions extends ClientOptionsBase {
   public static final int DEFAULT_MAX_REDIRECTS = 16;
 
   /**
-   * Default max redirect buffer size = 1024 * 1024 bytes (1MB)
+   * Default max redirect buffer size = 4 * 1024 bytes (4KB)
    */
-  public static final int DEFAULT_MAX_REDIRECT_BUFFER_SIZE = 1024 * 1024;
+  public static final int DEFAULT_MAX_REDIRECT_BUFFER_SIZE = 4 * 1024;
 
   /*
    * Default force SNI = {@code false}
@@ -905,7 +906,7 @@ public class HttpClientOptions extends ClientOptionsBase {
   }
 
   /**
-   * @return the maximum size of the redirect buffer when redirecting QUERY requests
+   * @return the maximum size in bytes of the redirect buffer when redirecting QUERY requests
    */
   public int getMaxRedirectBufferSize() {
     return maxRedirectBufferSize;
@@ -918,6 +919,7 @@ public class HttpClientOptions extends ClientOptionsBase {
    * @return a reference to this, so the API can be used fluently
    */
   public HttpClientOptions setMaxRedirectBufferSize(int maxRedirectBufferSize) {
+    Arguments.require(maxRedirectBufferSize >= 0, "Max redirect buffer size must be >= 0");
     this.maxRedirectBufferSize = maxRedirectBufferSize;
     return this;
   }
--- a/vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
@@ -133,6 +133,11 @@ public class HttpClientOptions extends ClientOptionsBase {
    */
   public static final int DEFAULT_MAX_REDIRECTS = 16;
 
+  /**
+   * Default max redirect buffer size = 1024 * 1024 bytes (1MB)
+   */
+  public static final int DEFAULT_MAX_REDIRECT_BUFFER_SIZE = 1024 * 1024;
+
   /*
    * Default force SNI = {@code false}
    */
@@ -171,6 +176,7 @@ public class HttpClientOptions extends ClientOptionsBase {
   private int defaultPort;
   private HttpVersion protocolVersion;
   private int maxRedirects;
+  private int maxRedirectBufferSize;
   private boolean forceSni;
 
   private TracingPolicy tracingPolicy;
@@ -211,6 +217,7 @@ public class HttpClientOptions extends ClientOptionsBase {
     this.defaultPort = other.defaultPort;
     this.protocolVersion = other.protocolVersion;
     this.maxRedirects = other.maxRedirects;
+    this.maxRedirectBufferSize = other.maxRedirectBufferSize;
     this.forceSni = other.forceSni;
     this.tracingPolicy = other.tracingPolicy;
     this.shared = other.shared;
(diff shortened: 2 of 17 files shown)


vietj:
I think that the redirect support is not correct because on a redirect the client needs to buffer the entire payload and send it again on a redirection.

vietj:
I believe that to implement this correctly we would need the extra following features for the redirection

1. on a QUERY with redirect enabled, the HTTP client needs to buffer the query payload entirely
2. when the redirect handler responds a QUERY method with a non null request options, the buffered payload should be sent again

To make this more efficient, we should improve the client to make buffering efficient than it currently is, I am thinking of the HttpClientRequestImpl that should be optimized when we send a single buffer.

We would also need to make this buffering have an upper bound and take another path (e.g. avoid redirection) when the payload exceeds the max buffer size.

vietj:
I think this is also relevant : https://github.com/netty/netty/pull/16978/changes

desiderantes:
> I believe that to implement this correctly we would need the extra following features for the redirection
> 
>     1. on a QUERY with redirect enabled, the HTTP client needs to buffer the query payload entirely
> 
>     2. when the redirect handler responds a QUERY method with a non null request options, the buffered payload should be sent again

Redirection in the RFC talks more about the client sending a QUERY and being redirected to do a GET (same or different endpoint). Redirecting a QUERY to another endpoint that handles QUERY would be the same as redirecting a PUT or a POST to another PUT/POST endpoint.

> 
> To make this more efficient, we should improve the client to make buffering efficient than it currently is, I am thinking of the HttpClientRequestImpl that should be optimized when we send a single buffer.
>
> We would also need to make this buffering have an upper bound and take another path (e.g. avoid redirection) when the payload exceeds the max buffer size.

I think this is a further optimization that, based on the prior point, would be disconnected from QUERY support

vietj:
> > I believe that to implement this correctly we would need the extra following features for the redirection
> > ```
> > 1. on a QUERY with redirect enabled, the HTTP client needs to buffer the query payload entirely
> > 
> > 2. when the redirect handler responds a QUERY method with a non null request options, the buffered payload should be sent again
> > ```
> 
> Redirection in the RFC talks more about the client sending a QUERY and being redirected to do a GET (same or different endpoint). Redirecting a QUERY to another endpoint that handles QUERY would be the same as redirecting a PUT or a POST to another PUT/POST endpoint.

as I read it, this is only valid for HTTP response status 303, the QUERY turns into a GET:

_A response to QUERY with the status code 303 (See Other, [Section 15.4.4](https://rfc-editor.org/rfc/rfc9110#section-15.4.4) of [[HTTP](https://www.rfc-editor.org/info/rfc10008/#RFC9110)]) indicates that the original query can be accomplished via a normal retrieval request on the URI referenced by the Location response field ([[HTTP](https://www.rfc-editor.org/info/rfc10008/#RFC9110)], [Section 10.2.2](https://rfc-editor.org/rfc/rfc9110#section-10.2.2)). For HTTP, this means sending a GET request to the new target URI, as illustrated by the example in [Appendix A.4.3](https://www.rfc-editor.org/info/rfc10008/#example.indirect)._

For other 4 status codes, I read:

_In all four cases, the server is suggesting that the user agent can accomplish its original QUERY request by sending a similar QUERY request to the new target URI referenced by the Location._

> 
> > To make this more efficient, we should improve the client to make buffering efficient than it currently is, I am thinking of the HttpClientRequestImpl that should be optimized when we send a single buffer.
> > We would also need to make this buffering have an upper bound and take another path (e.g. avoid redirection) when the payload exceeds the max buffer size.
> 
> I think this is a further optimization that, based on the prior point, would be disconnected from QUERY support

Right.

desiderantes:
> 
> as I read it, this is only valid for HTTP response status 303, the QUERY turns into a GET:
> 
> _A response to QUERY with the status code 303 (See Other, [Section 15.4.4](https://rfc-editor.org/rfc/rfc9110#section-15.4.4) of [[HTTP](https://www.rfc-editor.org/info/rfc10008/#RFC9110)]) indicates that the original query can be accomplished via a normal retrieval request on the URI referenced by the Location response field ([[HTTP](https://www.rfc-editor.org/info/rfc10008/#RFC9110)], [Section 10.2.2](https://rfc-editor.org/rfc/rfc9110#section-10.2.2)). For HTTP, this means sending a GET request to the new target URI, as illustrated by the example in [Appendix A.4.3](https://www.rfc-editor.org/info/rfc10008/#example.indirect)._
> 
> For other 4 status codes, I read:
> 
> _In all four cases, the server is suggesting that the user agent can accomplish its original QUERY request by sending a similar QUERY request to the new target URI referenced by the Location._

Yeah, I see what you mean, sorry. I'll drop the part related to this, and just have the test for redirection to GET. Then I'll take a look at making the QUERY resend possible

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -41,6 +41,7 @@ public class HttpClientRequestImpl extends HttpClientRequestBase implements Http
 
   static final Logger log = LoggerFactory.getLogger(HttpClientRequestImpl.class);
 
+  private static final int MAX_REDIRECT_BUFFER_SIZE = 1024 * 1024;
```
vietj: we should make this configurable in HttpClientConfig instead

vertx-core/src/main/java/io/vertx/core/http/HttpHeaders.java
```diff
@@ -71,6 +71,12 @@ public interface HttpHeaders {
   @GenIgnore(GenIgnore.PERMITTED_TYPE)
   CharSequence ACCEPT_PATCH = HttpHeaderNames.ACCEPT_PATCH;
 
+  /**
+   * Accept-Query header name
```
vietj, line 75: Add a note to use HttpHeaderNames.QUERY when bumping to the next Netty release

vertx-core/src/main/java/io/vertx/core/http/HttpMethod.java
```diff
@@ -198,6 +203,7 @@ public class HttpMethod {
     TRACE = new HttpMethod(io.netty.handler.codec.http.HttpMethod.TRACE);
     CONNECT = new HttpMethod(io.netty.handler.codec.http.HttpMethod.CONNECT);
     PATCH = new HttpMethod(io.netty.handler.codec.http.HttpMethod.PATCH);
+    QUERY = new HttpMethod(io.netty.handler.codec.http.HttpMethod.valueOf("QUERY"));
```
vietj, line 207: Add a note to use Netty HttpMethod.QUERY in the next Netty bump

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -505,6 +514,28 @@ private Future<Void> doWrite(Buffer buff, boolean end, boolean connect) {
...
+            } else {
+              bodyBuffer = buff;
+            }
+          } else {
+            if (bodyBuffer.length() + buff.length() > MAX_REDIRECT_BUFFER_SIZE) {
+              followRedirects = false;
+              bodyBuffer = null;
+            } else {
+              if (!bodyBufferCopied) {
+                bodyBuffer = Buffer.buffer().appendBuffer(bodyBuffer);
```
vietj: instead of copying buffers, instead we should keep a list of buffered buffers.

when we use them we should unwrap them and create a composite bytebuf that we wrap in a vertx buffer

vietj (changes requested):
a few comments.

We also need to test the behavior when the buffered payload exceeds the max payload length (check we don't get a redirect I think and the reply is processed as is)

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
vietj: we should expose it as well in `HttpClientOptions`

desiderantes: done

vertx-core/src/main/java/io/vertx/core/http/HttpClientConfig.java
```diff
@@ -451,6 +455,24 @@ public HttpClientConfig setMaxRedirects(int maxRedirects) {
     return this;
   }
 
+  /**
+   * @return the maximum size of the redirect buffer when redirecting QUERY requests
```
vietj:
```suggestion
   * @return the maximum size in bytes of the redirect buffer when redirecting QUERY requests
```

vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
```diff
@@ -133,6 +133,11 @@ public class HttpClientOptions extends ClientOptionsBase {
    */
   public static final int DEFAULT_MAX_REDIRECTS = 16;
 
+  /**
+   * Default max redirect buffer size = 1024 * 1024 bytes (1MB)
```
vietj: that seems too much, we are talking here of a payload that should replace a query string , so rather I would say 4KB by default ?

vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java
```diff
@@ -3388,6 +3423,95 @@ private void testFollowRedirect(
...
+        .setFollowRedirects(true)
+        .send(expected)
+        .expecting(HttpResponseExpectation.SC_OK))
+      .await();
+  }
+
+  @Test
+  public void testFollowRedirectQueryWithBodyExceedingLimit() throws Exception {
+    Buffer expected = TestUtils.randomBuffer(1024 * 1024 + 1); // 1MB + 1B
+    AtomicBoolean redirected = new AtomicBoolean();
```
vietj: instead of using a redirect boolean we should rather do a switch on the request path

vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java
```diff
@@ -3388,6 +3423,95 @@ private void testFollowRedirect(
       ).await();
   }
 
+  @Test
+  public void testFollowRedirectQueryWithBody() throws Exception {
+    Buffer expected = TestUtils.randomBuffer(2048);
+    AtomicBoolean redirected = new AtomicBoolean();
+    server.requestHandler(req -> {
+      if (redirected.compareAndSet(false, true)) {
```
vietj: instead of using a redirect boolean, we should instead do a switch on the request path

vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java
```diff
@@ -3388,6 +3423,95 @@ private void testFollowRedirect(
...
+        .expecting(HttpResponseExpectation.status(307)))
+      .await();
+  }
+
+  @Test
+  public void testFollowRedirectQueryWithConfiguredBodyExceedingLimit() throws Exception {
+    Buffer expected = TestUtils.randomBuffer(2048);
+    AtomicBoolean redirected = new AtomicBoolean();
+    server.requestHandler(req -> {
+      if (redirected.compareAndSet(false, true)) {
```
vietj: instead of using a redirect boolean we should rather do a switch on the request path

vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java
```diff
@@ -3388,6 +3423,95 @@ private void testFollowRedirect(
...
+        assertEquals(HttpMethod.QUERY, req.method());
+        req.bodyHandler(body -> {
```
vietj: instead of using bodyHandler we should use the body() future which will accurately report a failure for this test

vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java
```diff
@@ -3388,6 +3423,95 @@ private void testFollowRedirect(
...
+          assertEquals(body, expected);
+          String scheme = req.connection().isSsl() ? "https" : "http";
+          req.response().setStatusCode(307).putHeader(HttpHeaders.LOCATION, scheme + "://" + config.host() + ":" + config.port() + "/whatever").end();
+        });
+      } else {
+        assertEquals(HttpMethod.QUERY, req.method());
+        req.bodyHandler(body -> {
```
vietj: instead of using bodyHandler we should use the body() future which will accurately report a failure for this test

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -384,13 +403,29 @@ private void handleNextRequest(HttpClientRequest next, Promise<HttpClientRespons
...
         if (timeoutMs > 0) {
           next.idleTimeout(timeoutMs);
         }
-        next.end();
+        if (next.getMethod() == HttpMethod.QUERY && bodyBuffer != null && !bodyBuffer.isEmpty()) {
+          Buffer redirectBody;
+          if (bodyBuffer.size() == 1) {
+            redirectBody = bodyBuffer.get(0);
+          } else {
+            CompositeByteBuf composite = Unpooled.compositeBuffer();
```
vietj, line 417: this does not seem tested because we always send a single buffer in the tests.

we need a test that sends multiple buffers to ensure the behavior is correct, that would use `io.vertx.test.fakestream.FakeStream` for instance

vertx-core/src/main/java/io/vertx/core/http/HttpClientConfig.java
```diff
@@ -451,6 +455,24 @@ public HttpClientConfig setMaxRedirects(int maxRedirects) {
...
+  }
+
+  /**
+   * Set the maximum size of the redirect buffer in bytes when redirecting QUERY requests.
+   *
+   * @param maxRedirectBufferSize the maximum buffer size
+   * @return a reference to this, so the API can be used fluently
+   */
+  public HttpClientConfig setMaxRedirectBufferSize(int maxRedirectBufferSize) {
+    this.maxRedirectBufferSize = maxRedirectBufferSize;
```
vietj, line 474: we should check this value is >= 0

vietj (changes requested):
a few changes in the comments.

In addition, we need proper documentation in `http.adoc`

vertx-core/src/main/java/io/vertx/core/http/HttpClientOptions.java
desiderantes: I disagree, QUERY can be more than just replacing a query string, and there are use cases where you'd want a beefier buffer size. 4k is, however, enough for casual users I guess.

vertx-core/src/main/java/io/vertx/core/http/HttpClientRequest.java
```diff
@@ -106,6 +106,20 @@ public interface HttpClientRequest extends WriteStream<Buffer> {
    */
   int getMaxRedirects();
 
+  /**
+   * @return the maximum size in bytes of the redirect buffer when redirecting QUERY requests
+   */
+  int getMaxRedirectBufferSize();
```
vietj: we should name that `maxRedirectBufferSize()` instead

vertx-core/src/main/java/io/vertx/core/http/HttpClientRequest.java
```diff
@@ -106,6 +106,20 @@ public interface HttpClientRequest extends WriteStream<Buffer> {
...
+
+  /**
+   * Set the maximum size of the redirect buffer in bytes when redirecting QUERY requests.
+   *
+   * @param maxRedirectBufferSize the maximum buffer size
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  HttpClientRequest setMaxRedirectBufferSize(int maxRedirectBufferSize);
```
vietj: we should name that `maxRedirectBufferSize(int)` instead

vietj (changes requested):
A couple of renaming (at some point all accessors should follow this pattern and get/set methods will be deprecated for removal)

desiderantes:
> A couple of renaming (at some point all accessors should follow this pattern and get/set methods will be deprecated for removal)

Wouldn't that make it inconsistent in the meantime?

vietj:
> > A couple of renaming (at some point all accessors should follow this pattern and get/set methods will be deprecated for removal)
> 
> Wouldn't that make it inconsistent in the meantime?

yes but it's already globally inconsistent in this class with properties like authority, we could actually deprecate the get/set pattersn in this class and use what should be the default pattern I think

desiderantes:
> yes but it's already globally inconsistent in this class with properties like authority, we could actually deprecate the get/set pattersn in this class and use what should be the default pattern I think

That's OK, just wanted to be sure. I applied the changes.

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
vietj: @desiderantes this line should be removed

vietj (changes requested):
minor change to remove a constant leftover

vietj:
@desiderantes thanks for this contribution

PR #6214, opened 2026-06-23
Allow zero HTTP/2 keep-alive timeout

shblue21:
Motivation:

`Http2ClientConfig#setKeepAliveTimeout(Duration.ZERO)` currently throws.
The method documents zero and `null` as no timeout, and HTTP/1 and HTTP/3
client configs already accept zero.  

This change allows zero for HTTP/2 keep-alive timeout while still rejecting
negative durations, and adds a regression test.

Conformance:

I have signed the ECA.


shblue21:
This looks related to the recent HTTP client config changes.

PR #6215, opened 2026-06-23
Add https proxy support

### LINKED ISSUE #6207: HTTP client forward proxy traffic over SSL
## Summary
Vert.x currently provides proxy configuration for HTTP client usage, but there does not appear to be a clear option for HTTPS proxy setup. We need to route outbound secure traffic through an HTTPS proxy in our environment.

## Problem Statement
In restricted corporate networks, outbound traffic is commonly required to pass through a proxy endpoint, including secure connections. At the moment, we are unable to find a reliable Vert.x configuration path that supports HTTPS proxy behavior as required by our environment.

## Expected Behavior
Vert.x should provide a documented and supported way to configure HTTPS proxy for outbound client connections (including TLS traffic), with behavior comparable to enterprise proxy requirements. Currently the vertx io.vertx.core.net.ProxyType class doesn't have HTTPS option. 

## Environment
- Vert.x version: 5.0.0
- Java version: openjdk 17.0.13
- OS: Windows \ Linux
- Proxy type required: `HTTPS proxy`

## What We Tried
- Reviewed Vert.x proxy-related options and docs.
- Attempted available proxy configuration patterns for HTTP client.
- Confirmed network requirement is specifically HTTPS proxy routing in our setup.

Apache client has explicit option i.e. Scheme parameter in HttpHost
```
org.apache.http.HttpHost proxy = new org.apache.http.HttpHost(proxyHost,  proxyPort, scheme);

// the scheme parameter can be either http or https, which gives us option to specify whether proxy is secure or insecure.
```

Vertx io.vertx.core.net.ProxyType doesn't have a HTTPS type definition
```
io.vertx.core.net.ProxyType

io.vertx.core.net.ProxyType#HTTP // this works for a non-secure HTTP proxy, but doesn't work for HTTPS proxy.

```

## Request
Please confirm:
1. Whether HTTPS proxy is currently supported for this scenario.
2. If supported, the recommended configuration pattern (with sample).
3. If not supported, consider this a feature request and advise possible roadmap/workaround.

### Contribution

_No response_

gaganis:
Motivation:

Closes: #6207

# Add HTTPS proxy support (`ProxyType.HTTPS`)

## What this adds

Support for connecting to a forward proxy **over TLS** — i.e. the connection to
the proxy itself (leg 1) is encrypted. Enabled via:

```java
new ProxyOptions()
  .setType(ProxyType.HTTPS)
  .setHost(...).setPort(...)
  .setSslOptions(new ClientSSLOptions()...);   // TLS options for the proxy connection
```

The proxying semantics are unchanged from `ProxyType.HTTP`:

- **Plain origin → forward mode** (absolute-URI `GET`, single TLS leg to the
  proxy).
- **TLS origin → `CONNECT` tunnel** (leg-1 TLS to the proxy, then a second
  nested TLS to the origin). HTTP/2 over the tunnel negotiates via ALPN
  end-to-end. Works for both `HttpClient` and `NetClient` (`upgradeToSsl`).

## Key design decisions

1. **`ConnectionBase.isSsl()` keeps meaning "the origin is TLS", not "the wire
   is TLS".** Existing code and the public contract rely on
   `ConnectionBase.isSsl()` reflecting the origin. Rather than overload it,
   transport-level TLS (TLS to an HTTPS proxy) is computed separately via the
   `transportSsl`/`proxyHttps` helpers in `TcpHttpClientTransport`. The new
   override `Http1ClientConnection.isSsl()` returns the origin flag.

2. **Transport-level "how to connect" calculations live in the transport.**
   `HttpConnectParams` carries intent (`forwardProxy` flag, origin `ssl`,
   original `proxyOptions`); `TcpHttpClientTransport` derives peer/SNI, TLS
   options, ALPN eligibility, and proxy hostname verification at the point of
   use. The forward-vs-CONNECT decision stays at endpoint-creation time because
   it selects the pooling/connect target (`EndpointKey`).

3. **Zero-copy file region is gated on the actual pipeline, not
   `ConnectionBase.isSsl()`.** Because `ConnectionBase.isSsl()` now strictly
   means "origin", `VertxConnection.supportsFileRegion()` checks for an
   `SslHandler` in the pipeline instead of calling the inherited
   `ConnectionBase.isSsl()` — zero-copy must be disabled whenever the wire is
   encrypted (e.g. leg-1 TLS to the proxy), regardless of the logical origin
   flag.

4. **Two `SslHandler`s in the CONNECT-tunnel pipeline.** `NetSocketImpl` inserts
   the origin (leg-2) handler after the proxy (leg-1) handler (`proxy-ssl` →
   `ssl`), and `applicationLayerProtocol()` reads the **innermost**
   `SslHandler`, since ALPN is negotiated on the origin handshake.

5. **Secure defaults for the proxy leg.** Hostname verification defaults to
   `"HTTPS"` for the proxy connection; ALPN is never offered on the leg-1
   (proxy) handshake — it belongs to the origin; there is **no plaintext
   fallback** for an HTTPS proxy (a TLS handshake failure against a plaintext
   proxy fails cleanly rather than downgrading). Mutual TLS to the proxy is
   supported.

6. **Proxy SSL options participate in pool identity.** `EndpointKey` includes
   `ProxyOptions.getSslOptions()` in `equals`/`hashCode` so connections with
   different proxy TLS config aren't pooled together.

## Tests

- New `HttpsProxyTest`: forward (plain origin), `CONNECT` tunnel (HTTP/1 and
  HTTP/2 origins), proxy auth, mutual TLS, and negative cases (untrusted proxy
  cert, hostname mismatch, plaintext-proxy no-downgrade).
- New `NetTest` cases for `CONNECT` over a TLS proxy (`NetClient` +
  `upgradeToSsl`).
- `ProxyOptionsTest` for the new `sslOptions` field.

## Issues encountered & how they were resolved

- **Flaky empty body on the HTTP/2-over-CONNECT test**
  (`testHttpsProxy_Http2Origin_tunnelled`): the response body was read across
  the `await()` boundary, so body chunks could arrive with no handler attached.
  Fixed by reading the body inside the same future composition
  (`resp.body().map(...)`), which also folds the negotiated version into the
  asserted value. (Commit `2b8895f51`.)
- **Port leak in `HttpsProxyTest`**: fixed (commit `93b7ef36b`).

## Note on CI failures

CI runs on this branch intermittently failed on tests that I have seen failing
also in unrelated PRs suggesting these are flaky tests.

- `HttpSendFileTest.testSendOpenRangeFileFromClasspath` (empty HTTP body), and
- `NetTest.testListenDomainSocketAddress` (empty receive, `-PNativeEpoll`).

Conformance:

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
@@ -150,11 +150,6 @@ public class Http1ClientConnection extends Http1Connection implements io.vertx.c
   }
 
   @Override
-  public boolean isSsl() {
-    return ssl;
-  }
-
-  @Override
   public io.vertx.core.http.impl.HttpClientConnection evictionHandler(Handler<Void> handler) {
     evictionHandler = handler;
     return this;
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
@@ -150,6 +150,11 @@ public class Http1ClientConnection extends Http1Connection implements io.vertx.c
   }
 
   @Override
+  public boolean isSsl() {
+    return ssl;
+  }
+
+  @Override
   public io.vertx.core.http.impl.HttpClientConnection evictionHandler(Handler<Void> handler) {
     evictionHandler = handler;
     return this;
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
@@ -150,6 +150,13 @@ public class Http1ClientConnection extends Http1Connection implements io.vertx.c
   }
 
   @Override
+  public boolean isSsl() {
+    // Report the origin ssl, not the pipeline: in forward-proxy mode the TLS leg to an HTTPS proxy is
+    // itself the pipeline "ssl" handler.
+    return ssl;
+  }
+
+  @Override
   public io.vertx.core.http.impl.HttpClientConnection evictionHandler(Handler<Void> handler) {
     evictionHandler = handler;
     return this;
--- a/vertx-core/src/main/java/io/vertx/core/net/ProxyOptions.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/ProxyOptions.java
@@ -264,7 +264,7 @@ public class ProxyOptions {
   }
 
   /**
-   * Get the SSL options used for the connection to the proxy itself.
+   * Get the SSL options used between the client and the proxy.
    * <p>
    * Only relevant when {@link #getType()} is {@link ProxyType#HTTPS}.
    *
--- a/vertx-core/src/main/java/io/vertx/core/net/ProxyOptions.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/ProxyOptions.java
@@ -56,6 +56,7 @@ public class ProxyOptions {
   private String proxyAuthorization;
   private ProxyType type;
   private Duration connectTimeout;
+  private ClientSSLOptions sslOptions;
 
   /**
    * Default constructor.
@@ -80,6 +81,7 @@ public class ProxyOptions {
     proxyAuthorization = other.getProxyAuthorization();
     type = other.getType();
     connectTimeout = other.getConnectTimeout();
+    sslOptions = other.sslOptions != null ? other.sslOptions.copy() : null;
   }
 
   /**
@@ -258,6 +260,34 @@ public class ProxyOptions {
    */
   public ProxyOptions setConnectTimeout(Duration connectTimeout) {
     this.connectTimeout = connectTimeout;
+    return this;
+  }
+
+  /**
+   * Get the SSL options used for the connection to the proxy itself.
+   * <p>
+   * Only relevant when {@link #getType()} is {@link ProxyType#HTTPS}.
+   *
+   * @return the proxy SSL options, or {@code null}
+   */
+  public ClientSSLOptions getSslOptions() {
+    return sslOptions;
+  }
+
+  /**
+   * Set the SSL options used for the connection to the proxy itself, when the proxy is reached over
+   * SSL/TLS ({@link ProxyType#HTTPS}).
+   * <p>
+   * These options are resolved independently of the origin SSL options, so the proxy presents and
+   * is validated against its own certificate and trust store. When left unset for an
+   * {@link ProxyType#HTTPS} proxy, the connection is still established over SSL/TLS using the default
+   * trust source, with hostname verification enabled against the proxy host.
+   *
+   * @param sslOptions the proxy SSL options
+   * @return a reference to this, so the API can be used fluently
+   */
+  public ProxyOptions setSslOptions(ClientSSLOptions sslOptions) {
+    this.sslOptions = sslOptions;
     return this;
   }
 }
--- a/vertx-core/src/main/java/io/vertx/core/net/impl/VertxConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/impl/VertxConnection.java
@@ -12,6 +12,7 @@ package io.vertx.core.net.impl;
 
 import io.netty.buffer.Unpooled;
 import io.netty.channel.*;
+import io.netty.handler.ssl.SslHandler;
 import io.netty.handler.timeout.IdleStateEvent;
 import io.netty.util.ReferenceCountUtil;
 import io.netty.util.ReferenceCounted;
@@ -141,7 +142,8 @@ public class VertxConnection extends ConnectionBase {
   }
 
   protected boolean supportsFileRegion() {
-    return vertx.transport().supportFileRegion() && !isSsl() &&!isTrafficShaped();
+    boolean sslChannel = chctx.pipeline().get(SslHandler.class) != null;
+    return vertx.transport().supportFileRegion() && !sslChannel && !isTrafficShaped();
   }
 
   /**
--- a/vertx-core/src/main/java/io/vertx/core/net/impl/tcp/ChannelProvider.java
+++ b/vertx-core/src/main/java/io/vertx/core/net/impl/tcp/ChannelProvider.java
@@ -35,6 +35,7 @@ import io.vertx.core.internal.ContextInternal;
 import io.vertx.core.internal.VertxInternal;
 import io.vertx.core.internal.net.SslChannelProvider;
 import io.vertx.core.net.impl.SslEngineUtils;
+import io.vertx.core.internal.tls.ClientSslContextManager;
 import io.vertx.core.internal.tls.SslContextProvider;
 import io.vertx.core.net.ClientSSLOptions;
 import io.vertx.core.net.HostAndPort;
@@ -59,16 +60,19 @@ public final class ChannelProvider {
 
   private final Bootstrap bootstrap;
   private final SslContextProvider sslContextProvider;
+  private final ClientSslContextManager sslContextManager;
   private final ContextInternal context;
   private ProxyOptions proxyOptions;
   private String applicationProtocol;
 
   public ChannelProvider(Bootstrap bootstrap,
                          SslContextProvider sslContextProvider,
+                         ClientSslContextManager sslContextManager,
                          ContextInternal context) {
     this.bootstrap = bootstrap;
     this.context = context;
     this.sslContextProvider = sslContextProvider;
+    this.sslContextManager = sslContextManager;
   }
 
   /**
@@ -204,6 +208,7 @@ public final class ChannelProvider {
         switch (proxyType) {
           default:
           case HTTP:
+          case HTTPS:
             proxy = createHttpProxyHandler(proxyAddr, proxyUsername, proxyPassword, proxyAuthorization);
             break;
           case SOCKS5:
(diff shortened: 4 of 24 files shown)


vertx-core/src/test/java/io/vertx/tests/http/HttpsProxyTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+/**
+ * Tests for {@link ProxyType#HTTPS}: the connection to the proxy itself (leg 1) is established over
+ * TLS. The proxying semantics (absolute-URI {@code GET} forwarding for plain origins, {@code CONNECT}
+ * tunnel for TLS origins) are unchanged from {@link ProxyType#HTTP}.
+ */
+public class HttpsProxyTest extends HttpTestBase {
+
+  private Vertx proxyVertx;
+  private HttpProxy proxy;
```
vietj: instead of manually setting up the proxy, we should instead support that in the `io.vertx.test.proxy.Proxy` rule

vertx-core/src/main/java/io/vertx/core/net/impl/VertxConnection.java
```diff
@@ -141,7 +142,8 @@ protected void handleIdle(IdleStateEvent event) {
   }
 
   protected boolean supportsFileRegion() {
-    return vertx.transport().supportFileRegion() && !isSsl() &&!isTrafficShaped();
+    boolean sslChannel = chctx.pipeline().get(SslHandler.class) != null;
```
vietj, line 145: what is the motivation of this change ?

vertx-core/src/main/java/io/vertx/core/net/ProxyOptions.java
```diff
@@ -260,4 +262,32 @@ public ProxyOptions setConnectTimeout(Duration connectTimeout) {
     this.connectTimeout = connectTimeout;
     return this;
   }
+
+  /**
+   * Get the SSL options used for the connection to the proxy itself.
```
vietj:
```suggestion
   * Get the SSL options used between the client and the proxy.
```

vertx-core/src/test/java/io/vertx/tests/http/HttpsProxyTest.java
```diff
@@ -0,0 +1,274 @@
...
+      .setProxyOptions(new ProxyOptions().setType(ProxyType.HTTPS).setHost("localhost").setPort(proxy.port())));
+
+    try {
+      client.request(new RequestOptions().setMethod(HttpMethod.GET)
+          .setHost(DEFAULT_HTTP_HOST).setPort(DEFAULT_HTTP_PORT).setURI("/"))
+        .compose(req -> req.send())
+        .await();
+      fail("Expected the connection to the untrusted HTTPS proxy to fail");
+    } catch (Exception expected) {
+      // connect failed rather than hanging or silently downgrading
```
vietj: can we assert the failure type ?

vertx-core/src/test/java/io/vertx/tests/http/HttpsProxyTest.java
```diff
@@ -0,0 +1,274 @@
...
+        .setSslOptions(trustingProxySsl())));
+
+    try {
+      client.request(new RequestOptions().setMethod(HttpMethod.GET)
+          .setHost(DEFAULT_HTTP_HOST).setPort(DEFAULT_HTTP_PORT).setURI("/"))
+        .compose(req -> req.send())
+        .await();
+      fail("Expected hostname verification against the proxy to fail");
+    } catch (Exception expected) {
+      // verification rejected the mismatched proxy hostname
```
vietj: can we assert the failure type ?

vertx-core/src/test/java/io/vertx/tests/http/HttpsProxyTest.java
```diff
@@ -0,0 +1,274 @@
...
+    client = vertx.createHttpClient(new HttpClientOptions().setProxyOptions(httpsProxy(proxy.port())));
+
+    try {
+      client.request(new RequestOptions().setMethod(HttpMethod.GET)
+          .setHost(DEFAULT_HTTP_HOST).setPort(DEFAULT_HTTP_PORT).setURI("/"))
+        .compose(req -> req.send())
+        .await();
+      fail("Expected the TLS handshake against a plaintext proxy to fail");
+    } catch (Exception expected) {
+      // no silent downgrade: a HTTPS proxy type never falls back to plaintext
```
vietj: can we assert the failure type ?

vietj (changes requested):

vertx-core/src/main/java/io/vertx/core/net/impl/VertxConnection.java
gaganis: This function is used in sendFile to decide whether the connection is encrypted and if the connection is encrypted then it desides not to use zero-copy. The reasoning to my understanding is the the way zero copy writes to the socket bypasses the encryption which is why it is disabled.

In my PR I have opted for `isSsl()` to refer to how we connect to the origin and not if the connection is encrypted. 
So the addition covers the case where we have an SSL connection to the proxy but the origin is plain http. In this case we would have encryption on the pipeline. In that case `isSsl()` would be false but encryption would exist in the pipeline.

Maybe we should write a test to actually see how this is working. @vietj Are you aware of any tests that exercise this zero-copy function?

gaganis: Thank you for your comment @vietj. It led to finding multiple problems with my implementation that I have now hopefully fixed.  I have added a test to verify my motivation for the zero copy `io.vertx.tests.http.HttpsProxyTest#testHttpsProxy_sendFileThroughTunnel` and I have found the way that I addressed it above was not correct. Instead the fix should have been in `NetClientImpl.initChannel`. I have fixed that in the latest commited code.

This work also motivated me that I needed to be more precise with how I address my intention to keep `isSsl()` independent of the existence of an https proxy. The way this was working for not consistent before. I have added `io.vertx.tests.http.ProxyIsSslConsistencyTest` to confirm this and fixed the cases that were not working correctly.

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
```diff
@@ -149,6 +149,13 @@ public HostAndPort authority() {
     return authority;
   }
 
+  @Override
+  public boolean isSsl() {
+    // Report the origin ssl, not the pipeline: in forward-proxy mode the TLS leg to an HTTPS proxy is
```
vietj, line 154: so it means it should return false if the connection to the server uses a plain text connection over an HTTPS proxy ?

can we ensure that for both case there is an assertion for this ? I am thinking of

1. a plain text connection through an HTTPS proxy
2. an SSL connection through an HTTP proxy

vertx-core/src/main/java/io/vertx/core/net/impl/tcp/ChannelProvider.java
```diff
@@ -222,43 +227,83 @@ private void handleProxyConnect(SocketAddress remoteAddress, HostAndPort peerAdd
...
+          proxySslOptions = resolveProxySslOptions();
+          proxySslProviderFuture = sslContextManager.resolveSslContextProvider(proxySslOptions, context)
+            .map(p -> new SslChannelProvider(vertx, p, false, proxySslOptions.isUseHybridKeyExchangeProtocol()));
+        } else {
+          proxySslOptions = null;
+          proxySslProviderFuture = context.succeededFuture(null);
+        }
+
+        proxySslProviderFuture.onComplete(ar -> {
+          if (ar.failed()) {
```
vietj: we should encapsulate this code in a method instead of channelling it through the `proxySslProviderFuture` that is a succeeded future when there is no proxy to reduce the callback complexity in the normal case.

so when there is no https proxy we do call directly the method that setup the bootstrap, otherwise we do resolve the proxySslProviderFuture and continue from this.

vertx-core/src/main/java/io/vertx/core/net/impl/tcp/NetClientImpl.java
```diff
@@ -106,7 +107,8 @@ protected void initChannel(ChannelPipeline pipeline, boolean ssl) {
     if (logging != null) {
       pipeline.addLast("logging", new LoggingHandler(logging));
     }
-    if (ssl || !vertx.transport().supportFileRegion()) {
+    boolean sslChannel = pipeline.get(SslHandler.class) != null;
```
vietj, line 112: if we use this, shouldn't we remove the boolean ssl from the initChannel method ?

vertx-core/src/test/java/io/vertx/tests/http/HttpsProxyTest.java
```diff
@@ -0,0 +1,277 @@
...
+    server.requestHandler(req -> req.response().end("Hello from origin"));
+    startServer();
+  }
+
+  private void startHttpsOrigin(HttpVersion version) throws Exception {
+    server = vertx.createHttpServer(createBaseServerOptions()
+      .setSsl(true)
+      .setUseAlpn(version == HttpVersion.HTTP_2)
+      .setKeyCertOptions(Cert.SERVER_JKS.get()));
+    server.requestHandler(req -> req.response().end("Hello from origin"));
```
vietj: I think the server should respond with the boolean ssl status of its connection   in its message and we should assert it in tests.

vietj (changes requested):
A few comments have been made to this PR.

vietj:
@gaganis ping, can you also look at the conflict to resolve ?

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
gaganis: Yes, the idea is that a proxy should not impact the answer. In my opinion however I connect I should get the same answer. ie direct, http or https proxy. 

I was addressing this by testing all the permutations in `ProxyIsSslConsistencyTest` though I was not explicit on the answer. The idea in the beginning was to assert if the proxied connections report the same to direct connections. 

In this version I have made the assertions to the desired value and not compare to what the direct connection would report.

The cases you mention are covered by 
1. ProxyIsSslConsistencyTest#netSocket_plainOrigin_httpsProxy
    ProxyIsSslConsistencyTest#http1_plainOrigin_httpsProxy
    ProxyIsSslConsistencyTest#http2_plainOrigin_httpsProxy
     ProxyIsSslConsistencyTest#webSocket_plainOrigin_httpsProxy
2. ProxyIsSslConsistencyTest#netSocket_tlsOrigin_httpProxy
    ProxyIsSslConsistencyTest#http1_tlsOrigin_httpProxy
    ProxyIsSslConsistencyTest#http2_tlsOrigin_httpProxy
    ProxyIsSslConsistencyTest#webSocket_tlsOrigin_httpProxy

vertx-core/src/main/java/io/vertx/core/net/impl/tcp/ChannelProvider.java
gaganis: I have introduced a function. Please check if this is what you intended. Thanks

vertx-core/src/test/java/io/vertx/tests/http/ProxyIsSslConsistencyTest.java
```diff
@@ -0,0 +1,311 @@
...
+ * connection is the control: on failure it tells a proxy regression apart from one that was already
+ * broken without a proxy. {@code @WithProxy} selects one proxy kind per method, so there are two tests
+ * per case (HTTP proxy and HTTPS proxy). Origins are cleaned up by {@code tearDown} (vertx close), like
+ * other {@link HttpTestBase} tests; the per-type client logic ({@link #connectNetSocket},
+ * {@link #connectHttp}, {@link #connectWebSocket}) is reused across cases.
+ *
+ * <p>Covers the four main proxy-capable connection types (NetSocket, HTTP/1, HTTP/2, WebSocket)
+ * against a plain and a TLS origin.
+ */
+public class ProxyIsSslConsistencyTest extends HttpTestBase {
```
vietj: this test need to extend `HttpTestBase2` instead as we are trying to get away from `HttpTestBase`.

In addition we should be using the new config classes over the options classes whenever possible.

vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
vietj: that is clear thanks

vertx-core/src/test/java/io/vertx/tests/http/HttpsProxyTest.java
```diff
@@ -0,0 +1,281 @@
...
+import java.nio.file.Files;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Tests for {@link ProxyType#HTTPS}: the connection to the proxy itself (leg 1) is established over
+ * TLS. The proxying semantics (absolute-URI {@code GET} forwarding for plain origins, {@code CONNECT}
+ * tunnel for TLS origins) are unchanged from {@link ProxyType#HTTP}.
+ */
+public class HttpsProxyTest extends HttpTestBase {
```
vietj: we should extend `HttpTestBase2`

vietj (changes requested):
Tests should be modified

vertx-core/src/main/java/io/vertx/core/net/impl/tcp/ChannelProvider.java
vietj: yes that's more readable like that, thanks

vietj:
@gaganis a recent fix introduced a conflict, do you mind fixing it please ?

vietj (approved):

vietj:
thanks for the contribution @gaganis this is a great achievement!

gaganis:
Thank you too @vietj for your work to get this merged. I enjoyed working with you!

PR #6216, opened 2026-06-25
Update Post Quantum Cryptography (PQC) API

anavarr:
The single `useHybridKeyExchange` is removed in favor of two properties:

- `key-exchange-groups`: the user can specify which key exchange groups they want to use
- `pqc-enforcement-policy`: the user can specify how PQC will be enforced by Vert.x. The policies are:
	- **STRICT**: only PQC can be used to establish secure connections. If no SSL engine supports it, application won't start. If clients don't support it, SSL handshake fails
	- **CLIENT_NEGOTIATED**: if no SSL engine supports PQC, application won't start. If clients don't support it, falls back to other key exchange group
	- **RELAXED**: no enforcement whatsoever, traditional TLS negotiation

When PQC is required (STRICT or CLIENT_NEGOTIATED), Vert.x auto-selects a PQC-capable engine (JDK or OpenSSL) if none is explicitly configured. If none is available, application fails to start with a VertxException ("PQC enforcement policy requires X25519MLKEM768 but neither JDK nor OpenSSL supports it".

The resolution of key exchange groups and PQC enforcement policy (selecting an available SSL engine, adding X25519MLKEM768 to the list of supported groups for CLIENT_NEGOTIATED or restricting the list of supported groups for STRICT) is performed once in `SslContextManager`.

### DIFF
--- a/vertx-core/src/main/asciidoc/http.adoc
+++ b/vertx-core/src/main/asciidoc/http.adoc
@@ -82,14 +82,46 @@ To handle `h2` requests, TLS must be enabled:
 ----
 
 The rise of quantum computers will make key exchange protocols such as x25519 obsolete as they will be able to "crack" secret keys quickly.
-Vert.x proposes a quantum-safe key exchange protocol, x25519MLKEM768 (official recommendation of NIST) to ensure sessions over TLS are safe against quantum computers.
+Vert.x supports post-quantum cryptography (PQC) key exchange via X25519MLKEM768 (official recommendation of NIST) to ensure sessions over TLS are safe against quantum computers.
 
-Hybrid key exchange must be enabled with {@link io.vertx.core.net.SSLOptions#setUseHybridKeyExchangeProtocol(boolean)} and **requires OpenSSL** — it does not work with the JDK SSL engine. You must configure {@link io.vertx.core.net.OpenSSLEngineOptions} and add the following dependencies:
+PQC is configured with two properties:
+
+- {@link io.vertx.core.net.SSLOptions#setKeyExchangeGroups} — a list of key exchange group names (e.g. `["X25519MLKEM768", "X25519"]`)
+- {@link io.vertx.core.net.SSLOptions#setPqcEnforcementPolicy} — the enforcement policy: `RELAXED` (default), `CLIENT_NEGOTIATED`, or `STRICT`
+
+The three enforcement modes behave as follows:
+
+- **RELAXED**: uses the specified key exchange groups as-is; if none are specified, the SSL engine negotiates normally. No PQC enforcement.
+- **CLIENT_NEGOTIATED**: the server requires PQC support (X25519MLKEM768 is prepended to the groups if not present), but tolerates clients that do not support PQC.
+- **STRICT**: both server and client must support PQC. The key exchange groups are replaced with only X25519MLKEM768. Non-PQC clients will fail the TLS handshake.
+
+For both `STRICT` and `CLIENT_NEGOTIATED`, the application fails with either "PQC enforcement policy requires X25519MLKEM768 but the configured SSL engine does not support it" or "PQC enforcement policy requires X25519MLKEM768 but neither JDK nor OpenSSL support it" if pqc is not supported by the SSL engine.
+
+For `STRICT` and `CLIENT_NEGOTIATED` modes, you need an SslEngine supporting X25519MLKEM768. `JdkSslEngine` supports it starting with JDK 27 (preview feature), and OpenSsl supports it starting with OpenSsl 3.5. If no `SslEngine` is specified by the user, the first engine supporting pqc will be selected, starting with `JdkSslEngine`
+
+Using OpenSsl >= 3.5 requires the following dependencies:
 
 - `io.netty:netty-tcnative-classes` (version managed by the Netty BOM)
 - An OpenSSL provider such as `io.smallrye:smallrye-openssl`
 
-If OpenSSL is not available at runtime, the TLS handshake will fail rather than silently falling back to a non-quantum-safe key exchange.
+**PQC enforcement policy summary:**
+[cols="1,1,1,2", options="header"]
+|===
+| pqc-enforcement-policy | server-pqc-supports | client-pqc-supports | Result
+.3+| strict
+| false | required / true / false | **app fails** with error "PQC enforcement policy STRICT requires X25519MLKEM768 but the configured SSL engine does not support it"
+| true  | true / required         | app starts and ssl handshake succeeds
+| true  | false                   | app starts but **ssl handshake fails**
+
+.2+| client_negotiated
+| false | required / true / false | **app fails** with error "PQC enforcement policy STRICT requires X25519MLKEM768 but the configured SSL engine does not support it"
+| true  | required / true / false | app starts and ssl handshake succeeds
+
+.3+| relaxed
+| true  | required / true / false | app starts and ssl handshake succeeds
+| false | true / false            | app starts and ssl handshake succeeds
+| false | required                | app starts but **ssl handshake fails**
+|===
 
 With plain text (TLS is disabled), the server handles `h2c` requests that wants to upgrade connections presenting an
 HTTP/1.1 upgrade request to HTTP/2. It also accepts direct `h2c` (with prior knowledge) connections beginning with
--- a/vertx-core/src/main/asciidoc/http.adoc
+++ b/vertx-core/src/main/asciidoc/http.adoc
@@ -81,48 +81,6 @@ To handle `h2` requests, TLS must be enabled:
 {@link examples.HttpExamples#configurationOfAnH2Server}
 ----
 
-The rise of quantum computers will make key exchange protocols such as x25519 obsolete as they will be able to "crack" secret keys quickly.
-Vert.x supports post-quantum cryptography (PQC) key exchange via X25519MLKEM768 (official recommendation of NIST) to ensure sessions over TLS are safe against quantum computers.
-
-PQC is configured with two properties:
-
-- {@link io.vertx.core.net.SSLOptions#setKeyExchangeGroups} — a list of key exchange group names (e.g. `["X25519MLKEM768", "X25519"]`)
-- {@link io.vertx.core.net.SSLOptions#setPqcEnforcementPolicy} — the enforcement policy: `RELAXED` (default), `CLIENT_NEGOTIATED`, or `STRICT`
-
-The three enforcement modes behave as follows:
-
-- **RELAXED**: uses the specified key exchange groups as-is; if none are specified, the SSL engine negotiates normally. No PQC enforcement.
-- **CLIENT_NEGOTIATED**: the server requires PQC support (X25519MLKEM768 is prepended to the groups if not present), but tolerates clients that do not support PQC.
-- **STRICT**: both server and client must support PQC. The key exchange groups are replaced with only X25519MLKEM768. Non-PQC clients will fail the TLS handshake.
-
-For both `STRICT` and `CLIENT_NEGOTIATED`, the application fails with either "PQC enforcement policy requires X25519MLKEM768 but the configured SSL engine does not support it" or "PQC enforcement policy requires X25519MLKEM768 but neither JDK nor OpenSSL support it" if pqc is not supported by the SSL engine.
-
-For `STRICT` and `CLIENT_NEGOTIATED` modes, you need an SslEngine supporting X25519MLKEM768. `JdkSslEngine` supports it starting with JDK 27 (preview feature), and OpenSsl supports it starting with OpenSsl 3.5. If no `SslEngine` is specified by the user, the first engine supporting pqc will be selected, starting with `JdkSslEngine`
-
-Using OpenSsl >= 3.5 requires the following dependencies:
-
-- `io.netty:netty-tcnative-classes` (version managed by the Netty BOM)
-- An OpenSSL provider such as `io.smallrye:smallrye-openssl`
-
-**PQC enforcement policy summary:**
-[cols="1,1,1,2", options="header"]
-|===
-| pqc-enforcement-policy | server-pqc-supports | client-pqc-supports | Result
-.3+| strict
-| false | required / true / false | **app fails** with error "PQC enforcement policy STRICT requires X25519MLKEM768 but the configured SSL engine does not support it"
-| true  | true / required         | app starts and ssl handshake succeeds
-| true  | false                   | app starts but **ssl handshake fails**
-
-.2+| client_negotiated
-| false | required / true / false | **app fails** with error "PQC enforcement policy STRICT requires X25519MLKEM768 but the configured SSL engine does not support it"
-| true  | required / true / false | app starts and ssl handshake succeeds
-
-.3+| relaxed
-| true  | required / true / false | app starts and ssl handshake succeeds
-| false | true / false            | app starts and ssl handshake succeeds
-| false | required                | app starts but **ssl handshake fails**
-|===
-
 With plain text (TLS is disabled), the server handles `h2c` requests that wants to upgrade connections presenting an
 HTTP/1.1 upgrade request to HTTP/2. It also accepts direct `h2c` (with prior knowledge) connections beginning with
 the `PRI * HTTP/2.0\r\nSM\r\n` preface.
--- a/vertx-core/src/main/asciidoc/ssl.adoc
+++ b/vertx-core/src/main/asciidoc/ssl.adoc
@@ -66,3 +66,46 @@ OpenSSL provided much better performances and CPU usage than the JDK engine.
 {@link examples.SslExamples#exampleSSLEngine}
 ----
 
+=== Post Quantum Cryptography
+
+The rise of quantum computers will make key exchange protocols such as x25519 obsolete as they will be able to "crack" secret keys quickly.
+Vert.x supports post-quantum cryptography (PQC) key exchange via X25519MLKEM768 (official recommendation of NIST) to ensure sessions over TLS are safe against quantum computers.
+
+PQC is configured with two properties:
+
+- {@link io.vertx.core.net.SSLOptions#setKeyExchangeGroups} — a list of key exchange group names (e.g. `["X25519MLKEM768", "X25519"]`)
+- {@link io.vertx.core.net.SSLOptions#setPqcEnforcementPolicy} — the enforcement policy: `RELAXED` (default), `CLIENT_NEGOTIATED`, or `STRICT`
+
+The three enforcement modes behave as follows:
+
+- **RELAXED**: uses the specified key exchange groups as-is; if none are specified, the SSL engine negotiates normally. No PQC enforcement.
+- **CLIENT_NEGOTIATED**: the server requires PQC support (X25519MLKEM768 is prepended to the groups if not present), but tolerates clients that do not support PQC.
+- **STRICT**: both server and client must support PQC. The key exchange groups are replaced with only X25519MLKEM768. Non-PQC clients will fail the TLS handshake.
+
+For both `STRICT` and `CLIENT_NEGOTIATED`, the application fails with either "PQC enforcement policy requires X25519MLKEM768 but the configured SSL engine does not support it" or "PQC enforcement policy requires X25519MLKEM768 but neither JDK nor OpenSSL support it" if pqc is not supported by the SSL engine.
+
+For `STRICT` and `CLIENT_NEGOTIATED` modes, you need an SslEngine supporting X25519MLKEM768. `JdkSslEngine` supports it starting with JDK 27 (preview feature), and OpenSsl supports it starting with OpenSsl 3.5. If no `SslEngine` is specified by the user, the first engine supporting pqc will be selected, starting with `JdkSslEngine`
+
+Using OpenSsl >= 3.5 requires the following dependencies:
+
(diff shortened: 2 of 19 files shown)


vertx-core/src/main/asciidoc/http.adoc
```diff
@@ -82,14 +82,46 @@ To handle `h2` requests, TLS must be enabled:
 ----
 
 The rise of quantum computers will make key exchange protocols such as x25519 obsolete as they will be able to "crack" secret keys quickly.
-Vert.x proposes a quantum-safe key exchange protocol, x25519MLKEM768 (official recommendation of NIST) to ensure sessions over TLS are safe against quantum computers.
+Vert.x supports post-quantum cryptography (PQC) key exchange via X25519MLKEM768 (official recommendation of NIST) to ensure sessions over TLS are safe against quantum computers.
 
-Hybrid key exchange must be enabled with {@link io.vertx.core.net.SSLOptions#setUseHybridKeyExchangeProtocol(boolean)} and **requires OpenSSL** — it does not work with the JDK SSL engine. You must configure {@link io.vertx.core.net.OpenSSLEngineOptions} and add the following dependencies:
+PQC is configured with two properties:
+
+- {@link io.vertx.core.net.SSLOptions#setKeyExchangeGroups} — a list of key exchange group names (e.g. `["X25519MLKEM768", "X25519"]`)
```
vietj: I think overall we need to move this hybrid section section to `ssl.adoc` instead of being part of HTTP

vertx-core/src/main/java/io/vertx/core/internal/tls/SslContextManager.java
```diff
@@ -110,6 +146,48 @@ public static SSLEngineOptions resolveEngineOptions(SSLEngineOptions engineOptio
     return engineOptions;
   }
 
+  private static final String X25519MLKEM768 = "X25519MLKEM768";
+
+  // If the user didn't specify any group we return a default list, with X25519MLKEM768 prepended.
+  // This list is the default when weak named curves are removed (see https://www.java.com/en/configure_crypto.html)
+  private static final List<String> DEFAULT_KEY_EXCHANGE_GROUPS = List.of(X25519MLKEM768, "X25519", "secp256r1", "x448",
```
NilsRenaud: I'm wondering whether we should maintain this list, or reuse the one defined by java.security or the `java.tls.namedGroup` variable, adding PQ algorithms first when needed.

vertx-core/src/main/java/io/vertx/core/internal/tls/SslContextManager.java
```diff
@@ -110,6 +146,48 @@ public static SSLEngineOptions resolveEngineOptions(SSLEngineOptions engineOptio
     return engineOptions;
   }
 
+  private static final String X25519MLKEM768 = "X25519MLKEM768";
```
NilsRenaud: There are more PQ key exchanges than this one, Java 27 [have](https://openjdk.org/jeps/527): 
- X25519MLKEM768: Hybrid scheme combining ECDHE with [X25519](https://www.rfc-editor.org/rfc/rfc7748) and ML-KEM-768,
- SecP256r1MLKEM768: Hybrid scheme combining ECDHE using the secp256r1 curve with ML-KEM-768, and
- SecP384r1MLKEM1024: Hybrid scheme combining ECDHE using the secp384r1 curve with ML-KEM-1024.

That being said, the JEP says `X25519MLKEM768` should be the default and does not enable other algorithms by default. So we're probably good with this one by default, but IMO it can't be the only one checked by `isPqcAvailable()`. And maybe a better strategy would be to ask the SslEngine which are its supported algorithms, ranking first the one to use by default ?

vertx-core/src/main/java/io/vertx/core/internal/tls/SslContextManager.java
```diff
@@ -110,6 +146,48 @@ public static SSLEngineOptions resolveEngineOptions(SSLEngineOptions engineOptio
...
+   */
+  public static List<String> resolveKeyExchangeGroups(List<String> groups, PqcEnforcementPolicy pqcPolicy) {
+    if (pqcPolicy == null) {
+      pqcPolicy = PqcEnforcementPolicy.RELAXED;
+    }
+    switch (pqcPolicy) {
+      case STRICT:
+        if (groups != null && !groups.isEmpty()) {
+          if (!(groups.size() == 1 && groups.contains(X25519MLKEM768))) {
+            log.warn("PQC enforcement policy is STRICT: overriding key exchange groups " + groups + " with [" + X25519MLKEM768 + "]");
```
NilsRenaud: One could only expose `SecP256r1MLKEM768`, or  `X25519MLKEM768, SecP256r1MLKEM768` for some reasons and should not have its choice overriden

NilsRenaud:
As I understand the idea of having a `pqc-enforcement-policy`, I think it's quire redundant with `key-exchange-groups`, and makes the configuration tricky (I can have my key exchange groups property overriden, or the startup failed because of a incompatible configuration).

If one wants a strict `pqc-enforcement-policy`, then they only have to list the key exchange groups they want, such as : `X25519MLKEM768, SecP256r1MLKEM768`

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
```diff
@@ -57,6 +58,24 @@ public static synchronized boolean isAlpnAvailable() {
     return jdkAlpnAvailable;
   }
 
+  /**
+   * @return if PQC key exchange (X25519MLKEM768) is available via the JDK SSL engine
```
NilsRenaud, line 62: This method is trickier than that, what if one wants to use a PQ algorithm that's only implemented by a custom Security Provider ?

IMO we should not guess, and simply let the user write the ordered key exchange algorithm list they want to use and react on algorithm not available (or check their presence upfront, but not sure it's possible 🤔 )

vertx-core/src/main/java/io/vertx/core/internal/net/SslChannelProvider.java
```diff
@@ -111,19 +111,20 @@ private SslHandler createServerSslHandler(List<String> applicationProtocols, lon
...
+      keyExchangeGroups, remoteAddress);
   }
 
-  static void applyHybridCurves(SslHandler sslHandler) {
+  static void applyKeyExchangeGroups(SslHandler sslHandler, List<String> groups) {
+    String curvesList = String.join(":", groups);
     try {
       long sslPtr = ((ReferenceCountedOpenSslEngine) sslHandler.engine()).sslPointer();
-      boolean success = SSL.setCurvesList(sslPtr, "X25519MLKEM768");
+      boolean success = SSL.setCurvesList(sslPtr, curvesList);
```
NilsRenaud: This is a `netty-tc-native` method, will this work with JDK implementation ?
I've created a Netty's issue to have the curve list setter as first class citizen in Netty (https://github.com/netty/netty/issues/16999).
Maybe we should wait for their answer first ?

anavarr: you're absolutely right I forgot this path. Since the api was added in JDK 21 we have to use reflection or method handles. If the user uses a version of JdkSSL inferior, setNamedGroups won't be available and we should either crash or warn them. 

We have the pqc enforcement policy to decide what to do if the user specifically wants to use pqc, but in the case they simply provide a list of named groups and doesn't use a compatible ssl engine I think we should always fail right?

@vietj

anavarr:
Thank you for your comments @NilsRenaud!
Correct me if I'm wrong but I think your points boil down to 2 problems:

1. We focus entirely on X25519MLKEM768 as the sole hybrid key exchange protocol but should rather use an ordered list (easy fix).
2. PQC enforcement policy leaves a lot of guess work when the user doesn't specify neither an ssl engine nor a list of named groups.

Item **1** can be solved easily by maintaining a list and updating it as ssl engines versions increase and new hybrid key exchange protocols appear. It would require little work.
Now for item **2**. I still think having this property is a good thing as it allows for a very easy experience: without having set an ssl engine or a list of named groups, the user can set a PQC enforcement policy and Vert.x will enumerate the engines and look for available hybrid key exchange protocols if needed. I don't think it is redundant. However, as one of your comment pointed out, setting this policy along with a set of named groups (especially custom ones) can create friction if some groups are not known by Vert.x to be PQC compliant. I reckon this situation would be rare, and the user would simply have to turn off PQC enforcement. In my mind it is not a strong enough argument to dismiss the pros of having a PQC enforcement policy property. 

wdyt @NilsRenaud, @vietj and @cescoffier

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
anavarr: see https://github.com/eclipse-vertx/vert.x/pull/6216#issuecomment-4831434469. We should definitely add more PQC compliant named groups and not check only on MLKEM, but if a user wants to use obscure PQ algorithm only implemented by custom security providers and not known by Vert.x to be PQ compliant, they should disable PQ enforcement policy and rely on their own vigilance.

vertx-core/src/main/java/io/vertx/core/internal/net/SslChannelProvider.java
vietj: we can have code in the src/main/java21 tree that provides this if that is JDK 21, without using reflection.

vertx-core/src/main/java/io/vertx/core/internal/tls/SslContextManager.java
vietj: this list would be available for all JDK ?

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
vietj: in which JDK is the class `KEM` available ?

vertx-core/src/main/java/io/vertx/core/net/PqcEnforcementPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+
+package io.vertx.core.net;
+
+import io.vertx.codegen.annotations.VertxGen;
+
+/**
+ * Policy for enforcing post-quantum cryptography (PQC) key exchange.
+ */
+@VertxGen
+public enum PqcEnforcementPolicy {
```
vietj, line 22: please add `@Unstable`

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
anavarr: 27+ I believe

vertx-core/src/main/java/io/vertx/core/internal/tls/SslContextManager.java
anavarr: it is null by default. The simplest option is to get the SSL engine's default `namedGroups` instead.

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
NilsRenaud: @vietj : The KEM class has been added in 21: https://docs.oracle.com/en/java/javase/21/docs/api//java.base/javax/crypto/KEM.html,
the ML-KEM algorithm in java 24: https://openjdk.org/jeps/496
The support of ML-KEM algorithms in TLS 1.3 in java 27: https://openjdk.org/jeps/527

@anavarr 
> but if a user wants to use obscure PQ algorithm only implemented by custom security providers and not known by Vert.x to be PQ compliant, they should disable PQ enforcement policy and rely on their own vigilance.

Bypassing vert.x PQ enforcement to use their own PQC algorithms, which could be a future well-known NIST approved algorithm, is not very user-friendly in my opinion. And I think either : 
- user don't care about PQC and rely on JDK's default settings, PQ algorithm will be the first choice in JDK 27 anyway
- user care about PQC and will be able to set the list of algorithms they want.

So we could replace the enforcement option by a note in the doc instead.

anavarr: By the time it becomes a well-known NIST approved algorithm it will be known as a PQC algorithm in Vert.x.
I see this property as a complement to the list of named groups. If you know what you're doing you just leave it to its default RELAXED mode. If you don't, you set it to STRICT and are forced to use an SSL engine providing PQ compliant algorithms.

NilsRenaud:
Thanks @anavarr you summarized my points perfectly. 

I understand your thinking, but I still think the PQC enforcement policy is not necessary, makes the API unclear for future use case, and forces Vert.x to maintain a list of PQC-resistant alogrithms as new spawn, or become broken, in the future.
For example, what if we found a vulnerability in `X25519MLKEM768` making it NOT PQC-resistant ? Users will either have a false sense of security because they rely on the PQC enforcement policy, or disable it since it prevents them to set a really PQC-resistant algorithm.

I'm not maintainer though, and this PR is still valuable since the PQC enforcement is optional ! So thanks for this work that everyone will benefit of anyway :)

vietj:
> Thanks @anavarr you summarized my points perfectly.
> 
> I understand your thinking, but I still think the PQC enforcement policy is not necessary, makes the API unclear for future use case, and forces Vert.x to maintain a list of PQC-resistant alogrithms as new spawn, or become broken, in the future. For example, what if we found a vulnerability in `X25519MLKEM768` making it NOT PQC-resistant ? Users will either have a false sense of security because they rely on the PQC enforcement policy, or disable it since it prevents them to set a really PQC-resistant algorithm.
> 
> I'm not maintainer though, and this PR is still valuable since the PQC enforcement is optional ! So thanks for this work that everyone will benefit of anyway :)

if we can afford to introduce an API that is not strictly necessary, we should do it

cescoffier:
Ignore the JDK for a bit, as we are hoping for Java 29 (as LTS) but we need to deploy quantum safe TLS before that time. So, we will rely on OpenSSL, but it requires your operating system to provide the right version. Without this enforcement we could lead to an environment where the app will be running but fail every handshake, or degrade to non quantum safe. The enforcement will check during the startup that you can handle the handshake. The list of supported key exchange algorithms will be strictly the ones from the NIST. 

When the JDK support will be available, we would verify that the key exchange algorithms are available (because it may depends on your distribution. For example, Semuru already provide what we would need).

At some point in the future (hard to say when), this enforcement will become unnecessary because all environments and all JVMs will provide what is required. But we are not there yet. Lots of system will not be quantum safe capable for a bit, and the enforcement is there to make sure you don't fall into that trap.

anavarr:
Well I believe everything is ready for PQC in the JDK anyway. The only modification to do will be to change `isPqcAvailable` in `JdkSslOptions` (it returns false currently). But every path has a branch for both openssl and jdkssl engines. I added the two other pq-compliant key exchange protocols to the mix. The default list of named groups is that from the NIST. I used multi-release for specific methods to set named groups. Users are warned if they try to set named groups for jdk <20.

vertx-core/src/main/asciidoc/ssl.adoc
```diff
@@ -66,3 +66,46 @@ OpenSSL provided much better performances and CPU usage than the JDK engine.
...
 ----
 
+=== Post Quantum Cryptography
+
+The rise of quantum computers will make key exchange protocols such as x25519 obsolete as they will be able to "crack" secret keys quickly.
+Vert.x supports post-quantum cryptography (PQC) key exchange via X25519MLKEM768 (official recommendation of NIST) to ensure sessions over TLS are safe against quantum computers.
+
+PQC is configured with two properties:
+
+- {@link io.vertx.core.net.SSLOptions#setKeyExchangeGroups} — a list of key exchange group names (e.g. `["X25519MLKEM768", "X25519"]`)
```
NilsRenaud, line 76: maybe `An ordered list` instead of `A list`, or anything that hint that there is a priority order

vertx-core/src/main/asciidoc/ssl.adoc
```diff
@@ -66,3 +66,46 @@ OpenSSL provided much better performances and CPU usage than the JDK engine.
...
+
+The three enforcement modes behave as follows:
+
+- **RELAXED**: uses the specified key exchange groups as-is; if none are specified, the SSL engine negotiates normally. No PQC enforcement.
+- **CLIENT_NEGOTIATED**: the server requires PQC support (X25519MLKEM768 is prepended to the groups if not present), but tolerates clients that do not support PQC.
+- **STRICT**: both server and client must support PQC. The key exchange groups are replaced with only X25519MLKEM768. Non-PQC clients will fail the TLS handshake.
+
+For both `STRICT` and `CLIENT_NEGOTIATED`, the application fails with either "PQC enforcement policy requires X25519MLKEM768 but the configured SSL engine does not support it" or "PQC enforcement policy requires X25519MLKEM768 but neither JDK nor OpenSSL support it" if pqc is not supported by the SSL engine.
+
+For `STRICT` and `CLIENT_NEGOTIATED` modes, you need an SslEngine supporting X25519MLKEM768. `JdkSslEngine` supports it starting with JDK 27 (preview feature), and OpenSsl supports it starting with OpenSsl 3.5. If no `SslEngine` is specified by the user, the first engine supporting pqc will be selected, starting with `JdkSslEngine`
```
NilsRenaud, line 87: Not sure it's worth adding this, but one can add BouncyCastle's `org.bouncycastle.bcprov-jdk18on` + `org.bouncycastle.bctls-jdk18on` dependancies and add `BCJSSE` and `BC` as primary security providers to have a transparent support of PQ-resistant algorithms as would JDK 27+ do. This works today, on java 21 (haven't tried earlier versions)

vertx-core/src/main/java/io/vertx/core/internal/net/SslEngineHelper.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Pre Java 21 implementation: named groups not supported.
+ */
+public class SslEngineHelper {
+
+  private static final Logger log = LoggerFactory.getLogger(SslEngineHelper.class);
+
+  public static void applyNamedGroups(SSLEngine engine, List<String> groups) {
+    log.warn("Cannot apply key exchange groups " + groups + " on JDK SSL engine: requires JDK 20+");
```
NilsRenaud, line 27: That's not true, the method you call in the java 21 version exists since [Java 1.6](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/javax/net/ssl/SSLEngine.html#setSSLParameters(javax.net.ssl.SSLParameters)) that's only the support of hybrid PQ key exchange algorithms which landed in java 27.
And even this support depends of the SecurityProvider installed at the JVM level (this can be achieved with BouncyCastle for older version of java !)

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
```diff
@@ -57,6 +57,14 @@ public static synchronized boolean isAlpnAvailable() {
     return jdkAlpnAvailable;
   }
 
+  /**
+   * TODO implement it when JDK add supports
```
NilsRenaud, line 61: As said earlier, it depends on the security provider installed :s

vertx-core/src/main/java/io/vertx/core/internal/net/SslEngineHelper.java
anavarr: Hm no `setNamedGroups` definitely does not compile before jdk 20
see https://docs.oracle.com/en/java/javase/19/docs/api/java.base/javax/net/ssl/SSLParameters.html and https://docs.oracle.com/en/java/javase/21/docs/api/java.base/javax/net/ssl/SSLParameters.html

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
anavarr: The method is in the `JdkSSLEngineOptions` file, I think it is clear enough we are referring to `JdkSSLEngine` and thus the `SunJSSE` provider. I can refine the comment but imo it is explicit enough

vertx-core/src/main/java/io/vertx/core/internal/net/SslEngineHelper.java
NilsRenaud: Woops you're right, I got mixed up with another class ><'

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
NilsRenaud: > JdkSSLEngine and thus the SunJSSE provider.

That's where hides the devil: one can change the provider, so BouncyCastle works transparently through the JdkSSLEngine when installed as Security Provider, because Java will delegate TLS work to it as soon as it implements the JSSE interface (which it does).

vertx-core/src/main/asciidoc/ssl.adoc
vietj: this information is interesting but would quickly become stale as technology evolves.

vertx-core/src/main/java/io/vertx/core/net/JdkSSLEngineOptions.java
vietj: I think we can add a comment that the method returns false currently but might change in the future.

vietj:
See #6236 instead.

PR #6231, opened 2026-07-08
Add a test verifying that when an HTTP server request is paused, the server will stop reading bytes from the network when too many requests are cumulated.

vietj:


vietj:
the new test seems to have issues

Issue #6234, opened 2026-07-08
Add per-connection attachment storage to HttpConnectionInternal


Harshal96:
### Describe the feature

vert-x3/vertx-web#2910 needs per-connection state (announced Alt-Svc origins) that
must not leak after the connection closes. Current workaround lives entirely in
vertx-web: a `WeakHashMap<HttpConnection, T>` (see HttpConnectionLocal.java in that PR).
https://github.com/vert-x3/vertx-web/pull/2910#discussion_r3542072153 (comment by @vietj):
"actually I meant I would add a map like structure on vertx internal HTTP connection
object in vertx-core which would solve the issue more directly"

#### Proposal
Add attachment storage to the `HttpConnectionInternal` contract (new interface,
following the existing `HttpServerRequestInternal` / `HttpClientRequestInternal`
pattern in `io.vertx.core.internal.http`), scoped to the connection's lifecycle so
entries are cleared automatically on close:
```
    <T> T get(Object key);
    void set(Object key, Object value);
```

Downstream modules (vertx-web, etc.) can then stash per-connection state via this
contract instead of each reinventing weak-map/close-handler bookkeeping.

### Contribution

Myself

PR #6235, opened 2026-07-08
feat: add per-connection attachment storage to HttpConnectionInternal

Harshal96:
Motivation:

vert-x3/vertx-web#2910 needs per-connection state (announced Alt-Svc origins) scoped to the connection's lifecycle. The current workaround in vertx-web is a `WeakHashMap<HttpConnection, T>`, which can't reliably clean up stale connections (see discussion in that PR).

## Changes

New `HttpConnectionInternal` interface in `io.vertx.core.internal.http`, following the existing `HttpServerRequestInternal` / `HttpClientRequestInternal` pattern:
  ```java
  <T> T get(Object key);
  void set(Object key, Object value);
  ```
  
Backed by a lazily-created map stored directly on the connection object (not an external map keyed by connection), so entries are naturally scoped to the connection's own lifecycle — no leak risk, no weak references needed.

Wired into all four connection implementations:
ConnectionBase (shared by Http1Connection, Http2ConnectionImpl, Http2MultiplexConnection), cleared in handleClosed(). Http3Connection, which doesn't extend ConnectionBase, gets its own field and clears it in its own handleClosed().

Fixes [#6234](https://github.com/Harshal96/vert.x/issues/6234)

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
+++ b/vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2011-2026 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.internal.http;
+
+import io.vertx.core.http.HttpConnection;
+
+/**
+ * Extends to expose internal methods that are necessary for integration.
+ */
+public interface HttpConnectionInternal extends HttpConnection {
+
+  /**
+   * Get an attachment previously stored with {@link #set(Object, Object)}.
+   * <p/>
+   * Attachments are scoped to the lifecycle of the connection: they are discarded when the
+   * connection closes and do not need to be cleaned up manually.
+   *
+   * @param key the attachment key
+   * @return the attachment value, or {@code null} if absent
+   */
+  <T> T get(Object key);
+
+  /**
+   * Store an attachment on this connection, associated with {@code key}.
+   *
+   * @param key the attachment key
+   * @param value the attachment value
+   */
+  void set(Object key, Object value);
+
+}
--- a/vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
+++ b/vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
@@ -18,7 +18,7 @@ import io.vertx.core.http.HttpConnection;
 public interface HttpConnectionInternal extends HttpConnection {
 
   /**
-   * Get an attachment previously stored with {@link #set(Object, Object)}.
+   * Get an attachment previously stored with {@link #attach(Object, Object)}.
    * <p/>
    * Attachments are scoped to the lifecycle of the connection: they are discarded when the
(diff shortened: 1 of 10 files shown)


Harshal96:
@vietj something like this?

vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Get an attachment previously stored with {@link #set(Object, Object)}.
+   * <p/>
+   * Attachments are scoped to the lifecycle of the connection: they are discarded when the
+   * connection closes and do not need to be cleaned up manually.
+   *
+   * @param key the attachment key
+   * @return the attachment value, or {@code null} if absent
+   */
+  <T> T get(Object key);
```
vietj: name it `attachment`

vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Store an attachment on this connection, associated with {@code key}.
+   *
+   * @param key the attachment key
+   * @param value the attachment value
+   */
+  void set(Object key, Object value);
```
vietj: name it `attach`

vietj (changes requested):

vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
Harshal96: done

vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
Harshal96: done

vertx-core/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
```diff
@@ -74,6 +75,7 @@ public abstract class ConnectionBase {
   private Future<Void> closeFuture;
   private long remainingBytesRead;
   private long remainingBytesWritten;
+  private Map<Object, Object> attachments;
```
vietj: we should not add this to connection base because it will affect also other classes like QUIC streams

vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+ */
+package io.vertx.core.internal.http;
+
+import io.vertx.core.http.HttpConnection;
+
+/**
+ * Extends to expose internal methods that are necessary for integration.
+ */
+public interface HttpConnectionInternal extends HttpConnection {
```
vietj, line 18: this interface should be extended by:

- io.vertx.core.net.impl.HttpClientConnection
- io.vertx.core.http.impl.HttpServerConnection

vietj (changes requested):

vietj:
ping @Harshal96

vertx-core/src/main/java/io/vertx/core/internal/http/HttpConnectionInternal.java
Harshal96: done

vertx-core/src/main/java/io/vertx/core/net/impl/ConnectionBase.java
Harshal96: done

vietj (changes requested):
Can you instead of using a map with synchronization around instead use a ConcurrentHashMap implementation ?

Harshal96:
@vietj done

PR #6240, opened 2026-07-10
#6238 Improve HTTP/2 flow control defaults

### LINKED ISSUE #6238: Large HTTP/2 request bodies (gRPC 10MB+) are extremely slow
### Version

5.0.0

### Describe the bug

When serving large payloads over HTTP/2 (e.g. gRPC unary RPC with 10MB+ messages), requests are extremely slow or appear to hang. The server-side handler may not be invoked until the full body slowly trickles in.

This affects applications using Vert.x HTTP server + vertx-grpc unified mode (same port as HTTP). Legacy gRPC-Java/Netty server with a larger flow-control window completes the same workload in ~1s vs ~25s+ with Vert.x HTTP/2.

Related Quarkus issue: https://github.com/quarkusio/quarkus/issues/48526  

### Root cause (investigation)

The default HTTP/2 connection window is too small for large messages:

1. **`Http2ServerConfig.java`** — `connectionWindowSize` defaults to `-1`:
   - `vertx-core/src/main/java/io/vertx/core/http/Http2ServerConfig.java:38`

2. **`HttpServerOptions.java`** — `DEFAULT_HTTP2_CONNECTION_WINDOW_SIZE = -1`:
   - `vertx-core/src/main/java/io/vertx/core/http/HttpServerOptions.java:127`

3. **`Http2Settings.java`** — stream initial window = **65535 bytes (64KB)**:
   - `vertx-core/src/main/java/io/vertx/core/http/Http2Settings.java:56`

4. **`HttpServerConnectionHandler.java`** — larger window is only applied when `connectionWindowSize > 0`:
   - `vertx-core/src/main/java/io/vertx/core/http/impl/tcp/HttpServerConnectionHandler.java:92-94`
   - With default `-1`, `setWindowSize()` is never called.

5. **Flow-control path** — WINDOW_UPDATE only after app consumes bytes:
   - `DefaultHttp2Stream.java:91-94` → `Http2ConnectionImpl.java:464-469` → `VertxHttp2ConnectionHandler.java:274-278`

With a 64KB window, a 10MB message requires hundreds of WINDOW_UPDATE round-trips before the application receives the full body.

arnabnandy7:
Motivation:

Large HTTP/2 request and response bodies, such as 10 MiB+ gRPC unary messages, can be extremely slow because Vert.x uses the HTTP/2 protocol-default 65,535-byte stream and connection flow-control windows. Without dynamic window auto-tuning, small static windows repeatedly stall large transfers while waiting for WINDOW_UPDATE frames.

Modifications:

- Increase the default initial HTTP/2 stream window from 64 KiB to 1 MiB for HTTP servers and clients.
- Increase the default HTTP/2 connection window to 16 MiB for HTTP servers and clients.
- Document stream-level and connection-level flow control defaults and their memory/throughput trade-off.
- Add tests for the server and client default window sizes.

Result:

Large HTTP/2 transfers and concurrent streams can make progress with fewer flow-control update cycles.

Fixes #6238

Conformance:

I have signed the Eclipse Contributor Agreement and confirm that this contribution adheres to the Vert.x code style guidelines.

### DIFF
--- a/vertx-core/src/main/asciidoc/http.adoc
+++ b/vertx-core/src/main/asciidoc/http.adoc
@@ -92,7 +92,17 @@ WARNING: browsers do not support `h2c`, for serving websites you should use `h2`
 When a server accepts an HTTP/2 connection, it sends to the client its {@link io.vertx.core.http.Http2ServerConfig#getInitialSettings initial settings}. These settings define how the client can use the connection, the default initial settings for a server are:
 
 - {@link io.vertx.core.http.Http2Settings#getMaxConcurrentStreams}: `100` as recommended by the HTTP/2 RFC
+- {@link io.vertx.core.http.Http2Settings#getInitialWindowSize}: `1 MiB`
 - the default HTTP/2 settings values for the remaining settings
+
+HTTP/2 flow control uses two independent receive windows. The initial window size in
+{@link io.vertx.core.http.Http2ServerConfig#getInitialSettings} applies to each stream, while
+{@link io.vertx.core.http.Http2ServerConfig#getConnectionWindowSize} applies to the combined data received by all
+streams on a connection. The connection window defaults to `16 MiB`.
+
+Applications receiving large request bodies, such as large gRPC messages, can tune both values. Larger windows improve
+throughput on connections with significant bandwidth or latency, at the cost of allowing more data to be buffered before
+back-pressure is applied.
 
 === Configuring an HTTP/3 server
 
--- a/vertx-core/src/main/asciidoc/http.adoc
+++ b/vertx-core/src/main/asciidoc/http.adoc
@@ -989,6 +989,13 @@ To perform `h2` requests, TLS must be enabled:
 
 `{@link io.vertx.core.http.Http2ClientConfig}` captures the configuration of HTTP/2 specific aspects.
 
+The HTTP/2 client receive window defaults to `1 MiB` per stream and `16 MiB` for the connection. Applications receiving
+large response bodies can tune the stream window with
+{@link io.vertx.core.http.Http2ClientConfig#getInitialSettings} and the connection window with
+{@link io.vertx.core.http.Http2ClientConfig#setConnectionWindowSize(int)}. Larger windows improve throughput on
+connections with significant bandwidth or latency, at the cost of allowing more data to be buffered before back-pressure
+is applied.
+
 === Configuring an HTTP/3 client
 
 Vert.x supports HTTP/3 over QUIC, we recommend reading the <<quic_client,QUIC client section>>.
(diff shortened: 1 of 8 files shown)


vietj:
can you check what are the defaults in other libraries ?

arnabnandy7:
> can you check what are the defaults in other libraries ?

checking

arnabnandy7:
@vietj I checked a few other implementations:

- Netty uses the HTTP/2 default of 64 KiB.
- gRPC-Java starts at 1 MiB and automatically adjusts the window using BDP estimation.
- gRPC-Go also uses dynamic flow-control tuning by default.
- HAProxy defaults to 64 KiB but recommends 1 MiB for large uploads over higher-latency connections.

Since Vert.x does not currently auto-tune the window, I used a larger static default to avoid repeatedly stalling on 64 KiB updates. The 1 MiB stream window is consistent with gRPC-Java.

The 16 MiB connection window was intended to allow several concurrent streams to make progress, but it might be better to use 1 MiB there as well for closer alignment with the other implementations. What do you think?

vietj (changes requested):
can you also apply this for HTTP client ?

arnabnandy7:
> can you also apply this for HTTP client ?

I'll try

arnabnandy7:
> can you also apply this for HTTP client ?

It's done, please review

vietj:
it seems that the changes introduce regressions, e.g. in Http2Test.testClientReadStreamInWorker

arnabnandy7:
> it seems that the changes introduce regressions, e.g. in Http2Test.testClientReadStreamInWorker

looking into it

arnabnandy7:
@vietj please retrigger the CI's

vietj:
@arnabnandy7 thanks for investigating the issue with the flow control and providing a tentative fix.

I have reviewed more in depth this change and provide a different fix for both master and 5.1 because it is a bug independent of this change that should be fixed in both branches and also this bug does not happen with the multiplex implementation.

It can be fixed by changing the behavior of `writeData` implementation instead, this PR https://github.com/eclipse-vertx/vert.x/pull/6360 is now merged in master and fixes the bug.

Can you update your PR on top of master accordingly so we only keep the flow control related changes ?

arnabnandy7:
> @arnabnandy7 thanks for investigating the issue with the flow control and providing a tentative fix.
> 
> I have reviewed more in depth this change and provide a different fix for both master and 5.1 because it is a bug independent of this change that should be fixed in both branches and also this bug does not happen with the multiplex implementation.
> 
> It can be fixed by changing the behavior of `writeData` implementation instead, this PR https://github.com/eclipse-vertx/vert.x/pull/6360 is now merged in master and fixes the bug.
> 
> Can you update your PR on top of master accordingly so we only keep the flow control related changes ?

@vietj I'll do in sometime. Thank you for updating me.

vietj:
@arnabnandy7 thanks, I think after this change we are good to go and ship it for 5.2 release

arnabnandy7:
@vietj it's done, please check and let me know if any further changes required here.

PR #6270, opened 2026-08-01
Fix HTTP client load balancing for multi-homed hosts

### LINKED ISSUE #6200: HttpClient with a LoadBalancer uses only one IP for a multi-homed host
### Version

5.1.2

### Context

An `HttpClient` configured with a `LoadBalancer` is designed to resolve **all** of a host's A-records and balance requests across them (`HttpClientImpl`: `resolveAll = loadBalancer != null`; `OriginResolver` builds an `OriginServer` per resolved IP, all `available`). In practice it connects to a **single** IP for a multi-homed host.

### Notes
- `((VertxInternal) vertx).nameResolver().resolveAll("www.google.com")` returns all 8 addresses, so DNS resolution is fine — the issue is downstream in the built-in origin-resolution → endpoint → `selectServer` path.
- Workaround: supplying an explicit `AddressResolver` (e.g. `AddressResolver.mappingResolver(...)` with the resolved IPs) + the same `LoadBalancer` **does** round-robin across all IPs. So only the built-in (no-custom-resolver) path is affected.

### Reproducer
```java
Vertx vertx = Vertx.vertx();
HttpClient client = vertx.httpClientBuilder()
    .with(new HttpClientOptions().setProtocolVersion(HttpVersion.HTTP_1_1).setSsl(true))
    .withLoadBalancer(LoadBalancer.ROUND_ROBIN)   // no custom AddressResolver
    .build();

Set<String> ips = ConcurrentHashMap.newKeySet();
CountDownLatch latch = new CountDownLatch(24);
for (int i = 0; i < 24; i++) {
  client.request(new RequestOptions().setMethod(HttpMethod.GET)
          .setHost("www.google.com").setPort(443).setURI("/").setSsl(true))   // 8 A-records
      .compose(req -> req.send().compose(HttpClientResponse::body).map(b -> req.connection()))
      .onComplete(ar -> { if (ar.succeeded()) ips.add(ar.result().remoteAddress().hostAddress()); latch.countDown(); });
}
latch.await(40, TimeUnit.SECONDS);
System.out.println("distinct remote IPs = " + ips.size());   // prints 1, despite 8 A-records
```

### Expected
Requests spread across the host's resolved IPs per the policy (≈8 distinct remote IPs).

### Actual
All requests hit one IP. Same with `ROUND_ROBIN`, `RANDOM`, and `LEAST_REQUESTS` — and `RANDOM` is stateless, so the load balancer is selecting among a *single* server; the resolved address set never reaches the selector.

### Do you have a reproducer?

_No response_

arnabnandy7:
Motivation:

An `HttpClient` configured with a `LoadBalancer` resolved every address for a multi-homed host, but requests still connected to only one IP.

The resolved `InetAddress` instances retained the original hostname. As a result, the generated `SocketAddress` instances compared equally and shared the same connection-pool key, despite representing different IP addresses.

This change creates socket addresses using the resolved numeric IP while retaining the original `InetAddress`. Each resolved IP therefore receives a distinct connection pool and can be selected by the configured load balancer.

The existing DNS client-side load-balancing test now also verifies that DNS results produce distinct server addresses.

Fixes #6200

Conformance:

I have signed the Eclipse Contributor Agreement as explained in https://github.com/eclipse/vert.x/blob/master/CONTRIBUTING.md

I have followed the Vert.x code style guidelines: https://github.com/vert-x3/wiki/wiki/Vert.x-code-style-guidelines

Verification:

```
mvn -pl vertx-core -Dtest=Http1xTest#testDnsClientSideLoadBalancingEnabled test

Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/EndpointKey.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/EndpointKey.java
@@ -23,21 +23,27 @@ public final class EndpointKey {
   // Todo : that should become a scheme ???
   final boolean ssl;
   final HttpVersion protocol;
-  final SocketAddress server;
+  private final SocketAddress server; // For now don't expose
+  private final ProxyOptions proxyOptions; // For now don't expose
   final HostAndPort authority;
-  final ProxyOptions proxyOptions;
   final ClientSSLOptions sslOptions;
 
-  public EndpointKey(boolean ssl, HttpVersion protocol, ClientSSLOptions sslOptions, ProxyOptions proxyOptions, SocketAddress server, HostAndPort authority) {
-    if (server == null) {
-      throw new NullPointerException("No null server address");
-    }
+  public EndpointKey(boolean ssl, HttpVersion protocol, ClientSSLOptions sslOptions, SocketAddress server, HostAndPort authority) {
+    this.ssl = ssl;
+    this.protocol = protocol;
+    this.sslOptions = sslOptions;
+    this.proxyOptions = null;
+    this.authority = authority;
+    this.server = server.hostAddress() != null ? SocketAddress.inetSocketAddress(server.port(), server.hostAddress()) : server;
+  }
+
+  public EndpointKey(boolean ssl, HttpVersion protocol, ClientSSLOptions sslOptions, ProxyOptions proxyOptions, HostAndPort authority) {
     this.ssl = ssl;
     this.protocol = protocol;
     this.sslOptions = sslOptions;
     this.proxyOptions = proxyOptions;
     this.authority = authority;
-    this.server = server;
+    this.server = null;
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


vietj:
can you think of a test that would show the actual issue with a pool that is described ?

I don't think it is wrong that the resolvers returns a socket address with an host and an IP address field.

Instead I am thinking that the issue might not be in the resolver but instead in the client itself: I am thinking the EndpointKey server field should discard the hostname to implement the correct behavior.

arnabnandy7:
> can you think of a test that would show the actual issue with a pool that is described ?
> 
> I don't think it is wrong that the resolvers returns a socket address with an host and an IP address field.
> 
> Instead I am thinking that the issue might not be in the resolver but instead in the client itself: I am thinking the EndpointKey server field should discard the hostname to implement the correct behavior.

Sure, I'll think of something. It might take sometime.

vietj:
@arnabnandy7 sure

arnabnandy7:
> @arnabnandy7 sure

@vietj I've updated with the test cases, please review at your convenience.

```
HTTP/1 integration, EndpointKey, HTTP/1 metrics:
Tests run: 3, Failures: 0, Errors: 0

HTTP/2 metrics lifecycle:
Tests run: 8, Failures: 0, Errors: 0

HTTP/3 metrics lifecycle:
Tests run: 1, Failures: 0, Errors: 0
```

vietj:
thanks, this has been fixed

PR #6271, opened 2026-08-01
Fix HTTP client close hook leak

### LINKED ISSUE #6268: Closed HttpClient is never removed from the context CloseFuture, leaking every client built with HttpClientBuilder
### Version

5.1.5 (also present in the 5.x sources on master as far as I can tell)

### Context

Our application rebuilds its Druid HTTP client every 60 seconds inside a verticle and closes the previous client once its in-flight requests are done. Despite the close, every client ever built stays reachable.

Rebuilding a client that frequently is admittedly not a common pattern (we do it because of https://github.com/eclipse-vertx/vert.x/issues/6255). But the leak is not tied to that pattern: any client built with `httpClientBuilder()` and then closed is retained for the lifetime of the creating context.

There is no workaround through the public API either afaik: the object registered in the CloseFuture is the internal `HttpClientImpl`, while `build()` returns a `CleanableHttpClient` wrapper, so `ContextInternal.removeCloseHook(returnedClient)` does not match.

### Expected behavior

`client.close()` releases the client

### Steps to reproduce

```java
import io.vertx.core.AbstractVerticle;
import io.vertx.core.Vertx;
import io.vertx.core.http.HttpClientAgent;

public class Repro {

    public static void main(String[] args) throws Exception {
        Vertx vertx = Vertx.vertx();
        vertx.deployVerticle(new AbstractVerticle() {
            @Override
            public void start() {
                for (int i = 0; i < 10_000; i++) {
                    HttpClientAgent client = vertx.httpClientBuilder().build();
                    client.close();
                }
            }
        }).await();
        System.gc();
        Thread.sleep(2_000);
        System.gc();
        System.out.println("pid " + ProcessHandle.current().pid());
        Thread.sleep(120_000);
    }
}
```

```
$ jcmd <pid> GC.class_histogram | grep -E 'LegacyHttpClient|NetClientImpl'
   6:         10000        1280000  io.vertx.core.http.impl.HttpClientBuilderInternal$LegacyHttpClient
  17:         10000         720000  io.vertx.core.net.impl.tcp.NetClientImpl
  22:         10000         640000  io.vertx.core.net.impl.tcp.NetClientImpl$1
```

### Do you have a reproducer?

_No response_

arnabnandy7:
Motivation:

HTTP clients created with `httpClientBuilder()` register their internal closeable with the creating context's `CloseFuture`. Explicitly closing a client did not remove this hook, causing the context to retain the closed client and its resources for the context's lifetime.

This change removes the hook when the underlying HTTP client's close future completes. It also adds a regression test verifying that an explicitly closed client is detached from the context's `CloseFuture`.

Fixes #6268.

Conformance:

I have signed the Eclipse Contributor Agreement and adhered to the Vert.x code style guidelines.

Testing:

```text
mvn -pl vertx-core '-Dtest=io.vertx.tests.vertx.VertxTest#testCloseHttpClientRemovesCloseHook' test

Result: 1 test run, 0 failures, 0 errors.
```

### DIFF
--- a/vertx-core/src/main/generated/io/vertx/core/DeploymentOptionsConverter.java
+++ b/vertx-core/src/main/generated/io/vertx/core/DeploymentOptionsConverter.java
@@ -12,21 +12,6 @@ public class DeploymentOptionsConverter {
    static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json, DeploymentOptions obj) {
     for (java.util.Map.Entry<String, Object> member : json) {
       switch (member.getKey()) {
-        case "config":
-          if (member.getValue() instanceof JsonObject) {
-            obj.setConfig(((JsonObject)member.getValue()).copy());
-          }
-          break;
-        case "threadingModel":
-          if (member.getValue() instanceof String) {
-            obj.setThreadingModel(io.vertx.core.ThreadingModel.valueOf((String)member.getValue()));
-          }
-          break;
-        case "ha":
-          if (member.getValue() instanceof Boolean) {
-            obj.setHa((Boolean)member.getValue());
-          }
-          break;
         case "instances":
           if (member.getValue() instanceof Number) {
             obj.setInstances(((Number)member.getValue()).intValue());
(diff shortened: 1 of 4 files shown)


vietj:
thanks @arnabnandy7 , the fix is good but only works for non shared client, I made a PR here to fix a shared client bug and this bug at the same time : https://github.com/eclipse-vertx/vert.x/pull/6269

arnabnandy7:
> thanks @arnabnandy7 , the fix is good but only works for non shared client, I made a PR here to fix a shared client bug and this bug at the same time : https://github.com/eclipse-vertx/vert.x/pull/6269

Okay, thanks.

PR #6274, opened 2026-08-03
Preserve current context in Promise.promise

### LINKED ISSUE #4978: Future#fromCompletionStage(CompletionStage) with Context if it exists
Current version:
```
static <T> Future<T> fromCompletionStage(CompletionStage<T> completionStage) {
  Promise<T> promise = Promise.promise();
  completionStage.whenComplete((value, err) -> {
```
isn't it better to supply Vert.x context if it "exists"
```
static <T> Future<T> fromCompletionStage(CompletionStage<T> completionStage) {
  Context ctx = Vertx.currentContext();
  Promise<T> promise = (ctx instanceof ContextInternal) // ⇒ ≠ null
    ? ((ContextInternal)ctx).promise() 
    : Promise.promise();
  completionStage.whenComplete((value, err) -> {

```

leviYX:
Motivation:

Promise.promise()  currently creates a context-free promise even when called from a Vert.x context. As a result, handlers can run on the foreign thread that completes the promise. This also affects Future.fromCompletionStage(CompletionStage).

Changes:

- create a context-bound promise when a Vert.x context is current
- preserve the existing context-free behavior outside Vert.x
- keep timeout and Vert.x shutdown boundaries explicitly context-free
- document the conditional callback affinity
- add context, no-context, CompletionStage, timeout, shutdown, and network regression coverage

Fixes #4978

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/Future.java
+++ b/vertx-core/src/main/java/io/vertx/core/Future.java
@@ -666,7 +666,8 @@ public interface Future<T> extends AsyncResult<T> {
   /**
    * Bridges a {@link CompletionStage} object to a Vert.x future instance.
    * <p>
-   * The Vert.x future handling methods will be called from the thread that completes {@code completionStage}.
+   * When called from a Vert.x context, the Vert.x future handling methods will be called on that context.
+   * Otherwise, they will be called from the thread that completes {@code completionStage}.
    *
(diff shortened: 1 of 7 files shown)


vietj:
that is actually delibarate, most users should not need to create a promise and use an API that returns a future.

PR #6281, opened 2026-08-04
fix: resolve double-slash path in FileResolverImpl when fileName is root

### LINKED ISSUE #6280: Resolve files when absolute path is used, returns exception
### Version

io.vertx:vertx-core:4.5.30

### Context

When using Vertx, with PemTrustOptions for example, during the execution of the code, Vertx tries to load all the application properties from the quarkus classpath using the FileResolverImpl.

When this file is an absolute path reference, an exception is thrown. When a relative file is used, it works, but will represent other issues in the code.

Refering to line [147](https://github.com/eclipse-vertx/vert.x/blame/5a9bafbc965e5082ad6e6fdb2867fea2d66a0226/vertx-core/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java#L147)

This method also scans the entire classloader and starts to resolve all files.
But the resolve will generate an incorrect `subResource` reference in line [267](https://github.com/eclipse-vertx/vert.x/blame/5a9bafbc965e5082ad6e6fdb2867fea2d66a0226/vertx-core/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java#L267)

### Steps to reproduce

(minimal DemoResolve project in attachment [demoResolve.zip](https://github.com/user-attachments/files/30701396/demoResolve.zip))

1. Project containing only 1 resource file `application.yml` in `src/main/resources`
2. Test containing 1 `@QuarkusTest`
3. The test will try to resolve the (absolute) `/application.yml` file
4. Run test > gives exception

When changing 
`fileResolver.resolveFile("/application.yml");`
to 
`fileResolver.resolveFile("application.yml");`
it does not give an exception.

Problem lies in the fact that in this example, QuarkusTest writes an empty application.properties to a temp workfolder (e.g. `%TEMP%/startup-override9955225914473382599`)
The absolutePath will then find the startup folder (url) but reference is with a fileName '/' instead of `url.getPath()` which results in:
`//application.properties` > does not exist
instead of

`%TEMP%/startup-override9955225914473382599/application.properties` 

### Do you have a reproducer?

_No response_

waterWang:
## Description

Fixes #6280

When `FileResolverImpl.resolve()` is called with an absolute path (e.g., `/application.yml`), the `parentFile` loop in `resolveFile2()` reaches the root path `/` and calls `unpackUrlResource(url, "/", cl, true)`. In `unpackFromFileURL`, when iterating over the directory listing, the `subResource` construction concatenates `fileName + "/" + file`, producing `//application.yml` (double-slash prefix) instead of `/application.yml`.

This causes the subsequent `getValidClassLoaderResource(cl, subResource)` call to fail because `//application.yml` is not a valid classpath resource path.

### Root cause
`unpackFromFileURL` unconditionally adds `/` between `fileName` and `file`, but when `fileName` is already `/` (the root path), this produces a double-slash prefix.

### Fix
Use a conditional separator: append `/` only when `fileName` does not already end with `/`.

```java
String subResource = fileName + (fileName.endsWith("/") ? "" : "/") + file;
```

### Steps to reproduce
1. Call `fileResolver.resolveFile("/application.yml")` when the classloader has a resource root at a temp directory
2. Observe the exception caused by `//application.yml` not being found

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/file/impl/FileResolverImpl.java
@@ -264,7 +264,7 @@ public class FileResolverImpl implements FileResolver {
       String[] listing = resource.list();
       if (listing != null) {
         for (String file : listing) {
-          String subResource = fileName + "/" + file;
+          String subResource = fileName + (fileName.endsWith("/") ? "" : "/") + file;
           URL url2 = getValidClassLoaderResource(cl, subResource);
           unpackFromFileURL(url2, subResource, cl);
         }
@@ -410,12 +410,12 @@ public class FileResolverImpl implements FileResolver {
   /**
    * bundle:// urls are used by OSGi implementations to refer to a file contained in a bundle, or in a fragment. There
    * is not much we can do to get the file from it, except reading it from the url. This method copies the files by
-   * reading it from the url.
-   *
-   * @param url the url
-   * @param  fileName the file name used to cache the content
-   * @return the extracted file
-   */
+  * reading it from the url.
+  *
+  * @param url the url
+  * @param  fileName the file name used to cache the content
+  * @return the extracted file
+  */
   private File unpackFromBundleURL(URL url, String fileName, boolean isDir) {
     try {
       if ((getClassLoader() != null && isBundleUrlDirectory(url)) || isDir) {


timastered:
the file listing is done on resource (which potentially can be a new path on the classpath)
Doesn't this need to be taken into account if you want to set the name of the resource?
e.g. 
`String subResource = fileName + (fileName.endsWith("/") ? "" : "/") + file;`
change to
`String subResource = resource.getPath() + (fileName.endsWith("/") ? "" : "/") + file;`

Otherwise it will try to locate the file in the root of your application/running instance?


example in code:
* URL = `%TMP%/properties`
* listing on this resource gives back `application.properties`
* subResource should be = `%TMP%/properties/application.properties`?
  * instead of `/application.properties` in your example?

tsegismont:
Good points @timastered 

@waterWang can you add tests for this change? Thanks

PR #6283, opened 2026-08-05
Stop using deprecated NoStackTraceThrowable in vertx-core internals

### LINKED ISSUE #5012: Consider removing NoStackTraceThrowable
There are several places in the codebase where `NoStackTraceThrowable` is used to convey a failure which does not need a stacktrace.

This isn't a big issue with the callback or futurized API, but when using the virtual threads threading model or Kotlin coroutines, it forces users to catch `Throwable` in their applications.

`VertxException` extends `RuntimeException` and can be created without filling the stacktrace. In fact, we already have a `NoStackTraceException` which extends `VertxException`.

In Vert.x 5, we should consider removing `NoStackTraceThrowable`.

jnbdz:
## Motivation

Fixes #5012.

`NoStackTraceThrowable` and `NoStackTraceException` are `@Deprecated(forRemoval = true)`, and 17bad9d7e introduced the transition path: users should catch/observe `VertxException` instead, with the class removal itself planned for Vert.x 6.

However vertx-core internals still instantiate `NoStackTraceThrowable`, so failures created from a message (`Promise#fail(String)`, `Promise#tryFail(String)`, `Future#failedFuture(String)`, `Completable#fail(String)`) or from a `null` cause still surface the deprecated type at runtime.

## Changes

- Replace the internal instantiations in `Completable`, `Promise`, `FailedFuture` and `FutureImpl` with the equivalent `VertxException.noStackTrace(...)` factory. Message and no-stack-trace behaviour are unchanged; the runtime type broadens to `VertxException` (the direct superclass and the documented replacement). `NoStackTraceException` has no remaining usage.
- The deprecated classes themselves are left untouched for removal in Vert.x 6, per 17bad9d7e. Happy to remove them in this PR instead if that is preferred for 5.2.
- `NoStackTraceTimeoutException` is intentionally out of scope: it is not deprecated and must remain a `java.util.concurrent.TimeoutException` for the HTTP client timeout contract.
- Tests: `FutureTest` now pins the failure contract (`VertxException` exact class, message preserved, empty stack trace) for the message/null-cause paths.
- Additional test coverage: the shared `HttpClientTimeoutTest` suite now also runs over HTTP/3 (`Http3ClientTimeoutTest` wired through `Http3Configurator`) and asserts that client request timeout failures are `TimeoutException` instances carrying no stack trace across HTTP/1, HTTP/2 and HTTP/3. HTTP/3 tests relying on behaviours not yet available (pool saturation semantics, client re-creation from `HttpClientOptions`, request idle timeout on streams) are `@Ignore`d with documented reasons.

## Compatibility

Code catching `VertxException` (the documented replacement) is unaffected since `NoStackTraceThrowable extends VertxException`. Only `instanceof NoStackTraceThrowable` checks observe the change, which is the pattern the deprecation instructs migrating away from.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/Completable.java
+++ b/vertx-core/src/main/java/io/vertx/core/Completable.java
@@ -10,8 +10,6 @@
  */
 package io.vertx.core;
 
-import io.vertx.core.impl.NoStackTraceThrowable;
-
 /**
  * A view of something that can be completed with a success or failure.
  *
@@ -57,7 +55,7 @@ public interface Completable<T> {
    * @throws IllegalStateException when this instance is already completed or failed
    */
   default void fail(String message) {
-    complete(null, new NoStackTraceThrowable(message));
+    complete(null, VertxException.noStackTrace(message));
   }
 
   /**
--- a/vertx-core/src/main/java/io/vertx/core/Promise.java
+++ b/vertx-core/src/main/java/io/vertx/core/Promise.java
@@ -13,7 +13,6 @@ package io.vertx.core;
 import io.vertx.codegen.annotations.CacheReturn;
 import io.vertx.codegen.annotations.GenIgnore;
 import io.vertx.codegen.annotations.VertxGen;
-import io.vertx.core.impl.NoStackTraceThrowable;
 import io.vertx.core.impl.future.PromiseImpl;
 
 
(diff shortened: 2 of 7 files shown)


jnbdz:
Re-running ECA validation after linking Eclipse and GitHub accounts.

PR #6285, opened 2026-08-06
fix: avoid blocking WebSocket accept handshake on worker context

### LINKED ISSUE #6284: acceptHandshake blocks 10s waiting on a task it queued to its own executor (worker context), and the failure is unreportable
### Version

5.1.4 (regression; not present in 4.5.13)

### Context

Environment: vertx-core 5.1.4, Java 21.0.6, Linux x86_64.

On a `ThreadingModel.WORKER` verticle, `HttpServerRequest.toWebSocket()` can block its carrier thread for exactly 10s and then fail. It waits on a task it has just scheduled onto the context that is currently executing it — on a serialized executor that task cannot run until the wait returns.

The only symptom in the logs is a secondary `IllegalStateException: Response head already sent`; the real cause is discarded. Both problems are inside vertx-core — the application only calls `toWebSocket()`.

## Defect 1 — self-deadlock

`Http1ServerRequest.toWebSocket()` is internally `webSocketHandshake().compose(h -> h.accept())`. When that continuation is not inline it is delivered as a `FutureBase$EmitResultTask` through the context executor, so `accept()` runs *as a task on the serialized `TaskQueue`*.

`ServerWebSocketHandshaker.acceptHandshake()`, bytecode offsets in 5.1.4:

```
 58: WebSocketServerHandshaker.handshake(...)   // writes the 101
 64: Http1ServerResponse.completeHandshake()    // head COMMITTED
137: ContextInternal.execute(Runnable)          // queues the Netty pipeline swap
150: CompletableFuture.get(10, SECONDS)         // blocks waiting for the task queued at 137
159: new RuntimeException(cause)                // on timeout
```

`ContextBase.execute(Runnable)` runs inline only `if (executor().inThread())`, otherwise it enqueues. `WorkerExecutor.inThread()` is a per-executor `ThreadLocal`, so from within an `EmitResultTask` it takes the enqueue branch — appending the pipeline swap behind the task that is about to block on it.

The 10s value is a literal with no config key. Every observed failure consumed the *full* timeout (measured 10.003s and 10.069s), which is what distinguishes this from a race.

## Defect 2 — the error cannot report itself

The head is committed at offset **64**; the failure happens at **150**. `accept()` catches it and calls `rejectHandshake(BAD_REQUEST)` → `setStatusCode()` on an already-committed response:

```
java.lang.IllegalStateException: Response head already sent
    at io.vertx.core.http.impl.HttpServerResponseImpl.checkHeadWritten(HttpServerResponseImpl.java:120)
    at io.vertx.core.http.impl.HttpServerResponseImpl.setStatusCode(HttpServerResponseImpl.java:148)
```

The timeout `RuntimeException` is lost. **This stands independently of the threading issue** — any failure occurring between the commit point and the reject is currently unreportable.

## Observed stack

Captured live while blocked, via `jcmd Thread.print`:

```
"wp.APIServer.1-0"  java.lang.Thread.State: TIMED_WAITING (parking)
    at java.util.concurrent.CompletableFuture.timedGet(java.base@21.0.6)
    at java.util.concurrent.CompletableFuture.get(java.base@21.0.6)
    at io.vertx.core.http.impl.websocket.ServerWebSocketHandshaker.acceptHandshake(ServerWebSocketHandshaker.java:209)
    at io.vertx.core.http.impl.websocket.ServerWebSocketHandshaker.accept(ServerWebSocketHandshaker.java:106)
    at io.vertx.core.http.impl.http1.Http1ServerRequest.lambda$toWebSocket$0(Http1ServerRequest.java:414)
    at io.vertx.core.impl.future.Composition.complete(Composition.java:40)
    at io.vertx.core.impl.future.FutureBase$EmitResultTask.run(FutureBase.java:125)
    at io.vertx.core.impl.WorkerExecutor$1.execute(WorkerExecutor.java:68)
    at io.vertx.core.impl.WorkerTask.run(WorkerTask.java:57)
    at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:86)
```

`Composition.complete` → `EmitResultTask` → `TaskQueue.run` is the point: the handshake is running as a queued continuation, not inline. Two dumps taken 2s apart held different `CompletableFuture$Signaller` instances, so these are separate handshakes each blocking rather than one long wait.

## 4.5.13 for comparison

`Http1xServerConnection`, `ServerWebSocketImpl` and `Http1xServerRequest` in 4.5.13 contain **zero** `CompletableFuture.get`/`await`. `doHandshake()` commits *after* a successful handshake and removes the pipeline handler inline; `createWebSocket()` is fully promise-based. Both regressions arrived together: the commit point moved ahead of the failure point, and an inline pipeline swap became a cross-thread dispatch plus a blocking wait.

## Impact

WebSocket upgrades intermittently stall for 10s and then fail, so a SockJS event-bus bridge never connects and the application appears to hang rather than error. Measured on one server: 76 failed upgrades in ~35 minutes of normal traffic; zero on the same build with the verticle deployed as `EVENT_LOOP`.

`setWorkerPoolSize(N)` does not help — a verticle instance is serialized by its `TaskQueue` regardless of pool size.

## Calling code

Reduced from the application where this occurs. The only Vert.x entry point involved is `request.toWebSocket()`, reached through `SockJSHandler`.

Deployment — N instances, each with its own single-thread worker pool:

```java
for (int i = 0; i < instances; i++) {                       // 4 in our deployment
    vertx.deployVerticle(new ApiServerVerticle(), new DeploymentOptions()
            .setThreadingModel(ThreadingModel.WORKER)
            .setWorkerPoolSize(1)
            .setWorkerPoolName("wp.APIServer." + i));
}
```

Route — SockJS event-bus bridge, WEBSOCKET the only enabled transport:

```java

SockJSHandler sockJs = SockJSHandler.create(vertx, new SockJSHandlerOptions()
        .setHeartbeatInterval(60000)
        .addDisabledTransport(Transport.XHR.name())
        .addDisabledTransport(Transport.JSON_P.name())
        .addDisabledTransport(Transport.EVENT_SOURCE.name())
        .addDisabledTransport(Transport.HTML_FILE.name()));

router.route("/eventbus/*")
        .handler(CorsHandler.create())
        .failureHandler(ctx -> {
            // guarded, because the upgrade completes asynchronously and ending an
            // already-ended response here throws "Response head already sent" as well
            if (!ctx.response().ended()) {
                ctx.response().setStatusCode(ctx.statusCode() > 0 ? ctx.statusCode() : 403).end();
            }
        })
        .subRouter(sockJs.bridge(new SockJSBridgeOptions()
                .addInboundPermitted(new PermittedOptions().setAddress("server"))
                .addOutboundPermitted(new PermittedOptions().setAddressRegex("user\\..*"))));

vertx.createHttpServer(new HttpServerOptions().setSsl(true))
        .requestHandler(router)
        .listen(443);
```

`SockJSHandler`'s WebSocket transport calls `routingContext.request().toWebSocket()`, which is where the stack above originates.

Full thread dumps available on request.

### Steps to reproduce

1. Deploy a verticle with `ThreadingModel.WORKER`, `setWorkerPoolSize(1)`, a named worker pool, and several instances (reproduced with 4).
2. Have it serve a SockJS event-bus bridge with only the WEBSOCKET transport enabled (see "Calling code" above).
3. Drive WebSocket upgrades against `/eventbus/websocket` concurrently with other HTTP traffic.
4. Upgrades intermittently stall for exactly 10s, then fail with `IllegalStateException: Response head already sent`. A thread dump taken during the stall shows the carrier thread parked in `CompletableFuture.timedGet` inside `ServerWebSocketHandshaker.acceptHandshake`.

Note on reproducing this: the trigger is `webSocketHandshake()` completing asynchronously, so that `.compose(...)` delivers the continuation via `FutureBase$EmitResultTask` rather than running it inline. Under light synthetic load the future is usually already complete, the continuation runs inline, `inThread()` is true, the pipeline swap executes immediately and nothing blocks — a standalone harness driving 2,000 upgrades at 25-way concurrency did **not** reproduce it. It reproduces reliably under real mixed traffic.

### Do you have a reproducer?

_No response_

waterWang:
Fixes #6284

## Problem
On a `ThreadingModel.WORKER` verticle, `HttpServerRequest.toWebSocket()` can block its carrier thread for exactly 10s and then fail. `ServerWebSocketHandshaker.acceptHandshake()` queues the Netty pipeline-swap task onto the context executor and then blocks on `CompletableFuture.get(10, SECONDS)` waiting for it. On a serialized worker executor the queued task cannot run until the wait returns, causing a self-deadlock. The subsequent `IllegalStateException: Response head already sent` hides the real cause.

## Fix
Make `acceptHandshake()` asynchronous: it now returns a `Future<ServerWebSocket>` and the pipeline swap + websocket setup runs on the event-loop context via a `Promise`, eliminating the blocking wait entirely. `accept()` chains on the result with `onSuccess`/\/`onFailure` instead of blocking.

Affects WebSocket/SockJS upgrades on worker verticles — the worker thread is never blocked, so the 10s stall and the unreportable failure are both removed.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/websocket/ServerWebSocketHandshaker.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/websocket/ServerWebSocketHandshaker.java
@@ -37,9 +37,7 @@ import java.security.cert.Certificate;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.RejectedExecutionException;
-import java.util.concurrent.TimeUnit;
 
 import static io.netty.handler.codec.http.HttpResponseStatus.BAD_REQUEST;
 import static io.vertx.core.spi.metrics.Metrics.METRICS_ENABLED;
@@ -101,21 +99,12 @@ public class ServerWebSocketHandshaker extends FutureImpl<ServerWebSocket> imple
       }
       done = true;
     }
-    ServerWebSocket ws;
-    try {
-      ws = acceptHandshake();
-    } catch (Exception e) {
-      return rejectHandshake(BAD_REQUEST.code())
-        .transform(ar -> {
-          if (ar.succeeded()) {
-            return request.context().failedFuture(e);
-          } else {
-            // result is null
-            return (Future) ar;
-          }
-        });
-    }
-    tryComplete(ws);
+    acceptHandshake()
+      .onSuccess(ws -> tryComplete(ws))
+      .onFailure(e -> {
+        tryFail(e);
+        rejectHandshake(BAD_REQUEST.code());
+      });
     return this;
   }
 
@@ -168,7 +157,7 @@ public class ServerWebSocketHandshaker extends FutureImpl<ServerWebSocket> imple
     return response.setStatusCode(sc).end(status.reasonPhrase());
   }
 
-  private ServerWebSocket acceptHandshake() {
+  private Future<ServerWebSocket> acceptHandshake() {
     Http1ServerConnection httpConn = (Http1ServerConnection) request.connection();
     ChannelHandlerContext chctx = httpConn.channelHandlerContext();
     Channel channel = chctx.channel();
@@ -198,25 +187,23 @@ public class ServerWebSocketHandshaker extends FutureImpl<ServerWebSocket> imple
       webSocketConn.metric(httpConn.metric());
       return webSocketConn;
     });
-    CompletableFuture<Void> latch = new CompletableFuture<>();
+    Promise<ServerWebSocket> promise = request.context().promise();
     httpConn.context().execute(() -> {
-      // Must be done on event-loop
-      pipeline.replace(VertxHandler.class, "handler", handler);
-      latch.complete(null);
+      try {
+        // pipeline swap must happen on the event-loop context
+        pipeline.replace(VertxHandler.class, "handler", handler);
+        ServerWebSocketImpl webSocket = (ServerWebSocketImpl) handler.getConnection().webSocket();
+        if (METRICS_ENABLED && httpConn.httpMetrics != null) {
+          httpConn.httpMetrics.requestUpgraded(request.metric());
+          webSocket.setMetric(httpConn.httpMetrics.connected(request));
+        }
+        webSocket.registerHandler(httpConn.context().owner().eventBus());
+        promise.complete(webSocket);
+      } catch (Exception e) {
+        promise.fail(e);
+      }
     });
-    // This should actually only block the thread on a worker thread
-    try {
-      latch.get(10, TimeUnit.SECONDS);
-    } catch (Exception e) {
-      throw new RuntimeException(e);
-    }
-    ServerWebSocketImpl webSocket = (ServerWebSocketImpl) handler.getConnection().webSocket();
-    if (METRICS_ENABLED && httpConn.httpMetrics != null) {
-      httpConn.httpMetrics.requestUpgraded(request.metric());
-      webSocket.setMetric(httpConn.httpMetrics.connected(request));
-    }
-    webSocket.registerHandler(httpConn.context().owner().eventBus());
-    return webSocket;
+    return promise.future();
   }
 
   @Override


vietj:
we need a test for this

vietj:
See #6306

PR #6287, opened 2026-08-07
Improve pqc

anavarr:
- `STRICT` PQC policy now doesn't override user provided named groups if they're all PQ compliant
- groups comparisons, detection was case sensitive in SslEngineUtils and 21/.../JdkDependent, it is no longer
- logs describing the choices made by Vert.x regarding Ssl Engine and named groups selection are now 'warn' instead of 'debug': they will thus appear in Quarkus

added unit tests for SslEngineUtils checking the insensitiveness of the comparisons and the behaviour


anavarr:
I don't think my changes are involved in CI failure: 

--- 
```
[main] INFO io.vertx.test.core.AsyncTestBase - Starting test: HttpPoolCapacityTest#testCapacity3
[ERROR] Errors: 
[ERROR]   Http2ClientTest.testClientResetServerStream2 ? Bind Address already in use
[ERROR] Tests run: 6952, Failures: 0, Errors: 1, Skipped: 141
```

vietj:
I fixed the build and upgraded smallrye-openssl, you should rebase this PR on latest master

anavarr:
@vietj the -PNativeIoUring fails but others pass, is it a true fail?

anavarr:
we need this fix for https://github.com/quarkusio/quarkus/issues/56262

vertx-core/src/main/java/io/vertx/core/net/impl/SslEngineUtils.java
```diff
@@ -42,19 +47,28 @@ public static List<String> resolveKeyExchangeGroups(List<String> groups, PqcEnfo
...
+          return PQ_COMPLIANT_GROUPS;
         }
-        return PQ_COMPLIANT_GROUPS;
+        // the user specified a set of PQ compliant groups, we can use it
+        return groups;
       case CLIENT_NEGOTIATED:
+        // we check if the user provided a set of named groups
         if (groups == null || groups.isEmpty()) {
-          log.debug("No key exchange groups list was specified, a default list containing X25519MLKEM768 is selected");
+          log.warn("No key exchange groups list was specified, the default list "+DEFAULT_KEY_EXCHANGE_GROUPS+" is used");
```
vietj: can you avoid changing the log to warn ?

vertx-core/src/main/java/io/vertx/core/net/impl/SslEngineUtils.java
```diff
@@ -42,19 +47,28 @@ public static List<String> resolveKeyExchangeGroups(List<String> groups, PqcEnfo
...
           return DEFAULT_KEY_EXCHANGE_GROUPS;
         }
-        if (groups.stream().noneMatch(PQ_COMPLIANT_GROUPS::contains)) {
-          log.debug("PQC enforcement policy is CLIENT_NEGOTIATED: prepending " + PQ_COMPLIANT_GROUPS + " to key exchange groups " + groups);
+        // if they did, we check that the set they provided contains at least one PQ compliant group
+        if (groups.stream().noneMatch(g -> PQ_COMPLIANT_GROUPS_UPPER.contains(g.toUpperCase()))) {
+          log.warn("PQC enforcement policy is CLIENT_NEGOTIATED: prepending " + PQ_COMPLIANT_GROUPS + " to key exchange groups " + groups);
```
vietj: please use log.debug

vertx-core/src/main/java/io/vertx/core/net/impl/SslEngineUtils.java
anavarr: The goal is that these logs appear in Quarkus, hence the switch to warn

vertx-core/src/main/java/io/vertx/core/net/impl/SslEngineUtils.java
anavarr: same rationale as earlier

vertx-core/src/main/java/io/vertx/core/net/impl/SslEngineUtils.java
vietj: I really want to avoid using warn here, isn't there a way to configure quarkus to display them ?

vertx-core/src/main/java/io/vertx/core/net/impl/SslEngineUtils.java
vietj: as I said, I think the purpose of logging is configuration and the default should be debug. I believe quarkus can be configured to display them if the category of the log is known.

anavarr: my bad I didn't see you were refering to those specific messages, you're right

PR #6291, opened 2026-08-08
Clarify raw vs normalized path usage in Vert.x Core documentation

mathias82:
## Motivation

Developers may assume that `HttpServerRequest#path()` / `HttpRequestHead#path()` returns a normalized version of the request path. In practice, the value reflects the path from the request URI and may contain repeated separators or path traversal markers such as `//` or `..`.

Making this behavior explicit helps avoid incorrect assumptions when implementing path-based logic, particularly when using Vert.x Web routing.

## What changed

- Clarify the `HttpRequestHead#path()` Javadoc to explicitly document that the returned path is not normalized.
- Reapply the clarification cleanly on top of the current `master`.

## Previous PR

This supersedes #5824, which became outdated and developed conflicts after changes in `master`.

The original feedback from #5824 has been taken into account while rebasing the change onto the current codebase.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/HttpRequestHead.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/HttpRequestHead.java
@@ -27,7 +27,16 @@ public interface HttpRequestHead {
   String uri();
 
   /**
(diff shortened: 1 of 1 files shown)


mathias82:
Please can you review it @vietj

PR #6295, opened 2026-08-08
fix: don't close pooled connection when resetting a completed HTTP/1.1 request (#6294)

### LINKED ISSUE #6294: Resetting a completed HTTP/1.1 request closes a pooled connection after reuse
### Version

4.4.1, 4.5.31

### Context

An HTTP/1.1 HttpClientRequest can retain access to its former connection after the request has completed. If the connection returns to the pool and is assigned to a newer request, calling reset() on the completed request closes the physical connection now serving the newer request.

Expected behavior

Calling reset() on an already completed request should be a no-op and must not affect a subsequent request using the pooled connection.

Actual behavior

The subsequent request fails with io.vertx.core.http.HttpClosedException: Connection was closed.

Probable cause

Http1xClientConnection.reset() uses responses.contains(stream) || stream.responseEnded to decide that a stream is in flight. responseEnded remains true after a completed stream has left the connection tracking queues. A late reset therefore marks the physical connection for closure after it has been reused.

The attached focused Http1xTest uses one keep-alive connection with pipelining disabled. Request A completes, Request B reuses the connection, and reset() is called on A while B is active. Expect: 100-continue makes the sequence deterministic; the underlying issue is not specific to 100 Continue.

Environment: macOS 26.5.2, OpenJDK 17.0.19.

### Steps to reproduce

1. Check out the Vert.x 4.5.31 tag.
2. Apply the attached test patch.
3. Run: mvn -Dtest=io.vertx.core.http.Http1xTest#testResetCompletedRequestDoesNotCloseReusedExpectContinueRequest test
4. Observe that the test fails with HttpClosedException: Connection was closed.

The equivalent behavior passes on the 4.2.6 code line and fails on the 4.4.1 and 4.5.31 code lines.

[vertx-4531-reset-completed-request-reproducer.patch.txt](https://github.com/user-attachments/files/30863244/vertx-4531-reset-completed-request-reproducer.patch.txt)

### Do you have a reproducer?

_No response_

waterWang:
## Fixes

Closes #6294

## Problem

`Http1ClientConnection.reset(Stream)` calls `close()` on the physical connection when the stream is not found in the `pending` queue. However, a fully completed stream (both request and response ended) is no longer tracked in `pending` — it has already been removed after the response was fully received. If the connection has been returned to the pool and reused by a subsequent request, the late `reset()` call on the completed stream incorrectly closes the physical connection, killing the in-flight request that is now using it.

## Root cause

```java
private Boolean reset(Stream stream) {
    if (stream.reset) {
      return null;
    }
    stream.reset = true;
    boolean removed = pending.remove(stream);
    if (!removed) {
      close();  // closes connection even for fully-completed streams
    }
    return removed;
  }
```

When `pending.remove(stream)` returns `false`, the code unconditionally calls `close()`. But a stream that has already completed (response received, request ended, `onClose()` called) is not in `pending` — this is the normal case, not an error.

## Fix

Added `!stream.closed` guard: only close the connection when the stream is still in-flight (not in `pending` and not already completed). A completed stream that receives a late `reset()` call is treated as a no-op.

```java
if (!removed && !stream.closed) {
    close();
}
```

The `closed` flag is set by `Stream.onClose()`, which is called when the stream has fully completed (both request and response ended). This is the definitive signal that a stream is done and should not trigger connection closure.

## Test

The attached test from the issue (Http1xTest.testResetCompletedRequestDoesNotCloseReusedExpectContinueRequest) reproduces the scenario: Request A completes, Request B reuses the connection, reset() is called on A while B is active. Without the fix, Request B fails with `HttpClosedException`. With the fix, the late reset is a no-op and Request B completes normally.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ClientConnection.java
@@ -1,1462 +1,506 @@
-/*
- * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
- *
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License 2.0 which is available at
- * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
- * which is available at https://www.apache.org/licenses/LICENSE-2.0.
- *
- * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
- */
-
-package io.vertx.core.http.impl.http1;
-
-import io.netty.buffer.ByteBuf;
-import io.netty.buffer.Unpooled;
-import io.netty.channel.*;
-import io.netty.handler.codec.DecoderResult;
-import io.netty.handler.codec.compression.Brotli;
-import io.netty.handler.codec.compression.ZlibCodecFactory;
-import io.netty.handler.codec.compression.Zstd;
-import io.netty.handler.codec.http.*;
-import io.netty.handler.codec.http.HttpHeaders;
-import io.netty.handler.codec.http.websocketx.*;
-import io.netty.handler.codec.http.websocketx.WebSocketFrame;
-import io.netty.handler.codec.http.websocketx.extensions.WebSocketClientExtensionHandler;
-import io.netty.handler.codec.http.websocketx.extensions.WebSocketClientExtensionHandshaker;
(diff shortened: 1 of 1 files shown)


vietj:
can you add a test for this ?

vietj:
the new content of the file in YML does not look like what it is supposed to do

PR #6303, opened 2026-08-10
fix: CleanableObject.shutdown() returns null when resource was GC'd

### LINKED ISSUE #6302: CleanableObject.shutdown() returns a null Future when the underlying resource was garbage collected (HttpClient.close() returning null)
### Version

5.1.5 — still present on current master (46fa0929b).

### Context

An `HttpClient` built with `vertx.httpClientBuilder().build()` is wrapped in a `CleanableHttpClient`, which references the underlying `HttpClientImpl` only through the `WeakReference` inside `CleanableObject.Action`. If the underlying resource has become unreachable and was garbage collected — for example after the Vert.x instance shut down and released its own references, while the caller still holds the wrapper — then `close()` / `shutdown(Duration)` on the wrapper returns **`null`** instead of a `Future<Void>`:

```java
private static class Action<T> extends WeakReference<CloseableResource<? extends T>> implements Runnable {

    private Future<Void> closeFuture;

    @Override
    public void run() {
      CloseableResource<? extends T> d = get();
      if (d != null) {                       // referent collected -> closeFuture stays null
        closeFuture = d.shutdown(timeout);
      }
    }
}

public final Future<Void> shutdown(Duration timeout) {
    ...
    if (action != null) {
      assert cleanable != null;
      action.timeout = timeout;
      cleanable.clean();
      return action.closeFuture;             // <- null in that case
    } else {
      return Future.succeededFuture();       // the repeated-shutdown branch never returns null
    }
}
```

The asymmetry with the `else` branch suggests the API is intended to never return `null`, and callers reasonably chain `close().onSuccess(...)` directly, which then throws `NullPointerException`.

This was hit in the wild by Quarkus: the OpenTelemetry OTLP exporter races its `shutdown()` against Vert.x termination, and on a GC-timing-sensitive JVM (JDK 25 Semeru in CI) `client.close()` returned `null`, producing an NPE and a flaky test. Full analysis in quarkusio/quarkus#55946.

Possibly related to the lifecycle machinery discussed in #6268 (context CloseFuture retention of clients) — that issue is about the resource being retained too strongly; this one is about what happens once it is no longer retained.

### Do you have a reproducer?

Not yet as a standalone project. It should reproduce by building a client, letting the underlying resource become weakly reachable (e.g. closing the Vert.x instance), forcing a GC until the `WeakReference` is cleared, and then calling `client.close()` — I can put together a test case as part of a fix PR.

The fix itself looks one-line: have `shutdown(Duration)` fall back to `Future.succeededFuture()` when `action.closeFuture` is `null` after `clean()` (matching the semantics of the already-shut-down branch). I'm happy to submit that PR if the approach sounds right.

### Extra

Observed on OpenJDK 25 (IBM Semeru) in Quarkus CI; the mechanism is GC-timing dependent rather than JVM-specific.

waterWang:
## Description

When `CleanableObject.shutdown(Duration)` is called after the underlying resource has been garbage collected (e.g., after Vert.x instance shutdown while the caller still holds a wrapper), the method returns `null` instead of a `Future<Void>`, causing `NullPointerException` in callers that chain `close().onSuccess(...)`.

### Root Cause

`CleanableObject.Action` extends `WeakReference<CloseableResource>`. When the `Cleanable.clean()` triggers `Action.run()`, the reference may already be cleared by GC — in that case `closeFuture` is never set and stays `null`. The `shutdown()` method then returns this `null` instead of a completed future.

### Fix

Return `Future.succeededFuture()` when `action.closeFuture` is `null` after `clean()`, matching the semantics of the already-shut-down branch (the `else` block).

This was hit in the wild by Quarkus: the OpenTelemetry OTLP exporter races its `shutdown()` against Vert.x termination, and on a GC-timing-sensitive JVM (JDK 25 Semeru in CI), `client.close()` returned `null`, producing an NPE and a flaky test (see quarkusio/quarkus#55946).

Closes #6302

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/CleanableObject.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/CleanableObject.java
@@ -81,7 +81,8 @@ public class CleanableObject<T> {
       assert cleanable != null;
       action.timeout = timeout;
       cleanable.clean();
-      return action.closeFuture;
+      Future<Void> closeFuture = action.closeFuture;
+      return closeFuture != null ? closeFuture : Future.succeededFuture();
     } else {
       return Future.succeededFuture();
     }


jnbdz:
This matches the fix suggested in #6302, and the placement in `shutdown()` looks right to me — it covers any path that leaves `action.closeFuture` unassigned, consistent with the already-shut-down branch.

Two things seem to be missing to get it over the line: the ECA check needs signing, and a regression test for the collected-referent path. I'm happy to contribute the test if that helps — it should be doable deterministically by constructing a `CleanableObject` over a resource, dropping the strong reference, forcing GC until the weak reference clears, and asserting `shutdown()` returns a non-null succeeded future.

vietj:
@jnbdz a test is important to cover this case, without it, merging this PR is speculative fix

vietj:
thanks for the contribution, here is a fix with a test : I opted for a different fix though

PR #6308, opened 2026-08-12
Support host names in DNS client setup

gsmet:
Motivation:

Vert.x 4 was supporting using a host name in the DNS setup (typically localhost, which is what Testcontainers returns in the Stork quickstart).

I'm not sure exactly if there's a good reason to break this behavior so I prepared a patch that restores the existing behavior, and is arguably more correct as only IP addresses are supported after this call and getHost() can return a host name.

Conformance:

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/dns/impl/DnsClientImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/dns/impl/DnsClientImpl.java
@@ -56,7 +56,7 @@ public class DnsClientImpl implements DnsClient {
     }
 
     DnsAddressResolverProvider provider = DnsAddressResolverProvider.create(vertx, vertx.nameResolver().options()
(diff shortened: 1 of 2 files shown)


vietj:
it must be a regression due to a reimplementation of the DNS client in Vert.x 5, thank you for the contribution.

gsmet:
Ideally, I would like it backported to 5.1.x if it gets in (at least if we don't upgrade to 5.2 soon in Quarkus main).

vietj:
I'll review it by end of this week, no worries for backport

gsmet:
Awesome, thanks!

vietj:
Actually your change implies using a blocking Java API that resolves the server host to an IP address. So it's less obvious than it appears first.

I made this change to let the server address resolution be made lazily at runtime instead and made it resolved with the vertx name resolver (that will understand localhost): #6309

Issue #6218, opened 2026-06-28
Summary of latest 60(to 28/6/2026) CI failures.


gaganis:
### Version

5.x 5.x-stable

### Context

I had trouble getting a green CI build of my https proxy PR #6215. In my effort to determine whether the failures I was seeing were due to my changes I did an analysis of the past failures to see if these failure preexisted due to flaky tests.

I decided to publish the report in case it is useful and to serve as a meta ticket to tackle these flaky test failures.  I also added a column to link related PRs or issues. I have posted a PR with a suggested fix for `NetTest.testListenDomainSocketAddress[Native]` #6217

| Frequency | Test | Error | Related issues / PRs |
|---|---|---|---|
| 9/60 (15%) | `NetTest.testEndHandlerCalledAfterAllEmissions` | IllegalState "Timed out" | #6219 |
| 7/60 (11.7%) | Http2 `*HeadersPutAll` / `testResponseTrailers*` | Illegal 'TE' header value / Stream reset |  |
| 4/60 (6.7%) | `Http2*.testSslHandshakeTimeout` | Timeout Unsatisfied checkpoint |  |
| 4/60 (6.7%) | `NetTest.testListenDomainSocketAddress[Native]` | leftover `.sock` / native timeout | #6217 |
| 3/60 (5%) | `it.networklogging.NetTest.testServerLogging` / `testNoLogging` | Timeout (120 s) |  |
| 2/60 (3.3%) | `WebSocketTest.*` (4 methods across 2 runs) | Timeout Unsatisfied checkpoint |  |
| 2/60 (3.3%) | `Http1xProxyTest.testNonProxyHosts3` | Bind "Address already in use" |  |
| 2/60 (3.3%) | `Http1xSendFileTest.testSendOpenRangeFileFromClasspath` |  |  |
| 2/60 (3.3%) | `MetricsTest.testThreadPoolMetrics*` (windows) |  |  |
| 2/60 (3.3%) | `NetTest.testServerWithIdleTimeoutSendChunkedFile` |  |  |
| 1/60 (1.7%) | `NetBandwidthLimitingTest.fileUploadIsThrottled` | Timeout (120 s) |  |
| 1/60 (1.7%) | `DNSTest.*` (windows) | Bind "Address already in use" |  |
| 1/60 (1.7%) | `SynchronizationTest.*` (4.x, windows) |  |  |
| 1/60 (1.7%) | `NetTest.testCloseUnwritableSocket` |  |  |
[...]

PR #6317, opened 2026-08-15
Keep the NetClient referenced in tests that rely on the connection staying open

jnbdz:
## Motivation

`Http2Test#testSslHandshakeTimeout` is listed in #6218 as failing in 4 out of 60 CI runs with `Timeout Unsatisfied checkpoint`.

The test opens a connection with `vertx.createNetClient().connect(...).await()`, keeps neither the client nor the socket referenced and returns, then the checkpoint waits for the server to report `handshake timed out after 1234ms`. Since a client that is no longer referenced is closed when it is garbage collected, under memory pressure the connection is closed *before* the handshake timeout happens: the server sees a plain close, never reports the `SSLHandshakeException`, and the checkpoint is never satisfied. Adding a `System.gc()` loop after the connect reproduces the CI failure deterministically:

```
java.util.concurrent.TimeoutException: Unsatisfied checkpoint
```

`HttpConnectionEarlyResetTest#testExceptionCaught` has the same pattern: it expects a connection reset (`SO_LINGER=0`) two seconds after connecting, but a client closed by the garbage collector closes the connection gracefully instead and the latch never counts down.

## Changes

Keep the `NetClient` in a field in both tests so it stays referenced for the duration of the test (test-only change). `Http2Test#testSslHandshakeTimeout` passes with the forced garbage collection in place once the client is referenced.

Related to #6218.

### DIFF
--- a/vertx-core/src/test/java/io/vertx/tests/http/Http2Test.java
+++ b/vertx-core/src/test/java/io/vertx/tests/http/Http2Test.java
@@ -19,6 +19,7 @@ import io.vertx.core.*;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.*;
 import io.vertx.core.net.JdkSSLEngineOptions;
+import io.vertx.core.net.NetClient;
 import io.vertx.core.net.NetSocket;
 import io.vertx.core.net.OpenSSLEngineOptions;
 import io.vertx.core.net.SSLEngineOptions;
@@ -52,6 +53,8 @@ import static org.junit.Assert.*;
  * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
  */
 public class Http2Test extends HttpTest {
+
+  private NetClient netClient;
 
   public Http2Test() {
     this(false);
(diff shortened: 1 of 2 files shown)


tsegismont (approved):
LGTM, thank you

tsegismont:
Please rebase and run `mvn spotless:apply` to apply formatting changes

vertx-core/src/test/java/io/vertx/tests/http/Http2Test.java
```diff
@@ -735,7 +738,10 @@ public void testSslHandshakeTimeout(Checkpoint checkpoint) throws Exception {
         }
       });
     startServer();
-    NetSocket so = vertx.createNetClient().connect(config.port(), config.host()).await();
+    // Keep a reference to the client: an unreferenced client is closed when it is garbage collected, which
+    // would close the connection before the handshake timeout happens
+    netClient = vertx.createNetClient();
```
vietj, line 743: I think that the netClient does not need to be a field on the class, it is fine to be a local variable as long as we use checkpoint.awaitSuccess()

vietj (changes requested):

vietj (changes requested):
added feedback

vietj:
fixed directly

Issue #5987, opened 2026-02-19
Unpooled HTTP connection closed unexpectedly


doxlik:
### Version

5.0.6

### Context

Hi, dear Vert.x team. 

I wanted to use one pipelined HTTP 1.1 connection per verticle approach, without using pool. So I got such server

```    
val httpServer = vertx
      .createHttpServer()
      .requestHandler { req ->
        req.response()
          .putHeader("content-type", "text/plain")
          .end("Hello from Vert.x!")
      }
      .listen(port)
      .coAwait()
```

and such client (actually proxy) 

```
    val client: HttpClientAgent = vertx.createHttpClient(
      HttpClientOptions().setPipelining(true).setPipeliningLimit(4).setKeepAlive(true).setKeepAliveTimeout(0)
    )

    val connection = client.connect(HttpConnectOptions().setHost("127.0.0.1").setPort(8080)).coAwait()
    connection.closeHandler { println("Client connection closed") }
    connection.exceptionHandler { println("Client connection got error: $it") }
    

    val httpServer = vertx
      .createHttpServer()
      .requestHandler { serverReq ->
        launch {
          val clientReq = connection.request(HttpMethod.GET, "/").coAwait()
          val clientResponse = clientReq.send().coAwait()
          val body = clientResponse.body().coAwait()
          serverReq.response().end(body)
        }
      }
      .listen(port)
      .coAwait()

```

And having multiple cURL requests is okay. But once having some load, let's say with **wrk** it start to behave randomly. If loading time is short like 10 seconds, it in mostly cases closes as soon as load stopped, but if load is longer like 60 second then it is closed in 30-40 seconds. But using pool with 1 connection of course solves this problem as it probably reconnect under the hood.

But I want to understand such behavior. I saw in docs 

```
public interface HttpClientAgent
...
When pipe-lining is enabled the connection will be automatically closed when all in-flight responses have returned and there are no outstanding pending requests to write.
...

```

but thus it either must be closed after the first curl request finished or not closed during 60 second load.

I would be very grateful for any assistance and suggestions. I am asking not because I need to close some small business task, but I want in future to become contributor to Vert.x and thus want to understand all its edge cases.

doxlik:
I think found the issue. 

The instance of returned http client is 
```
/**
 * A lightweight proxy of Vert.x {@link HttpClient} that can be collected by the garbage collector and release
 * the resources when it happens with a {@code 30} seconds grace period.
 *
[...]

Issue #5916, opened 2026-01-29
Breaking Change: NetClient.connect() connection timeout behavior changed in Vert.x 5


yinjun622:
### Version

Vert.x 4.x → Vert.x 5.0.7

### Context

When migrating from Vert.x 4 to Vert.x 5, I encountered unexpected connection timeout behavior with `NetClient.connect()`.

## Code that worked in Vert.x 4.x

```java
NetClientOptions options = new NetClientOptions()
    .setIdleTimeout(0)
    .setTcpKeepAlive(true)
    .setTcpNoDelay(true);

NetClient netClient = vertx.createNetClient(options);

// This worked fine in Vert.x 4.x - connection stayed open indefinitely
netClient.connect(socketaddress).onComplete(ar -> {
    // handle connection
});
```

## Problem in Vert.x 5

After upgrading to Vert.x 5, the connection automatically disconnects after **60 seconds**, even though `setIdleTimeout(0)` was configured.

## Root Cause

In `NetClientImpl.java` (Vert.x 5.0.7, lines 296-299):

```java
int connectTimeout = connectOptions.getTimeout();
if (connectTimeout < 0) {
  connectTimeout = options.getConnectTimeout();
}
```

**The issue:**
1. When using `netClient.connect(SocketAddress)`, an internal `ConnectOptions` is created
2. `ConnectOptions.getTimeout()` defaults to `-1`
3. When timeout is `-1`, it falls back to `NetClientOptions.getConnectTimeout()` which defaults to **60000ms**
4. This applies a 60-second timeout to the connection, regardless of `idleTimeout` settings

## Workaround

To maintain the Vert.x 4.x behavior (no connection timeout), you must explicitly create `ConnectOptions`:

```java
ConnectOptions conOptions = new ConnectOptions();
conOptions.setRemoteAddress(SocketAddress.inetSocketAddress(port, host));
conOptions.setTimeout(0);  // Explicitly disable connection timeout

netClient.connect(conOptions).onComplete(ar -> {
    // handle connection
});
```

## Expected Behavior

When `NetClientOptions.setIdleTimeout(0)` is configured, the `netClient.connect(SocketAddress)` convenience method should not apply an implicit 60-second connection timeout. The behavior should be consistent with Vert.x 4.x.

**Alternative:** If this is intentional behavior, it should be documented in:
1. The [Vert.x 4 to 5 Migration Guide](https://vertx.io/docs/guides/vertx-5-migration-guide/)
[...]

Issue #5853, opened 2025-12-22
Unstable test Http2Test.testHttpConnect, possibly due to garbage collected NetClient


ahus1:
### Version

master

### Context

The test sometimes fails, with a percentage of 20% (?) when running it locally in my IDE.
It also failed in the CI

See https://github.com/eclipse-vertx/vert.x/actions/runs/20429375873/job/58696486105

### Steps to reproduce

Run the test. 

When debugging this, I found that the following snippet connects to NetServer, but then disconnects immediately without sending any data. I assume that the NetClient is garbage collected. 

I'll prepare a PR where this no longer happens by keeping a reference. 

https://github.com/eclipse-vertx/vert.x/blob/785b9cfdcc1020aeb4f125b63f411e4e6a35150b/vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java#L4782-L4801

PR #6318, opened 2026-08-16
Document that clients must be kept referenced to avoid being shut down when garbage collected

jnbdz:
## Motivation

`HttpClient`, `NetClient` and `WebSocketClient` are lightweight proxies (`CleanableHttpClient`, `CleanableNetClient`, `CleanableWebSocketClient`) that are shut down when they are garbage collected. This is intentional, but it is not mentioned anywhere in the documentation, and users who create a client inline without keeping a reference to it regularly observe connections closed unexpectedly, requests refused with `Client is closed`, or WebSocket handshakes aborted, and report it as a bug: #5839 (WebSocket handshake), #5916 and #5987 (TCP / unpooled HTTP connections), #5577 (`Pool closed`). Even the test suite has been bitten (#5853, #6317).

## Changes

Docs only.

- `http.adoc`, "Creating an HTTP client": `IMPORTANT` admonition explaining that a client must stay referenced for as long as it is used, that an unreferenced client is shut down when garbage collected as if `HttpClient#shutdown()` had been called (new requests refused, connections closed), the failures this surfaces as, and where to keep the reference.
- `http.adoc`, "WebSockets on the client": the same note, mentioning that WebSockets with a handshake in progress are closed as well.
- `tcp.adoc`, "Creating a TCP client": the same note pointing at `NetClient#shutdown()` and the existing "TCP graceful shut down" section (sockets without a shutdown handler are closed immediately).

While there, in the WebSocket client section:
- fixed two truncated sentences (`e Vert.x ...` → `The Vert.x ...`, `en connecting ...` → `When connecting ...`) and stray leading spaces;
- the `clientWebSocketConfigurationAndConnect` example block was indented by one space, so it was rendered as a literal paragraph rather than a source block; it now renders as `[source,java]`.

Verified with `mvn compile` (docgen): the generated `http.adoc` / `tcp.adoc` contain the admonitions with resolved API links and the fixed source block.

### DIFF
--- a/vertx-core/src/main/asciidoc/http.adoc
+++ b/vertx-core/src/main/asciidoc/http.adoc
@@ -922,6 +922,12 @@ You create an {@link io.vertx.core.http.HttpClient} instance with the default co
 
 By default, the client supports HTTP/1, HTTP/2 in plain text.
 
+IMPORTANT: Keep a reference to the client for as long as you use it. A client that is no longer referenced is
+shut down when it is garbage collected, as if {@link io.vertx.core.http.HttpClient#shutdown()} had been called: new
+requests are refused and connections are closed, which can surface as unexpected `Connection was closed` or
+`Client is closed` failures. Typically you store the client in a field of the verticle or of the component that uses it,
+instead of creating it inline in the code that performs a request.
+
 === Configuring an HTTP client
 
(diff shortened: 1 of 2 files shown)


tsegismont (approved):
Looks good to me, thanks

PR #6329, opened 2026-08-22
fix: HTTP/1 active-request metrics leak when response ends before request body completes

waterWang:
Fix HTTP/1 active-request metrics leak when connection closes before request body completes

## Problem

When an HTTP/1 server ends the response before the request body has fully arrived, and the client then closes the connection without completing the declared body, Micrometer receives `requestBegin` and `responseEnd`, but **neither `requestEnd` nor `requestReset`**. The `vertx_http_server_active_requests` gauge therefore retains the request forever, increasing monotonically with each connection.

Example: a client sends `GET /` with `Content-Length: 1`, the server responds `401` early, the client closes without sending the byte.

## Root cause

`Http1ServerConnection.handleClosed()` only cleans up metrics for `responseInProgress` (via `handleException`), but when the response has already been sent the `responseInProgress` field is `null`. The `requestInProgress` still references the request whose body never completed (`requestEnd` was never reported), and when the connection closes the request metric is never reset.

## Fix

In `handleClosed()`, after handling `responseInProgress`, also report `requestReset` on `requestInProgress` when:
- it is a different request from `responseInProgress`
- its body did not complete (`!isEnded()`)

This matches the existing metric invariants: every `requestBegin` must be paired with exactly one terminal event (`requestEnd` or `requestReset`).

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerConnection.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/http1/Http1ServerConnection.java
@@ -513,6 +513,14 @@ public class Http1ServerConnection extends Http1Connection implements HttpServer
     if (responseInProgress != null) {
       responseInProgress.handleException(HttpUtils.CONNECTION_CLOSED_EXCEPTION);
     }
+    Http1ServerRequest requestInProgress = this.requestInProgress;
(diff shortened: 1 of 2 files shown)


tsegismont:
Superseded by https://github.com/eclipse-vertx/vert.x/pull/6336

PR #6332, opened 2026-08-25
Add a ChunkedWriteHandler when a socket is upgraded to SSL

### LINKED ISSUE #6331: ChunkedWriteHandler is not added to the pipeline on a runtime TLS upgrade
### Version

5.2.0

### Context

`sendFile` on a `NetSocket` upgraded to TLS at runtime (STARTTLS-style) always fails: the future
completes with `UnsupportedMessageTypeException` and the peer receives zero bytes. Works fine when
TLS is on from the start (`setSsl(true)`) or on a plain socket — only the *upgraded* socket breaks.

## Reproducer

Drop into `io.vertx.tests.net.NetTest`:

```java
@Test
public void testSendFileAfterTlsUpgrade() throws Exception {
  File dir = testFolder.newFolder();
  int size = 64 * 1024;
  File f = setupFile(dir.toString(), "up.dat", String.valueOf('a').repeat(size));

  server.connectHandler(socket -> {
    socket.upgradeToSsl(new ServerSSLOptions().setKeyCertOptions(Cert.SERVER_JKS.get()))
      .onComplete(onSuccess(v -> socket.sendFile(f.getAbsolutePath())
        .onComplete(onSuccess(v2 -> testComplete()))));
  });
  server.listen(1234, "localhost").await();

  NetSocket socket = client.connect(new ConnectOptions()
    .setPort(1234).setHost("localhost").setSsl(true)
    .setSslOptions(new ClientSSLOptions()
      .setHostnameVerificationAlgorithm("").setTrustAll(true))).await();

  AtomicInteger received = new AtomicInteger();
  socket.handler(b -> received.addAndGet(b.length()));
  await();
  assertEquals(size, received.get());
}
```

Observed — upgrade succeeds, `chunkedWriter` absent, `sendFile` fails, 0 of 65536 bytes received:

```
io.netty.handler.codec.UnsupportedMessageTypeException:
  io.vertx.core.net.impl.UncloseableChunkedNioFile (expected: io.netty.buffer.ByteBuf)
```

## Root cause

`ChunkedWriteHandler` is installed once at pipeline init and only if TLS is already on, but
`supportsFileRegion()` is evaluated at write time against the live pipeline. A runtime upgrade
flips the second condition without satisfying the first:

1. `NetServerImpl:295-297` (same in `NetClientImpl:111-112`) — `chunkedWriter` added only when
   `ssl || !fileRegionEnabled || !transport.supportFileRegion() || trafficShaping`. Plain server ⇒ not added.
2. `NetSocketImpl:165-171` — the upgrade adds `ssl` + `handshaker`, never `chunkedWriter`.
3. `VertxConnection:144-148` — `supportsFileRegion()` returns `false` once it sees the `SslHandler`:
   ```java
   boolean sslChannel = chctx.pipeline().get(SslHandler.class) != null;
   return vertx.transport().supportFileRegion() && !sslChannel && !isTrafficShaped();
   ```
4. `VertxConnection:549-556` — so `sendFile` takes the chunked branch:
   ```java
   if (!supportsFileRegion()) {
     writeToChannel(new UncloseableChunkedNioFile(fc, offset, length), writeFuture);
   } else {
     sendFileRegion(fc, offset, length, writeFuture);
   }
   ```

The `UncloseableChunkedNioFile` reaches the tail unconsumed and Netty rejects it.

## Suggested fix

Add `chunkedWriter` when the upgrade completes, if absent — in `NetSocketImpl.sslUpgrade` alongside
the `ssl`/`handshaker` insertion. Position matters: it must sit where the init-time one does so
chunks are encrypted. Alternative: install it unconditionally at init, costing one pass-through
handler per plain socket. Neither variant tested.

## Notes

- Reproduced server-side. `NetClientImpl:111-112` has the identical guard, so a client socket
  upgraded via `upgradeToSsl` is likely affected too — not verified.
- Silent from the peer's side: connection stays open, delivers nothing.

doxlik:
The ChunkedWriteHandler is installed once when the pipeline is created and only when the channel is encrypted at that point, whereas supportsFileRegion() is evaluated at write time against the live pipeline. A socket upgraded with upgradeToSsl() therefore stops using the file region transfer without gaining the handler that consumes the chunked file NetSocket#sendFile writes: the ChunkedNioFile reaches the tail of the pipeline and Netty rejects it with an UnsupportedMessageTypeException, the future fails and the peer receives no data.

  Install the handler when the upgrade completes, unless the pipeline already has one. It is
  inserted before the connection handler, so that the chunks it produces are written through the
  SSL handler and encrypted. This is the position the pipeline gives it when the channel is
  created encrypted.

Closes https://github.com/eclipse-vertx/vert.x/issues/6331


doxlik:
@vietj please have a look once you have time

vertx-core/src/test/java/io/vertx/tests/net/NetTest.java
```diff
@@ -2480,6 +2480,70 @@ public void sendFileServerToClient(Checkpoint checkpoint) throws Exception {
...
+    server.listen(1234, "localhost").await();
+    NetSocket socket = client.connect(new ConnectOptions()
+      .setPort(1234)
+      .setHost("localhost")
+      .setSsl(true)
+      .setSslOptions(new ClientSSLOptions()
+        .setHostnameVerificationAlgorithm("")
+        .setTrustAll(true))).await();
+    Buffer received = Buffer.buffer();
+    Promise<Void> done = Promise.promise();
```
vietj, line 2504: I am afraid this test can be racy because the handshake might happen before the client set the socket handler and receive data.

Instead the server should send the file after for instance receiving data from the client, I would suggest:

1. client connects
2. client set connect handler and end handler
3. client write a ping message to the server

The server should set a handler and perform send file operation when it receives the client signal.

vertx-core/src/test/java/io/vertx/tests/net/NetTest.java
```diff
@@ -2480,6 +2480,70 @@ public void sendFileServerToClient(Checkpoint checkpoint) throws Exception {
     sock.write("foo");
   }
 
+  @Test
+  public void testSendFileAfterTlsUpgrade() throws Exception {
+    File dir = testFolder.newFolder();
+    int size = 64 * 1024;
+    String content = String.valueOf('a').repeat(size);
```
vietj, line 2487: instead you can use TestUtils create random alpha string method

vertx-core/src/test/java/io/vertx/tests/net/NetTest.java
```diff
@@ -2480,6 +2480,70 @@ public void sendFileServerToClient(Checkpoint checkpoint) throws Exception {
...
+  }
+
+  @Test
+  public void testSendFileFromClientAfterTlsUpgrade() throws Exception {
+    File dir = testFolder.newFolder();
+    int size = 64 * 1024;
+    String content = String.valueOf('a').repeat(size);
+    File f = setupFile(dir.toString(), "upgraded-client.dat", content);
+    Buffer received = Buffer.buffer();
+    Promise<Void> done = Promise.promise();
```
vietj, line 2525: instead I think this should be a checkpoint object instead of a promise

vertx-core/src/test/java/io/vertx/tests/net/NetTest.java
```diff
@@ -2480,6 +2480,70 @@ public void sendFileServerToClient(Checkpoint checkpoint) throws Exception {
     sock.write("foo");
   }
 
+  @Test
+  public void testSendFileAfterTlsUpgrade() throws Exception {
+    File dir = testFolder.newFolder();
+    int size = 64 * 1024;
+    String content = String.valueOf('a').repeat(size);
+    File f = setupFile(dir.toString(), "upgraded.dat", content);
+    Promise<Void> sent = Promise.promise();
```
vietj, line 2489: this should be a checkpoint instead of a promise

vietj (changes requested):

doxlik:
@vietj thank you for comments. I tried to fix them, can you please check my changes.

vietj:
@doxlik can you provide a PR for 5.1 as well ?

PR #6334, opened 2026-08-26
fix: handle null address in NioTransport.convert to avoid NPE

### LINKED ISSUE #6333: NPE on NioTransport.convert(null)
### Version

5.0.12

### Context

I'll start by saying I am not sure how to reproduce this, but I have seen it occur in my production system.

```
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "address" is null
at io.vertx.core.impl.transports.UnixDomainSocketNioTransport.isUnixDomainSocketAddress(UnixDomainSocketNioTransport.java:57)
at io.vertx.core.impl.transports.NioTransport.convert(NioTransport.java:49)
at io.vertx.core.datagram.impl.DatagramSocketImpl.localAddress(DatagramSocketImpl.java:302)
at (my code)
```

DatagramSocketImpl does
```
  @Override
  public SocketAddress localAddress() {
    return context.owner().transport().convert(channel.localAddress());
  }
```

The Javadoc for `DatagramChannel.localAddress()` indicates that it can return null if the channel is not bound. But when you pass that to `NioTransport.convert` it throws NullPointerException. 

### Steps to reproduce

_No response_

### Do you have a reproducer?

_No response_

waterWang:
Fixes #6333

`DatagramChannel.localAddress()` returns null when the channel is not bound, and passing null to `NioTransport.convert(SocketAddress)` causes a NullPointerException in `isUnixDomainSocketAddress` (which calls `address.getClass()`).

Epoll/KQueue transports already handle this safely via `instanceof` checks (returning null for non-InetSocketAddress). This aligns NioTransport by returning null for null input.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/impl/transports/NioTransport.java
+++ b/vertx-core/src/main/java/io/vertx/core/impl/transports/NioTransport.java
@@ -48,6 +48,9 @@ public class NioTransport implements Transport {
 
   @Override
   public io.vertx.core.net.SocketAddress convert(SocketAddress address) {
+    if (address == null) {
+      return null;
+    }
     if (unixDomainSocketNioTransport != null && unixDomainSocketNioTransport.isUnixDomainSocketAddress(address)) {
       return unixDomainSocketNioTransport.convert(address);
     }


tsegismont:
Superseded by https://github.com/eclipse-vertx/vert.x/pull/6335

PR #6335, opened 2026-08-27
NPE thrown by isUnixDomainSocketAddress

tsegismont:
See #6333

NPE thrown when the provided address is null (may happen if checking the local address of an unbounded server).

Other transports are protected from this because they use instanceof. This transport can't use it because it uses method lookups.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/datagram/impl/DatagramSocketImpl.java
@@ -328,7 +328,8 @@ public class DatagramSocketImpl implements DatagramSocket, MetricsProvider, Clos
 
   @Override
   public SocketAddress localAddress() {
-    return context.owner().transport().convert(channel.localAddress());
+    InetSocketAddress localAddress = channel.localAddress();
+    return localAddress != null ? context.owner().transport().convert(localAddress) : null;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


vertx-core/src/main/java/io/vertx/core/impl/transports/UnixDomainSocketNioTransport.java
```diff
@@ -54,7 +54,7 @@ SocketAddress convert(io.vertx.core.net.SocketAddress address) {
   }
 
   boolean isUnixDomainSocketAddress(SocketAddress address) {
-    return unixDomainSocketAddressClass.isAssignableFrom(address.getClass());
+    return address != null && unixDomainSocketAddressClass.isAssignableFrom(address.getClass());
```
vietj: I think this should never be called with a null argument, the caller must check

vietj (changes requested):

vertx-core/src/main/java/io/vertx/core/impl/transports/UnixDomainSocketNioTransport.java
tsegismont: I kept this change for safety and update the datagram impl.

vietj: please don't, a null argument is considered as a programming error and should not be silently ignored, instead we should be aware of it and fix the caller.

tsegismont: PR updated

PR #6336, opened 2026-08-28
Fix HTTP/1 active-request metric lifecycle when connection closes before request body completes

tsegismont:
When the server sends an early response before the request body is fully received and the client then closes the connection, responseEnd() is already called by reportResponseComplete() at response-write time. The additional requestReset() call in handleClosed() violates the SPI contract of exactly one terminal call per request.

Fix handleClosed() by delegating to requestInProgress.handleException() behind a requestInProgress.response != null guard, which skips pipelined requests where handleBegin() was never called (avoiding a null metric NPE) and relies on the existing !response.ended() check to prevent a double terminal call.

Apply the same response != null guard to handleException() to prevent the equivalent NPE under IO errors.

Add two regression tests using Vert.x NetClient: one verifying that responseEnd is the terminal call when an early 401 is sent and the client closes before delivering the body; one verifying that requestReset is the terminal call when the client closes before the body with no response sent.

Assisted-by: Anthropic Claude (claude-sonnet-4-6)

### DIFF
--- a/vertx-core/src/test/java/io/vertx/tests/metrics/Http1xMetricsTest.java
+++ b/vertx-core/src/test/java/io/vertx/tests/metrics/Http1xMetricsTest.java
@@ -13,12 +13,17 @@ package io.vertx.tests.metrics;
 import io.vertx.core.ThreadingModel;
 import io.vertx.core.http.HttpClientOptions;
 import io.vertx.core.http.HttpVersion;
+import io.vertx.core.net.NetClient;
+import io.vertx.test.fakemetrics.FakeHttpServerMetrics;
+import io.vertx.test.fakemetrics.FakeMetricsBase;
+import io.vertx.test.fakemetrics.HttpServerMetric;
 import io.vertx.test.http.HttpConfigurator;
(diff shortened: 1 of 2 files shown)


vertx-core/src/test/java/io/vertx/tests/metrics/Http1xMetricsTest.java
```diff
@@ -46,4 +51,45 @@ public void testAllocatedStreamResetShouldNotCallMetricsLifecycle() throws Excep
...
+    server.requestHandler(req -> {
+      FakeHttpServerMetrics metrics = FakeMetricsBase.httpMetricsOf(server);
+      metricRef.set(metrics.getRequestMetric(req));
+      // Early response: end before the request body is consumed
+      req.response().setStatusCode(401).end();
+    });
+    startServer(testAddress);
+
+    // Send headers with Content-Length: 1 but close the connection without sending the body byte
+    try (Socket socket = new Socket("127.0.0.1", testAddress.port())) {
```
vietj: I'd prefer we use vertx NetClient for this and not java socket API

PR #6348, opened 2026-09-07
Send the request body again when a redirection keeps the request method

jnbdz:
### Motivation

Since #6212 the body of a `QUERY` request is buffered (within `ClientRedirectConfig#setMaxBufferedSize`) and sent again when the request is redirected. The body of any other request is still dropped: a custom redirect handler that follows a `307` or `308` redirection of a `POST` or `PUT` request, as RFC 9110 requires for those status codes, ends up sending the redirected request with an empty body. This is what quarkusio/quarkus#41751 reports for the Quarkus REST Client, and @vietj said there that an extension of the redirect handling in Vert.x 5 would be welcome if there was demand for it.

### Changes

- the body of any request whose method can carry one (everything but `GET`, `HEAD` and `CONNECT`) is buffered while `followRedirects` is set, within the existing max buffered size, and sent again whenever the redirected request keeps the method. This is the QUERY mechanism generalised, nothing else changes for QUERY;
- when the body exceeds the limit, redirections other than `303` are not followed, as already documented for QUERY, since they may keep the method and need the body. A `303` still turns the request into a bodyless `GET` and is followed;
- the redirected request now buffers its own body instead of sharing the buffer list of the previous request: sharing it appended the replayed body to the same list, so a request redirected twice sent the body twice;
- the default redirect handler policy is unchanged: `307` and `308` redirections of methods other than `GET`, `HEAD` and `QUERY` are still only followed by a custom handler. Javadoc and the `http.adoc` redirection section are updated accordingly.

### Tests

Five tests in `HttpTest` use a redirect handler that keeps the method on `307`/`308`: `POST` on `307` and `PUT` on `308` with a body, a chunked `POST` written in two buffers, a body over the limit on `307` (the `307` is returned to the caller) and a body over the limit on `303` (followed as `GET`). The first four fail on `master`, all pass with this change, and the existing `testFollowRedirect*` tests as well as `Http1xTest`, `Http2Test` and `Http1xTLSTest` (1024 tests) pass on both HTTP versions.

### DIFF
--- a/vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
+++ b/vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
@@ -63,6 +63,7 @@ public class HttpClientRequestImpl extends HttpClientRequestBase implements Http
   private String traceOperation;
   private int maxRedirectBufferSize = HttpClientOptions.DEFAULT_MAX_REDIRECT_BUFFERED_SIZE;
   private List<Buffer> bodyBuffer;
+  private boolean bodyBufferDiscarded;
 
   public HttpClientRequestImpl(HostAndPort authority, HttpConnection connection, HttpClientStream stream) {
     super(authority, connection, stream, stream.context().promise(), HttpMethod.GET, "/");
@@ -403,31 +404,46 @@ public class HttpClientRequestImpl extends HttpClientRequestBase implements Http
     next.setMaxRedirects(maxRedirects);
     HttpClientRequestImpl nextImpl = (HttpClientRequestImpl) next;
     nextImpl.numberOfRedirections = numberOfRedirections + 1;
-    nextImpl.bodyBuffer = bodyBuffer;
     endFuture.onComplete(ar -> {
       if (ar.succeeded()) {
         if (timeoutMs > 0) {
           next.idleTimeout(timeoutMs);
         }
-        if (next.getMethod() == HttpMethod.QUERY && bodyBuffer != null && !bodyBuffer.isEmpty()) {
-          Buffer redirectBody;
-          if (bodyBuffer.size() == 1) {
-            redirectBody = bodyBuffer.get(0);
-          } else {
-            CompositeByteBuf composite = Unpooled.compositeBuffer();
-            for (Buffer b : bodyBuffer) {
-              composite.addComponent(true, ((BufferInternal) b).getByteBuf());
-            }
-            redirectBody = BufferInternal.buffer(composite);
+        if (getMethod().equals(next.getMethod()) && canRedirectBody(getMethod())) {
+          // The redirection keeps the method, so the body is sent again
+          if (bodyBufferDiscarded) {
+            handler.tryFail(new VertxException("Cannot follow the " + getMethod() + " redirection: the request body " +
+              "exceeds the maximum size that can be buffered for redirections (" + maxRedirectBufferSize + " bytes)", true));
+            next.reset(0);
+            return;
           }
-          next.end(redirectBody);
-        } else {
-          next.end();
+          if (bodyBuffer != null && !bodyBuffer.isEmpty()) {
+            Buffer redirectBody;
+            if (bodyBuffer.size() == 1) {
+              redirectBody = bodyBuffer.get(0);
+            } else {
+              CompositeByteBuf composite = Unpooled.compositeBuffer();
+              for (Buffer b : bodyBuffer) {
+                composite.addComponent(true, ((BufferInternal) b).getByteBuf());
+              }
+              redirectBody = BufferInternal.buffer(composite);
+            }
+            next.end(redirectBody);
+            return;
+          }
         }
+        next.end();
       } else {
         next.reset(0);
       }
     });
+  }
+
+  /**
+   * @return whether the body of a request using {@code method} is sent again when a redirection keeps the method
+   */
+  private static boolean canRedirectBody(HttpMethod method) {
+    return method != HttpMethod.GET && method != HttpMethod.HEAD && method != HttpMethod.CONNECT;
   }
 
   private void handleContinue(Void v) {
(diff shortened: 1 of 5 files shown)


vietj:
@jnbdz are you using an LLM to assist your contributions ?

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -403,33 +404,48 @@ private void handleNextRequest(HttpClientRequest next, Promise<HttpClientRespons
...
       } else {
         next.reset(0);
       }
     });
   }
 
+  /**
+   * @return whether the body of a request using {@code method} is sent again when a redirection keeps the method
+   */
+  private static boolean canRedirectBody(HttpMethod method) {
```
vietj, line 445: I think this should be the reverse, we should only redirect bodies when the method is QUERY or POST or PUT

vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java
```diff
@@ -3518,6 +3518,165 @@ public void testFollowRedirectQueryWithConfiguredBodyExceedingLimit() throws Exc
...
+        .setFollowRedirects(true)
+        .send(expected)
+        .expecting(HttpResponseExpectation.SC_OK))
+      .await();
+  }
+
+  /**
+   * A client following 307 and 308 redirections with the same method, like an HTTP client compliant with RFC 9110 would do.
+   */
+  private HttpClientAgent keepMethodRedirectClient() {
```
vietj, line 3665: should be named preservingMethodRedirectClient

vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java
```diff
@@ -3518,6 +3518,165 @@ public void testFollowRedirectQueryWithConfiguredBodyExceedingLimit() throws Exc
...
+      .setHost(config.host())
+      .setPort(config.port());
+    client.request(opts).compose(req -> req
+        .setFollowRedirects(true)
+        .send(expected)
+        .expecting(HttpResponseExpectation.SC_OK))
+      .await();
+  }
+
+  private void testFollowRedirectWithBodyKeepingMethod(HttpMethod method, int statusCode, int bodySize) throws Exception {
```
vietj, line 3630: Preserving not Keeping

vertx-core/src/test/java/io/vertx/tests/http/HttpTest.java
```diff
@@ -3518,6 +3518,165 @@ public void testFollowRedirectQueryWithConfiguredBodyExceedingLimit() throws Exc
...
+    startServer(testAddress);
+    client = keepMethodRedirectClient();
+    RequestOptions opts = new RequestOptions()
+      .setMethod(POST)
+      .setHost(config.host())
+      .setPort(config.port());
+    client.request(opts).compose(req -> {
+        req.setFollowRedirects(true);
+        req.setChunked(true);
+        req.write(chunk1);
```
vietj, line 3562: the first write and end should be composed futures

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -403,33 +404,48 @@ private void handleNextRequest(HttpClientRequest next, Promise<HttpClientRespons
...
-            CompositeByteBuf composite = Unpooled.compositeBuffer();
-            for (Buffer b : bodyBuffer) {
-              composite.addComponent(true, ((BufferInternal) b).getByteBuf());
+        if (getMethod().equals(next.getMethod()) && canRedirectBody(getMethod())) {
+          // The redirection keeps the method, so the body is sent again
+          if (bodyBufferDiscarded) {
+            handler.tryFail(new VertxException("Cannot follow the " + getMethod() + " redirection: the request body " +
+              "exceeds the maximum size that can be buffered for redirections (" + maxRedirectBufferSize + " bytes)", true));
+            next.reset(0);
+            return;
```
vietj, line 418: I am not fan of this return, we can avoid it ?

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -403,33 +404,48 @@ private void handleNextRequest(HttpClientRequest next, Promise<HttpClientRespons
...
-        if (next.getMethod() == HttpMethod.QUERY && bodyBuffer != null && !bodyBuffer.isEmpty()) {
-          Buffer redirectBody;
-          if (bodyBuffer.size() == 1) {
-            redirectBody = bodyBuffer.get(0);
-          } else {
-            CompositeByteBuf composite = Unpooled.compositeBuffer();
-            for (Buffer b : bodyBuffer) {
-              composite.addComponent(true, ((BufferInternal) b).getByteBuf());
+        if (getMethod().equals(next.getMethod()) && canRedirectBody(getMethod())) {
+          // The redirection keeps the method, so the body is sent again
```
vietj, line 413: "preserves" the method, not "keeps"

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -63,6 +63,7 @@ public class HttpClientRequestImpl extends HttpClientRequestBase implements Http
   private String traceOperation;
   private int maxRedirectBufferSize = HttpClientOptions.DEFAULT_MAX_REDIRECT_BUFFERED_SIZE;
   private List<Buffer> bodyBuffer;
```
vietj, line 65: that should be renamed "bufferedBody" instead

vertx-core/src/main/java/io/vertx/core/http/impl/HttpClientRequestImpl.java
```diff
@@ -538,23 +556,22 @@ private Future<Void> doWrite(Buffer buff, boolean end, boolean connect) {
...
-            followRedirects = false;
-            bodyBuffer = null;
-          } else {
-            if (bodyBuffer == null) {
-              bodyBuffer = new ArrayList<>();
-            }
-            bodyBuffer.add(buff.copy());
+        }
+        if (currentLen + buff.length() > maxRedirectBufferSize) {
+          bodyBufferDiscarded = true;
```
vietj, line 568: why do we need to introduce bodyBufferDiscarded insted of using followRedirects ? it is not clear, adding another variable increases the complexity

vietj:
if we only add support for new methods, I don't get why the implementation of HttpClientRequest needs so many changes, can you avoid changing the implementation to keep changes to the strict minimum ?

vietj:
I also believe that this should be configurable and we should think more about the security implications of such redirections. QUERY introduces a clear semantic of redirecting a payload, for other HTTP methods it is not clearly defined, specially regarding cross origin redirection.

So until we have clearly defined and analyzed the security implications of this change, I am not in favor to implement and make this available out of the box.

PR #6361, opened 2026-09-14
Add a ChunkedWriteHandler when a socket is upgraded to SSL

doxlik:
The ChunkedWriteHandler is installed once when the pipeline is created and only when the channel is encrypted at that point, whereas supportsFileRegion() is evaluated at write time against the live pipeline. A socket upgraded with upgradeToSsl() therefore stops using the file region transfer without gaining the handler that consumes the chunked file NetSocket#sendFile writes: the ChunkedNioFile reaches the tail of the pipeline and Netty rejects it with an UnsupportedMessageTypeException, the future fails and the peer receives no data.

Install the handler when the upgrade completes, unless the pipeline already has one. It is inserted before the connection handler, so that the chunks it produces are written through the SSL handler and encrypted. This is the position the pipeline gives it when the channel is created encrypted.


doxlik:
@vietj hi. here is backport.

PR #6368, opened 2026-09-18
Support strict thread mode for HTTP/2 server streams

### LINKED ISSUE #6369: Support strict thread mode for HTTP/2 server streams
### Describe the feature

`HttpServerOptions.setStrictThreadMode(true)` only benefits HTTP/1 today: HTTP/2 server streams
ignore the option and every write goes through the per-stream outbound message queue, even when the
stream is bound to the connection event-loop and can only be written from it. Such a stream could
relay its writes to the connection directly, like HTTP/1 does, and reject writes made from another
thread. This applies to both the codec and the multiplex server implementations.

### Contribution

_No response_

doxlik:
When the server runs in strict thread mode with the event-loop threading model, a stream bound to the connection event-loop is exclusively written from that thread, therefore the outbound message queue is not needed: messages can be relayed to the connection directly, like HTTP/1 does.

The outbound queue of DefaultHttp2Stream becomes an OutboundWriteQueue with two implementations: QueuedWriteQueue, the existing OutboundMessageQueue, and DirectWriteQueue, which writes without queueing and rejects writes performed from another thread. Frames the stream window cannot accommodate are held by the Netty flow controller instead, which also fails their promises when the stream is reset.

HttpServerResponseImpl checks the calling thread before it mutates any state, so that a rejected write leaves the response usable.

Measured on a hello world server:

| Setup | h2load | Before | After |
| --- | --- | --- | --- |
| 8 event loops | `-c8 -m128` | 2.14M req/s | 2.48M req/s |
| 16 event loops | `-c16 -m128` | 2.42M req/s | 2.8-3M req/s |

Resolves https://github.com/eclipse-vertx/vert.x/issues/6369


doxlik:
@vietj Hi. Please have a look once you have time.

vietj:
actually I am not sure to keep this feature in vertx because I think it will raise too many issues on the long term to support it.

vietj:
also if I would merge that, before I would like to perform some work on the HTTP stream API to make more efficient ByteBuf <-> Object conversion for gRPC which means in practice rewriting some internals. 

So allow me for now to park this PR until further notice.

